PR #18, opened 2020-10-07
EVAKA-3882 Better E2E test development flow

teemu47:
#### Summary

Sometimes it was painful to develop E2E tests against dev DB because it was easiest to just clean the database after failed E2E test in order to avoid duplicate data entries when running the test again. But that strategy may have lost some valuable development data (e.g. fee decisions invoices, placements). So:
- Added possibility to run dev server against integration test database
- Added comments to `gradle.properties` to easily configure flyway to take effect in integration test DB

So if you want to develop E2E tests against integration test DB just start the server with `./gradlew bootRunTest`

To clean integration test database:
- edit `gradle.properties` (configs are commented out in the file)
- run `./gradlew flywayClean`
- restart the server

### DIFF
--- a/service/README.md
+++ b/service/README.md
@@ -55,6 +55,13 @@ Run integration tests (needs to have docker daemon running):
 ./gradlew integrationTest
 ```
 
+To start dev server connected to integration test database (suitable for e.g. running E2E tests without cluttering
+ development data):
+```sh
+./gradlew bootRunTest
+```
+Edit `gradle.properties` to run flyway commands against integration test database.
+
 Run linter autofix:
 
 ```sh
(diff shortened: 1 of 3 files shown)


service/README.md
```diff
@@ -55,6 +55,13 @@ Run integration tests (needs to have docker daemon running):
 ./gradlew integrationTest
 ```
 
+To start dev server connected to integration test database (suitable for e.g. running E2E tests without cluttering
+ development data):
+```sh
+./gradlew bootRunTest
+```
+Edit `gradle.properties` to run flyway commands against integration test database.
```
reynders, line 63: Enable running flyway commands against integration test database in `gradle.properties´" (I think this is better because it gives the proper impression that dev does not need to know how to do the db config change, just needs to use existing configuration)

reynders (approved):

PR #27, opened 2020-10-08
EVAKA-HOTFIX fixed area filter in occupancy report

massproduction:
#### Summary
Fixed area filter in occupancy report.


kustaahuhtala (approved):
Tested locally, works now 👍

PR #42, opened 2020-10-12
EVAKA-3893 preparation for value decisions

jaahon:
#### Summary

Merge fee decisions and invoices pages into one finance page that has its own tab navigation. Also use a better path for fee decisions api.

### DIFF
--- a/frontend/packages/employee-frontend/src/App.tsx
+++ b/frontend/packages/employee-frontend/src/App.tsx
@@ -7,7 +7,8 @@ import {
   BrowserRouter as Router,
   Route,
   Switch,
-  Redirect
+  Redirect,
+  useParams
 } from 'react-router-dom'
 import ChildInformation from '~/components/ChildInformation'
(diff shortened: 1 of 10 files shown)


frontend/packages/employee-frontend/src/App.tsx
```diff
@@ -157,28 +141,15 @@ function App() {
...
-              component={ensureAuthenticated(FeeDecisionDetailsPage)}
-            />
-            <RouteWithTitle
+            <Route
               exact
-              path="/invoices"
-              component={ensureAuthenticated(InvoicesPage)}
-              title={i18n.titles.invoices}
+              path="/finance/fee-decisions/:id"
+              component={FeeDecisionDetailsPage}
```
hruoho, line 147: Hmm shouldn't we have `ensureAuthenticated` here too?

jaahon: Yeah, good catch.

hruoho (approved):

PR #53, opened 2020-10-15
Value decisions

jaahon:
#### Summary

Generate value decisions similarly to fee decisions on same events. Add an elementary list view for value decisions.


frontend/packages/employee-frontend/src/api/invoicing.ts
```diff
@@ -16,6 +16,7 @@ import {
   FeeDecision,
   FeeDecisionDetailed,
   FeeDecisionSummary,
+  ValueDecisionSummary,
```
reynders, line 19: Globally: value decision -> voucher value decision

frontend/packages/employee-frontend/src/assets/i18n/fi.ts
```diff
@@ -1237,6 +1240,32 @@ export const fi = {
...
+      children: 'Lapset',
+      validity: 'Arvopäätös voimassa',
+      number: 'Numero',
+      status: 'Tila',
+      createdAt: 'Luotu',
+      sentAt: 'Lähetetty',
+      paging: 'Sivu'
+    },
+    buttons: {
+      checked: (count: number) =>
```
reynders, line 1258: No need to fix this, mut a simpler solution: checked: (count: number) => `Maksupäätöksiä valittu ${count}` ?

reynders (approved):

frontend/packages/employee-frontend/src/assets/i18n/fi.ts
jaahon: Yeah, but then it's not as fluent 😌

PR #56, opened 2020-10-16
EVAKA-3886 Purchased units to Varda

teemu47:
#### Summary

Update varda_child to use daycare organizer OID instead of basic numeric ID. Prepare for the case that daycare's organizer OID changes. Better support deleting data from Varda: handle soft delete coherently.


service/src/main/kotlin/fi/espoo/evaka/varda/VardaChildren.kt
```diff
@@ -93,35 +94,44 @@ private fun insertPerson(vardaPersonResponse: VardaPersonResponse, personId: UUI
...
+        WITH child_organizer AS (
+            SELECT p.child_id, d.oph_organizer_oid FROM placement p JOIN daycare d ON p.unit_id = d.id AND daterange(p.start_date, p.end_date, '[]') @> current_date
+        )
+        UPDATE varda_child vc SET oph_organizer_oid = (SELECT oph_organizer_oid FROM child_organizer WHERE child_id = vc.person_id)
+        """.trimIndent()
+
+    h.createUpdate(sql)
+        .execute()
+}
+
```
teemu47, line 109: What's the idea behind this? We'd want to populate `oph_organizer_id` column in `varda_child` with data from Varda at least once after we migrate the new schema. My hunch is that we should populate the column with Espoo's OID since it looks like every child in Varda now is under the Espoo `vakatoimija`. Of course we need to be sure about it before doing it.

service/src/main/kotlin/fi/espoo/evaka/varda/VardaChildren.kt
```diff
@@ -93,35 +94,44 @@ private fun insertPerson(vardaPersonResponse: VardaPersonResponse, personId: UUI
...
 private fun toVardaChildRequest(
-    getPersonUrl: (Long) -> String,
-    getOrganizerUrl: (Long) -> String
+    getPersonUrl: (Long) -> String
 ): (ResultSet, StatementContext) -> VardaChildRequest =
     { rs, _ ->
         VardaChildRequest(
             id = rs.getUUID("id"),
             personUrl = getPersonUrl(rs.getLong("varda_person_id")),
-            organizerUrl = getOrganizerUrl(rs.getLong("varda_organizer_id"))
```
teemu47, line 124: We can remove `getOrganizerUrl()` function from `VardaClient` after this

hruoho (approved):

PR #59, opened 2020-10-19
EVAKA-3942 Improve SSN sanitizing

teemu47:
#### Summary

- no sanitization with SSNs inside UUIDs
- add case insensitivity in order to catch typo'd SSNs

### DIFF
--- a/service-lib/src/main/resources/fi/espoo/voltti/logging/logback/default-appender-sanitized.xml
+++ b/service-lib/src/main/resources/fi/espoo/voltti/logging/logback/default-appender-sanitized.xml
@@ -38,8 +38,8 @@ Sanitized default appender logback configuration provided for import
           <pattern>
             {
             <!--Hide social security numbers in message and in stack trace-->
-            "message": "%replace(%message){'(^|[^-])(\\d{2})(\\d{2})(\\d{2})[Aa+-](\\d{3})[\\dA-z](.{0}$|[^-])', 'REDACTED-SSN'}",
-            "stackTrace": "%replace(%stacktrace{60,40,32000,rootFirst,inlineHash}){'(^|[^-])(\\d{2})(\\d{2})(\\d{2})[Aa+-](\\d{3})[\\dA-z](.{0}$|[^-])', 'REDACTED-SSN'}",
+            "message": "%replace(%message){'(?&lt;!-|[\\dA-z])(\\d{2})(\\d{2})(\\d{2})[Aa+-](\\d{3})[\\dA-z](?!-)', 'REDACTED-SSN'}",
+            "stackTrace": "%replace(%stacktrace{60,40,32000,rootFirst,inlineHash}){'(?&lt;!-|[\\dA-z])(\\d{2})(\\d{2})(\\d{2})[Aa+-](\\d{3})[\\dA-z](?!-)', 'REDACTED-SSN'}",
             "type": "app-misc",
             "appBuild": "${APP_BUILD}",
             "appCommit": "${APP_COMMIT}",
--- a/service-lib/src/main/resources/fi/espoo/voltti/logging/logback/default-appender-sanitized.xml
+++ b/service-lib/src/main/resources/fi/espoo/voltti/logging/logback/default-appender-sanitized.xml
@@ -38,8 +38,8 @@ Sanitized default appender logback configuration provided for import
           <pattern>
             {
             <!--Hide social security numbers in message and in stack trace-->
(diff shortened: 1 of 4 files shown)


service-lib/src/test/kotlin/fi/espoo/voltti/logging/logback/AppenderTest.kt
```diff
@@ -80,7 +80,7 @@ class AppenderTest {
             it.assertSanitized().containsExactly(event1.tuple())
             it.withLatestSanitized { actual ->
                 defaultErrorAssertions(actual, meta, exception)
-                assertThat(actual["stackTrace"] as String).doesNotContain(testSSN)
+                assertThat(actual["stackTrace"] as String).doesNotContain(UUIDWithSSN)
```
mikkopiu, line 83: Should assert that `UUIDWithSSN` _is_ in the log message.

mikkopiu (changes requested):

service-lib/src/test/kotlin/fi/espoo/voltti/logging/logback/AppenderTest.kt
teemu47: That should've asserted `message` instead of `stackTrace` since the UUID is there 😢 
Fixed the `stackTrace` assertions and added a couple for `message`

service-lib/src/main/resources/fi/espoo/voltti/logging/logback/default-appender-sanitized.xml
```diff
@@ -38,8 +38,8 @@ Sanitized default appender logback configuration provided for import
           <pattern>
             {
             <!--Hide social security numbers in message and in stack trace-->
-            "message": "%replace(%message){'(\\d{2})(\\d{2})(\\d{2})[A+-](\\d{3})[\\dA-Z]', 'REDACTED-SSN'}",
-            "stackTrace": "%replace(%stacktrace{60,40,32000,rootFirst,inlineHash}){'(\\d{2})(\\d{2})(\\d{2})[A+-](\\d{3})[\\dA-Z]', 'REDACTED-SSN'}",
+            "message": "%replace(%message){'(\\d{2})(\\d{2})(\\d{2})[Aa+-](\\d{3})[\\dA-z](.{0}$|[^-])', 'REDACTED-SSN'}",
+            "stackTrace": "%replace(%stacktrace{60,40,32000,rootFirst,inlineHash}){'(\\d{2})(\\d{2})(\\d{2})[Aa+-](\\d{3})[\\dA-z](.{0}$|[^-])', 'REDACTED-SSN'}",
```
mikkopiu, line 42: Don't yet have a good solution for this but the UUID "e1130765-2925-4549-b395-922222a8208a" would still be match (the last section).

teemu47: Good catch. We could do the same check for the start of the SSN as to the end of SSN: Not treating the SSN as SSN if there is a `-` at the start or at the end of the SSN.

In that case for example these string are not redacted: `-130732-555A`. `130732-555A-`. What do you think?

mikkopiu: I would be fine with that, there's no perfect solution for this. It could be that the syntax here doesn't support negative lookbehinds but lets hope not.

teemu47: Checking for anything before, even no character at all, but `-` character can be done with `(.{0}^|[^-])` regex so adding it to the beginning of the regex did the trick

mikkopiu: Is there any functional difference between `.{0}^` and `^`? Quick regex101 check would suggest no (in this case).

Also: not sure how the replacing function works but if it processes the fields as multiline strings, this would eat the newline from a previous line like:

```txt
Some text on first line
000000-0000A
```
would become: `Some text on first lineREDACTED-SSN` if I'm not totally mistaken (regex is hard).

mikkopiu: I unfortunately don't have the time this week but maybe a brainstorming session would be good; this feels like a brittle direction we are heading towards.

teemu47: Yeah you are right and I'm out of ideas. Let's brainstorm when you have time.

mikkopiu (approved):

PR #67, opened 2020-10-21
[EVAKA-3961] Fix Koski study right voiding

Gekkio:
#### Summary

Apparently voiding actually removes the study right, so the study right OID can't be used afterwards. On our side, the study right key (child+unit+type) will be reused if the study right is activated (= a new placement is created), which led to update requests to voided (= removed) study rights.

* make sure voiding is a one-shot operation (never void a voided study right)
* if a voided study right is reactivated on our side, make sure we send a new study right instead of trying to update an old one
* update mock Koski server to be more realistic in these scenarios


hruoho (approved):
:tada:

PR #70, opened 2020-10-22
Voucher value decisions part 2

jaahon:
#### Summary

* Add view for a single voucher value decision and its details.
* Add a way to send voucher value decisions.
* Add very basic e2e tests and an integration test that tests finance decisions for a family with children in service voucher and municipal units.


service/src/main/kotlin/fi/espoo/evaka/invoicing/client/S3DocumentClient.kt
```diff
@@ -58,13 +58,20 @@ class S3DocumentClient(
         return documentKey
     }
 
+    fun getPdf(bucket: String, key: String): ByteArray {
+        val s3Object = s3Client.getObject(bucket, key)
+        return IOUtils.toByteArray(s3Object.objectContent)
```
Gekkio, line 63: Could maybe use Kotlin builtin `s3Object.objectContent.readAllBytes()`? IOUtils seems like something more intended for AWS SDK internal use even though it's public

service/src/main/kotlin/fi/espoo/evaka/invoicing/client/S3DocumentClient.kt
```diff
@@ -58,13 +58,20 @@ class S3DocumentClient(
...
+
     fun getFeeDecisionPdf(documentKey: String): ByteArray {
         logger.debug { "Getting fee decision PDF for $documentKey." }
         val s3Object = s3Client.getObject(feeDecisionBucket, documentKey)
         return IOUtils.toByteArray(s3Object.objectContent)
     }
 
     fun getFeeDecisionDocumentUri(documentKey: String): String = "s3://$feeDecisionBucket/$documentKey"
+
+    fun getDocumentUri(bucket: String, key: String): String = "s3//$bucket/$key"
```
Gekkio, line 74: URI is missing `:` character

service/src/main/kotlin/fi/espoo/evaka/invoicing/controller/VoucherValueDecisionController.kt
```diff
@@ -52,6 +82,81 @@ class VoucherValueDecisionController(private val jdbi: Jdbi) {
...
+        }
+        return ResponseEntity(pdf, headers, HttpStatus.OK)
+    }
+
+    private fun sendVoucherValueDecisions(h: Handle, user: AuthenticatedUser, ids: List<UUID>) {
+        h.lockValueDecisions(ids)
+        val decisions = h.getValueDecisionsByIds(objectMapper, ids)
+        if (decisions.isEmpty()) return
+
+        val notDrafts = decisions.filterNot { it.status == VoucherValueDecisionStatus.DRAFT }
```
Gekkio, line 124: Personal preference: I think `any`/`all` are more readable than filtering the actual list elements and checking the resulting length:

```
if (decisions.any { it.status != VoucherValueDecisionStatus.DRAFT }) {
  throw BadRequest("Some voucher value decisions were not drafts")
}
```

error message matches the check exactly

service/src/main/kotlin/fi/espoo/evaka/invoicing/controller/VoucherValueDecisionController.kt
```diff
@@ -52,6 +82,81 @@ class VoucherValueDecisionController(private val jdbi: Jdbi) {
...
+        if (notDrafts.isNotEmpty()) {
+            throw BadRequest("Some voucher value decisions were not drafts")
+        }
+
+        val notYetValidDecisions = decisions.filter { it.validFrom > LocalDate.now() }
```
Gekkio, line 129: Personal preference: I think `any`/`all` are more readable than filtering the actual list elements and checking the resulting length:

```
if (decisions.any { it.validFrom > LocalDate::now() }) {
  throw BadRequest("Some of the voucher value decisions are not valid yet")
}
```

error message matches the check exactly

service/src/main/kotlin/fi/espoo/evaka/invoicing/data/VoucherValueDecisionQueries.kt
```diff
@@ -331,12 +355,136 @@ fun Handle.searchValueDecisions(
...
+}
+
+fun Handle.getVoucherValueDecisionDocumentKey(id: UUID): String? {
+    // language=sql
+    val sql = "SELECT document_key FROM voucher_value_decision WHERE id = :id"
+
+    return createQuery(sql)
+        .bind("id", id)
+        .mapTo<String>()
+        .firstOrNull()
```
Gekkio, line 453: Personal preference: if you're expecting only one result, prefer `singleOrNull`

Gekkio (approved):

frontend/packages/employee-frontend/src/components/voucher-value-decision/VoucherValueDecisionChildSection.tsx
```diff
@@ -0,0 +1,78 @@
...
+}
+
+const ChildSection = React.memo(function ChildSection({
+  child,
+  placement,
+  placementUnit
+}: Props) {
+  const { i18n } = useTranslation()
+
+  if (!child || !placement) {
```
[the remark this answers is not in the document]
jaahon: The check is unnecessary. I wonder how long the useless check has been in the original component 😄

service/src/main/kotlin/fi/espoo/evaka/invoicing/data/VoucherValueDecisionQueries.kt
jaahon: Can you elaborate why? Because `first` implies that there are more elements?

jaahon: Nevermind, I read the function's java doc.
> Returns single element, or `null` if the list is empty or has more than one element.

jaahon: Actually, it would be nicer if it threw an exception when the list has more than one element... 😞

Gekkio: TIL: `singleOrNull` returns null when there's more than one :cry: I thought it would throw an exception (like `single` does for more than one element cases)

Gekkio: The original point was to enforce the assumption "0-1 elements", but it doesn't matter whether you use `firstOrNull` or `singleOrNull` since neither does exactly the right thing

PR #73, opened 2020-10-23
Bump target Java version to 11

Gekkio:
#### Summary

This simply changes the generated bytecode version, but doesn't
otherwise affect the compilation. But this should provide a clear error
if somebody tries to compile with a too old Java version


mikkopiu (approved):
Verified locally with:
```txt
java -version
openjdk version "11.0.2" 2019-01-15
OpenJDK Runtime Environment 18.9 (build 11.0.2+9)
OpenJDK 64-Bit Server VM 18.9 (build 11.0.2+9, mixed mode)
```

PR #89, opened 2020-10-27
Hotfix: finance decision generation

jaahon:
#### Summary

Only fetch family relations that are relevant, otherwise `Period` initialization fails as the code expects that all of the relations are within the decision generation period.

### DIFF
--- a/service/src/main/kotlin/fi/espoo/evaka/pis/ParentshipQueries.kt
+++ b/service/src/main/kotlin/fi/espoo/evaka/pis/ParentshipQueries.kt
@@ -10,6 +10,7 @@ import fi.espoo.evaka.pis.service.PersonJSON
 import fi.espoo.evaka.shared.db.bindNullable
 import fi.espoo.evaka.shared.db.getUUID
 import fi.espoo.evaka.shared.domain.BadRequest
+import fi.espoo.evaka.shared.domain.Period
 import org.jdbi.v3.core.Handle
 import org.jdbi.v3.core.kotlin.mapTo
 import org.jdbi.v3.core.statement.StatementContext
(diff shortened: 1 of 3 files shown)


service/src/main/kotlin/fi/espoo/evaka/pis/ParentshipQueries.kt
```diff
@@ -56,12 +58,15 @@ fun Handle.getParentships(
         JOIN person head ON fc.head_of_child = head.id
         WHERE (:headOfChild::uuid IS NULL OR head_of_child = :headOfChild)
         AND (:child::uuid IS NULL OR child_id = :child)
+        AND daterange(fc.start_date, fc.end_date, '[]') && daterange(:from, :to, '[]')
```
Gekkio, line 61: Pro tip: you can bind period objects directly as parameters, so you can do `.bindNullable("period", period")` and use `&& :period` in the query

service/src/main/kotlin/fi/espoo/evaka/pis/PartnershipQueries.kt
```diff
@@ -76,11 +78,14 @@ fun Handle.getPartnersForPerson(personId: UUID, includeConflicts: Boolean): List
         JOIN fridge_partner partner ON fp.partnership_id = partner.partnership_id AND fp.indx != partner.indx
         JOIN person p ON partner.person_id = p.id
         WHERE fp.person_id = :personId
+        AND daterange(fp.start_date, fp.end_date, '[]') && daterange(:from, :to, '[]')
```
Gekkio, line 81: You could bind period directly here too

Gekkio (approved):

service/src/main/kotlin/fi/espoo/evaka/pis/ParentshipQueries.kt
jaahon: Yeah, but `Period` doesn't support null start dates and I'm a bit afraid to change it 😅

PR #106, opened 2020-10-30
Add suomi.fi staging certificate

Gekkio:
#### Summary

This only makes suomi.fi staging certificate known to the application, and infra changes are needed to make the application accept it.


mikkopiu (approved):
:+1:

PR #108, opened 2020-10-30
Upgrade apigw dependencies

Gekkio:
#### Summary

Remaining outdated dependencies:

- @types/node: intentionally locked to v12 (because we use Node 12)
- passport-saml: intentionally locked to 1.3.5 (because passport-saml 1.4.x has serious issues with typings right now)

**Needs manual testing in staging before production deployment** (double-check at least logging and suomi.fi / AD logins)


jaahon (approved):
I tested this in staging and both logins work 👌

PR #110, opened 2020-10-30
EVAKA-FIX Varda decision part time bug

teemu47:
#### Summary

Decision should be part time only if the care hours per week are under 25 hours - 25 hour care should yield full time decision


service/src/integrationTest/kotlin/fi/espoo/evaka/varda/VardaDecisionsIntegrationTest.kt
```diff
@@ -577,11 +577,11 @@ class VardaDecisionsIntegrationTest : FullApplicationTest() {
     }
 
     @Test
-    fun `a decision is full day with more than 25 hours`() {
+    fun `a decision is full day with 25 or more hours`() {
```
massproduction, line 580: Should we leave the 25.5 test case and add a new 25.0 test case?

teemu47: Yeah why not, tests are a little bit clearer that way. Fixed.

hruoho (approved):

PR #115, opened 2020-11-03
Evaka 3968 refactor mobile attendances

massproduction:
#### Summary
Refactored mobile attendances page. Missing search and transition animations. These will come in the next pull requests.


frontend/packages/employee-frontend/src/components/attendances/AttendancePageWrapper.tsx
```diff
@@ -0,0 +1,319 @@
...
+  const [freeText, setFreeText] = useState<string>('')
+  const [searchResults, setSearchResults] = useState<ChildInGroup[]>([])
+  const debouncedSearchTerms = useDebounce(freeText, 500)
+
+  useEffect(() => {
+    void getUnitData(id, LocalDate.today(), LocalDate.today()).then(setUnitData)
+    void getDaycare(id).then(setUnit)
+    groupid === 'all'
+      ? void getDaycareAttendances(id).then(setGoupAttendances)
+      : void getChildrenInGroup(groupid).then(setGoupAttendances)
```
Joosakur, line 115: I wonder if we should only use `getDaycareAttendances` and ditch the other one. Fetching only a single group sometimes may have slightly better performance, but having only one DB-query and endpoint would simplify the code enough to possibly be a worthy tradeoff.

Joosakur (approved):

PR #117, opened 2020-11-03
Use unit operational days in invoice generation

jaahon:
#### Summary

Instead of just assuming that invoiced units are operational from monday until friday unless they are a round the clock unit, use the new `operation_days` field in `daycare` table.


reynders (approved):

jaahon:
`espoo-dev-data.sql` is broken btw

PR #132, opened 2020-11-05
Remove jdbc template

jaahon:
#### Summary

Remove jdbc template and replace it with jdbi. I try to pass a handle instead of using `withSpringHandle` as much as feels sensible.


service/src/main/kotlin/fi/espoo/evaka/daycare/service/CaretakerService.kt
```diff
@@ -30,25 +26,24 @@ class CaretakerService(
...
+            .toList()
     }
 
-    @Transactional
-    fun insert(groupId: UUID, startDate: LocalDate, endDate: LocalDate?, amount: Double) {
+    fun insert(h: Handle, groupId: UUID, startDate: LocalDate, endDate: LocalDate?, amount: Double) {
         if (endDate != null && endDate.isBefore(startDate)) throw BadRequest("End date cannot be before start")
 
-        endPreviousRow(groupId, startDate)
+        endPreviousRow(h, groupId, startDate)
```
reynders, line 46: Should this endPreviousRow and the later update be inside h.transaction to ensure there is a transaction when touching these two?

reynders (approved):

service/src/main/kotlin/fi/espoo/evaka/daycare/service/CaretakerService.kt
reynders: Discussed this with the team, this is ok

PR #135, opened 2020-11-06
Fix invalid async job runner usage

Gekkio:
#### Summary

* `scheduleImmediateRun` must be called after the transaction planning the jobs has been committed
* `runAfterCommit` only works inside Spring transactions, and is no-op otherwise

### DIFF
--- a/service/src/main/kotlin/fi/espoo/evaka/application/DecisionMessageProcessor.kt
+++ b/service/src/main/kotlin/fi/espoo/evaka/application/DecisionMessageProcessor.kt
@@ -9,7 +9,6 @@ import fi.espoo.evaka.shared.async.AsyncJobRunner
 import fi.espoo.evaka.shared.async.NotifyDecisionCreated
 import fi.espoo.evaka.shared.async.SendDecision
 import fi.espoo.evaka.shared.config.Roles
-import fi.espoo.evaka.shared.db.runAfterCommit
 import mu.KotlinLogging
 import org.jdbi.v3.core.Handle
 import org.springframework.stereotype.Component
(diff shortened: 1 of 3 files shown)


service/src/main/kotlin/fi/espoo/evaka/pis/controllers/ParentshipController.kt
```diff
@@ -43,6 +48,7 @@ class ParentshipController(private val parentshipService: ParentshipService, pri
         with(body) {
             return jdbi.transaction {
                 parentshipService.createParentship(it, childId, headOfChildId, startDate, endDate)
+                    .also { asyncJobRunner.scheduleImmediateRun() }
```
jaahon, line 51: Isn't this `also` in the wrong place? Now it's called inside `transaction`

service/src/main/kotlin/fi/espoo/evaka/decision/DecisionService.kt
```diff
@@ -60,13 +59,12 @@ class DecisionService(
...
-            withSpringHandle(dataSource) {
-                asyncJobRunner.plan(it, decisionIds.map { NotifyDecisionCreated(it, user, sendAsMessage) })
+            withSpringHandle(dataSource) { h ->
+                asyncJobRunner.plan(h, decisionIds.map { NotifyDecisionCreated(it, user, sendAsMessage) })
             }
-            runAfterCommit { asyncJobRunner.scheduleImmediateRun() }
 
             decisionIds
-        }
+        }.also { asyncJobRunner.scheduleImmediateRun() }
```
jaahon, line 67: Does this guarantee that the transaction has been commited?

Gekkio: :facepalm: need more :coffee:

service/src/main/kotlin/fi/espoo/evaka/pis/controllers/ParentshipController.kt
Gekkio: :facepalm: need more :coffee:

jaahon (approved):

PR #144, opened 2020-11-10
Evaka 4002 improve dvv logging

reynders:
#### Checklist for pull request creator

- [ ] The code is consistent with the existing code base
- [ ] Tests have been written for the change (e2e, integration tests)
- [ ] The change has been tested locally
- [ ] The change conforms to the UX specifications
- [ ] The code is self-documenting or has been documented sufficiently, e.g., in the README
- [ ] The branch has been rebased against master before the PR was created

#### Checklist for pull request reviewer (copy to review text box)

```
- [ ] All changes in all changed files have been reviewed
- [ ] The code is consistent with the existing code base
- [ ] Tests have been written for the change (e2e, integration tests)
- [ ] The change has been tested locally
- [ ] The change conforms to the UX specifications
- [ ] The code is self-documenting or has been documented sufficiently, e.g., in the README
- [ ] The PR branch has been rebased against master and force pushed if necessary before merging
```

### DIFF
--- a/service/src/main/kotlin/fi/espoo/evaka/dvv/DvvModificationsBatchRefreshService.kt
+++ b/service/src/main/kotlin/fi/espoo/evaka/dvv/DvvModificationsBatchRefreshService.kt
@@ -30,8 +30,9 @@ class DvvModificationsBatchRefreshService(
     }
 
     fun doDvvModificationsRefresh(msg: DvvModificationsRefresh) {
-        logger.info("DvvModificationsRefresh: updating ${msg.ssns.size} ssns")
-        dvvModificationsService.updatePersonsFromDvv(msg.ssns)
+        logger.info("DvvModificationsRefresh: starting to process ${msg.ssns.size} ssns")
+        val modificationCount = dvvModificationsService.updatePersonsFromDvv(msg.ssns)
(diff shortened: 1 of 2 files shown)


mikkopiu (commented):
Is the value in having these logs _temporary_ or permanent?

If it's only temporary (i.e. to debug for a while as this is a new feature), you could also just change the logging level of this particular logger, similar to:

```
service/src/main/resources/application-local.properties:logging.level.fi.espoo.evaka.invoicing.integration.IntegrationClient=DEBUG
```

reynders:
> Is the value in having these logs _temporary_ or permanent?
> 
> If it's only temporary (i.e. to debug for a while as this is a new feature), you could also just change the logging level of this particular logger, similar to:
> 
> ```
> service/src/main/resources/application-local.properties:logging.level.fi.espoo.evaka.invoicing.integration.IntegrationClient=DEBUG
> ```

Yes, I thought of this, but current assumption is that the logging level must be info so that we have trace where a change came from. At some point we hopefully get a separate person change history log, but until then this is the way

mikkopiu (approved):
:white_check_mark: Could maybe be changed to proper audit logging later on if that's the real value here.

PR #148, opened 2020-11-10
Fix family initializer transactions

Gekkio:
#### Summary

Remove nested independent transactions


jaahon (approved):
This job will now create family relationships even if the whole job does not succeed, but I guess it's not a problem.

Gekkio:
Yeah, we should look into using savepoints if possible, but this is mostly a quickfix that removes some deadlock scenarios and tries to retain old semantics.

PR #160, opened 2020-11-12
Experimental database API

Gekkio:
#### Summary

Make transaction boundaries explicit -> functions either require a transaction to be active (`tx: Database.Transaction` / `tx: Database.Read`), or require a transaction to *not* be active (`db: Database.Connection` / `db: Database`).

* add "subtransaction" API using savepoints for scenarios where we need to try/catch an operation inside a transaction
* use the new API in FamilIyInitializerService / PersonService where we have the most complicated code involving transactions

### DIFF
--- a/service/src/main/kotlin/fi/espoo/evaka/shared/db/Database.kt
+++ b/service/src/main/kotlin/fi/espoo/evaka/shared/db/Database.kt
@@ -15,7 +15,7 @@ import org.jdbi.v3.core.statement.Update
 //
 //     fun doStuff(db: Database):
 //         To call this function, you need to have a database reference *without* an active connection or transaction.
-//         The can connect/disconnect from the database 0 to N times, and do whatever it wants using the connection(s).
+//         The function can connect/disconnect from the database 0 to N times, and do whatever it wants using the connection(s).
 //     fun doStuff(db: Database.Connection)
 //         To call this function, you need to have an active database connection *without* an active transaction.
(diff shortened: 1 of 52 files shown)


service/src/main/kotlin/fi/espoo/evaka/shared/db/Database.kt
```diff
@@ -0,0 +1,190 @@
...
+     */
+    class Transaction internal constructor(handle: Handle) : Database.Read(handle) {
+        private var savepointId: Long = 0
+
+        fun nextSavepoint(): String = "savepoint-${savepointId++}"
+        fun createUpdate(@Language("sql") sql: String): Update = handle.createUpdate(sql)
+        fun prepareBatch(@Language("sql") sql: String): PreparedBatch = handle.prepareBatch(sql)
+        fun execute(@Language("sql") sql: String, vararg args: Any): Int = handle.execute(sql, *args)
+
+        inline fun <reified T> subTransaction(crossinline f: () -> T): T {
```
jaahon, line 156: Can you explain why is this function `inline` by the way? What's the practical reason? I didn't really understand Kotlin's documentation about `crossinline` parameters.

jaahon: Nevermind, I got it now.

jaahon (approved):
Just merge it 😎

PR #167, opened 2020-11-16
EVAKA-3977: Make backup care placements use the regular group placement form

mapehe:
#### Summary

Filter missing group placements by time range. At the moment such filtering is done for missing group placements, but not for backup cares. This leads to an inconsistent UX due to [how](https://github.com/espoon-voltti/evaka/blob/a28f087bb88136e03ac392817cfb3223ea5c9e0c/frontend/packages/employee-frontend/src/components/unit/tab-groups/MissingGroupPlacements.tsx#L134-L136) missing group placements are mapped to backup cares in the frontend.

### DIFF
--- a/frontend/e2e-test/test/e2e/specs/5_employee/backup-care.spec.ts
+++ b/frontend/e2e-test/test/e2e/specs/5_employee/backup-care.spec.ts
@@ -16,7 +16,7 @@ import UnitPage, {
   missingPlacementElement
 } from '../../pages/employee/units/unit-page'
 import { logConsoleMessages } from '../../utils/fixture'
-import BackupCareGroupModal from '../../pages/employee/units/backup-care-group-modal'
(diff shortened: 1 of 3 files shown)


mapehe:
@Joosakur raised some concerns about this. Maybe marking this as ready for review is a good way to discuss?

mapehe:
Perhaps we could get rid of `BackupCareGroupModal` completely. However, [this](https://github.com/espoon-voltti/evaka/blob/a28f087bb88136e03ac392817cfb3223ea5c9e0c/frontend/packages/employee-frontend/src/components/unit/tab-groups/Groups.tsx#L223-L227) usage requires some refactoring. I don't have enough domain knowledge to make sure that the data types are refactoring correctly.

Joosakur:
Code looks good to me now! I'll still need to do some manual testing to ensure this change does not cause regression bugs.

Joosakur (approved):
Tested locally, could not find any issues, great work!

PR #169, opened 2020-11-16
Use new database API - part 2

Gekkio:
#### Summary

* use new Database API in controllers, services, and async jobs. Most database queries still use old Handle API
* pass Database.Connection to tests and controllers. This was the original idea, but previously I wasn't sure if connection cleanup was possible to implement with simple Spring MVC integration

### DIFF
--- a/service/src/main/kotlin/fi/espoo/evaka/placement/PlacementService.kt
+++ b/service/src/main/kotlin/fi/espoo/evaka/placement/PlacementService.kt
@@ -8,11 +8,11 @@ import fi.espoo.evaka.application.utils.exhaust
 import fi.espoo.evaka.daycare.dao.mapPSQLException
 import fi.espoo.evaka.daycare.getDaycareGroup
 import fi.espoo.evaka.shared.auth.AclAuthorization
+import fi.espoo.evaka.shared.db.Database
 import fi.espoo.evaka.shared.domain.BadRequest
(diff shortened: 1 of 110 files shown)


service/src/main/kotlin/fi/espoo/evaka/placement/PlacementService.kt
```diff
@@ -155,54 +145,53 @@ private fun clearOldPlacements(h: Handle, childId: UUID, from: LocalDate, to: Lo
                 // old placement is within new placement (or identical)
                 !old.startDate.isBefore(from) && !old.endDate.isAfter(to) -> {
                     checkAclAuth(aclAuth, old)
-                    h.cancelPlacement(old.id)
+                    val (_, _, _) = handle.cancelPlacement(old.id)
```
Joosakur, line 148: is `val (_, _, _) = ` necessary?

Joosakur (approved):

service/src/main/kotlin/fi/espoo/evaka/placement/PlacementService.kt
Gekkio: It's not the only option, but all branches of `when` need to return the same type `Unit`.
`val _ = ...` doesn't work, because Kotlin has a bit strange rules about where underscores can be used.

PR #189, opened 2020-11-20
Add the word Koski to scheduling log message

Gekkio:
#### Summary

This makes sure it's included if logs are searched with the word "Koski"


reynders (approved):

mikkopiu:
Alternatively, you could use the new and improved meta functionality of the loggers to do something like: `meta {"className":"KoskiSomething"}` but obviously that's a bigger change

Gekkio:
Having the logger name in meta would be very nice, and we should probably try to add it automatically to meta everywhere

PR #201, opened 2020-11-24
EVAKA-3976: Create Special Education Teacher ACL role

mapehe:
#### Summary

Create Special Education Teacher ACL role.

Some stuff is still unclear (_EDIT: not unclear anymore_):
* Should SET be able to get and create aplication notes. _EDIT: Yes_
* Should SET only be able to see applications that include assistance requirements. _EDIT: Yes_


service/src/main/kotlin/fi/espoo/evaka/application/ApplicationControllerV2.kt
```diff
@@ -208,7 +213,9 @@ class ApplicationControllerV2(
                 periodStart = periodStart,
                 periodEnd = periodEnd,
                 searchTerms = searchTerms ?: "",
-                transferApplications = transferApplications ?: TransferApplicationFilter.ALL
+                transferApplications = transferApplications ?: TransferApplicationFilter.ALL,
+                authorizedUnitFilterList = authorizedUnitFilterList,
```
Gekkio, line 217: I think it would be better to pass an `AclAuthorization` object here to the function (obtained with `acl.getAuthorizedUnits(user)`, because it *attaches meaning* to an optional list/set of IDs. Just like we prefer passing `AuthenticatedUser` instead of just `UUID` if a function needs a user ID.

Then, in the query binding you should be able to do `aclAuthorization.ids?.toTypedArray()`, and it handles all the details automatically

service/src/main/kotlin/fi/espoo/evaka/shared/auth/Acl.kt
```diff
@@ -53,7 +53,18 @@ WHERE employee_id = :userId AND daycare_id = :daycareId
     )
 
     fun getRolesForApplication(user: AuthenticatedUser, applicationId: UUID): AclAppliedRoles = AclAppliedRoles(
-        (user.roles - UserRole.ACL_ROLES) + Database(jdbi).read {
+        (user.roles - UserRole.ACL_ROLES) + Database(jdbi).read { it ->
+            val assistanceNeeded = it.createQuery(
+                // language=SQL
+                """
+SELECT (f.document -> 'careDetails' ->> 'assistanceNeeded')
+FROM application a inner join application_form f on a.id = f.application_id and f.latest = true
```
Gekkio, line 61: Since we already have an application ID, maybe the `application` table can be skipped?

```
...
FROM application_form f
WHERE f.application_id = :applicationId AND f.latest IS TRUE
```

service/src/main/kotlin/fi/espoo/evaka/application/ApplicationControllerV2.kt
```diff
@@ -189,7 +189,7 @@ class ApplicationControllerV2(
 
         val userHasLimitedPrivileges = !user.hasOneOfRoles(Roles.ADMIN, Roles.FINANCE_ADMIN, Roles.SERVICE_WORKER)
         val authorizedUnitFilterList = when {
-            userHasLimitedPrivileges -> acl.getAuthorizedUnits(user).ids?.toList() ?: listOf()
+            userHasLimitedPrivileges -> acl.getAuthorizedUnits(user)
```
Gekkio, line 192: You can even remove this `when` block completely :wink: 
`getAuthorizedUnits` returns either:

* `AclAuthorization.All`, which has null `.ids` property
* `AclAuthorization.Subset`, which has a non-null `.ids` property

So you can just pass a non-nullable `AclAuthorization`, which represents the idea of either allowing all or a subset of something

mapehe: Actually thought about this, but only later realised I also need to update the [null check](https://github.com/espoon-voltti/evaka/blob/ae73918c9fc3aaa8ccc4cda989943234c37ad9e5/service/src/main/kotlin/fi/espoo/evaka/application/ApplicationQueries.kt#L185) accordingly if the argument is non-nullable. 😄

reynders (approved):

mapehe:
Realized last minute that child age language report needs a filter too. Fixed in 7fbc7ce.

PR #214, opened 2020-11-26
EVAKA-4006: mobile devices backend

Joosakur:
#### Checklist for pull request creator

- [x] The code is consistent with the existing code base
- [x] Tests have been written for the change (e2e, integration tests)
- [ ] The change has been tested locally
- [x] The change conforms to the UX specifications
- [x] The code is self-documenting or has been documented sufficiently, e.g., in the README
- [x] The branch has been rebased against master before the PR was created

#### Checklist for pull request reviewer (copy to review text box)

```
- [ ] All changes in all changed files have been reviewed
- [ ] The code is consistent with the existing code base
- [ ] Tests have been written for the change (e2e, integration tests)
- [ ] The change has been tested locally
- [ ] The change conforms to the UX specifications
- [ ] The code is self-documenting or has been documented sufficiently, e.g., in the README
- [ ] The PR branch has been rebased against master and force pushed if necessary before merging
```


service/src/main/kotlin/fi/espoo/evaka/pairing/MobileDeviceQueries.kt
```diff
@@ -0,0 +1,68 @@
...
+import org.jdbi.v3.core.kotlin.mapTo
+import java.util.UUID
+
+fun insertDeviceData(tx: Database.Transaction, unitId: UUID): UUID {
+    val defaultName = "Nimeämätön laite"
+
+    // language=sql
+    val employeeInsert =
+        """
+            INSERT INTO employee (first_name, last_name, email, aad_object_id)
```
Gekkio, line 14: Personal preference: I'd combine these queries into one. Something like this:

```
INSERT INTO mobile_device (id, unit_id, name)
SELECT id, :unit_id, :name
FROM (
  INSERT INTO employee (...)
  VALUES (...)
  RETURNING id
) new_employee
```

service/src/main/kotlin/fi/espoo/evaka/pairing/MobileDeviceQueries.kt
```diff
@@ -0,0 +1,68 @@
+package fi.espoo.evaka.pairing
+
+import fi.espoo.evaka.shared.db.Database
+import fi.espoo.evaka.shared.domain.NotFound
+import org.jdbi.v3.core.kotlin.mapTo
+import java.util.UUID
+
+fun insertDeviceData(tx: Database.Transaction, unitId: UUID): UUID {
```
Gekkio, line 8: Didn't we decide to use the form `fun Database.Transaction.insertDeviceData(...)` at some point? Once again we're adding new code with two different styles :slightly_frowning_face:

service/src/main/kotlin/fi/espoo/evaka/pairing/MobileDeviceQueries.kt
```diff
@@ -0,0 +1,68 @@
...
+        .bind("unitId", unitId)
+        .bind("name", defaultName)
+        .execute()
+
+    return id
+}
+
+fun listDevices(tx: Database.Read, unitId: UUID): List<MobileDevice> {
+    // language=sql
+    val sql = "SELECT * FROM mobile_device WHERE unit_id = :unitId AND deleted = false ORDER BY created"
```
Gekkio, line 41: I'd much rather order in the frontend closer to where the data is used. A good convention is to use `ORDER BY` only when it's required for correctness, explicitly required somehow in the function signature, or if there's an obvious natural order which should always be used. In this case none of these guidelines apply, because ordering is just a visual decision.

service/src/main/kotlin/fi/espoo/evaka/pairing/MobileDeviceQueries.kt
```diff
@@ -0,0 +1,68 @@
...
+        .list()
+}
+
+fun renameDevice(tx: Database.Transaction, id: UUID, name: String) {
+    // language=sql
+    val deviceUpdate = "UPDATE mobile_device SET name = :name WHERE id = :id"
+    tx.createUpdate(deviceUpdate)
+        .bind("id", id)
+        .bind("name", name)
+        .execute().takeIf { it > 0 } ?: throw NotFound("Device $id not found")
```
Gekkio, line 54: This seems an overly complicated way of doing the check. We're conditionally converting an integer to null and throwing away the integer, so it's equivalent to:

```kotlin
val rows = foo.execute()
val rowsOrNull = if (rows > 0) rows else null
if (rowsOrNull == null) throw NotFound(...)
```

How about `.execute().let { if (it == 0) throw NotFound(..) }` if you want to do it without a temporary variable like `val rows = ...`?

service/src/main/kotlin/fi/espoo/evaka/pairing/MobileDevicesController.kt
```diff
@@ -0,0 +1,63 @@
...
+        @RequestBody body: RenameRequest
+    ): ResponseEntity<Unit> {
+        Audit.MobileDevicesRename.log(targetId = id)
+        acl.getRolesForMobileDevice(user, id).requireOneOfRoles(UserRole.ADMIN, UserRole.UNIT_SUPERVISOR)
+
+        db.transaction { tx -> renameDevice(tx, id, body.name) }
+        return ResponseEntity.noContent().build()
+    }
+
+    @DeleteMapping("/apigw/mobile-devices/{id}")
```
Gekkio, line 52: This path should probably be `/mobile-devices/{id}`

service/src/main/resources/db/migration/V23__mobile_device.sql
```diff
@@ -0,0 +1,10 @@
+CREATE TABLE mobile_device(
+    id uuid PRIMARY KEY REFERENCES employee(id),
+    created timestamp with time zone DEFAULT now() NOT NULL,
+    updated timestamp with time zone DEFAULT now() NOT NULL,
+    unit_id uuid NOT NULL REFERENCES daycare(id),
+    name text NOT NULL,
+    deleted boolean NOT NULL DEFAULT false
+);
+
```
Gekkio, line 9: Let's also add

`CREATE INDEX idx$mobile_device_unit ON mobile_device (unit_id) WHERE deleted IS FALSE`

as a sensible default

service/src/main/kotlin/fi/espoo/evaka/pairing/MobileDevicesController.kt
```diff
@@ -0,0 +1,63 @@
...
+    fun deleteMobileDevice(
+        db: Database.Connection,
+        user: AuthenticatedUser,
+        @PathVariable id: UUID
+    ): ResponseEntity<Unit> {
+        Audit.MobileDevicesDelete.log(targetId = id)
+
```
Gekkio, line 59: Needs ACL check

service/src/main/kotlin/fi/espoo/evaka/pairing/PairingsController.kt
```diff
@@ -122,8 +126,6 @@ class PairingsController(
 
         db.transaction { tx ->
             validatePairing(tx, id, body.challengeKey, body.responseKey)
-
-            // todo: insert into mobile_device and employee ?
```
Gekkio, line 126: One thing we're still missing is a link between a pairing and the mobile device created from it. Once the pairing process is done, the mobile-side frontend has only a pairing ID + the keys that are sent to the apigw which needs the mobile device ID so it can be stored to the session. I think this validation endpoint could return the mobile device ID, but we also need to have a link in the database to be able to find it. Maybe a nullable `mobile_device_id` to the pairing table?

service/src/main/kotlin/fi/espoo/evaka/pairing/MobileDeviceQueries.kt
Joosakur: couldn't get that syntax to work so went with
```
WITH new_employee AS (
    INSERT INTO employee (first_name, last_name, email, aad_object_id)
    VALUES (:name, (SELECT name FROM daycare WHERE id = :unitId), null, null)
    RETURNING id
)
INSERT INTO mobile_device (id, unit_id, name) 
SELECT new_employee.id, :unitId, :name
FROM new_employee
RETURNING id
```

service/src/main/kotlin/fi/espoo/evaka/pairing/MobileDeviceQueries.kt
Joosakur: changed these too

Gekkio (approved):

PR #217, opened 2020-11-26
EVAKA 3976: ACL permission fixes plus refactoring

mapehe:
#### Summary

Fix some permission issues. Unify code style a bit.

The diff is pretty messy: I did my best to help the reviewers by dividing this to neat commits. I suggest reviewing this on the commit level.


Gekkio:
I've got one potentially annoying comment :sweat_smile: 

Instead of importing from `Roles`, I think we should import from `UserRole`:

```
import fi.espoo.evaka.shared.config.Roles.ADMIN
```

vs

```
import fi.espoo.evaka.shared.auth.UserRole.ADMIN
```

The reason is that `Roles` basically just re-exports values from `UserRole`, which is already an enum so `Roles` adds no extra value at all. It only exists for legacy reasons, because all the roles used to be strings so `Roles` had all the constants in one place.

mapehe:
I planned to refactor the role imports a bit, but I eventually decided to do that in a separate pull request.

service/src/main/kotlin/fi/espoo/evaka/shared/auth/Acl.kt
```diff
@@ -140,6 +140,70 @@ WHERE employee_id = :userId AND decision.id = :decisionId
...
+                // language=SQL
+                """
+SELECT child_id
+FROM assistance_need
+WHERE id = :assistanceNeedId
+                """.trimIndent()
+            ).bind("assistanceNeedId", assistanceNeedId)
+                .mapTo<UUID>()
+                .map { childId -> getRolesForChild(user, childId).roles }
+                .fold(setOf<UserRole>()) { acc, rolesSet -> acc.intersect(rolesSet) }
```
Gekkio, line 155: Doesn't this always result in an empty set?
We start with an empty set and use intersection -> nothing ever intersects with an empty set so the result will be empty

mapehe: Great catch... We could use `union` here instead.

I don't really like this bit though, since we'd normally expect the set to have exactly one element and the fold feels a bit awkward. Would there be some nice way to express this?

Gekkio: I can think of two options:

a)
since the query only returns only one row (or null), you can use something like this:

```
.mapTo<UUID>()
.singleOrNull()
?.let { childId -> getRolesForChild(user, childId).roles }
?: emptySet()
```

b)
Instead of reusing the `getRolesForChild` function, create a reusable database view `child_acl` (employee_id, child_id, role) using a repeatable migration (`R__foo.sql`) and join to it in the query. Then you can have the main query return just `UserRole` instead of having to go through intermediate data

Gekkio: Personally I prefer b), because `getRolesForChild` is not really meant to be used by other functions like that. It does two "top-level function" things which are unnecessary when used as a part of another top-level function: it creates a new temporary database connection internally (`Database(jdbi)`) and does the all vs subset check.

mapehe: Thanks for the comments, I'll implement option b.

service/src/main/kotlin/fi/espoo/evaka/shared/auth/Acl.kt
```diff
@@ -114,13 +114,9 @@ WHERE employee_id = :userId AND placement.id = :placementId
...
-SELECT acl.role
-FROM person ch
-LEFT JOIN placement pl ON pl.child_id = ch.id AND pl.end_date > current_date - INTERVAL '1 month'
-LEFT JOIN application a ON a.child_id = ch.id AND a.status = ANY ('{SENT,WAITING_PLACEMENT,WAITING_CONFIRMATION,WAITING_DECISION,WAITING_MAILING,WAITING_UNIT_CONFIRMATION, ACTIVE}'::application_status_type[])
-LEFT JOIN placement_plan pp ON pp.application_id = a.id
-JOIN daycare_acl acl ON acl.daycare_id = pl.unit_id OR acl.daycare_id = pp.unit_id
-WHERE employee_id = :userId AND ch.id = :childId
+SELECT role
+FROM child_acl_view
+WHERE employee_id = :userId AND child_id = :childId
```
mapehe, line 119: Since we introduce `R__child_acl_view.sql`, I thought we might want to use that here too.

Gekkio (commented):
Looks good! :+1: 
I realized we need to implement one extra feature: child ACL in backup care cases. What I mean is not access to the `backup_care` row, but to a child who has been placed in backup care. This hasn't been implemented yet because the checks were imprecise so it didn't matter earlier...but now if we don't implement this, unit supervisors / staff lose access to backup care children in their unit.

Basically, in `child_acl_view` we need to also consider the *target unit*  in backup care. Backup care is kind of a "secondary placement" where a child is temporarily in another unit. In that case, there's no application/placement/placement_plan for the target unit, but just a backup_care row with `unit_id` = target unit. Now that we have a view for child ACL, it should be fairly straightforward to use UNION to handle this case too. Backup care should probably have the 1 month time extension just like normal placements have in the view.

So, if a child has a placement in unit A and is placed to temporary backup care to unit B, there will be:

* `placement` row with unit_id=A. ACL example: `UNIT_SUPERVISOR` in unit A has access to the child based on placement end_date
* `backup_care` row with unit_id=B. ACL example: `UNIT_SUPERVISOR` in unit B has access to the child based on backup_care end_date

mapehe:
Thanks again for the feedback @Gekkio!

Backup care adds some extra hassle: I was in touch with the domain expert and they said that, in case of a backup care, `SPECIAL_EDUCATION_TEACHER` should have `STAFF` level rights.

I think this could easily be overlooked in the future, since backup care seems a bit like an edge case. In d9e1c96 I'm proposing that, when granting rights based on backup care, we default to `STAFF` and manually whitelist the cases in which we want to grant elevated rights.

Gekkio (approved):
:+1:

PR #222, opened 2020-11-30
Refactor roles imports

mapehe:
#### Summary

* Replace `config.Roles` by `auth.UserRole` throughout.
* Remove the `config.Roles` enum.
* Instead of importing a role directly (e.g. `ADMIN`), import `auth.UserRole` and refer to e.g. `UserRole.ADMIN` in code throughout.


jaahon (approved):
Good job 👍 

> Instead of importing a role directly (e.g. ADMIN), import auth.UserRole and refer to e.g. UserRole.ADMIN in code throughout.

Although, I wonder why you did this. My first thought would be to do the opposite since it's usually very clear from the context which enum the value is a member of. I'm just curious, don't expect you to change anything 😄

mapehe:
Thanks for the feedback @jaahon.

> Although, I wonder why you did this. My first thought would be to do the opposite since it's usually very clear from the context which enum the value is a member of. I'm just curious, don't expect you to change anything 😄

The code wasn't consistent in using either, say, `ADMIN` or `UserRole.ADMIN` and I went back and forth between the two before deciding to go with option 2.

To be honest, a major reason is it's what IntelliJ defaults to in my setup :D I thought that other people probably have the same thing and consistency might be preserved better in case people use IDE quick fixes. I guess it's also kind of aesthetic that, even though the code becomes a bit more verbose, the imports look the same in all files.

PR #230, opened 2020-12-01
Evaka 4006 mobile device pairing frontend

massproduction:
#### Summary
Frontend for pairing mobile devices.


frontend/packages/employee-frontend/src/api/unit.ts
```diff
@@ -456,6 +456,149 @@ export async function removeDaycareAclStaff(
...
+  | 'PAIRED'
+
+export interface PairingResponse {
+  id: UUID
+  unitId: UUID
+  challengeKey: string
+  responseKey?: string
+  expires: Date
+  status: PairingStatus
+  mobileDeviceId?: UUID
```
Joosakur, line 481: because of jackson serialization more accurate would be
```
  responseKey: string | null
  mobileDeviceId: UUID | null
```

Joosakur (approved):

PR #235, opened 2020-12-02
EVAKA-3464 proxy max body size

jaahon:
#### Summary

Unset `client_max_body_size` in proxy for `*/attachments` paths.

### DIFF
--- a/proxy/nginx.conf.template
+++ b/proxy/nginx.conf.template
@@ -228,11 +228,21 @@ server {
   location /api/application {
     include     proxy_params;
     proxy_pass  $enduserUrl;
+
+    location /api/application/attachments {
+      client_max_body_size 100m;
+      proxy_pass  $enduserUrl;
+    }
   }
 
   location /api/internal {
     include     proxy_params;
     proxy_pass  $internalUrl;
+
+    location /api/internal/attachments {
+      client_max_body_size 100m;
+      proxy_pass  $internalUrl;
+    }
   }
 
   # Direct all CPS reports to internal endpoint /api/csp/csp-report
--- a/compose/proxy/nginx.conf
+++ b/compose/proxy/nginx.conf
@@ -92,6 +92,11 @@ server {
     proxy_no_cache 1;
 
     proxy_pass  $enduserUrl;
+
+    location /api/application/attachments {
+      client_max_body_size 100m;
+      proxy_pass  $enduserUrl;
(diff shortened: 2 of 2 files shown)


proxy/nginx.conf.template
```diff
@@ -228,11 +228,21 @@ server {
   location /api/application {
     include     proxy_params;
     proxy_pass  $enduserUrl;
+
+    location /api/application/attachments {
+      client_max_body_size 0;
```
mikkopiu, line 233: :grimacing: I'd much rather prefer _some_ reasonable limit instead of unlimited. Most optimally, I'd have it as a configurable parameter but can cope with something reasonable.

mikkopiu (changes requested):

proxy/nginx.conf.template
jaahon: I guess realistically evaka will never accept files over 100MB, but I just don't want this to be configured twice. How about the move the 10MB limit to proxy and remove it from service?

mikkopiu: I suppose we inevitably have to have some implementation details on the proxy level (i.e. the path `/api/application/attachments`), so I can live with that.

The only gripe I have is that IMO the service should still block unreasonably sized uploads even if it wasn't behind this particular proxy (and architecture-wise it shouldn't even know about its existence). Personally I'd just do the double configuration with a high limit at the proxy level and a "use case" limit at the service level but as said, I can live with it just being a proxy configuration. Practically the chance of us replacing the proxy with something else is rather low. Though, the other municipalities might want to use their own :thinking:.

proxy/nginx.conf.template
```diff
@@ -228,11 +228,21 @@ server {
...
+      proxy_pass  $enduserUrl;
+    }
   }
 
   location /api/internal {
     include     proxy_params;
     proxy_pass  $internalUrl;
+
+    location /api/internal/attachments {
```
mikkopiu, line 242: Q: do the employees upload files too?

jaahon: Not at the moment, no.

proxy/nginx.conf.template
jaahon: Alright, I'll just a large limit in the proxy and keep the service limit intact.

mikkopiu (approved):

PR #238, opened 2020-12-02
EVAKA-3859: Missing income days warning

mapehe:
#### Summary

Warn employee when changes on incomes leads to periods of unreported income.

Screenshot for reference.

image

### DIFF
--- a/frontend/packages/employee-frontend/src/assets/i18n/fi.ts
+++ b/frontend/packages/employee-frontend/src/assets/i18n/fi.ts
@@ -27,6 +27,7 @@ export const fi = {
   common: {
     yes: 'Kyllä',
     no: 'Ei',
+    and: 'Ja',
     loadingFailed: 'Tietojen haku epäonnistui',
     edit: 'Muokkaa',
     add: 'Lisää',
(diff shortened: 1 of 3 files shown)


frontend/packages/employee-frontend/src/components/person-profile/PersonIncome.tsx
```diff
@@ -54,6 +58,25 @@ const PersonIncome = React.memo(function PersonIncome({ id, open }: Props) {
   }
 
   useEffect(loadData, [id, setIncomes])
+  useEffect(() => {
+    if (incomeDataLoaded) {
```
hruoho, line 62: Usually we have used `isSuccess(incomes)` for the two first `if`'s here, wouldn't that do it?

frontend/packages/employee-frontend/src/assets/i18n/fi.ts
```diff
@@ -936,6 +936,12 @@ export const fi = {
           YEARLY: 'Vuosi'
         },
         updateError: 'Tulotietojen tallennus epäonnistui',
+        missingIncomeDaysWarningTitle: 'Tulotiedot puuttuvat joiltain päiviltä',
+        missingIncomeDaysWarningText: (missingIncomePeriodStrings: string[]) =>
```
hruoho, line 940: Sorry to annoy you but I wonder if we should, for the sake of simplicity, avoid explicitly listing the missing days but just inform the user that he/she should check the timespans... I know this suggestion is undermining much of your hard work but to me it just seems a slight overkill to introduce handful of new functions just to be able to format a warning message :grimacing:

mapehe: Thanks for the feedback @hruoho. I agree that this seems like a lot of complexity. I tried to implement this as precisely according to Figma as possible, though. In my opinion it might actually be worthwhile to list the missing income periods to improve the UX, since calculating the income gaps from the time periods can be pretty annoying.

Maybe we should refactor this though and move the concatenation logic away from `i18n/fi.ts`.

hruoho: Ah okay sorry I haven't seen that thing in Figma! In Jira they said nothing about displaying the missing dates so that's why I questioned this :shrug:

frontend/packages/employee-frontend/src/components/person-profile/PersonIncome.tsx
mapehe: Refactored this a bit

hruoho (approved):

PR #245, opened 2020-12-04
EVAKA-4006 e2e tests for mobile pairing flow

massproduction:
#### Summary
Added e2e tests for mobile pairing flow.

These tests only test only side of the pairing flow at a time and "mock" the other side with dev-api calls.

### DIFF
--- a/frontend/e2e-test/test/e2e/dev-api/index.ts
+++ b/frontend/e2e-test/test/e2e/dev-api/index.ts
@@ -580,3 +580,91 @@ export async function cleanUpApplicationEmails(): Promise<void> {
     throw new DevApiError(e)
   }
 }
+
+type PairingStatus =
+  | 'WAITING_CHALLENGE'
+  | 'WAITING_RESPONSE'
+  | 'READY'
+  | 'PAIRED'
+
+export interface PairingResponse {
+  id: UUID
+  unitId: UUID
+  challengeKey: string
+  responseKey: string | null
+  expires: Date
+  status: PairingStatus
+  mobileDeviceId: UUID | null
+}
+
+export async function postPairingChallenge(
+  challengeKey: string
+): Promise<PairingResponse> {
+  try {
+    const { data } = await devClient.post<PairingResponse>(
+      `/mobile/pairings/challenge`,
+      {
+        challengeKey
+      }
+    )
+    return data
+  } catch (e) {
+    throw new DevApiError(e)
+  }
+}
+
+export async function postPairingResponse(
+  pairingId: UUID,
+  challengeKey: string,
+  responseKey: string
+): Promise<PairingResponse> {
+  try {
+    const { data } = await devClient.post<PairingResponse>(
+      `/mobile/pairings/${pairingId}/response`,
+      {
+        challengeKey,
+        responseKey
+      }
+    )
+    return data
+  } catch (e) {
+    throw new DevApiError(e)
+  }
+}
+
+export async function postPairing(unitId: UUID): Promise<PairingResponse> {
+  try {
+    const { data } = await devClient.post<PairingResponse>(`/mobile/pairings`, {
+      unitId
+    })
+    return data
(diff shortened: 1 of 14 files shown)


frontend/e2e-test/test/e2e/specs/5_employee/unit-acl.spec.ts
```diff
@@ -72,6 +79,8 @@ fixture('Employee - Unit ACL')
   })
   .afterEach(logConsoleMessages)
   .after(async () => {
+    if (pairingId) await deletePairing(pairingId)
+    if (deviceId) await deleteMobileDevice(deviceId)
```
Joosakur, line 83: set these back to null after delete?

service/src/main/kotlin/fi/espoo/evaka/shared/dev/DevApi.kt
```diff
@@ -609,6 +618,64 @@ RETURNING id
...
+            val pairing = tx.initPairing(body.unitId)
+
+            asyncJobRunner.plan(
+                tx = tx,
+                payloads = listOf(GarbageCollectPairing(pairingId = pairing.id)),
+                runAt = ZonedDateTime
+                    .ofInstant(pairing.expires, zoneId)
+                    .plusDays(1)
+                    .toInstant()
+            )
```
Joosakur, line 662: not needed in dev-api imo

service/src/main/kotlin/fi/espoo/evaka/shared/dev/DevApi.kt
```diff
@@ -647,6 +714,14 @@ fun Handle.clearDatabase() = listOf(
     execute("DELETE FROM $it")
 }
 
+fun Handle.deletePairing(id: UUID) {
+    execute("DELETE FROM pairing WHERE id = ?", id)
+}
+
+fun Handle.deleteMobileDevice(id: UUID) {
```
Joosakur, line 721: replace deprecated `fun Handle.foo` with `fun Database.Transaction.foo`

frontend/e2e-test/test/e2e/dev-api/index.ts
```diff
@@ -580,3 +580,91 @@ export async function cleanUpApplicationEmails(): Promise<void> {
...
+  expires: Date
+  status: PairingStatus
+  mobileDeviceId: UUID | null
+}
+
+export async function postPairingChallenge(
+  challengeKey: string
+): Promise<PairingResponse> {
+  try {
+    const { data } = await devClient.post<PairingResponse>(
```
Joosakur, line 604: This is also fine, but I'm wondering if we could do these things through standard API without duplicating these to dev-api. Maybe @Gekkio can comment?

frontend/e2e-test/test/e2e/specs/6_mobile/pairing.spec.ts
```diff
@@ -0,0 +1,102 @@
...
+  const responseKey = await pairingFlow.responseKey.textContent
+  const pairingResponse = await postPairingResponse(
+    res.id,
+    res.challengeKey,
+    responseKey
+  )
+  pairingId = pairingResponse.id
+  deviceId = pairingResponse.mobileDeviceId
+
+  await t.expect(pairingFlow.mobilePairingTitle3.exists).ok()
```
Joosakur, line 101: I think we should verify that the session has been created correctly, i.e. go to the mobile app and verify that it opens the correct unit and is able to fetch some data.

Joosakur (changes requested):

frontend/e2e-test/test/e2e/specs/6_mobile/pairing.spec.ts
massproduction: Added check that in the group selector (the next page) we get the unit's name.

Joosakur (approved):

PR #248, opened 2020-12-04
EVAKA-3464 pending attachment download

hruoho:
#### Summary

When user uploads an attachment and tries to retrieve it too soon (or the file is irretrievable), show a modal saying that file processing is in progress.

### DIFF
--- a/frontend/packages/enduser-frontend/src/components/common/file-upload.vue
+++ b/frontend/packages/enduser-frontend/src/components/common/file-upload.vue
@@ -36,13 +36,13 @@ SPDX-License-Identifier: LGPL-2.1-or-later
         </div>
         <div class="file-details">
           <div class="file-header">
-            <a
+            <button
               v-if="file.id"
-              class="file-name"
+              class="file-name-button"
               @click="checkFileAvailability(file)"
             >
               {{ file.name }}
-            </a>
+            </button>
             <span v-else class="file-name">{{ file.name }}</span>
             <div>
               <button
@@ -104,7 +104,7 @@ SPDX-License-Identifier: LGPL-2.1-or-later
       :header="$t('file-upload.modal-header')"
       :message="$t('file-upload.modal-message')"
       :acceptText="$t('file-upload.modal-confirm')"
-      ref="infoUnavailableFile"
+      ref="fileUnavailableModal"
     >
     </confirm-modal>
   </div>
@@ -158,7 +158,7 @@ SPDX-License-Identifier: LGPL-2.1-or-later
         }
       },
       showFileUnavailableModal() {
-        this.$refs.infoUnavailableFile.open()
+        this.$refs.fileUnavailableModal.open()
       },
       deleteFile(file) {
         this.onDelete(file)
@@ -258,6 +258,7 @@ SPDX-License-Identifier: LGPL-2.1-or-later
         .file-icon-container {
           margin-right: 16px;
           flex: 0;
+          color: $blue-light;
         }
 
         .file-details {
@@ -296,6 +297,13 @@ SPDX-License-Identifier: LGPL-2.1-or-later
               white-space: nowrap;
               overflow: hidden;
               text-overflow: ellipsis;
+            }
+
+            .file-name-button {
+              border: none;
+              background: none;
+              cursor: pointer;
+              color: $blue-light;
             }
           }
 
--- a/frontend/packages/enduser-frontend/src/components/common/file-upload.vue
+++ b/frontend/packages/enduser-frontend/src/components/common/file-upload.vue
@@ -40,8 +40,6 @@ SPDX-License-Identifier: LGPL-2.1-or-later
               v-if="file.id"
               class="file-name"
               @click="deliverBlob(file)"
-              target="_blank"
-              rel="noreferrer"
             >
               {{ file.name }}
             </a>
@@ -101,16 +99,28 @@ SPDX-License-Identifier: LGPL-2.1-or-later
         </div>
       </div>
     </div>
+
+    <confirm-modal
+      :header="$t('file-upload.modal-header')"
+      :message="$t('file-upload.modal-message')"
+      :acceptText="$t('file-upload.modal-confirm')"
+      ref="infoUnavailableFile"
+    >
+    </confirm-modal>
   </div>
 </template>
 
 <script>
   import axios from 'axios'
+  import ConfirmModal from '@/components/modal/confirm.vue'
   export default {
     props: {
       files: Array,
       onUpload: Function,
       onDelete: Function
+    },
+    components: {
+      ConfirmModal
     },
     methods: {
       deliverBlob(file) {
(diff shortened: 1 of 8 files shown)


service/src/main/kotlin/fi/espoo/evaka/attachment/AttachmentsController.kt
```diff
@@ -114,6 +115,30 @@ class AttachmentsController(
...
+        user: AuthenticatedUser,
+        @PathVariable attachmentId: UUID
+    ): ResponseEntity<PreDownloadResponse> {
+        Audit.AttachmentsRead.log(targetId = attachmentId)
+        if (!user.hasOneOfRoles(UserRole.ADMIN, UserRole.SERVICE_WORKER, UserRole.FINANCE_ADMIN)) {
+            if (!db.read { it.isOwnAttachment(attachmentId, user) }) throw Forbidden("Permission denied")
+        }
+
+        return try {
+            s3Client.get(filesBucket, "123").let { _ ->
```
jaahon, line 134: Should probably use the real id here 😄

frontend/packages/enduser-frontend/src/localization/app/fi.json
```diff
@@ -1005,6 +1005,9 @@
         "Sallitut tiedostomuodot:",
         "PDF, JPEG/JPG, PNG ja DOC/DOCX"
       ]
-    }
+    },
+    "modal-header": "Käsittely kesken",
+    "modal-message": "Tiedosto ei ole juuri nyt ladattavissa. Kokeile hetken kuluttua uudelleen.",
+    "modal-confirm": "Sulje"
```
jaahon, line 1011: Should probably ask for these at least in Swedish as well if you haven't already.

frontend/packages/enduser-frontend/src/components/common/file-upload.vue
```diff
@@ -120,6 +157,9 @@ SPDX-License-Identifier: LGPL-2.1-or-later
           this.onUpload(event.dataTransfer.files[0])
         }
       },
+      showFileUnavailableModal() {
+        this.$refs.infoUnavailableFile.open().then(this.removeApplication, null)
```
jaahon, line 161: This probably doesn't belong here `.then(this.removeApplication, null)`

service/src/main/kotlin/fi/espoo/evaka/attachment/AttachmentsController.kt
```diff
@@ -114,6 +115,30 @@ class AttachmentsController(
...
+        user: AuthenticatedUser,
+        @PathVariable attachmentId: UUID
+    ): ResponseEntity<PreDownloadResponse> {
+        Audit.AttachmentsRead.log(targetId = attachmentId)
+        if (!user.hasOneOfRoles(UserRole.ADMIN, UserRole.SERVICE_WORKER, UserRole.FINANCE_ADMIN)) {
+            if (!db.read { it.isOwnAttachment(attachmentId, user) }) throw Forbidden("Permission denied")
+        }
+
+        return try {
+            s3Client.get(filesBucket, "$attachmentId").let { _ ->
```
mikkopiu, line 134: Shouldn't this be:
```suggestion
            s3Client.headObject(filesBucket, "$attachmentId").let { _ ->
```

mikkopiu (commented):
Would it be an option to just bake the availability checks straight into the download action and just catch the errors (and show the modal)? The availability check against S3 has to be done every time anyway, so there shouldn't a "performance" hit (not that that would ever be an issue here).

Also, supernit: shouldn't the link be a button instead when it's used as such (click handling and "complex" logic vs. static anchor)?

service/src/main/kotlin/fi/espoo/evaka/attachment/AttachmentsController.kt
hruoho: Yes it should... Thank you.

frontend/packages/enduser-frontend/src/localization/app/fi.json
hruoho: Yep looking for these atm

service/src/main/kotlin/fi/espoo/evaka/attachment/AttachmentsController.kt
```diff
@@ -114,6 +115,30 @@ class AttachmentsController(
...
+            if (!db.read { it.isOwnAttachment(attachmentId, user) }) throw Forbidden("Permission denied")
+        }
+
+        return try {
+            s3Client.headObject(filesBucket, "$attachmentId").let { _ ->
+                ResponseEntity.ok().body(PreDownloadResponse(true))
+            }
+        } catch (e: AmazonS3Exception) {
+            ResponseEntity.ok().body(PreDownloadResponse(false))
+        }
```
mikkopiu, line 139: I think it'd be better to explicitly only handle the forbidden case (looks like there isn't a specific class for that and might have to just check `.statusCode`). Currently we would swallow every other service exception (though not client exceptions, like connection errors) and it would make debugging harder.

frontend/packages/enduser-frontend/src/localization/app/fi.json
```diff
@@ -1005,6 +1005,9 @@
         "Sallitut tiedostomuodot:",
         "PDF, JPEG/JPG, PNG ja DOC/DOCX"
       ]
-    }
+    },
+    "modal-header": "File processing is ongoing",
```
mikkopiu, line 1009: Nit: "in progress" is better in this context. "Ongoing" is usually used to refer to something longer and more constant.
```suggestion
    "modal-header": "File processing is in progress",
```

mikkopiu:
@hruoho How do you feel about these:

> Would it be an option to just bake the availability checks straight into the download action and just catch the errors (and show the modal)? The availability check against S3 has to be done every time anyway, so there shouldn't a "performance" hit (not that that would ever be an issue here).
> 
> Also, supernit: shouldn't the link be a button instead when it's used as such (click handling and "complex" logic vs. static anchor)?

hruoho:
> @hruoho How do you feel about these:
> 
> > Would it be an option to just bake the availability checks straight into the download action and just catch the errors (and show the modal)? The availability check against S3 has to be done every time anyway, so there shouldn't a "performance" hit (not that that would ever be an issue here).
> > Also, supernit: shouldn't the link be a button instead when it's used as such (click handling and "complex" logic vs. static anchor)?

Well the first implementation was somewhat like this. Now it's splitted into two to keep those things apart and handle the pre-download check on it's own. To my mind the gain is in the separate return types. The pre-download endpoint returns a simple & extendable json which the frontend can manage as wished. And the other endpoint then returns the blob without any hassle.

But yes I guess this could still be streamlined a bit if necessary.

mikkopiu:
> Well the first implementation was somewhat like this. Now it's splitted into two to keep those things apart and to handle the pre-download check on it's own. To my mind the gain is in the separate return types. The pre-download endpoint returns a simple & extendable json which the frontend can manage as wished. And the other endpoint then returns the blob without any hassle.
> 
> But yes I guess this could still be streamlined a bit if necessary.

Your current solution is OK by me but as a not-really-a-link I feel it would be better as a button; users cannot interact with it like they would with a link (open in a new tab or not, copy the link etc.) and there's additional failure modes.

mikkopiu (approved):
:+1: and agreed to create tests for this in a separate PR later on

PR #249, opened 2020-12-04
Set max limit for file uploads per user

reynders:
#### Summary

Prevent uploading more than the specified amount of attachments. This should never happen to a non malicious user so no user UI is created for this.

### DIFF
--- a/service/src/integrationTest/kotlin/fi/espoo/evaka/attachments/AttachmentsControllerIntegrationTest.kt
+++ b/service/src/integrationTest/kotlin/fi/espoo/evaka/attachments/AttachmentsControllerIntegrationTest.kt
@@ -1,0 +1,66 @@
+package fi.espoo.evaka.attachments
+
+import com.github.kittinunf.fuel.core.FileDataPart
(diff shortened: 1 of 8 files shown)


service/src/integrationTest/kotlin/fi/espoo/evaka/attachments/AttachmentsControllerIntegrationTest.kt
```diff
@@ -0,0 +1,66 @@
...
+
+class AttachmentsControllerIntegrationTest : FullApplicationTest() {
+
+    private lateinit var user: AuthenticatedUser
+    private val applicationId = UUID.randomUUID()
+    private val pngFile = this::class.java.getResource("/attachments-fixtures/espoo-logo.png")
+
+    @BeforeEach
+    protected fun beforeEach() {
+        jdbi.handle { h ->
```
Gekkio, line 31: New code shouldn't use JDBI directly anymore. All the test superclasses expose a `Database.Connection` instance, so you can use `db.transaction { ... }` here

Gekkio (approved):

PR #252, opened 2020-12-07
KeyCloak to proxy

ilja-git:
#### Summary

Optional KeyCloak for proxy.

### DIFF
--- a/proxy/nginx.conf.template
+++ b/proxy/nginx.conf.template
@@ -123,7 +123,9 @@ server {
   add_header X-XSS-Protection '1; mode=block';
   add_header X-DNS-Prefetch-Control off;
   add_header Report-To '{"group","csp-endpoint","max_age":31536000,"endpoints":[{"url":"https://$host/api/csp"}]}';
-  add_header Content-Security-Policy "block-all-mixed-content; upgrade-insecure-requests; form-action 'self'; frame-ancestors 'none'; default-src 'self'; script-src 'self' https://maps.googleapis.com; font-src 'self' data: https://fonts.gstatic.com; style-src 'self' 'unsafe-inline' https://fonts.googleapis.com; img-src 'self' data: https://maps.gstatic.com https://maps.googleapis.com; connect-src 'self' https://sentry.io; object-src 'none'; report-uri /api/csp; report-to csp-endpoint";
+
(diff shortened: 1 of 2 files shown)


proxy/nginx.conf.template
```diff
@@ -251,6 +253,23 @@ server {
...
+
+  set $authUrl "<%= ENV["KEYCLOAK_URL"] %>";
+
+  location /auth {
+    <% if ENV["BASIC_AUTH_ENABLED"] == "true" %>
+    auth_basic "off";
+    <% end %>
+    include    proxy_params;
+    limit_req  zone=req_zone burst=100 nodelay;
+    add_header Content-Security-Policy "${contentSecurityPolicyBase}; script-src 'self' 'unsafe-inline' 'unsafe-eval'; frame-ancestors 'self'";
```
mikkopiu, line 266: We should investigate how to get rid of `unsafe-inline` and `unsafe-eval` here but I'm fine with moving this forward for finish the "PoC" -> if there isn't an easy solution for this now, make a task for it in the tech backlog so that we don't forget

proxy/nginx.conf.template
```diff
@@ -251,6 +253,23 @@ server {
...
   }
 
+  <% if ENV.key?("KEYCLOAK_URL") && ENV["KEYCLOAK_URL"] != "" %>
+
+  set $authUrl "<%= ENV["KEYCLOAK_URL"] %>";
+
+  location /auth {
+    <% if ENV["BASIC_AUTH_ENABLED"] == "true" %>
+    auth_basic "off";
+    <% end %>
```
mikkopiu, line 263: Why's this?

mikkopiu (approved):

proxy/nginx.conf.template
ilja-git: Some functionality did not work with basic authentication enabled and seems to problem with Authorization header that both functionality use.

mikkopiu: Ah right, and everything in keycloak is always behind authentication? I.e. it's safe to leave "more open"?

ilja-git: There is welcome page that is just links. But, we should not use dummy passwords in testing environments, but if/as we enforce password policy it should not be a problem. We could try to limit exceptions only to paths that are conflicting, but that would need to be tested with full flow.

mikkopiu (approved):
LGTM (with the note about `unsafe*`) :+1:

PR #254, opened 2020-12-08
[EVAKA-4006] Use a separate token in long term mobile cookie

Gekkio:
#### Summary

Instead of directly using the mobile device id in the cookie, use a separate token that can be rotated without having to remove+add actual rows in the mobile_device table.

### DIFF
--- a/apigw/src/shared/service-client.ts
+++ b/apigw/src/shared/service-client.ts
@@ -121,11 +121,16 @@ export interface ValidatePairingRequest {
   responseKey: string
 }
 
+export interface MobileDeviceIdentity {
+  id: UUID
+  longTermToken: UUID
+}
+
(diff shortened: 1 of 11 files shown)


apigw/src/shared/service-client.ts
```diff
@@ -146,6 +147,19 @@ export interface MobileDevice {
   unitId: UUID
 }
 
+export async function identifyMobileDevice(
+  req: express.Request,
+  token: UUID
+): Promise<MobileDeviceIdentity> {
+  const { data } = await client.get(
+    `/system/mobile-identity/${encodeURIComponent(token)}`,
```
Joosakur, line 155: safe enough to pass token in url?

Joosakur (approved):

PR #270, opened 2020-12-11
EVAKA-4018: 5 year old absences

Joosakur:

### DIFF
--- a/frontend/packages/employee-frontend/src/types/absence.ts
+++ b/frontend/packages/employee-frontend/src/types/absence.ts
@@ -53,10 +53,17 @@ export const billableCareTypes: CareType[] = ['PRESCHOOL_DAYCARE', 'DAYCARE']
 export function formatCareType(
   careType: CareType,
   placementType: PlacementType,
+  dateOfBirth: LocalDate,
   i18n: Translations
 ) {
   const isPreparatory =
     placementType === 'PREPARATORY' || placementType === 'PREPARATORY_DAYCARE'
+
+  const hasFree5YearsOldDaycare =
+    dateOfBirth.year === LocalDate.today().year - 5
+
+  if (careType === 'DAYCARE' && hasFree5YearsOldDaycare)
+    return i18n.common.types.DAYCARE_5YO_PAID
 
   if (isPreparatory && careType === 'PRESCHOOL')
     return i18n.common.types.PREPARATORY_EDUCATION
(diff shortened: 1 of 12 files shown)


frontend/packages/employee-frontend/src/types/absence.ts
```diff
@@ -53,11 +53,18 @@ export const billableCareTypes: CareType[] = ['PRESCHOOL_DAYCARE', 'DAYCARE']
...
   careType: CareType,
   placementType: PlacementType,
+  dateOfBirth: LocalDate,
   i18n: Translations
 ) {
   const isPreparatory =
     placementType === 'PREPARATORY' || placementType === 'PREPARATORY_DAYCARE'
 
+  const hasFree5YearsOldDaycare =
+    dateOfBirth.year === LocalDate.today().year - 5
```
jaahon, line 63: This is not how it's determined. Children that turn 5 in year xxxx get the discount during the "season" of 1.8.xxxx - 31.7.(xxxx + 1), so basically for the year before preschool.

Joosakur: Oh.. argh

service/src/main/kotlin/fi/espoo/evaka/attendance/ChildAttendanceController.kt
```diff
@@ -443,3 +436,17 @@ fun wasAbsentFromPreschoolDaycare(placementType: PlacementType, arrived: LocalTi
 
     return false
 }
+
+fun wasAbsentFromPaidDaycare(placementBasics: ChildPlacementBasics, arrived: LocalTime, departed: LocalTime): Boolean {
+    val (placementType, dateOfBirth) = placementBasics
+
+    if (!isEntitledToFreeFiveYearsOldDaycare(dateOfBirth)) {
+        return false
```
jaahon, line 444: Shouldn't this be `true` or am I missing something? I don't fully understand what's going on here.

jaahon (approved):

service/src/main/kotlin/fi/espoo/evaka/attendance/ChildAttendanceController.kt
Joosakur: No, but the function name may still be misleading. The function is used when child is leaving to determine if we should add an absence for the paid portion of the daycare. That scenario only applies to 5yo case, hence false. Maybe I'll rename it as `wasAbsentFrom5yoPaidDaycare`?

jaahon: Yeah that rename sounds like a good idea.

PR #272, opened 2020-12-14
Upgrade to kotlin 1.4.21

Gekkio:


mikkopiu (approved):
:+1: Verified locally. The reentrant lock (vs. non-reentrant mutex) change is also :white_check_mark:  in the current single-threaded & blocking Varda context (not sure why I didn't go with that originally).
EDIT: Might've been 'cause I was trying hard to make it non-blocking and thread-safe (without succeeding).

PR #273, opened 2020-12-14
Improve Varda fee data reliability

jaahon:
#### Summary

Base Varda fee data on the actual decision data in Varda instead of doing unreliable pairing with evaka placement data. Since Varda requires fee data to match its decision data this increases reliability of the fee data integration and also prevents the old bug where identical fee data was sent to Varda multiple times.

Clean up related fee data from Varda when deleting a Varda decision. Currently, deleting Varda decisions fails every time they have overlapping fee data in Varda.

Improve the mock Varda client so that the stored data matches the actual operations.

### DIFF
--- a/service/src/main/kotlin/fi/espoo/evaka/varda/integration/MockVardaIntegrationEndpoint.kt
+++ b/service/src/main/kotlin/fi/espoo/evaka/varda/integration/MockVardaIntegrationEndpoint.kt
@@ -169,24 +169,6 @@ class MockVardaIntegrationEndpoint(private val mapper: ObjectMapper) {
         ResponseEntity.noContent().build()
     }
 
-    @GetMapping("v1/lapset/{childId}/varhaiskasvatuspaatokset/")
-    fun getDecisions(
-        @PathVariable childId: Long,
-        @RequestHeader(name = "Authorization") auth: String
-    ): ResponseEntity<String> = lock.withLock {
-        val result =
-            """{
-            "results": 
-            ${mapper.writeValueAsString(
-                decisions.filter { (_, decision) ->
-                    decision.childUrl.contains(childId.toString())
-                }
-            )}
-            }
-            """.trimIndent()
-        ResponseEntity.ok(result)
-    }
-
     @PostMapping("/v1/varhaiskasvatussuhteet/")
     fun createPlacement(
         @RequestBody body: VardaPlacement,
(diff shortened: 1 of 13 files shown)


service/src/main/kotlin/fi/espoo/evaka/varda/integration/MockVardaIntegrationEndpoint.kt
```diff
@@ -32,177 +36,226 @@ private val logger = KotlinLogging.logger {}
 @RestController
 @RequestMapping("/mock-integration/varda/api")
 class MockVardaIntegrationEndpoint(private val mapper: ObjectMapper) {
-    val decisions = mutableListOf<VardaDecision>()
-    val feeData = mutableListOf<VardaFeeData>()
-    val children = mutableListOf<VardaChildRequest>()
+    private val mutex = Mutex()
```
Gekkio, line 39: kotlinx has been removed in master, so you should use standard library `ReentrantLock` instead.

See https://github.com/espoon-voltti/evaka/pull/272/commits/e6e9037085047b26a016f77f64f2ed42ec69acd2 for more info

service/src/main/kotlin/fi/espoo/evaka/varda/integration/MockVardaIntegrationEndpoint.kt
```diff
@@ -32,177 +36,226 @@ private val logger = KotlinLogging.logger {}
...
+    var decisionId = 0L
+    val decisions = mutableMapOf<Long, VardaDecision>()
+    var placementId = 0L
+    val placements = mutableMapOf<Long, VardaPlacement>()
+    var feeDataId = 0L
+    val feeData = mutableMapOf<Long, VardaFeeData>()
+    var childId = 0L
+    val children = mutableMapOf<Long, VardaChildRequest>()
+
+    fun cleanUp() {
```
Gekkio, line 59: `cleanUp` should also use the lock, because it mutates global state

service/src/main/kotlin/fi/espoo/evaka/varda/integration/MockVardaIntegrationEndpoint.kt
jaahon: 👍

service/src/main/kotlin/fi/espoo/evaka/varda/integration/MockVardaIntegrationEndpoint.kt
jaahon: Oh yeah, true!

Gekkio (approved):

PR #279, opened 2020-12-14
EVAKA-PATCH: do not expose error messages

Joosakur:


service/src/main/kotlin/fi/espoo/evaka/shared/controllers/ExceptionHandler.kt
```diff
@@ -116,26 +106,19 @@ class ExceptionHandler : ResponseEntityExceptionHandler() {
...
-            ErrorResponse(
-                message = ex.message ?: ""
-            )
+            ErrorResponse()
         )
     }
 
     @ExceptionHandler(value = [Throwable::class])
     fun unexpectedError(req: HttpServletRequest, ex: Throwable): ResponseEntity<ErrorResponse> {
         // Checked exceptions get wrapped in UndeclaredThrowableException in kotlin
```
Gekkio, line 117: This comment no longer makes sense since the code its referring to has been removed

Gekkio (approved):

PR #283, opened 2020-12-15
Evaka 3926: Finance decision manager

mapehe:
#### Summary

* Set decision manager in the unit editor.
* Filter fee decisions by decision manager


jaahon (commented):
I don't think `manager` is a very accurate term, maybe `handler` would be more fitting?

frontend/packages/employee-frontend/src/components/common/Filters.tsx
```diff
@@ -276,6 +276,38 @@ export const UnitFilter = React.memo(function UnitFilter({
...
 
+interface FinanceDecisionManagerFilterProps {
+  employees: { id: string; label: string }[]
+  selected?: { id: string; label: string }
+  select: (employee: string) => void
+}
+
+export const FinanceDecisionManagerFilter = React.memo(
+  function FinanceDecisionManagerFilter({
+    employees,
```
jaahon, line 287: I think a more descriptive name like `decisionHandlers` would be better than a very generic `employees`.

frontend/packages/employee-frontend/src/components/fee-decisions/FeeDecisionFilters.tsx
```diff
@@ -43,6 +52,10 @@ function FeeDecisionFilters() {
     void getUnits([], 'DAYCARE').then(setUnits)
   }, [])
 
+  useEffect(() => {
+    void getEmployees().then(setEmployees)
```
jaahon, line 56: You should not use a list of all employees in the filter, it's going to be very long and most of the values are irrelevant. You'll have to create a separate endpoint for that.

frontend/packages/employee-frontend/src/components/fee-decisions/FeeDecisionFilters.tsx
```diff
@@ -140,6 +159,32 @@ function FeeDecisionFilters() {
...
           />
+          <Gap size="L" />
+          <FinanceDecisionManagerFilter
+            employees={employees
+              .map((e) =>
+                e.map(({ id, firstName, lastName }) => ({
+                  id,
+                  label: [firstName, lastName].join(' ')
+                }))
+              )
```
jaahon, line 170: You could do this options preprosessing when you store the value in state instead of having to do it on every render.

frontend/packages/employee-frontend/src/components/fee-decisions/FeeDecisionFilters.tsx
```diff
@@ -140,6 +159,32 @@ function FeeDecisionFilters() {
...
+                    .map(({ id, firstName, lastName }) => ({
+                      id,
+                      label: [firstName, lastName].join(' ')
+                    }))
+                    .filter(
+                      (employee) =>
+                        employee.id === searchFilters.financeDecisionManagerId
+                    )[0]
+              )
+              .getOrElse(undefined)}
```
jaahon, line 185: Similarly here, you could store a selected value so that it can be passed to the component as it is, so you don't have to iterate through the whole list every time this component is rendered.

frontend/packages/employee-frontend/src/components/unit/unit-details/UnitEditor.tsx
```diff
@@ -896,6 +912,28 @@ export default function UnitEditor(props: Props): JSX.Element {
...
+        <div>{showRequired(i18n.unitEditor.label.financeDecisionManager)}</div>
+        {props.editable ? (
+          <Select
+            options={employeeOptions}
+            placeholder={i18n.unitEditor.placeholder.financeDecisionManager}
+            onChange={(value) =>
+              value && 'value' in value
+                ? updateForm({ financeDecisionManagerId: value.value })
+                : undefined
+            }
```
jaahon, line 925: You should pass the selected value to the component.

frontend/packages/employee-frontend/src/components/unit/unit-details/UnitEditor.tsx
```diff
@@ -514,6 +523,13 @@ export default function UnitEditor(props: Props): JSX.Element {
     label: name
   }))
 
+  const employeeOptions = props.employees.map(
+    ({ id, firstName, lastName }) => ({
+      value: id,
+      label: [firstName, lastName].join(' ')
+    })
+  )
```
jaahon, line 531: Same thing here as in the filter component.

frontend/packages/employee-frontend/src/components/unit/unit-details/UnitEditor.tsx
```diff
@@ -896,6 +912,28 @@ export default function UnitEditor(props: Props): JSX.Element {
...
+                : undefined
+            }
+          />
+        ) : (
+          props.employees
+            .map((employee) => ({
+              id: employee.id,
+              name: [employee.firstName, employee.lastName].join(' ')
+            }))
+            .find((employee) => employee.id === form.financeDecisionManagerId)
```
jaahon, line 933: It's usually preferable to first `find` and then `map` the value to whatever format you need.

frontend/packages/employee-frontend/src/components/voucher-value-decisions/VoucherValueDecisionFilters.tsx
```diff
@@ -33,6 +42,10 @@ export default React.memo(function VoucherValueDecisionFilters() {
     void getAreas().then(setAvailableAreas)
   }, [])
 
+  useEffect(() => {
+    void getEmployees().then(setEmployees)
```
jaahon, line 46: All the comments I left at the fee decisions components apply here as well.

frontend/packages/employee-frontend/src/state/invoicing-ui.tsx
```diff
@@ -90,6 +93,8 @@ interface InvoiceSearchFilterState {
 interface SharedState {
   units: Result<Unit[]>
   setUnits: Dispatch<SetStateAction<Result<Unit[]>>>
+  employees: Result<Employee[]>
```
jaahon, line 96: So, like I commented elsewhere, you could store the options objects here instead of full `Employee` objects.

frontend/packages/employee-frontend/src/types/invoicing.ts
```diff
@@ -265,6 +266,7 @@ export interface FeeDecisionDetailed {
   approvedAt: Date | null
   createdAt: Date
   sentAt: Date | null
+  financeDecisionManager: { employee: Employee } | null
```
jaahon, line 269: I don't think this field needs to be this detailed, something like `financeDecisionManager: string` with the value being the name would be enough.

service/src/main/kotlin/fi/espoo/evaka/invoicing/data/VoucherValueDecisionQueries.kt
```diff
@@ -342,6 +345,7 @@ fun Handle.searchValueDecisions(
                 AND youngest_child.area = ANY(:areas)
                 AND (:unit::uuid IS NULL OR part.placement_unit = :unit)
                 AND $freeTextQuery
+                ${if (financeDecisionManagerId != null) financeDecisionManagerWhere else ""}
```
jaahon, line 348: The idea nowadays is to use conditions like `AND (:unit::uuid IS NULL OR part.placement_unit = :unit)` with optional conditions so that the query string is as close to valid sql as possible. So in this case you could just write `AND (:financeDecisionManagerId::uuid IS NULL OR placement_unit.finance_decision_manager = :financeDecisionManagerId)`. The query in `FeeDecisionQueries.kt` is the old format.

service/src/main/kotlin/fi/espoo/evaka/invoicing/data/FeeDecisionQueries.kt
```diff
@@ -106,6 +114,7 @@ val feeDecisionDetailedQueryBase =
         LEFT JOIN daycare ON part.placement_unit = daycare.id
         LEFT JOIN care_area ON daycare.care_area_id = care_area.id
         LEFT JOIN employee as approved_by ON decision.approved_by = approved_by.id
+        LEFT JOIN employee as finance_decision_manager ON daycare.finance_decision_manager = finance_decision_manager.id
```
jaahon, line 117: Instead of just picking whatever person is the decision handler at the time of the query, I think the person should be stored in a `fee_decision` row so that the data matches reality later as well.

service/src/main/resources/db/migration/V28__create_finance_decision_manager_field.sql
```diff
@@ -0,0 +1,3 @@
+ALTER TABLE daycare ADD COLUMN finance_decision_manager uuid;
+ALTER TABLE daycare ADD FOREIGN KEY (finance_decision_manager)
+REFERENCES employee(id) ON DELETE SET null;
```
jaahon, line 3: I don't know if you did this in separate steps on purpose, but you can combine this to a single statement `ALTER TABLE daycare ADD COLUMN finance_decision_manager uuid REFERENCES employee(id) ON DELETE SET NULL;`

jaahon (commented):
I don't think `manager` is a very accurate term, maybe `handler` would be more fitting?

service/src/main/kotlin/fi/espoo/evaka/invoicing/data/FeeDecisionQueries.kt
jaahon: Oh, and before you change anything, the handler should be stored only for approved decisions. Drafts should still have it fetched dynamically. And since this is supposed to replace the area filter, the handler should be determined by the placement unit of the youngest child to match the behavior requested by the users. That is done so that decisions are split distinctly between handlers and the same person handles a family’s decisions for as long as possible.

service/src/main/kotlin/fi/espoo/evaka/pis/EmployeeQueries.kt
```diff
@@ -44,7 +44,21 @@ WHERE (:id::uuid IS NULL OR e.id = :id) AND md.id IS NULL
     .mapTo<Employee>()
     .asSequence()
 
+private fun Handle.searchFinanceDecisionHandlers(id: UUID? = null) = createQuery(
+    // language=SQL
+    """
+SELECT e.id, e.first_name, e.last_name, e.email, e.external_id, e.created, e.updated, e.roles
+FROM employee e
+JOIN daycare ON daycare.finance_decision_handler = e.id
+LEFT JOIN mobile_device md on e.id = md.id
```
jaahon, line 53: This part is unnecessary since mobile devices won't be picked as decision handlers anyway.

frontend/packages/employee-frontend/src/components/fee-decisions/FeeDecisionFilters.tsx
```diff
@@ -43,6 +60,17 @@ function FeeDecisionFilters() {
...
+  useEffect(() => {
+    void getFinanceDecisionHandlers().then(setFinanceDecisionHandlers)
+  }, [])
+
+  useEffect(() => {
+    const handler = financeDecisionHandlers
+      .getOrElse([])
+      .find((handler) => handler.id === searchFilters.financeDecisionHandlerId)
+    setSelectedFinanceDecisionHandler(handler)
+  }, [searchFilters.financeDecisionHandlerId])
```
jaahon, line 72: This and the `useState` hook could just be replaced by this
```
const selectedFinanceDecisionHandler = useMemo(
  () => financeDecisionHandlers
    .getOrElse([])
    .find((handler) => handler.id === searchFilters.financeDecisionHandlerId), 
  [searchFilters.financeDecisionHandlerId]
)
```

service/src/main/kotlin/fi/espoo/evaka/invoicing/data/FeeDecisionQueries.kt
```diff
@@ -503,13 +521,27 @@ fun findFeeDecisionsForHeadOfFamily(
...
+        )
         UPDATE fee_decision
         SET
             status = :status,
             decision_number = nextval('fee_decision_number_sequence'),
-            approved_by = :approvedBy,
+            approved_by = CASE
+                WHEN youngest_child.finance_decision_handler_id IS NOT NULL THEN youngest_child.finance_decision_handler_id
+                ELSE :approvedBy
+                END,
```
jaahon, line 540: I would rather store this data in a new column, so we have both the information of who sent the decision and who was the designated handler at the time.

service/src/main/kotlin/fi/espoo/evaka/invoicing/data/FeeDecisionQueries.kt
```diff
@@ -56,6 +56,16 @@ val feeDecisionQueryBase =
...
+    WITH finance_decision_handler_by_youngest_child AS (
+        SELECT
+            fee_decision_part.fee_decision_id AS decision_id,
+            finance_decision_handler.first_name AS first_name,
+            finance_decision_handler.last_name AS last_name,
+            row_number() OVER (PARTITION BY (fee_decision_id) ORDER BY date_of_birth DESC) AS rownum
+        FROM fee_decision_part
+        LEFT JOIN daycare ON fee_decision_part.placement_unit = daycare.id
+        LEFT JOIN employee AS finance_decision_handler ON finance_decision_handler.id = daycare.finance_decision_handler
+    )
```
jaahon, line 68: Decisions in draft state don't need to display this information, so just left join the `employee` table straight to the `fee_decision` table.

service/src/main/kotlin/fi/espoo/evaka/daycare/CareArea.kt
```diff
@@ -64,6 +68,12 @@ data class Daycare(
     val operationDays: Set<Int>
 )
 
+@PropagateNull("finance_decision_handler_id")
+data class financeDecisionHandler(
+    @Nested("")
+    val employee: Employee
+)
```
jaahon, line 75: Instead of having this nested structure that pulls data that is ultimately not needed, just change this into a simple data class. Something like this 
```
data class FinanceDecisionHandler(val id: UUID, val firstName: String, val lastName: String)
```

frontend/packages/employee-frontend/src/components/unit/unit-details/UnitEditor.tsx
jaahon: To elaborate since the original comment isn't visible anymore, instead of doing this data processing every time this component renders, you could just do it once where you store the values in state. So do this processing in the part where you call `setEmployees`.

jaahon (approved):

PR #284, opened 2020-12-15
Fix an accepted application's read rights

jaahon:
#### Summary

Previously if an application's placement plan was applied (i.e. its decisions were accepted) unit supervisors couldn't read the application unless the applicant had applied to the unit.


mikkopiu (commented):
Could this be regression tested with the unit supervisor role?

jaahon:
Yeah I wrote an e2e test for it now.

mikkopiu (approved):
:+1:

PR #286, opened 2020-12-15
EVAKA-HOTFIX hide mobile users from employees listing

massproduction:
#### Summary
This PR hides "mobile device users" from employee listings.

### DIFF
--- a/frontend/e2e-test/test/e2e/pages/employee/units/unit-page.ts
+++ b/frontend/e2e-test/test/e2e/pages/employee/units/unit-page.ts
@@ -4,6 +4,7 @@
 
 import { ClientFunction, Selector, t } from 'testcafe'
 import config from '../../../config'
+import { postPairingChallenge } from '../../../dev-api'
 import { UUID } from '../../../dev-api/types'
 
 export default class UnitPage {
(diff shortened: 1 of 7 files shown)


service/src/main/kotlin/fi/espoo/evaka/shared/dev/DevApi.kt
```diff
@@ -388,6 +389,12 @@ DELETE FROM attachment USING ApplicationsDeleted WHERE application_id = Applicat
         return ResponseEntity.ok().build()
     }
 
+    @DeleteMapping("/employee/all")
+    fun deleteEmployees(db: Database): ResponseEntity<Unit> {
+        db.transaction { it.handle.deleteEmployees() }
```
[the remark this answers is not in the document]
Joosakur: Btw are mobile devices cascaded or do we need to first delete them to avoid errors? This could also fail with any other references, like absences/decisions created by the employee. So now I'm starting to think this whole approach of deleting all employees does not work. Either we need to surgically delete correct things in correct order or go for the full db nuke. This level of partial nuking is likely to just cause more issues.

service/src/main/kotlin/fi/espoo/evaka/shared/dev/DevApi.kt
```diff
@@ -388,6 +389,12 @@ DELETE FROM attachment USING ApplicationsDeleted WHERE application_id = Applicat
         return ResponseEntity.ok().build()
     }
 
+    @DeleteMapping("/employee/all")
```
Joosakur, line 392: where is this used?

Joosakur (changes requested):

service/src/main/kotlin/fi/espoo/evaka/shared/dev/DevApi.kt
massproduction: It was used at some point. Deleted it now.

frontend/e2e-test/test/e2e/pages/employee/units/unit-page.ts
```diff
@@ -126,6 +127,31 @@ export default class UnitPage {
...
+    }
+    await t.expect(this.mobileDevicesNameInput.exists).ok()
+
+    const pairingId = res.id
+    const deviceId = await this.mobileDevicesNameInput.getAttribute('id')
+
+    await t.typeText(this.mobileDevicesNameInput, 'testphone')
+    await t.click(this.mobileDevicePairingDoneBtn)
+
+    return { pairingId, deviceId }
```
jaahon, line 153: I personally don't see the value in doing assertions like this `await t.expect(this.mobileDevicesResponseKeyInput.exists).ok()` since the `await t.typeText(this.mobileDevicesResponseKeyInput, res.responseKey)` kind of does the same assertion as well. 

Also, if a missing `res.responseKey` means that the test can't succeed, should you just throw an error? Now the test just skips inputting the value silently.

jaahon (approved):
Good job 👍

jaahon:
Sorry I should refresh before I write my reviews 😂

frontend/e2e-test/test/e2e/pages/employee/units/unit-page.ts
massproduction: Removed the extra `exists).ok` checks and added a `throw new Error` when not receiving the responseKey.

Joosakur (approved):

PR #287, opened 2020-12-15
EVAKA-4041: Add special education teacher view for admin

kustaahuhtala:
#### Summary

- Special education teachers can be added or removed from an unit
- E2E test for this case

#### How to test

- As an admin, go to unit page
- Add user to special education teacher role
- Login as that user and verify you can see the unit
- Login as an admin and remove the user from special education teacher role

### DIFF
--- a/frontend/e2e-test/test/e2e/pages/employee/units/unit-page.ts
+++ b/frontend/e2e-test/test/e2e/pages/employee/units/unit-page.ts
@@ -75,6 +75,11 @@ export default class UnitPage {
   readonly supervisorAcl = new AclTable(
     Selector('[data-qa="daycare-acl-supervisors"]')
   )
+
+  readonly specialEducationTeacherAcl = new AclTable(
+    Selector('[data-qa="daycare-acl-set"]')
+  )
+
   readonly staffAcl = new AclTable(Selector('[data-qa="daycare-acl-staff"]'))
 
   readonly placementProposalsAcceptButton = Selector(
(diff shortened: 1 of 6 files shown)


massproduction (approved):

service/src/main/kotlin/fi/espoo/evaka/daycare/controllers/UnitAclController.kt
```diff
@@ -65,6 +65,34 @@ class UnitAclController(private val acl: AccessControlList) {
         return ResponseEntity.noContent().build()
     }
 
+    @PutMapping("/daycares/{daycareId}/specialeducationteacher/{employeeId}")
```
kustaahuhtala, line 68: I wonder whether we should do this in a more generic way (since it's pretty much the same as with `/daycares/{daycareId}/supervisors/{employeeId}` mapping. We could have one generic endpoint for both, supervisor and specialeducationteacher. Even required roles are the same. But then again, I wonder if it's worth it since there are only two cases.

frontend/e2e-test/test/e2e/pages/employee/units/unit-page.ts
```diff
@@ -75,6 +75,11 @@ export default class UnitPage {
   readonly supervisorAcl = new AclTable(
     Selector('[data-qa="daycare-acl-supervisors"]')
   )
+
+  readonly specialEducationTeacherAcl = new AclTable(
+    Selector('[data-qa="daycare-acl-set"]')
```
Joosakur, line 80: set is slightly unfortunate acronym :D

Joosakur (approved):

PR #289, opened 2020-12-16
Fix Varda expired token refresh

jaahon:
#### Summary

Update the Varda token refresh error message check.

### DIFF
--- a/service/src/main/kotlin/fi/espoo/evaka/varda/integration/VardaClient.kt
+++ b/service/src/main/kotlin/fi/espoo/evaka/varda/integration/VardaClient.kt
@@ -413,8 +413,10 @@ class VardaClient(
                 .header(Headers.ACCEPT, "application/json")
                 .responseStringWithRetries(maxTries) { r, remainingTries ->
                     when (r.second.statusCode) {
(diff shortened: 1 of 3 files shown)


service/src/main/kotlin/fi/espoo/evaka/varda/integration/VardaClient.kt
```diff
@@ -423,7 +423,10 @@ class VardaClient(
                                     .authentication().token(newToken)
                                     .responseStringWithRetries(remainingTries)
                             }
-                            else -> r
+                            else -> {
+                                logger.error { "Varda request was unauthorized (${r.third.error.errorData})" }
```
mikkopiu, line 427: I'd rather add this to the call site, i.e. `getAllPages` error handler. All the other handler log like this:

```kotlin
logger.error(result.getException()) {
                    "Updating a placement to Varda failed (body: $updatedPlacement)." +
                        " message: ${String(result.error.errorData)}"
                }
```

mikkopiu (changes requested):

mikkopiu (approved):

PR #291, opened 2020-12-16
EVAKA-4098: Improve decision sending if non-disclosure order in place

kustaahuhtala:
#### Summary

If the applicant (guardian) has non-disclosure order, do not
automatically send decision to second guardian, even if he/she
didn't have non-disclosure order.

In this case, service worker contacts the guardian to determine where
and how to send the decision.

#### How to test

- Create an application with first guardian having non-disclosure order and second guardian not having one.
- Handle the application all the way to the decision phase
- Make sure the second guardian does not receive the decision as a Suomi.fi message

### DIFF
--- a/frontend/e2e-test/test/e2e/specs/5_employee/application-details.spec.ts
+++ b/frontend/e2e-test/test/e2e/specs/5_employee/application-details.spec.ts
@@ -11,6 +11,7 @@ import {
 import { applicationFixture } from '../../dev-api/fixtures'
 import { logConsoleMessages } from '../../utils/fixture'
 import {
+  cleanUpMessages,
   cleanUpInvoicingDatabase,
   createPlacementPlan,
   deleteAclForDaycare,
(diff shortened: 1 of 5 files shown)


frontend/e2e-test/test/e2e/specs/5_employee/application-details.spec.ts
```diff
@@ -74,6 +79,16 @@ fixture('Application - employee application details')
       ),
       id: '0c8b9ad3-d283-460d-a5d4-77bdcbc69374'
     }
+    application4 = {
```
mikkopiu, line 82: Supernit: all of these should be named so that their purpose/content is obvious; `application4` says nothing but `applicationRestrictedDetails` would be clearer.

mikkopiu (approved):

reynders (approved):

frontend/e2e-test/test/e2e/specs/5_employee/application-details.spec.ts
kustaahuhtala: You're correct, I improved this a little.

mikkopiu (approved):
LGTM with the fixes (and good catch, @reynders, on the message cleanup)

PR #292, opened 2020-12-16
EVAKA-4040: replace can apply flags with periods

Joosakur:


frontend/packages/employee-frontend/src/api/unit.ts
```diff
@@ -32,7 +33,31 @@ function convertUnitJson(unit: JsonOf<Unit>): Unit {
   return {
     ...unit,
     openingDate: unit.openingDate ? LocalDate.parseIso(unit.openingDate) : null,
-    closingDate: unit.closingDate ? LocalDate.parseIso(unit.closingDate) : null
+    closingDate: unit.closingDate ? LocalDate.parseIso(unit.closingDate) : null,
+    daycareApplyPeriod: unit.daycareApplyPeriod
+      ? {
```
Gekkio, line 38: It would be nicer to have a `parseOpenEndedPeriod` function and just use it here a couple of times

frontend/e2e-test/test/e2e/dev-api/types.ts
```diff
@@ -175,6 +177,10 @@ export interface CareArea {
   subCostCenter: string
 }
 
+interface OpenEndedPeriod {
```
Gekkio, line 180: IMHO it would be better to use the same name as in the service code

frontend/packages/employee-frontend/src/components/unit/unit-details/UnitEditor.tsx
```diff
@@ -437,9 +442,9 @@ function toFormData(unit: Unit | undefined): FormData {
       : type?.includes('CENTRE')
       ? 'CENTRE'
       : undefined,
-    canApplyDaycare: unit?.canApplyDaycare ?? false,
-    canApplyPreschool: unit?.canApplyPreschool ?? false,
-    canApplyClub: unit?.canApplyClub ?? false,
+    daycareApplyPeriod: unit?.daycareApplyPeriod ?? null,
```
Gekkio, line 445: Do we really need the default values here? "if value is null, return null instead"
There more of this later too...IMHO it makes the code less readable since it adds a pointless "branch" you need to think about when trying to understand the code

frontend/packages/employee-frontend/src/components/unit/unit-details/UnitEditor.tsx
```diff
@@ -687,27 +692,94 @@ export default function UnitEditor(props: Props): JSX.Element {
...
-          <Checkbox
-            disabled={!props.editable}
-            label={i18n.unitEditor.field.canApplyClub}
-            checked={form.canApplyClub}
-            onChange={(canApplyClub) => updateForm({ canApplyClub })}
-            dataQa="application-type-checkbox-CLUB"
-          />
+          {['daycare', 'preschool', 'club'].map((type) => {
+            const key = `${type}ApplyPeriod`
+            const period = form[key] as OpenEndedPeriod | null
```
Gekkio, line 697: This unchecked cast and the other one a couple of lines later will almost certainly cause a bug later and debugging+fixing it will take more time than implementing this correctly in the first place.

It's a bit more code, but actually spelling out the different names avoids "dynamic name generation" and loss of type safety. For example:

```
[
  {type: 'daycare', checkboxI18n: 'canApplyDaycare', field: 'daycareApplyPeriod'},
  ...
].map(...)
```

frontend/e2e-test/test/e2e/dev-api/types.ts
Joosakur: yes, but that is Period, which already exists here with a different meaning.. I suggest we fix this as part of the larger daterange refactoring

frontend/packages/employee-frontend/src/api/unit.ts
Joosakur: I suggest we fix this as part of the larger daterange refactoring, for example as a method in a new typescript class Period

frontend/e2e-test/test/e2e/dev-api/types.ts
Gekkio: OK, works for me

frontend/packages/employee-frontend/src/api/unit.ts
Gekkio: OK, works for me

frontend/packages/employee-frontend/src/components/unit/unit-details/UnitEditor.tsx
Joosakur: hmm I think the issue is that the value may be undefined and that needs to be defaulted to null instead

Gekkio: Hmm I think I picked a bad example...I think there was another place where there was no mixing of undefined and null.

Of course this raises another question: why do we mix undefined and null in the first place? (this is a bigger issue and not specific to this pull request)

frontend/packages/employee-frontend/src/components/unit/unit-details/UnitEditor.tsx
```diff
@@ -687,27 +692,107 @@ export default function UnitEditor(props: Props): JSX.Element {
...
+                                })
+                              }}
+                            />
+                          ) : (
+                            period.start.format()
+                          )}
+                          {' - '}
+                          {props.editable ? (
+                            <DatePickerClearable
+                              date={period?.end ?? null}
```
Joosakur, line 763: this maybe

frontend/packages/employee-frontend/src/components/unit/unit-details/UnitEditor.tsx
Joosakur: Removed those other cases I could find. Yeah our undefined/null usage would need general cleanup.

Gekkio (approved):

PR #294, opened 2020-12-17
EVAKA-4040: prevent preschool applications for next term before 8.1.2021

Joosakur:


Gekkio:
It's not critical, but the right terminology is "preschool term", not "preschool season"

kustaahuhtala (approved):
Tested locally, it works fine

PR #295, opened 2020-12-17
Avoid doing VTJ querys on varda fee data update

reynders:


jaahon (approved):
So I guess we should just try this.

PR #306, opened 2020-12-22
Evaka fix delete also application attachments

reynders:


service/src/main/kotlin/fi/espoo/evaka/application/ApplicationQueries.kt
```diff
@@ -596,36 +599,44 @@ fun Handle.getApplicationUnitSummaries(unitId: UUID): List<ApplicationUnitSummar
 @JsonIgnoreProperties(ignoreUnknown = true)
 data class FormWithType(val type: String)
 
-fun mapRequestedPlacementType(row: RowView, colName: String): PlacementType = when (row.mapJsonColumn<FormWithType>(colName).type) {
```
reynders, line 599: I do not understand why this non - change is in the diff

hruoho (approved):

PR #317, opened 2020-12-31
Unify DB connection pool configs, Spring configs

mikkopiu:
#### Summary
- Allow configuring DB connection pool sizes in services
   - As all of the services might use a single database instance, configuring a maximum database connection (Hikari) pool size is important to prevent services taking up too many connections from a shared pool
   - When the datasource configuration was refactored, the maximum pool size configuration was hard-coded instead of reading the `spring.datasource.hikari.maximumPoolSize` property -> re-utilize the property if set and set a reasonable default if not
      - Also explicitly configure maximum pool size in message-service
   - **NOTE:** Keycloak was skipped for now as we aren't yet running our own image of it and pool size configuration cannot be configured as environment variables, only in the `standalone.xml` file -> created separate task for this
      - Thankfully, the default for Keycloak's Postgres driver is a reasonable 20
- Convert all Spring props files to YAML
   - To unify configuration style across the board -- was already started during last Spring Boot version upgrade
- Drop duplicate properties files
   - Integration test properties for service and message-service is split unnecessarily into `application-integration-test.properties` and `integration-test-common.properties` files (artifact of a previous CI vs. local split) -> combine into a single file
      - This also allows dropping the hard-coded `@PropertySources`
- service-lib: drop [deprecated](https://github.com/spring-projects/spring-boot/wiki/Spring-Boot-2.0-Migration-Guide#security) security-properties (`security.basic.enabled` and `security.ignored`), i.e. the whole `application.properties` file
- Gradle: use v6.7 file watching configuration
   - Since Gradle 6.7, `org.gradle.unsafe.watch-fs` has been replaced by `org.gradle.vfs.watch`

#### Testing

In `master`:

1. Run app with `LOGGING_LEVEL_COM_ZAXXER_HIKARI=DEBUG SPRING_DATASOURCE_HIKARI_MAXIMUMPOOLSIZE=7 ./gradlew bootRun` -> Hikari will output its configuration and **ignore** the environment variable & property configs and use the hard-coded value of 20

In branch:

1. Run app with `LOGGING_LEVEL_COM_ZAXXER_HIKARI=DEBUG SPRING_DATASOURCE_HIKARI_MAXIMUMPOOLSIZE=7 ./gradlew bootRun` -> Hikari will now have maximumPoolSize correctly set to 7


jaahon (approved):
It seems like the `service-lib`s `application.properties` files is unnecessary as the properties configured there are deprecated in spring boot 2.

Apart from that everything else seems good.

mikkopiu:
> It seems like the `service-lib`s `application.properties` files is unnecessary as the properties configured there are deprecated in spring boot 2.
> 
> Apart from that everything else seems good.

Ooh, great catch! I'll delete that file completely.

PR #321, opened 2021-01-04
Use correct citizen frontend sentry dsn

jaahon:
#### Summary

The previous value was just copied from employee frontend config.

### DIFF
--- a/compose/proxy/nginx.conf
+++ b/compose/proxy/nginx.conf
@@ -41,7 +41,7 @@ server {
   add_header X-Content-Type-Options nosniff;
   add_header X-XSS-Protection '1; mode=block';
   add_header X-DNS-Prefetch-Control off;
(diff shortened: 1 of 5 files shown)


mikkopiu (changes requested):
Currently only `https://sentry.io` is allowed in our CSP (`connect-src`), it must now be extended to include this new organization-specific ingest domain, too. I'd amend the list with an explicit domain (and not just expand to a wildcard).

Also: looks like Sentry have changed the ingest addresses (`@sentry.io` -> `@o<org number>.ingest.sentry.io`), i.e. DSN's have changed ([source](https://blog.sentry.io/2020/10/27/notice-of-internet-address-change)) -- would you mind updating all the other DSN's too?

mikkopiu (approved):

PR #327, opened 2021-01-05
Citizen decisions list E2E test

reynders:

### DIFF
--- a/frontend/e2e-test/test/e2e/specs/0_citizen/citizen-decisions.spec.ts
+++ b/frontend/e2e-test/test/e2e/specs/0_citizen/citizen-decisions.spec.ts
@@ -5,19 +5,16 @@
 import { logConsoleMessages } from '../../utils/fixture'
 import { enduserRole } from '../../config/users'
 import CitizenHomePage from '../../pages/citizen/citizen-homepage'
-import { initializeAreaAndPersonData } from '../../dev-api/data-init'
 import {
-  deleteApplication,
-  deleteEmployeeById,
-  insertApplications,
-  insertEmployeeFixture
-} from '../../dev-api'
+  AreaAndPersonFixtures,
+  initializeAreaAndPersonData
+} from '../../dev-api/data-init'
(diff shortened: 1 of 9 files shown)


jaahon (commented):
The citizen tests don't clean up after themselves properly, you can't run the tests twice.

reynders:
> The citizen tests don't clean up after themselves properly, you can't run the tests twice.

I believe this is the case with all of our e2e tests? I could add the cleanup to this PR (basically a full db reset), but it is unrelated with the decision functionality. Or did you have a specific case in mind?

frontend/e2e-test/test/e2e/specs/0_citizen/citizen-decisions.spec.ts
```diff
@@ -0,0 +1,74 @@
...
+const citizenHomePage = new CitizenHomePage()
+const citizenDecisionsPage = new CitizenDecisionsPage()
+
+let applicationId: string
+let employee: EmployeeBuilder
+
+fixture('Citizen decisions')
+  .meta({ type: 'regression', subType: 'citizen-decisions' })
+  .before(async () => {
+    await initializeAreaAndPersonData()
```
jaahon, line 29: This data is not cleaned up afterwards. The function returns a function that does it.

reynders: Yes - which is "await Fixture.cleanup()" which is called in the suite cleanup. But agree that it is not obvious, so will add a redundant call via the function returned by initializeAreaAndPersonData 👍

jaahon: Oh, nevermind then. You don't need to add the function call, I just didn't know `Fixture.cleanup` does that. There's something else wrong with the clean up then. It definitely has worked for other test fixtures before.

jaahon (approved):

PR #330, opened 2021-01-07
EVAKA-HOTFIX: include file extension to attachment name

hruoho:
#### Summary

When user downloads an attachment, we use the file's UUID instead of it's original name and hence lose the file extension. Now it's parsed from the original name and appended to UUID. 

Also fixed some missing translations related to attachments.

### DIFF
--- a/service/src/main/kotlin/fi/espoo/evaka/attachment/AttachmentsController.kt
+++ b/service/src/main/kotlin/fi/espoo/evaka/attachment/AttachmentsController.kt
@@ -147,9 +147,11 @@ class AttachmentsController(
         val attachment =
             db.read { it.getAttachment(attachmentId) ?: throw NotFound("Attachment $attachmentId not found") }
 
+        val extension = parseExtension(attachment.name) ?: throw Forbidden("Couldn't parse valid file extension for attachment $attachmentId")
+
         return s3Client.get(filesBucket, "$attachmentId").let { document ->
             ResponseEntity.ok()
-                .header("Content-Disposition", "attachment;filename=${document.getName()}")
+                .header("Content-Disposition", "attachment;filename=${document.getName()}.$extension")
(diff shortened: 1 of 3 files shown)


service/src/main/kotlin/fi/espoo/evaka/attachment/AttachmentsController.kt
```diff
@@ -147,9 +147,11 @@ class AttachmentsController(
         val attachment =
             db.read { it.getAttachment(attachmentId) ?: throw NotFound("Attachment $attachmentId not found") }
 
+        val extension = parseExtension(attachment.name) ?: throw Forbidden("Couldn't parse valid file extension for attachment $attachmentId")
+
         return s3Client.get(filesBucket, "$attachmentId").let { document ->
             ResponseEntity.ok()
-                .header("Content-Disposition", "attachment;filename=${document.getName()}")
+                .header("Content-Disposition", "attachment;filename=${document.getName()}.$extension")
```
jaahon, line 154: `document.getName()` should probably be replaced with `attachment.name` as well, since it seems kind of silly to have people download files with uuids as their names.

service/src/main/kotlin/fi/espoo/evaka/attachment/AttachmentsController.kt
```diff
@@ -147,9 +147,11 @@ class AttachmentsController(
         val attachment =
             db.read { it.getAttachment(attachmentId) ?: throw NotFound("Attachment $attachmentId not found") }
 
+        val extension = parseExtension(attachment.name) ?: throw Forbidden("Couldn't parse valid file extension for attachment $attachmentId")
```
jaahon, line 150: I don't think it's necessary to throw on "invalid" extension since the validity should have been checked when the attachment was saved. At this point we should just assume that whatever the extension is, it should be valid.

jaahon (approved):

PR #332, opened 2021-01-07
EVAKA-HOTFIX: Fetch preschool term from db

hruoho:
#### Summary

Fetch preschool term from db instead of using hard coded terms

### DIFF
--- a/service/src/main/kotlin/fi/espoo/evaka/placement/PlacementPlanService.kt
+++ b/service/src/main/kotlin/fi/espoo/evaka/placement/PlacementPlanService.kt
@@ -8,8 +8,8 @@ import fi.espoo.evaka.application.ApplicationDetails
 import fi.espoo.evaka.application.ApplicationType
 import fi.espoo.evaka.application.DaycarePlacementPlan
 import fi.espoo.evaka.application.fetchApplicationDetails
+import fi.espoo.evaka.daycare.getActivePreschoolTermAt
 import fi.espoo.evaka.deriveClubTerm
-import fi.espoo.evaka.derivePreschoolTerm
 import fi.espoo.evaka.shared.async.AsyncJobRunner
 import fi.espoo.evaka.shared.async.NotifyPlacementPlanApplied
 import fi.espoo.evaka.shared.db.Database
@@ -47,7 +47,8 @@ class PlacementPlanService(
 
         when (application.type) {
             ApplicationType.PRESCHOOL -> {
-                val term = derivePreschoolTerm(startDate)
+                val term = tx.getActivePreschoolTermAt(startDate)?.extendedTerm
+                    ?: throw Exception("No suitable preschool term found for start date $startDate")
                 val period = FiniteDateRange(startDate, term.end)
                 val preschoolDaycarePeriod = if (type == PlacementType.PRESCHOOL_DAYCARE || type == PlacementType.PREPARATORY_DAYCARE) {
                     FiniteDateRange(startDate, LocalDate.of(term.end.year, 7, 31))
@@ -131,8 +132,10 @@ class PlacementPlanService(
                     requestedStartDate?.let { plan.preschoolDaycarePeriod?.copy(start = it) }
                         ?: plan.preschoolDaycarePeriod
                     )!!
-                // TODO: this should not be hard-coded
-                val term = derivePreschoolTerm(period.start)
+
+                val term = tx.getActivePreschoolTermAt(period.start)?.extendedTerm
+                    ?: throw Exception("No suitable preschool term found for start date ${period.start}")
+
                 // if the preschool daycare extends beyond the end of the preschool term, a normal daycare
                 // placement is used because invoices are handled differently
                 if (period.end.isAfter(term.end)) {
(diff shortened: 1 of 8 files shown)


service/src/main/kotlin/fi/espoo/evaka/placement/PlacementPlanService.kt
```diff
@@ -131,8 +132,10 @@ class PlacementPlanService(
                     requestedStartDate?.let { plan.preschoolDaycarePeriod?.copy(start = it) }
                         ?: plan.preschoolDaycarePeriod
                     )!!
-                // TODO: this should not be hard-coded
-                val term = derivePreschoolTerm(period.start)
```
jaahon, line 135: I think you might as well replace all usages of this function and just delete it completely.

service/src/main/kotlin/fi/espoo/evaka/placement/PlacementPlanService.kt
```diff
@@ -47,7 +47,8 @@ class PlacementPlanService(
 
         when (application.type) {
             ApplicationType.PRESCHOOL -> {
-                val term = derivePreschoolTerm(startDate)
+                val term = tx.getActivePreschoolTermAt(startDate)?.extendedTerm
```
jaahon, line 50: This should be either `finnishPreschool` or `swedishPreschool` depending on the unit instead of `extendedTerm`. The `extendedTerm` should be used when determining the `preschoolDaycarePeriod` below.

service/src/main/kotlin/fi/espoo/evaka/placement/PlacementPlanService.kt
```diff
@@ -131,8 +132,10 @@ class PlacementPlanService(
                     requestedStartDate?.let { plan.preschoolDaycarePeriod?.copy(start = it) }
                         ?: plan.preschoolDaycarePeriod
                     )!!
-                // TODO: this should not be hard-coded
-                val term = derivePreschoolTerm(period.start)
+
+                val term = tx.getActivePreschoolTermAt(period.start)?.extendedTerm
```
jaahon, line 136: Similarly, you should not use `extendedTerm` here.

service/src/main/kotlin/fi/espoo/evaka/placement/PlacementPlanService.kt
hruoho: At this point of the flow we don't yet have only one unit and hence the dates are like best guesses, aren't they? Maybe the `finnishPreschool` would be much better than `extendedTerm` though.

jaahon: Oh yeah, true. In either case, `extendedTerm` is incorrect. I guess `finnishPreschool` is a good default until Svebi decides otherwise 😄

jaahon (approved):
😎

service/src/main/kotlin/fi/espoo/evaka/placement/PlacementPlanService.kt
```diff
@@ -47,10 +47,12 @@ class PlacementPlanService(
 
         when (application.type) {
             ApplicationType.PRESCHOOL -> {
-                val term = derivePreschoolTerm(startDate)
-                val period = FiniteDateRange(startDate, term.end)
+                val preschoolTerms = tx.getActivePreschoolTermAt(startDate)
+                    ?: throw Exception("No suitable preschool term found for start date $startDate")
```
Joosakur, line 51: I wonder if we should default to the next active term (if such exists) in case none is active during the requested date?

service/src/main/kotlin/fi/espoo/evaka/placement/PlacementPlanService.kt
```diff
@@ -47,10 +47,12 @@ class PlacementPlanService(
...
+                val exactTerm = if (isSvebiUnit(tx, preferredUnits[0].id)) preschoolTerms.swedishPreschool else preschoolTerms.finnishPreschool
+                val period = FiniteDateRange(startDate, exactTerm.end)
                 val preschoolDaycarePeriod = if (type == PlacementType.PRESCHOOL_DAYCARE || type == PlacementType.PREPARATORY_DAYCARE) {
-                    FiniteDateRange(startDate, LocalDate.of(term.end.year, 7, 31))
+                    FiniteDateRange(startDate, LocalDate.of(preschoolTerms.extendedTerm.end.year, 7, 31))
```
Joosakur, line 55: hmm why do we hard code 31.7.

Joosakur: ah nevermind, I guess that is not available in the data

Joosakur (approved):

service/src/main/kotlin/fi/espoo/evaka/placement/PlacementPlanService.kt
hruoho: I'll ask around to figure how end date should be defined but will merge this now anyway

service/src/main/kotlin/fi/espoo/evaka/placement/PlacementPlanService.kt
hruoho: I'll ask around to figure how that end date should be defined but will merge this now anyway

service/src/main/kotlin/fi/espoo/evaka/placement/PlacementPlanService.kt
hruoho: I thought this too but then decided to minimize all the guessing for now... Maybe it was a stupid decision, we'll see.

PR #333, opened 2021-01-08
Citizen footer

reynders:

### DIFF
--- a/frontend/packages/citizen-frontend/src/Footer.tsx
+++ b/frontend/packages/citizen-frontend/src/Footer.tsx
@@ -1,0 +1,43 @@
+import React from 'react'
+import { useTranslation } from '~localization'
+import styled from 'styled-components'
+import colors from '@evaka/lib-components/src/colors'
+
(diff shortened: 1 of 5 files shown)


frontend/packages/citizen-frontend/src/Footer.tsx
```diff
@@ -0,0 +1,45 @@
...
+`
+
+const MunicipalityContainer = styled.div``
+
+const Nav = styled.nav`
+  & a {
+    display: inline-block;
+    margin: 0 40px 0 40px;
+    text-decoration: none;
+  }
```
jaahon, line 42: These styles could just be moved under `NavItem`, which could be named something else since like I mentioned in my previous comment, they're not really a part of the application's navigation.

jaahon (commented):
The styles are a bit off. The font size and colors are not quite the same (figma shows the exact values) and the layout does not work as it should when moving towards mobile view sizes.

jaahon (approved):

PR #334, opened 2021-01-08
Add support for temporary placements to Varda integration

jaahon:
#### Summary

Send temporary placements separately from other decisions derived from placements since their deduction is quite unreliable and they behave differently compared to regular placements. There is a real risk of the temporary decisions sent to Varda overlapping with regular decisions but it's difficult and unreliable to prevent it without implementing temporary placements differently.


mikkopiu (approved):
LGTM

PR #342, opened 2021-01-11
Smooth scroll polyfill for Safari & IE, update axios

mikkopiu:
#### Summary
- Safari, IE and older browsers (e.g. Chrome 45 on OS X) don't support `ScrollToOptions` (and equivalent for other scroll methods) at all or `behavior: smooth` when other options are available -> all the browser are still supported, so include a conditional polyfill
  - Include polyfill in enduser, employee & citizen frontends
  - Polyfill is only applied when `scrollBehavior` isn't supported (`'scrollBehavior' in document.documentElement.style`)
- [`seamless-scroll-polyfill`](https://github.com/magic-akari/seamless-scroll-polyfill) appears to have fixed most of [`smoothscroll-polyfill`](https://github.com/iamdustan/smoothscroll)'s open issues (like SVG support) + adds official TypeScript typings which is nice
- Also update axios to the latest version (0.21.1) that contains vulnerability fixes (and an official `isAxiosError` implementation)

#### Testing

With `master` version:

1. Open enduser-frontend in **Safari**
2. Start creating a new application
3. Click one of the section headers
4. -> browser should **immediately** (i.e. not smoothly) scroll the page so the header is on top

If you have an old enough Chrome available on OS X (e.g. v49 on Mountain Lion/10.7, not sure which version fixes this as MDN simply states this as fixed for "Chrome" and does not separate the OS X and Windows/Linux builds): repeat and scrolling should fail altogether. **NOTE:** I tested this manually with Browserstack (v49 on Mountain Lion) and could repeat the issue.

With branch version:

1. Repeat
2. -> browser should **smoothly** scroll into the same position
3. Compare to e.g. Firefox / Chrome behaviour

### DIFF
--- a/apigw/src/shared/routes/auth/saml/index.ts
+++ b/apigw/src/shared/routes/auth/saml/index.ts
@@ -68,7 +68,7 @@ function createLoginHandler({
       (err: any, user: SamlUser | undefined) => {
         if (err || !user) {
           const description =
-            parseDescriptionFromSamlError(err) || 'Could not parse SAML error.'
+            parseDescriptionFromSamlError(err) || 'Could not parse SAML error'
           logAuditEvent(
             `evaka.saml.${strategyName}.sign_in_failed`,
             req,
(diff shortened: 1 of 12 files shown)


jaahon (approved):
Works at least on newest Safari.

PR #347, opened 2021-01-12
Update ruby in proxy docker image

jaahon:
#### Summary

Current build is broken with old ruby version.

### DIFF
--- a/proxy/Dockerfile
+++ b/proxy/Dockerfile
@@ -18,7 +18,7 @@ ENV SECURITYTXT_CONTACTS 'mailto:fake@fake.test;mailto:another@fake.test'
 ENV SECURITYTXT_LANGUAGES fi,se,en
 ENV STATIC_FILES_ENDPOINT_URL http://fake.test
 
-RUN apk add --no-cache ruby=2.7.1-r3
+RUN apk add --no-cache 'ruby=~2.7'
 
 COPY proxy_params /etc/nginx/proxy_params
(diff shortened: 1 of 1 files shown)


reynders (approved):

proxy/Dockerfile
```diff
@@ -18,7 +18,7 @@ ENV SECURITYTXT_CONTACTS 'mailto:fake@fake.test;mailto:another@fake.test'
 ENV SECURITYTXT_LANGUAGES fi,se,en
 ENV STATIC_FILES_ENDPOINT_URL http://fake.test
 
-RUN apk add --no-cache ruby=2.7.1-r3
+RUN apk add --no-cache ruby=2.7.2-r0
```
mikkopiu, line 21: As suggested in Slack, I think the best strategy for us would be to instead use version ranges:

- Alpine main repository (officially) only keeps the latest stable releases -> pinning is actually not great / recommended ([ref](https://gitlab.alpinelinux.org/alpine/abuild/-/issues/9996#note_87135), [ref](https://gitlab.alpinelinux.org/alpine/abuild/-/issues/9996))
- Don't update to non-patch releases without us knowing about it

e.g.: `apk add --no-cache 'ruby=~2.7'`

The

mikkopiu (changes requested):

proxy/Dockerfile
mikkopiu: I actually didn't remember/realize this when I went around pinning these originally.

mikkopiu (approved):

PR #350, opened 2021-01-13
EVAKA-3899: Translations

mapehe:


Joosakur (approved):
No idea about the swedish and there's minor things on some english translation that I'd say differently, but I think this is good enough for now, since I assume even the finnish texts have not been fully finalized yet and we'll hopefully get official translations later on.

PR #355, opened 2021-01-14
Add a mechanism that can refresh mobile sessions

Gekkio:
#### Summary

This "idle tracker" triggers an /auth/status call if the browser window
becomes visible after being non-visible *and* there is 20 minutes or
more since the last request to backend.

On mobile devices where a valid long term token is available, this means
if you open the browser after it has been hidden, the session is
refreshed and the user can immediately continue (assuming network
connectivity is ok).

On desktop devices, this can do a "force logout" if the session has
already disappeared, but this can actually be better than the current
situation where the application just stops working or throws errors
everywhere.

I'm not 100% happy with the idleTracker abstraction, but it works and we can change it later if necessary.


Joosakur (approved):
Code looks very good and clear to me. 💎  I can test it in action tomorrow if you wish, but I think this should work nicely.

PR #359, opened 2021-01-14
Evaka hotfix sort attendances

massproduction:
#### Summary

Sort child attendances children by first name.

### DIFF
--- a/frontend/packages/employee-frontend/src/api/attendances.ts
+++ b/frontend/packages/employee-frontend/src/api/attendances.ts
@@ -2,7 +2,7 @@
 //
 // SPDX-License-Identifier: LGPL-2.1-or-later
 
-import { JsonOf } from '@evaka/lib-common/src/json'
+import { JsonOf, JsonOfObject } from '@evaka/lib-common/src/json'
 import { Failure, Result, Success } from '@evaka/lib-common/src/api'
 import { UUID } from '~types'
 import { AbsenceType, CareType } from '~types/absence'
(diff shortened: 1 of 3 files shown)


service/src/main/kotlin/fi/espoo/evaka/attendance/ChildAttendanceController.kt
```diff
@@ -341,7 +341,7 @@ private fun Database.Read.getAttendancesResponse(unitId: UUID): AttendanceRespon
         )
     }
 
-    return AttendanceResponse(unitInfo, children)
+    return AttendanceResponse(unitInfo, children.sortedBy { it.firstName })
```
Joosakur, line 344: I think others have encouraged sorting in frontend. You can use lodash `_,sortyBy(children, child => child.name)`

massproduction: Moved the sorting to the frontend.

frontend/packages/employee-frontend/src/api/attendances.ts
```diff
@@ -201,13 +201,26 @@ export async function postDeparture(
...
+}
+
 function deserializeAttendanceResponse(
   data: JsonOf<AttendanceResponse>
 ): AttendanceResponse {
   {
     return {
       unit: data.unit,
-      children: data.children.map((attendanceChild) => {
+      children: data.children.sort(compareFirstNames).map((attendanceChild) => {
```
Joosakur, line 223: Now two children with same first name will have undefined order, wouldn't it make more sense to then sort by last name next? Also a custom comparator for such a simple case is quite verbose compared to using lodash.

frontend/packages/employee-frontend/src/api/attendances.ts
```diff
@@ -201,13 +201,26 @@ export async function postDeparture(
     .catch((e) => Failure.fromError(e))
 }
 
+function compareFirstNames(
+  a: JsonOfObject<AttendanceChild>,
```
Joosakur, line 205: do we need `JsonOfObject` here?

massproduction: Yeah, it was giving a warning without it.

Joosakur (approved):
only minor comments 👍

frontend/packages/employee-frontend/src/api/attendances.ts
Joosakur: huh, weird

frontend/packages/employee-frontend/src/api/attendances.ts
```diff
@@ -220,20 +221,23 @@ function deserializeAttendanceResponse(
...
-                departed: attendanceChild.attendance.departed
-                  ? new Date(attendanceChild.attendance.departed)
-                  : null
-              }
-            : null
-        }
-      })
+      children: data.children
+        .sort((a, b) => compareByProperty(a, b, 'lastName'))
+        .sort((a, b) => compareByProperty(a, b, 'firstName'))
```
Joosakur, line 226: fwiw alternative option would be `_.sortBy(data.children, ['firstName', 'lastName']).map( ...`

PR #363, opened 2021-01-15
Log numbers for varda integration data

jaahon:
#### Summary

There's a mysterious problem in production and I hope this might help debug it. This isn't too spammy so keeping it at `info` level should be fine.


service/src/main/kotlin/fi/espoo/evaka/varda/VardaDecisions.kt
```diff
@@ -83,6 +86,7 @@ fun sendNewDecisions(db: Database.Connection, client: VardaClient) {
     }
 
     val newTemporaryDecisions = db.read { getNewTemporaryDecisions(it, client.getChildUrl) }
+    logger.info { "Varda: Creating ${newDecisions.size} new temporary decisions" }
```
reynders, line 89: newTemporaryDecisions

reynders (approved):

service/src/main/kotlin/fi/espoo/evaka/varda/VardaDecisions.kt
jaahon: Good catch! 😅

PR #364, opened 2021-01-15
EVAKA-HOTFIX Removed history manipulation from attendance marking

massproduction:
#### Summary
After these changes the back button should always return the user to the last page user was before coming to the child page. In practice what this accomplishes is that when user clicks on a child in the list view and marks them present (or does other state transitions) the `history.goBack()` call will return the user to the list page instead of returning to the child page.


jaahon (commented):
Works well 👍

jaahon:
By the way if the delay on the `AsyncButton` is too long it could be shortened as well.

Joosakur (approved):
+1 for considering shortening async button time

PR #375, opened 2021-01-20
Employee: prevent downloads of pending decision PDFs

mikkopiu:
#### Summary
- Employee: prevent download attempts for pending decision PDFs
   - Replace link with a warning message (see screenshot below)
   - The more robust solution would be to do a separate pre-download step (like in enduser-frontend file upload) but this is such an edge case, I felt this to be a satisfactory solution
   - Add a simple E2E test case for the new logic
- Service: return 404 instead of throwing an exception when attempting to download a pending decision PDF
  - A delay in processing is a valid state as PDF generation is done in a separate async job
  - 429 with some arbitrary retry time could've also worked but 404 feels more semantically correct as there's no way of knowing when the user should retry
- Service: create new Dev API endpoint for triggering decision PDF generation (allows for more stable testing of the above logic)
- Include document URI generation in decision integration tests
  - Ensure decisions have documentUri set (i.e. a PDF generated and uploaded to S3) once they've been created and async job queue is empty

image

### DIFF
--- a/frontend/e2e-test/test/e2e/specs/2_employee-2/application-transitions.spec.ts
+++ b/frontend/e2e-test/test/e2e/specs/2_employee-2/application-transitions.spec.ts
@@ -7,19 +7,20 @@ import {
   initializeAreaAndPersonData,
   AreaAndPersonFixtures
 } from '../../dev-api/data-init'
-import { applicationFixture } from '../../dev-api/fixtures'
+import { applicationFixture, decisionFixture } from '../../dev-api/fixtures'
 import { logConsoleMessages } from '../../utils/fixture'
 import { EmployeeDetail } from '../../dev-api/types'
 import {
   cleanUpMessages,
+  createDecisionPdf,
   deleteAclForDaycare,
   deleteApplication,
   deleteEmployeeFixture,
   execSimpleApplicationAction,
   execSimpleApplicationActions,
   insertApplications,
+  insertDecisionFixtures,
   insertEmployeeFixture,
-  runPendingAsyncJobs,
   setAclForDaycares as setAclForDaycare
 } from '../../dev-api'
 import { seppoAdminRole, seppoManagerRole } from '../../config/users'
@@ -46,6 +47,7 @@ let fixtures: AreaAndPersonFixtures
 let cleanUp: () => Promise<void>
 let applicationId: string
 let applicationId2: string
+let supervisorId: string
 
 fixture('Application-transitions')
   .meta({ type: 'regression', subType: 'daycare-application' })
@@ -55,7 +57,7 @@ fixture('Application-transitions')
       ...supervisor,
       email: `${Math.random().toString(36).substring(7)}@espoo.fi`
     }
-    await insertEmployeeFixture(uniqueSupervisor)
+    supervisorId = await insertEmployeeFixture(uniqueSupervisor)
     await setAclForDaycare(
       config.supervisorExternalId,
       fixtures.daycareFixture.id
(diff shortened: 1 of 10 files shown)


service/src/main/kotlin/fi/espoo/evaka/application/ApplicationControllerV2.kt
```diff
@@ -262,7 +262,7 @@ class ApplicationControllerV2(
             ResponseEntity.ok(
                 ApplicationResponse(
                     application = application,
-                    decisions = decisions.map { it.copy(documentUri = null) },
+                    decisions = decisions,
```
Joosakur, line 265: If I recall correctly, the point of this was to not expose s3 uri outside. Preferrably the field would not even exist in the response type, but having an almost identical type for it and mapping between them is also annoying, since in kotlin we dont have something like `({documentUri, ...rest}) => ({...rest})`

Joosakur: inheritance might be one way to solve this but then we can't use `data class`

frontend/e2e-test/test/e2e/specs/2_employee-2/application-transitions.spec.ts
```diff
@@ -295,3 +296,32 @@ test('Placement proposal flow', async (t) => {
...
+  await execSimpleApplicationActions(applicationId, [
+    'move-to-waiting-placement',
+    'create-default-placement-plan',
+    'send-decisions-without-proposal'
+  ])
+
+  await applicationReadView.openApplicationByLink(applicationId)
+  await applicationReadView.assertDecisionDownloadPending('DAYCARE')
+
+  await runPendingAsyncJobs()
```
Joosakur, line 324: I wonder if this will sometimes fail because the async job would happen to finish too soon. If I'm correct, `runPendingAsyncJobs` ensures that pending jobs have been run, but they _may_ have been run before that already anyway.

mikkopiu: Ah, that's true. I'll check if there's some way around this.

service/src/main/kotlin/fi/espoo/evaka/application/ApplicationControllerV2.kt
mikkopiu: I stumbled across this solution in `DecisionController` which actually feels like a nice middle ground between the current change and implementing a whole separate interface/sealed class just for `Decision`:

```kotlin
fun withPublicDocumentUri(decisions: List<Decision>) = decisions.map { decision ->
    decision.copy(documentUri = "/api/internal/decisions2/${decision.id}/download") // do not expose direct s3 uri
}
```

I'd maybe move that to the data class itself, so the full solution would be like:

```kotlin
data class Decision(...) {
    // other methods...
    fun withPublicDocumentUris() = this.copy(
            documentUri = if (this.documentUri != null) "/api/internal/decisions2/${this.id}/download" else null,
            otherGuardianDocumentUri = if (this.otherGuardianDocumentUri != null) "/api/internal/decisions2/${this.id}/download" else null
    ) // Do not expose direct S3 URIs
}

// Usage:
val decisions = getDecisionsByApplication(tx.handle, applicationId, acl.getAuthorizedUnits(user))
        .map { it.withPublicDocumentUris() }
```

-> no need for a separate mapping & solves the exposed S3 URL issue (though that's IMO more of a smell than an actual issue, as the bucket still has be properly access controlled).

mikkopiu: :thinking: Also, looks like the original S3 URIs are currently returned from quite a few endpoints (obviously as they're part of the data model and it requires explicitly obscuring them).

mikkopiu: Security-wise I don't feel exposing the URIs is a major issue so my suggestion is: go forward with this as-is (+E2E test fixes) and take tech debt for overhauling the decision data model so that it doesn't have the full URIs at all (just filenames/IDs and anything further is just download-time configuration). This current is issue is such an edge case, I don't think I should use too much time on this right now. WDYT?

Joosakur: Sounds good to me 👍

frontend/e2e-test/test/e2e/specs/2_employee-2/application-transitions.spec.ts
Joosakur: If it doesn't seem to be failing in practice you could also just leave a comment so if it later becomes unstable the reason is easy to figure out

mikkopiu: Decided to rewrite this test with different (and new) Dev API endpoints which should now properly control when the PDF generation is triggered, making this stable.

mikkopiu:
Issue created in tech backlog for the discussed S3 URI improvement/migration.

Joosakur (approved):

PR #377, opened 2021-01-21
Update custodian info on all DVV modification types

reynders:
Custodian information should be updated also when custodian relation is deleted and updated


kustaahuhtala (commented):
Looks good. I think we should unify naming regarding DVV/VTJ in the code, logs and comments. VTJ (and VTJ muutostietorajapinta) is the system we are using for refreshing customer information, DVV is the organisation. But this is so minor thing, that I don't think it needs to be changed just now.

kustaahuhtala (approved):

reynders:
> Looks good. I think we should unify naming regarding DVV/VTJ in the code, logs and comments. VTJ (and VTJ muutostietorajapinta) is the system we are using for refreshing customer information, DVV is the organisation. But this is so minor thing, that I don't think it needs to be changed just now.

Yes. For now the VTJ / DVV separation serves as a distinction between the old API and new API code

PR #383, opened 2021-01-22
EVAKA-4103: improved application update api

Joosakur:

### DIFF
--- a/service/src/main/kotlin/fi/espoo/evaka/application/ApplicationStateService.kt
+++ b/service/src/main/kotlin/fi/espoo/evaka/application/ApplicationStateService.kt
@@ -413,36 +413,44 @@ class ApplicationStateService(
 
         val updatedForm = ApplicationForm.fromV0(formV0, original.childRestricted, original.guardianRestricted)
 
-        if (listOf(SENT).contains(original.status)) {
-            original.form.preferences.preferredStartDate?.let { previousStartDate ->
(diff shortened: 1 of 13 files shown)


service/src/main/kotlin/fi/espoo/evaka/application/ApplicationStateService.kt
```diff
@@ -558,20 +506,20 @@ class ApplicationStateService(
...
     }
 
-    private fun validateApplication(tx: Database.Read, application: ApplicationDetails) {
+    private fun validateApplication(tx: Database.Read, type: ApplicationType, application: ApplicationForm, strict: Boolean) {
         val result = ValidationResult()
 
-        val preferredStartDate = application.form.preferences.preferredStartDate
-        if (application.type == ApplicationType.PRESCHOOL && preferredStartDate != null) {
+        val preferredStartDate = application.preferences.preferredStartDate
+        if (type == ApplicationType.PRESCHOOL && preferredStartDate != null) {
```
reynders, line 513: Small style suggestions: extract each validation as own function, for example fun validatePreschoolPreferredStartDate(...).

reynders (approved):

service/src/main/kotlin/fi/espoo/evaka/application/ApplicationStateService.kt
Joosakur: I think this function needs refactoring when we do frontend validation, so I'll try to remember this then

PR #387, opened 2021-01-25
Move employee mobile frontend into own project

jaahon:
#### Summary

In short, I moved all the code related to employee mobile into it's own yarn workspace, I moved webpack dependencies to workspace root and removed unused storybook configs from employee-frontend.
Webpack dependency had to be moved to workspace root because it does not work with multiple instances of webpack. Not sure why adding the new project triggered the problems when both citizen frontend and employee frontend played nice previously.

Tested that everything works by deploying to dev.


Gekkio (approved):
Webpack in workspace root is not optimal, but maybe we can get rid of it later when all the plugins/loaders stabilize a bit more

PR #389, opened 2021-01-25
Citizen save as draft -terms of service info modal

reynders:
Screenshot 2021-01-25 at 16 02 20


frontend/packages/lib-components/src/molecules/modals/InfoModal.tsx
```diff
@@ -171,7 +169,7 @@ function InfoModal({
                       <Button
                         onClick={reject}
                         dataQa="modal-cancelBtn"
-                        text={rejectLabel ?? i18n.common.cancel}
+                        text={rejectLabel ?? 'Peruuta'}
```
Joosakur, line 172: Maybe these should now be mandatory props. Otherwise some Swedish-speaking user will eventually get furious.

Joosakur (changes requested):

Joosakur (approved):

PR #390, opened 2021-01-25
EVAKA: application typings to lib commons + only uppercase application status

Joosakur:


frontend/packages/lib-common/src/api-types/application/ApplicationDetails.ts
```diff
@@ -1,46 +1,14 @@
-// SPDX-FileCopyrightText: 2017-2021 City of Espoo
-//
-// SPDX-License-Identifier: LGPL-2.1-or-later
```
massproduction, line 3: Removed the license

Joosakur: the whole file should be gone now

massproduction (approved):

PR #392, opened 2021-01-26
Application editor collapsible sections

jaahon:
#### Summary

Collapsible sections for citizen application editor sections. There is already support for validation error indicators.

Also fix `AddButton` styles when `reversed` prop is on.


Joosakur (approved):
💎

PR #393, opened 2021-01-26
button style update

Joosakur:

### DIFF
--- a/frontend/packages/lib-components/src/atoms/buttons/InlineButton.tsx
+++ b/frontend/packages/lib-components/src/atoms/buttons/InlineButton.tsx
@@ -45,6 +45,7 @@ const StyledButton = styled.button`
 
   svg {
     margin-right: ${defaultMargins.xs};
+    font-size: 20px;
   }
 
   ${defaultButtonTextStyle}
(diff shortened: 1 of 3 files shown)


frontend/packages/lib-components/src/atoms/buttons/button-commons.tsx
```diff
@@ -7,10 +7,9 @@ import colors from '../../colors'
 export const defaultButtonTextStyle = `
   color: ${colors.primary};
   font-family: 'Open Sans', sans-serif;
-  font-size: 0.875em;
-  line-height: 1em;
+  font-size: 16px;
```
jaahon, line 10: Use `em` units.

Joosakur: So `1em`?

jaahon: Yeah

frontend/packages/lib-components/src/atoms/buttons/InlineButton.tsx
```diff
@@ -45,6 +45,7 @@ const StyledButton = styled.button`
 
   svg {
     margin-right: ${defaultMargins.xs};
+    font-size: 20px;
```
jaahon, line 48: This should probably be a relative unit as well so that the icon grows with text size.

jaahon (approved):

PR #398, opened 2021-01-27
Fix absence colors

jaahon:
#### Summary

The removal of `EspooColours` broke them because of misleading naming.


Joosakur (approved):
oops.. should we rename absenceColours to absenceTextColours?

jaahon:
> oops.. should we rename absenceColours to absenceTextColours?

They're not actually used anywhere so I just deleted them and left just the "background" colors.

PR #407, opened 2021-01-28
Evaka feat datepicker 2

massproduction:
#### Summary
Adds a new DatePicker component that uses strings instead of dates. This might be used to replace the current DatePicker that is renamed as DatePickerDeprecated.


Joosakur (approved):
I will use and improve this further in the application validation PR.

PR #416, opened 2021-02-02
Pass user locale as suomi.fi auth request param

Gekkio:

### DIFF
--- a/apigw/src/shared/routes/auth/saml/index.ts
+++ b/apigw/src/shared/routes/auth/saml/index.ts
@@ -14,6 +14,7 @@ import { logoutExpress, saveLogoutToken } from '../../../session'
 import { fromCallback } from '../../../promise-utils'
 import { getEmployees } from '../../../dev-api'
 import _ from 'lodash'
+import { AuthenticateOptions } from 'passport-saml'
 
 const urlencodedParser = urlencoded({ extended: false })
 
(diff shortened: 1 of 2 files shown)


mikkopiu (approved):
LGTM but did you find any official documentation on this? I managed to only find some random repos with the implementation for `locale`.

apigw/src/shared/routes/auth/saml/index.ts
```diff
@@ -62,8 +63,13 @@ function createLoginHandler({
       req,
       'Login endpoint called'
     )
+    const locale = req.query.locale
+    const options: AuthenticateOptions = {
+      additionalParams: typeof locale === 'string' ? { locale } : {}
```
mikkopiu, line 68: More of a Q: will an incorrect locale trigger an error in `passport.authenticate` or will it fail silently? Just wondering whether we should explicitly catch that or not.

Gekkio:
https://palveluhallinta.suomi.fi/fi/tuki/artikkelit/59116c3014bbb10001966f70

> Taulukko 4. Käyttöliittymän kieli, tunnistuspyyntö.

apigw/src/shared/routes/auth/saml/index.ts
Gekkio: A non-string locale (= not present, or uses one of the exotic query parameter formats supported by express) -> skipped
A locale unsupported by suomi.fi -> ???

PR #417, opened 2021-02-02
Varda fee data update consistency

jaahon:
#### Summary

Do Varda fee data updates before sending new fee data. This should prevent fee data from overlapping in Varda even temporarily. Also, log request url, method, body and response message when request to Varda fails.

### DIFF
--- a/service/src/main/kotlin/fi/espoo/evaka/varda/integration/VardaClient.kt
+++ b/service/src/main/kotlin/fi/espoo/evaka/varda/integration/VardaClient.kt
@@ -7,6 +7,7 @@ package fi.espoo.evaka.varda.integration
 import com.fasterxml.jackson.annotation.JsonIgnoreProperties
 import com.fasterxml.jackson.databind.ObjectMapper
 import com.fasterxml.jackson.module.kotlin.readValue
+import com.github.kittinunf.fuel.core.FuelError
 import com.github.kittinunf.fuel.core.FuelManager
 import com.github.kittinunf.fuel.core.Headers
 import com.github.kittinunf.fuel.core.Request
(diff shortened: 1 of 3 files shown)


service/src/main/kotlin/fi/espoo/evaka/varda/integration/VardaClient.kt
```diff
@@ -432,3 +389,9 @@ class VardaClient(
                 }
         }
 }
+
+private fun logRequestError(request: Request, error: FuelError) {
+    logger.error(error.exception) {
+        "Varda request failed (method=${request.method}, url=${request.url}, body=${request.body.asString("application/json")}, error=${error.errorData.decodeToString()})"
+    }
```
mikkopiu, line 396: These fields would be better placed in the `meta` field of the structured `app-misc` logs as they would then be searchable in Kibana as `meta.url:"/something"` instead of the slow free text search. It also avoids having to deal with possible line breaks in the `message` field (in both alerts and searches).

It looks like our `KLogger` extensions are missing a usable extension for `logger.error` _with exceptions_ but that could be like (in `service-lib/src/main/kotlin/fi/espoo/voltti/logging/loggers/AppMiscLoggers.kt`):

```kotlin
fun KLogger.error(args: Map<String, Any?>, error: Any, m: () -> Any?) {
    if (isErrorEnabled) error(m.toStringSafe(), StructuredArguments.entries(args), error)
}
```

and then you'd use it here as:

```suggestion
    logger.error("Varda request failed", mapOf("meta" to mapOf(
            "method" to request.method,
            "url" to request.url,
            "body" to request.body.asString("application/json"),
            "error" to error.errorData.decodeToString()
    )), error)
```

mikkopiu (approved):

PR #419, opened 2021-02-02
Fix duplicate application warning bug

mapehe:
#### Summary

Warn user when they are attempting to create a duplicate application.

### DIFF
--- a/frontend/packages/citizen-frontend/src/applications/api.ts
+++ b/frontend/packages/citizen-frontend/src/applications/api.ts
@@ -138,8 +138,8 @@ export async function getDuplicateApplications(
     `/citizen/applications/duplicates/${childId}`
   )
   return Object.fromEntries(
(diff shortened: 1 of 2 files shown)


frontend/packages/citizen-frontend/src/applications/api.ts
```diff
@@ -138,8 +138,8 @@ export async function getDuplicateApplications(
     `/citizen/applications/duplicates/${childId}`
   )
   return Object.fromEntries(
-    Object.entries(data).map(([type, value]) => [type.toLowerCase(), value])
-  ) as Record<ApplicationType, boolean> // FIXME
+    Object.entries(data).map(([type, value]) => [type, value])
+  ) as Record<ApplicationType, boolean>
 }
```
jaahon, line 143: This whole data mapping now does nothing and you can just `return data`.

jaahon (approved):

PR #420, opened 2021-02-02
Pending decision reminder email

reynders:
https://voltti.atlassian.net/browse/EVAKA-4116

Related infra: https://github.com/espoon-voltti/evaka-infra/pull/351

* Email for pending decision
* Async job for sending it to all guardians with pending decisions older than week, 2 times one week apart
* Scheduled api endpoint to trigger daily the above job
* Refactored the email client to be content agnostic


service/src/main/kotlin/fi/espoo/evaka/application/PendingDecisionEmailService.kt
```diff
@@ -0,0 +1,200 @@
...
+                                decisionId.toString(),
+                                guardian.email,
+                                fromAddress,
+                                getSubject(lang),
+                                getHtml(lang),
+                                getText(lang)
+                            )
+
+                            // Mark as sent
+                            db.transaction { tx ->
```
jaahon, line 101: I don't see a reason to open a nested transaction here.

service/src/main/kotlin/fi/espoo/evaka/application/PendingDecisionEmailService.kt
```diff
@@ -0,0 +1,200 @@
...
+        }
+
+        asyncJobRunner.scheduleImmediateRun()
+
+        return jobCount
+    }
+
+    fun sendPendingDecisionEmail(db: Database, decisionId: UUID) {
+        db.read { tx ->
+            getDecision(tx.handle, decisionId)?.let { decision ->
```
jaahon, line 84: All this nesting is quite unpleasant. You could just log and return if something is missing:
```
val decision = getDecision(tx.handle, decisionId) ?: run {
    logger.warn("Could not send pending decision email for decision $decisionId: decision missing")
    return
}
val application = fetchApplicationDetails(tx.handle, decision.applicationId) ?: run {
    logger.warn("Could not send pending decision email for decision ${decision.id}: application could not be found")
    return
}
...
```
or even just throw and wrap it all in a try catch.

service/src/main/kotlin/fi/espoo/evaka/application/PendingDecisionEmailService.kt
```diff
@@ -0,0 +1,200 @@
...
+
+            val pendingDecisions = tx.createQuery(
+                """
+SELECT DISTINCT(d.id)
+FROM decision d
+WHERE d.status = 'PENDING'
+AND d.resolved IS NULL
+AND d.sent_date + INTERVAL '1 week' < current_date
+AND d.pending_decision_emails_sent_count < 2
+AND d.pending_decision_email_sent IS NULL or d.pending_decision_email_sent < current_date - INTERVAL '1 week'
```
jaahon, line 53: I think if both of these date comparisons were the "same way" this would be a bit easier to understand, i.e. `column + INTERVAL '1 week' < current_date` or `column < current_date - INTERVAL '1 week'`.

service/src/main/kotlin/fi/espoo/evaka/application/PendingDecisionEmailService.kt
```diff
@@ -0,0 +1,200 @@
...
+class PendingDecisionEmailService(
+    private val asyncJobRunner: AsyncJobRunner,
+    private val emailClient: IEmailClient,
+    private val env: Environment
+) {
+    init {
+        asyncJobRunner.sendPendingDecisionEmail = ::doSendPendingDecisionsEmail
+    }
+
+    val fromAddress = env.getProperty("mail_reply_to_address", "no-reply.evaka@espoo.fi")
```
jaahon, line 33: You could just use `getRequiredProperty` and drop the default or drop the default config from `application.yaml`. It's now duplicated for no reason.

service/src/main/kotlin/fi/espoo/evaka/application/PendingDecisionEmailService.kt
reynders: Honestly not sure which way is cleaner, but as I also was amused by the callback hellish kinda code I reformatted it.

service/src/main/kotlin/fi/espoo/evaka/application/PendingDecisionEmailService.kt
```diff
@@ -0,0 +1,208 @@
...
+                        decisionId.toString(),
+                        guardian.email,
+                        fromAddress,
+                        getSubject(lang),
+                        getHtml(lang),
+                        getText(lang)
+                    )
+
+                    // Mark as sent
+                    tx.handle.createUpdate(
```
jaahon, line 111: You should use `db.transaction` so you can call `createUpdate` on the `Transaction` object directly, using the `handle` to circumvent that is a no-no.

service/src/main/kotlin/fi/espoo/evaka/application/PendingDecisionEmailService.kt
```diff
@@ -0,0 +1,208 @@
...
+
+        asyncJobRunner.scheduleImmediateRun()
+
+        return jobCount
+    }
+
+    fun sendPendingDecisionEmail(db: Database, decisionId: UUID) {
+        try {
+            db.read { tx ->
+                val decision = getDecision(tx.handle, decisionId) ?: run {
```
jaahon, line 85: If you just throw an exception `run` is unnecessary. You can just do `?: throw ...`.

service/src/main/kotlin/fi/espoo/evaka/application/PendingDecisionEmailService.kt
```diff
@@ -0,0 +1,208 @@
...
+    WHERE id = :id
+                        """.trimIndent()
+                    )
+                        .bind("id", decisionId)
+                        .execute()
+                } else {
+                    throw Exception("Could not send pending decision email to guardian ${guardian.id}: invalid email")
+                }
+            }
+        } catch (e: Exception) {
```
jaahon, line 124: Catching everything that extends `Exception` is kind of dangerous since really we only want to catch exceptions _we_ threw.

service/src/main/kotlin/fi/espoo/evaka/application/PendingDecisionEmailService.kt
jaahon: Please replace this with `tx.createUpdate`.

jaahon (approved):

PR #423, opened 2021-02-03
Make SSN nullable

reynders:

### DIFF
--- a/service/src/main/kotlin/fi/espoo/evaka/dvv/DvvModificationsService.kt
+++ b/service/src/main/kotlin/fi/espoo/evaka/dvv/DvvModificationsService.kt
@@ -181,15 +181,19 @@ class DvvModificationsService(
         caretakerLimitedDvvInfoGroup: CaretakerLimitedDvvInfoGroup
     ) {
         val user = AuthenticatedUser.machineUser
-        db.transaction { tx ->
-            personService.getOrCreatePerson(
-                tx,
-                user,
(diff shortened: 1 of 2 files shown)


service/src/main/kotlin/fi/espoo/evaka/dvv/DvvModificationsService.kt
```diff
@@ -181,15 +181,19 @@ class DvvModificationsService(
...
+                    tx,
+                    user,
+                    ExternalIdentifier.SSN.getInstance(caretakerLimitedDvvInfoGroup.huoltaja.henkilotunnus)
+                )
+            }?.let {
+                logger.info("Dvv modification ${caretakerLimitedDvvInfoGroup.muutosattribuutti} for ${it.id}: refreshing all caretaker/custodian info from DVV")
+                fridgeFamilyService.doVTJRefresh(db, VTJRefresh(it.id, user.id))
+            }
+        } else {
+            logger.info("Dvv modification ignored for caretaker: ssn is null")
```
mikkopiu, line 196: Is there any ID etc. available that could be logged here (if this ever comes up)?

mikkopiu (approved):

service/src/main/kotlin/fi/espoo/evaka/dvv/DvvModificationsService.kt
reynders: We have the caretaker ssn but we cannot log those

PR #426, opened 2021-02-03
EVAKA-4115: nav-headers, route, initial layout and auth changes for citizen map

Joosakur:


Gekkio (approved):
LGTM, but maybe some "map person" should review this too

PR #429, opened 2021-02-04
Different texts for different application types

mapehe:


frontend/packages/citizen-frontend/src/applications/editor/FeeSection.tsx
```diff
@@ -34,7 +37,7 @@ export default React.memo(function FeeSection({
     >
       <P
         dangerouslySetInnerHTML={{
```
mikkopiu, line 39: Looks like this is already a convention but to me this smells a bit: using `dangerouslySetInnerHTML` requires constant vigilance that only trusted sources are used (to avoid any possibility of user-provided HTML being set). Looking at the current translations that include HTML, most of them could just as well be small components -- which wouldn't require doing anything special while making any future accidents less likely.

Multi-line strings could be `\n` in translation + `white-space: pre-wrap` or even separate translations and handle the layout _in the layout_ instead of the translations. Obviously not as simple but a lot more robust.

I won't block for this as the current usage is safe enough but at least something to keep in mind, IMO.

frontend/packages/citizen-frontend/src/applications/editor/unit-preference/SiblingBasisSubSection.tsx
```diff
@@ -40,7 +51,12 @@ export default React.memo(function SiblingBasisSubSection({
             <>
               <Gap />
               <Label>
-                {t.applications.editor.unitPreference.siblingBasis.radioLabel} *
+                {
+                  t.applications.editor.unitPreference.siblingBasis.radioLabel[
+                    applicationType
+                  ]
+                }{' '}
+                *
```
mikkopiu, line 59: Supernit: why not include the asterisk in the string:
```suggestion
                }{' *'}
```

mikkopiu (approved):
OK but see comments

PR #430, opened 2021-02-04
EVAKA-4115: listing units

Joosakur:
#### Summary
list units, filter by language, improve layout

desktop:
Screenshot 2021-02-04 at 14 32 56

mobile:
Screenshot 2021-02-04 at 14 34 13

### DIFF
--- a/frontend/packages/citizen-frontend/src/App.tsx
+++ b/frontend/packages/citizen-frontend/src/App.tsx
@@ -7,7 +7,6 @@ import { BrowserRouter, Route, Switch } from 'react-router-dom'
 import { Localization } from '~localization'
 import Header from '~header/Header'
 import Decisions from '~decisions/decisions-page/Decisions'
-import Footer from '~Footer'
 import DecisionResponseList from '~decisions/decision-response-page/DecisionResponseList'
 import ApplicationEditor from '~applications/editor/ApplicationEditor'
 import GlobalErrorDialog from '~overlay/Error'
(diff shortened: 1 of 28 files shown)


frontend/packages/citizen-frontend/src/App.tsx
```diff
@@ -61,7 +60,6 @@ export default function App() {
               />
               <Route path="/" component={RedirectToEnduser} />
             </Switch>
-            <Footer />
```
mapehe, line 64: My first thought would be that this belongs here (but removing it is probably intentional and for a good reason).

frontend/packages/citizen-frontend/src/localization/en.ts
```diff
@@ -16,7 +16,22 @@ const en: Translations = {
         PRIVATE: 'Private service',
         MUNICIPAL_SCHOOL: 'Municipal',
         PRIVATE_SERVICE_VOUCHER: 'Private service (service voucher)'
+      },
+      careTypes: {
+        CLUB: 'Kerho',
```
mapehe, line 21: Should we make ticket about translations so that the are double checked at some point?

mapehe (approved):

frontend/packages/citizen-frontend/src/App.tsx
Joosakur: Yeah the issue is that the map page wants to render footer differently than other pages. Couldn't come up with a better way to solve that, but ideas are welcome

PR #432, opened 2021-02-04
EVAKA-3956 transfer application improvements

jaahon:
#### Summary

Show current (or future if none exist) placement unit in applications list and add an endpoint for a scheduled job that cancels all outdated transfer applications.

### DIFF
--- a/service/src/main/kotlin/fi/espoo/evaka/shared/controllers/ScheduledOperationController.kt
+++ b/service/src/main/kotlin/fi/espoo/evaka/shared/controllers/ScheduledOperationController.kt
@@ -5,6 +5,7 @@
 package fi.espoo.evaka.shared.controllers
 
 import fi.espoo.evaka.Audit
+import fi.espoo.evaka.application.cancelOutdatedTransferApplications
 import fi.espoo.evaka.application.removeOldDrafts
 import fi.espoo.evaka.attachment.AttachmentsController
 import fi.espoo.evaka.dvv.DvvModificationsBatchRefreshService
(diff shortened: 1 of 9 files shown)


service/src/main/kotlin/fi/espoo/evaka/shared/controllers/ScheduledOperationController.kt
```diff
@@ -63,6 +67,16 @@ class ScheduledOperationController(
...
 
+    @PostMapping("/application/cancel-outdated-transfer-applications")
+    fun cancelOutdatedTransferApplications(db: Database.Connection): ResponseEntity<Unit> {
+        Audit.ApplicationsCancelOutdatedTransferApplications.log()
+        val canceledApplications = db.transaction { it.cancelOutdatedTransferApplications() }
+        logger.info {
+            "Canceled ${canceledApplications.size} outdated transfer applications (ids: ${canceledApplications.joinToString(", ")})"
+        }
+        return ResponseEntity.noContent().build()
+    }
```
mikkopiu, line 78: Is there a likely risk that the rows this needs access to would be locked by another query for over 12s (the scheduling Lambda timeout, obviously adjustable)? If yes, this endpoint should only schedule an async job instead of doing the work immediately like e.g. the Varda update.

jaahon: I don't think it's likely, especially since this would be run sometime during midnight. I wanted to keep it simple and not add a new async job for it, but maybe I should anyway.

mikkopiu: I'm fine with going either way (the Koski and DVV jobs are a bit special, and slow, which is why I wanted them to be schedule-only) and this should be easy to fix later on if it becomes a problem.

mikkopiu (approved):
LGTM with comments

PR #435, opened 2021-02-05
EVAKA-3956 transfer application cancelling

jaahon:
#### Summary

Do not cancel transfer applications that are waiting for confirmation or that are already accepted or rejected. Also clean up any related placement plans and decision drafts.

### DIFF
--- a/service/src/integrationTest/kotlin/fi/espoo/evaka/shared/controllers/ScheduledOperationControllerIntegrationTest.kt
+++ b/service/src/integrationTest/kotlin/fi/espoo/evaka/shared/controllers/ScheduledOperationControllerIntegrationTest.kt
@@ -5,14 +5,17 @@
 package fi.espoo.evaka.shared.controllers
 
 import fi.espoo.evaka.FullApplicationTest
+import fi.espoo.evaka.application.ApplicationStateService
 import fi.espoo.evaka.application.ApplicationStatus
 import fi.espoo.evaka.application.ApplicationType
+import fi.espoo.evaka.application.DaycarePlacementPlan
 import fi.espoo.evaka.application.fetchApplicationDetails
 import fi.espoo.evaka.application.persistence.daycare.DaycareFormV0
 import fi.espoo.evaka.attachment.getApplicationAttachments
 import fi.espoo.evaka.insertApplication
 import fi.espoo.evaka.insertGeneralTestFixtures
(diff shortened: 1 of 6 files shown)


service/src/integrationTest/kotlin/fi/espoo/evaka/shared/controllers/ScheduledOperationControllerIntegrationTest.kt
```diff
@@ -242,16 +251,74 @@ class ScheduledOperationControllerIntegrationTest : FullApplicationTest() {
         assertEquals(ApplicationStatus.SENT, applicationStatus)
     }
 
+    @Test
+    fun `transfer application cancelling cleans up placement plans and decision drafts`() {
+        val preferredStartDate = LocalDate.now()
```
Gekkio, line 256: Is it possible to avoid `LocalDate.now()` here or do we depend on today's date in some of the other test data already?

service/src/main/kotlin/fi/espoo/evaka/application/ApplicationQueries.kt
```diff
@@ -857,7 +857,10 @@ fun removeOldDrafts(db: Database.Transaction, deleteAttachment: (db: Database.Tr
 fun Database.Transaction.cancelOutdatedTransferApplications(): List<UUID> = createUpdate(
     """
 UPDATE application SET status = :cancelled
-WHERE transferapplication AND NOT EXISTS (
+WHERE transferapplication
+-- only include applications that don't have decisions
```
Gekkio, line 861: Comments in SQL are problematic with the current JDBI version because its parser does not understand them...it might work or might not work, but I recommend not having any comments. I don't remember the exact details but we had one at some point a query that did a wrong thing because of some SQL comments that confused the parser.

JDBI 3.18.1 is supposed to fix this...

service/src/integrationTest/kotlin/fi/espoo/evaka/shared/controllers/ScheduledOperationControllerIntegrationTest.kt
jaahon: The behavior depends on the current date, there's no dependency on other temporal data.

Gekkio (approved):

PR #437, opened 2021-02-08
EVAKA-4103 translate verification view

hruoho:
#### Summary

Add translations to verification view

### DIFF
--- a/frontend/packages/citizen-frontend/src/localization/en.ts
+++ b/frontend/packages/citizen-frontend/src/localization/en.ts
@@ -147,8 +147,8 @@ const en: Translations = {
       verification: {
         title: {
           DAYCARE: 'Verifying the application',
-          PRESCHOOL: 'Esiopetushakemuksen tarkistaminen',
-          CLUB: 'Kerhohakemuksen tarkistaminen'
+          PRESCHOOL: 'Verifying the application',
+          CLUB: 'Verifying the application'
         },
         notYetSent:
           '<strong>The application has not yet been sent.</strong> Please check the information you provided and send the application by clicking "Send the application" button in the end of the form.',
@@ -158,16 +158,16 @@ const en: Translations = {
           modified: 'Last modified'
         },
         attachmentBox: {
-          nb: 'Huom!',
+          nb: 'Please note!',
           headline:
-            'Jos lisäät liitteet seuraaviin kohtiin sähköisesti, hakemuksesi käsitellään nopeammin, sillä käsittelyaika alkaa liitteiden saapumisesta.',
-          urgency: 'Hakemus on kiireellinen',
-          shiftCare: 'Ilta- ja vuorohoito',
-          goBackLinkText: 'Palaa takaisin hakemusnäkymään',
-          goBackRestText: 'lisätäksesi liitteet hakemukseen.'
+            ' If you add attachments to the following sections online, your application will be processed more quickly as the processing period will begin once we have received the required attachments.',
+          urgency: 'Application is urgent',
+          shiftCare: 'Evening and round-the-clock care',
+          goBackLinkText: 'Return back to application',
+          goBackRestText: ' to add the attachments.'
         },
         serviceNeed: {
-          title: 'Palveluntarve',
+          title: 'Service need',
           wasOnDaycare: 'Varhaiskasvatuksessa ennen kerhoa',
           wasOnDaycareYes:
             'Lapsi, jolle haetaan kerhopaikkaa, on varhaiskasvatuksessa ennen kerhon toivottua aloituspäivää.',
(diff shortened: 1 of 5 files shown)


frontend/packages/citizen-frontend/src/localization/en.ts
```diff
@@ -158,66 +158,66 @@ const en: Translations = {
...
         attachmentBox: {
-          nb: 'Huom!',
+          nb: 'Please note!',
           headline:
-            'Jos lisäät liitteet seuraaviin kohtiin sähköisesti, hakemuksesi käsitellään nopeammin, sillä käsittelyaika alkaa liitteiden saapumisesta.',
-          urgency: 'Hakemus on kiireellinen',
-          shiftCare: 'Ilta- ja vuorohoito',
-          goBackLinkText: 'Palaa takaisin hakemusnäkymään',
-          goBackRestText: 'lisätäksesi liitteet hakemukseen.'
+            ' If you add attachments to the following sections online, your application will be processed more quickly as the processing period will begin once we have received the required attachments.',
```
mikkopiu, line 163: Extra whitespace in the beginning of the string, intentional?
EDIT: same for the rest of these with extra whitespace; would it not be better handled in layout instead of the translations?

frontend/packages/citizen-frontend/src/localization/en.ts
```diff
@@ -158,66 +158,66 @@ const en: Translations = {
           modified: 'Last modified'
         },
         attachmentBox: {
-          nb: 'Huom!',
+          nb: 'Please note!',
```
mikkopiu, line 161: Would just "Note!" be better as it uses a similar space to both "Huom!" and "Obs!"? Or is extra clarity preferable in this context?

frontend/packages/citizen-frontend/src/localization/en.ts
```diff
@@ -236,63 +236,66 @@ const en: Translations = {
...
         contactInfo: {
-          title: 'Henkilötiedot',
+          title: 'Personal information',
           child: {
-            title: 'Lapsen tiedot',
-            name: 'Lapsen nimi',
-            ssn: 'Lapsen henkilötunnus',
-            streetAddress: 'Kotiosoite',
-            isAddressChanging: 'Osoite muuttunut / muuttumassa',
+            title: 'Child´s information',
```
mikkopiu, line 241: Incorrect punctuation, should be:
```suggestion
            title: "Child’s information",
```

frontend/packages/citizen-frontend/src/localization/en.ts
```diff
@@ -236,63 +236,66 @@ const en: Translations = {
...
+            name: 'Child’s name',
+            ssn: 'Child’s personal identity code',
```
mikkopiu, line 243: Aren't both of these "apostrophes" actually ["Right Single Quotation Marks"](https://www.codetable.net/decimal/8217), i.e. should be:
```suggestion
            name: "Child's name",
            ssn: "Child's personal identity code",
```

mikkopiu (changes requested):

frontend/packages/citizen-frontend/src/localization/en.ts
mikkopiu: Actually, I remembered this whole ordeal wrong: right single quotation mark is the "preferred apostrophe" in Unicode but keyboards usually only have the ASCII-compatible straight apostrophe -- so leaving this one as-is is actually more correct.

frontend/packages/citizen-frontend/src/localization/en.ts
hruoho: I dunno, this originated from the old vue translations. But I guess we could use the shorter form here.

frontend/packages/citizen-frontend/src/localization/en.ts
hruoho: Sure it would! I'd go for `&nbsp;` but do you see a better way?

mikkopiu: If it doesn't explicitly have to be non-breaking, there's always `{' '}`

hruoho: OK perhaps we would be better off with `{' '}`

frontend/packages/citizen-frontend/src/localization/en.ts
mikkopiu: Seems like you missed this one

frontend/packages/citizen-frontend/src/localization/en.ts
mikkopiu: Seems like you missed these (unless you decided to go with this apostrophe in which case you should update the rest to match)

hruoho: Let's go with the single quote for simplicity

mikkopiu (approved):
:+1: and sry for the hassle

PR #438, opened 2021-02-08
Cleanup Varda and Koski error logs, KLogger extensions

mikkopiu:
#### Summary
- Log Varda and Koski error metadata correctly
    - #417 added metadata to Varda logs but the missing `mapOf("meta" to meta)` was missed, resulting in logs where the meta-fields were actually top-level fields
        - The KLogger extensions don't actually map directly to the meta-field in our JSON logs but need an extra mapOf(meta to arg) wrapped around them
    - Unify Koski error logging with Varda and instead of outputting the whole multi-line error response, log the relevant fields in "meta"
- As a "lesson learned" from the above, map `KLogger` meta-arguments directly to meta-fields
    - There isn't a use case for overwriting any other fields than the freeform "meta" (as everything else is provided by the loggers and/or is structured data)
    - Handle the extra meta-mapping inside the logger extensions and rename the field (`args -> meta`) to better reflect its use case
    - Add documentation for logger extensions existence to make them more widely known
- service-lib: test for error logger with exception
   - New logger method was missing tests
   - It's not a great test but more of a sanity check

Snippet of incorrect log output:
```jsonc
    "logLevel": "ERROR",
    "hostIp": "10.230.25.167",
    "exception": "BubbleFuelError",
    "url": "https://backend.varda-db.csc.fi/api/v1/maksutiedot/", // <-- This shouldn't be here
    "version": 1
  }
```

and correct:

```jsonc
    "logLevel": "ERROR",
    "hostIp": "10.230.25.167",
    "exception": "BubbleFuelError",
    "meta": {
        "url": "https://backend.varda-db.csc.fi/api/v1/maksutiedot/", // <-- But here instead
        // ...
    }
    "version": 1
  }
```

### DIFF
--- a/service-lib/src/main/kotlin/fi/espoo/voltti/logging/loggers/AppMiscLoggers.kt
+++ b/service-lib/src/main/kotlin/fi/espoo/voltti/logging/loggers/AppMiscLoggers.kt
@@ -8,30 +8,32 @@ import mu.KLogger
 import net.logstash.logback.argument.StructuredArguments
 
 /**
- * Extensions to allow adding meta-fields (and technically any other) to app-misc logs
- * without requiring StructuredArguments wrapper and a dependency to net.logstash.logback.
+ * Extensions to allow adding meta-fields to app-misc logs without requiring StructuredArguments wrapper
+ * and an unnecessary dependency to net.logstash.logback.
  */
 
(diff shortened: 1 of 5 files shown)


jaahon (approved):

mikkopiu:
Fixed tests, missing String conversion

PR #443, opened 2021-02-09
EVAKA-4103 indicate transfer applications

hruoho:
#### Summary

For transfer applications, show piece of text next to the title on application list & editor

### DIFF
--- a/frontend/packages/citizen-frontend/src/applications/ApplicationCreation.tsx
+++ b/frontend/packages/citizen-frontend/src/applications/ApplicationCreation.tsx
@@ -114,6 +114,7 @@ export default React.memo(function ApplicationCreation() {
               <Gap size="L" />
               <AlertBox
                 thin
+                data-qa={'duplicate-application-notification'}
                 message={t.applications.creation.duplicateWarning}
               />
             </>
@@ -123,6 +124,7 @@ export default React.memo(function ApplicationCreation() {
               <Gap size="L" />
               <InfoBox
                 thin
+                data-qa={'transfer-application-notification'}
                 message={
                   t.applications.creation.transferApplicationInfo[
                     selectedType === 'DAYCARE' ? 'DAYCARE' : 'PRESCHOOL'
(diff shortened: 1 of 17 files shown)


jaahon (commented):
Should the transfer application part be in lower case? I forgot, but there is an info box when creating a transfer application. That should probably added as well.

Also, clean up the git history.

image

frontend/packages/citizen-frontend/src/applications/ApplicationCreation.tsx
```diff
@@ -98,6 +115,19 @@ export default React.memo(function ApplicationCreation() {
               />
             </>
           ) : null}
+          {!duplicateExists && shouldUseTransferApplication && (
+            <>
+              <Gap size="L" />
+              <AlertBox
```
jaahon, line 121: This should be an `InfoBox`.

service/src/main/kotlin/fi/espoo/evaka/application/ApplicationQueries.kt
```diff
@@ -115,6 +115,35 @@ fun duplicateApplicationExists(
         .isNotEmpty()
 }
 
+fun activePlacementExists(
+    h: Handle,
+    childId: UUID,
+    type: ApplicationType
+): Boolean {
+    // language=sql
```
jaahon, line 123: Move this to correct place or just remove it.

service/src/main/kotlin/fi/espoo/evaka/application/ApplicationQueries.kt
```diff
@@ -115,6 +115,35 @@ fun duplicateApplicationExists(
         .isNotEmpty()
 }
 
+fun activePlacementExists(
+    h: Handle,
```
jaahon, line 119: You shouldn't use `Handle` directly anymore, prefer `Database.Read` and `Database.Transaction`.

service/src/main/kotlin/fi/espoo/evaka/application/ApplicationQueries.kt
```diff
@@ -115,6 +115,35 @@ fun duplicateApplicationExists(
...
+            SELECT 1
+            FROM placement
+            WHERE 
+                child_id = :childId AND
+                type IN (<types>) AND 
+                current_date <= end_date
+        """.trimIndent()
+    return h.createQuery(sql)
+        .bind("childId", childId)
+        .bindList("types", placementTypes)
```
jaahon, line 141: The `bindList` method has some quirks so regular `bind` with an array is preferable, you'll have to change the condition to `type = ANY(:types)` as well.

jaahon (approved):
🏆

PR #445, opened 2021-02-09
Round the clock flag

mapehe:
Summary:

Re-enable the legacy round the clock checkbox from unit editor for front end filtering purposes.


Joosakur (approved):
Could also add the field to `PublicUnit` class and to relevant queries returning that

Joosakur (approved):
Could also add the field to `PublicUnit` class and to relevant queries returning that

mapehe:
Thanks for the feedback @Joosakur. Updated accordingly!

PR #449, opened 2021-02-10
EVAKA-4135: Restrict certain access rights from service worker

kustaahuhtala:
#### Summary

 Service worker should not be allowed to
 - Create or modify groups in a daycare
 - Place children into groups (or transfer them within groups)
 - Retrieve family contact details report
 - Add or edit service needs

 All these actions are unit supervisor's duty. However, service worker is allowed to list groups,
 children in groups, service need etc.

 For some endpoints, there was even wider set of roles allowed, cleaned those as well.

### DIFF
--- a/frontend/packages/employee-frontend/src/components/unit/tab-groups/Groups.tsx
+++ b/frontend/packages/employee-frontend/src/components/unit/tab-groups/Groups.tsx
@@ -159,7 +159,7 @@ export default React.memo(function Groups({
             dataQa="toggle-all-groups-collapsible"
           />
         </TitleContainer>
-        {requireRole(roles, 'ADMIN', 'SERVICE_WORKER', 'UNIT_SUPERVISOR') && (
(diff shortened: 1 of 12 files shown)


frontend/packages/employee-frontend/src/components/unit/tab-groups/Groups.tsx
```diff
@@ -159,7 +159,7 @@ export default React.memo(function Groups({
             dataQa="toggle-all-groups-collapsible"
           />
         </TitleContainer>
-        {requireRole(roles, 'ADMIN', 'SERVICE_WORKER', 'UNIT_SUPERVISOR') && (
+        {canManageGroups && (
```
jaahon, line 162: Using `canManageGroups` here might not work in the future, but since the endpoint for the data is secured separately, it's not very risky.

jaahon (approved):
💪

PR #454, opened 2021-02-11
E2E test input wrappers

jaahon:
#### Summary

* Use a `Checkbox` wrapper for checkbox and radio inputs in e2e tests. The wrapper makes it easy to click the correct visible element and also check the `checked` status of the invisible input.
* Create a function for picking an element from a react select component.
* Add missing props to the `Select` component in group placement modal.
* Remove unused e2e test code that was previously used with the old "admin frontend".
* Make a timeout shorter since the change is instantaneous anyway.


Joosakur (approved):
🚗 💨

PR #455, opened 2021-02-11
EVAKA-4134: add care area to starting placements report

Joosakur:


kustaahuhtala (commented):
This is a minor, but an annoying this that is bugging in other places too: When changing month range, the selected care area is not persisted, but always reverts back to "Kaikki".

Joosakur:
> This is a minor, but an annoying this that is bugging in other places too: When changing month range, the selected care area is not persisted, but always reverts back to "Kaikki".

That's because it's a frontend filter where the options are derived from results, which change when changing the month. Not sure if the issue is important enough to fix by loading area options from api 🤷‍♂️

kustaahuhtala (approved):

PR #457, opened 2021-02-11
Use placement type only for daycare occupancy

jaahon:
#### Summary

This changes the coefficients to what they were before. Part week children caused the occupancy percentages to be off.


mikkopiu (approved):

hruoho (approved):
Cheers man! I think it would be worthwhile to go through this occupancy thing again with dev, just to spread the knowledge.

PR #459, opened 2021-02-12
EVAKA-HOTFIX update unit filter list when changing the area filter

massproduction:
#### Summary

Updates the unit list when selecting an area. This is to help user not to select an area/unit combination that does not exist.


jaahon (changes requested):
Instead of adding a new `useEffect` hook you should just take `area` into account in the hook below.

massproduction:
> Instead of adding a new `useEffect` hook you should just take `area` into account in the hook below.

Fixed.

jaahon (approved):
🏅

PR #461, opened 2021-02-12
EVAKA-4103: include quick selection children in verification view

Joosakur:


jaahon (commented):
The verification view seems to work correctly now, but there is some kind of bug when marking vtj siblings and adding more children as fridge children. Could you look into it now as well?

When I made an application for Jari-Petteri and chose Veera as a sibling and added another child, the additional child does not get saved to the draft nor are they shown in the verification view.

jaahon (approved):
You can merge this, the form bugs can be fixed later.

PR #462, opened 2021-02-12
lib-components: fix AsyncButton spinner clipping

mikkopiu:
#### Summary
- The actual height of the spinner is `height` + (`border-width` x 2), i.e. not 24px (which is the container's height/width) but 24px + 2px + 2px = 28px
- Instead of increasing the containers size and therefore altering the other elements of the button (namely the text/content), reduce spinner size by 4px so it actually is 24px

Before:

async-button-before

After:

async-button-after


jaahon (approved):
I wonder when this broke 🤔

PR #464, opened 2021-02-12
[EVAKA-4115] Map

Gekkio:
#### Summary

* add a map that uses `react-leaflet`
* don't send empty graphql requests in map view unit list
* E2E tests will be added in a later pull request


Joosakur (approved):
🎖️

PR #465, opened 2021-02-15
UI polishing

reynders:
#### Summary
- Make modals fit mobile screens
- Split "remove application" link on another line on mobile
- Do not show Gap on mobile
- Fix english mailto address
- Make "you have 1 pending applications" spacing bigger
- Fix application status icon & text left spacing

### DIFF
--- a/frontend/packages/lib-components/src/layout/flex-helpers.ts
+++ b/frontend/packages/lib-components/src/layout/flex-helpers.ts
@@ -11,14 +11,14 @@ interface FixedSpaceRowProps {
   justifyContent?: Property.JustifyContent
   alignItems?: Property.AlignItems
   marginBottom?: SpacingSize | string
-  wrap?: boolean
+  flexWrap?: boolean
 }
 export const FixedSpaceRow = styled.div<FixedSpaceRowProps>`
   display: flex;
   flex-direction: row;
   ${(p) => (p.justifyContent ? `justify-content: ${p.justifyContent};` : '')}
   ${(p) => (p.alignItems ? `align-items: ${p.alignItems};` : '')}
-  ${(p) => (p.wrap ? 'flex-wrap: wrap;' : '')}
+  ${(p) => (p.flexWrap ? 'flex-wrap: wrap;' : '')}
 
   ${(p) =>
     p.marginBottom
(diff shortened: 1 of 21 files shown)


frontend/packages/lib-components/src/white-space.ts
```diff
@@ -47,10 +47,16 @@ export const defaultMargins: Record<SpacingSize, string> = {
 type GapProps = {
   horizontal?: boolean
   size?: SpacingSize
+  hideOnMobile?: boolean
```
Joosakur, line 50: I'm wondering if `sizeMobile?: SpacingSize` would be more useful, given that `'zero'` is a valid `SpacingSize` and sometimes you want the gap to be smaller/bigger on mobile, not just to hide it.

frontend/packages/lib-components/src/white-space.ts
```diff
@@ -47,10 +47,16 @@ export const defaultMargins: Record<SpacingSize, string> = {
...
 }
 
 export const Gap = styled.div<GapProps>`
   display: ${(p) => (p.horizontal ? 'inline-block' : 'block')};
   ${(p) => (p.horizontal ? 'width' : 'height')}: ${(p) =>
     defaultMargins[p.size || 'm']};
+
+  @media (max-width: 600px) {
```
Joosakur, line 58: should we import the breakpoint from constants? it was `tabletMin` or something

frontend/packages/lib-components/src/layout/flex-helpers.ts
```diff
@@ -11,14 +11,14 @@ interface FixedSpaceRowProps {
   justifyContent?: Property.JustifyContent
   alignItems?: Property.AlignItems
   marginBottom?: SpacingSize | string
-  wrap?: boolean
+  flexWrap?: boolean
```
Joosakur, line 14: Oh we already have wrap prop here.. at some point we should probably remove it since I don't think it really works as intended. `FixedSpaceFlexWrap` and `AdaptiveFlex` are better options although also imperfect.

Joosakur: But no need to do anything now

Joosakur (approved):
Pre-approved with a couple minor comments

PR #466, opened 2021-02-15
Evaka 4103 remove old datepicker

massproduction:
#### Summary
Changed the now deprecated old datepicker to the new datepicker when accepting a decision.

### DIFF
--- a/frontend/packages/citizen-frontend/src/decisions/decision-response-page/DecisionResponse.tsx
+++ b/frontend/packages/citizen-frontend/src/decisions/decision-response-page/DecisionResponse.tsx
@@ -4,7 +4,7 @@
 
 import { Decision } from '~decisions/types'
 import React, { useContext, useState } from 'react'
-import { useTranslation } from '~localization'
+import { useLang, useTranslation } from '~localization'
 import LocalDate from '@evaka/lib-common/src/local-date'
 import { OverlayContext } from '~overlay/state'
 import { H2, H3, Label, P } from '@evaka/lib-components/src/typography'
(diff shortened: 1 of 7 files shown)


reynders (approved):

frontend/packages/citizen-frontend/src/decisions/decision-response-page/DecisionResponse.tsx
```diff
@@ -134,16 +136,15 @@ export default React.memo(function DecisionResponse({
...
-                type="short"
-                minDate={startDate}
-                maxDate={
-                  ['PRESCHOOL', 'PREPARATORY_EDUCATION'].includes(decisionType)
-                    ? startDate
-                    : startDate.addDays(14)
+              <DatePicker
+                date={requestedStartDate.format()}
+                onChange={(date: string) =>
+                  setRequestedStartDate(LocalDate.parseFi(date))
```
Joosakur, line 142: I don't think this is ok. The type must be string all the way in state.

Joosakur (changes requested):

frontend/packages/citizen-frontend/src/decisions/decision-response-page/DecisionResponse.tsx
Joosakur: Also there's another ticket that for preschool/preparatory we should not use date picker since changing the date is not allowed anyway. It might make sense to implement that at the same time.

massproduction: Changed the startDate and endDate to be strings all the way from the `deserializeDecision` function. Also changed the startDate to be a string on the page in `PRESCHOOL` and `PREPARATORY_EDUCATION` cases.

Joosakur: great! I'll test tomorrow

Joosakur: Does't quite work, tested with connected daycare

1. In datepicker all dates look disabled whether valid or not.
2. In datepicker any date can be selected whether valid or not.
3. Any date can be written (as should be), but no validation warning is shown if the date is outside the accepted range, nor is submit button disabled.

massproduction: Fixed issues with users being able to select non-valid dates and added warnings on disallowed dates.

Joosakur (approved):
Now it works nicely! Only one minor issue which happens here (not in scope of this PR) is that after opening a date picker near the bottom of the screen you need to scroll down to see it, since it always opens downwards.

PR #469, opened 2021-02-16
Filter empty other children on submit citizen application

mapehe:
Summary:

Prevent form check error if the user selects `Children under 18 living in the same address` and immediately deselects.

### DIFF
--- a/frontend/packages/citizen-frontend/src/applications/editor/validations.ts
+++ b/frontend/packages/citizen-frontend/src/applications/editor/validations.ts
@@ -221,7 +221,7 @@ export const validateApplication = (
       otherChildren: {
         arrayErrors: undefined,
         itemErrors: form.contactInfo.otherChildren.map((child) =>
-          requireFullFamily
+          requireFullFamily && form.contactInfo.otherChildrenExists
             ? {
                 firstName: validate(child.firstName, required),
                 lastName: validate(child.lastName, required),


frontend/packages/citizen-frontend/src/applications/editor/contact-info/OtherChildrenSubSection.tsx
```diff
@@ -64,15 +64,26 @@ export default React.memo(function OtherChildrenSubSection({
...
-                  firstName: '',
-                  lastName: '',
-                  socialSecurityNumber: ''
-                }
-              ]
+              otherChildren: formData.otherChildren.filter(
+                (child) =>
+                  child.firstName.length > 0 ||
+                  child.lastName.length > 0 ||
+                  child.socialSecurityNumber.length > 0
```
Joosakur, line 85: If you fill one of these fields and then uncheck the same issue will still happen? Should there be `&&` instead of `||`?

Joosakur: Although still there would be case where you could fill in invalid ssn and have it fail while hidden. So actually the solution is not to remove rows when unchhecking, but rather edit the validation function so that it only validates those children if `otherChildrenExists === true`. That is one of the reasons why the checkbox boolean is part of the FormData even though it is not sent to backend.

Joosakur (changes requested):

frontend/packages/citizen-frontend/src/applications/editor/contact-info/OtherChildrenSubSection.tsx
mapehe: Thanks for the feedback @Joosakur. I'll make changes accordingly.

Joosakur (approved):
well that became a small change 😄

mapehe:
> well that became a small change 😄

The simpler the better!

PR #471, opened 2021-02-16
[EVAKA-4138] Fix attachment downloads with pre-download checks

mikkopiu:
#### Summary

- Create `FileDownloadButton` for use in file downloads that might not be immediately available (e.g. attachments)
  - Use in all frontends for attachment downloads to ensure that attachment downloads during virus scanning don't lead to ugly failures/exceptions
  - As file download APIs are implementation specific, the component only takes the fetch and error handlers as function inputs
  - Not terribly satisfied with the component UI/UX-wise but decided to just follow the existing styles for now (button is styled as a link)
  - Use component in all frontends; fixes attachment downloads in citizen application summary view
- Replace attachments' explicit pre-download checks with a 404 response and error handling in the new component
- Attachment download: prefer IDs to `FileObject`, more shared code
- Auto-update year in license helper tool
   - Default to "2017-$CURRENT_YEAR" and allow overriding with `REUSE_YEARS` environment variable
   - Not totally sure if this is the correct format (or whether it should just be "2021" for any new files) but lets start with something

#### About tests

Unfortunately none of the existing S3 mock services support bucket policies -> really difficult to test the "in progress" error handling cases this actually solves without creating complex mock APIs -> decided to just create E2E tests for happy paths and stories for component error cases to have at least something.


frontend/packages/lib-components/src/molecules/FileDownloadButton.tsx
```diff
@@ -0,0 +1,62 @@
...
+import { Attachment } from '@evaka/lib-common/src/api-types/application/ApplicationDetails'
+import { espooBrandColors } from '@evaka/lib-components/src/colors'
+import { UUID } from '@evaka/lib-common/src/types'
+
+const DownloadButton = styled.button`
+  border: none;
+  text-align: start;
+  padding: 0;
+  background: none;
+  color: ${espooBrandColors.espooTurquoise};
```
jaahon, line 19: This color should be `blueColors.medium` if the idea is to make it look like an `a` element.

frontend/e2e-test/test/e2e/specs/0_citizen/citizen-daycare-application-create.spec.ts
```diff
@@ -26,17 +27,35 @@ import {
...
+)
+
+const testFilePath = '../../assets/test_file.jpg'
+const testFileName = path.basename(testFilePath)
+const testFileDispositionRegex = new RegExp(
+  `attachment;\\s*filename=${testFileName.replace(
+    /[|\\{}()[\]^$+*?.]/g,
+    '\\$&' // Escape characters in filename to make a valid regular expression
+  )}`
+)
```
jaahon, line 52: Instead of this pretty complicated regex would a simple string comparison or a `String.includes` be enough?

jaahon (commented):
Good stuff! Just a couple minor things.

frontend/packages/lib-components/src/molecules/FileDownloadButton.tsx
mikkopiu: :+1: Not sure where I picked this one from.

frontend/e2e-test/test/e2e/specs/0_citizen/citizen-daycare-application-create.spec.ts
mikkopiu: Yeah, sounds reasonable -- I have a bit of a blind spot for complicated regexes.

jaahon (approved):

PR #493, opened 2021-02-22
Disable pending decision email sending for testing

reynders:
Disable pending decision email sending by default to allow production dry run testing


service/src/main/kotlin/fi/espoo/evaka/application/PendingDecisionEmailService.kt
```diff
@@ -137,10 +144,12 @@ WHERE id = :id
     }
 
     private fun getSubject(language: Language): String {
+        val postfix = if (System.getenv("VOLTTI_ENV") == "staging") " [staging]" else ""
```
Joosakur, line 147: or maybe `val postfix = System.getenv("VOLTTI_ENV").let { if(it == "prod") "" else " [$it]" }`

Joosakur (approved):

PR #497, opened 2021-02-22
Debounce digitransit autocomplete requests

mapehe:
#### Summary

Avoid spamming digitransit API with autocomplete requests on each keystroke by introducing a timeout between the requests.


reynders (approved):
Tested locally as ok

mapehe:
Thanks for the comment @reynders. @Joosakur pointed out that `useDebounce` would probably belong to `lib-commons`. Added a small refactoring commit where I moved it there.

PR #510, opened 2021-02-23
EVAKA-3952: citizen reads bulletins

Joosakur:
#### Summary
Desktop:
Screenshot 2021-02-24 at 11 41 07

Mobile 1:
Screenshot 2021-02-24 at 11 44 58

Mobile 2:
Screenshot 2021-02-24 at 11 45 14

### DIFF
--- a/frontend/packages/citizen-frontend/src/messages/MessageListItem.tsx
+++ b/frontend/packages/citizen-frontend/src/messages/MessageListItem.tsx
@@ -59,7 +59,7 @@ const Container = styled.div<{ isRead: boolean; active: boolean }>`
   `
       : ''}
 
-  ${(p) => (p.active ? `background-color: #E9F5FF;` : '')}
+  ${(p) => (p.active ? `background-color: ${colors.brandEspoo.espooTurquoiseLight};` : '')}
 `
 
 const Header = styled.div`
--- a/frontend/packages/citizen-frontend/src/messages/MessageListItem.tsx
+++ b/frontend/packages/citizen-frontend/src/messages/MessageListItem.tsx
@@ -16,7 +16,12 @@ type Props = {
   active: boolean
   onClick: () => void
 }
-function MessageListItem({ bulletin, active, onClick }: Props) {
+
+export default React.memo(function MessageListItem({
+  bulletin,
+  active,
+  onClick
+}: Props) {
   return (
     <Container
       isRead={bulletin.isRead}
@@ -44,7 +49,7 @@ function MessageListItem({ bulletin, active, onClick }: Props) {
       </FixedSpaceColumn>
     </Container>
   )
-}
+})
 
 const Container = styled.div<{ isRead: boolean; active: boolean }>`
   background-color: ${colors.greyscale.white};
@@ -84,5 +89,3 @@ const ContentSummary = styled.div`
   overflow: hidden;
   text-overflow: ellipsis;
 `
-
-export default React.memo(MessageListItem)
--- a/frontend/packages/citizen-frontend/src/messages/MessageListItem.tsx
+++ b/frontend/packages/citizen-frontend/src/messages/MessageListItem.tsx
@@ -1,0 +1,58 @@
+import React from 'react'
+
+import styled from 'styled-components'
+import colors from '@evaka/lib-components/src/colors'
+import { defaultMargins } from '@evaka/lib-components/src/white-space'
+import { ReceivedBulletin } from '~messages/types'
+import { formatDate } from '~util'
+import { FixedSpaceColumn } from '@evaka/lib-components/src/layout/flex-helpers'
+
+type Props = {
+  bulletin: ReceivedBulletin
+  onClick: () => void
+}
+function MessageListItem({ bulletin, onClick }: Props) {
+  return (
+    <Container onClick={onClick} isRead={bulletin.isRead}>
+      <FixedSpaceColumn>
+        <Header>
+          <span>{bulletin.sender}</span>
+          <span>{formatDate(bulletin.sentAt, 'dd.MM')}</span>
+        </Header>
+        <Title>{bulletin.title}</Title>
+        <ContentSummary>{bulletin.content.split('\n')[0]}</ContentSummary>
+      </FixedSpaceColumn>
+    </Container>
+  )
+}
+
+const Container = styled.div<{ isRead: boolean }>`
+  background-color: ${colors.greyscale.white};
+  padding: ${defaultMargins.m};
+
+  border: 1px solid ${colors.greyscale.lighter};
+
+  ${(p) =>
+    !p.isRead
+      ? `
+    border-left-color: ${colors.brandEspoo.espooTurquoise};
+    border-left-width: 6px;
+  `
+      : ''}
+`
+
+const Header = styled.div`
+  display: flex;
+  justify-content: space-between;
+  font-weight: 600;
+`
+
+const Title = styled.div``
+
+const ContentSummary = styled.div`
+  white-space: nowrap;
+  overflow: hidden;
+  text-overflow: ellipsis;
+`
+
+export default React.memo(MessageListItem)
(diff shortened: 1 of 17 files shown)


frontend/packages/citizen-frontend/src/messages/MessageListItem.tsx
```diff
@@ -0,0 +1,85 @@
...
+  onClick: () => void
+}
+function MessageListItem({ bulletin, active, onClick }: Props) {
+  return (
+    <Container
+      isRead={bulletin.isRead}
+      active={active}
+      onClick={onClick}
+      tabIndex={0}
+      onKeyPress={(e) => e.key === 'Enter' && onClick()}
```
jaahon, line 23: I don't know if just using a `button` instead of manually setting these whould be a better idea in terms of accessibility. Seems kind of weird in a way, though.

frontend/packages/citizen-frontend/src/messages/MessageReadView.tsx
```diff
@@ -0,0 +1,67 @@
+import React from 'react'
```
jaahon, line 1: Same thing here, add a license.

frontend/packages/citizen-frontend/src/messages/MessageReadView.tsx
```diff
@@ -0,0 +1,67 @@
...
+
+  @media (max-width: ${messagesBreakpoint}) {
+    padding: ${defaultMargins.s};
+  }
+`
+
+const Header = styled.div`
+  display: flex;
+  justify-content: space-between;
+  align-items: baseline;
```
jaahon, line 51: I think `flex-end` would be safer here, `baseline` works differently depending on browser.

frontend/packages/citizen-frontend/src/messages/MessagesList.tsx
```diff
@@ -0,0 +1,97 @@
+import React from 'react'
```
jaahon, line 1: Same thing here, add a license.

frontend/packages/citizen-frontend/src/messages/MessagesPage.tsx
```diff
@@ -0,0 +1,59 @@
+import React, { useEffect, useState } from 'react'
```
jaahon, line 1: Same thing here, add a license.

frontend/packages/citizen-frontend/src/messages/MessagesPage.tsx
```diff
@@ -0,0 +1,59 @@
...
+      isRead: true
+    })
+
+    if (bulletins.isSuccess) {
+      setBulletins(
+        Success.of(
+          bulletins.value.map((b) =>
+            b.id === bulletin.id ? { ...b, isRead: true } : b
+          )
+        )
```
jaahon, line 38: You can just do `bulletins.map((b) => ...)` here.

frontend/packages/citizen-frontend/src/messages/MessageListItem.tsx
```diff
@@ -0,0 +1,85 @@
...
+  font-weight: 600;
+`
+
+const ContentSummary = styled.div`
+  white-space: nowrap;
+  overflow: hidden;
+  text-overflow: ellipsis;
+`
+
+export default React.memo(MessageListItem)
```
jaahon, line 85: I prefer doing the export where you define the component so you see right there what's the actual default export.

frontend/packages/citizen-frontend/src/messages/api.ts
```diff
@@ -0,0 +1,15 @@
...
+
+export async function getBulletins(): Promise<Result<ReceivedBulletin[]>> {
+  return client
+    .get<JsonOf<ReceivedBulletin[]>>('/citizen/bulletins')
+    .then((res) => Success.of(res.data.map(deserializeReceivedBulletin)))
+    .catch((e) => Failure.fromError(e))
+}
+
+export function markBulletinRead(id: string) {
+  void client.put(`/citizen/bulletins/${id}/read`)
```
jaahon, line 14: Why not return the promise?

frontend/packages/citizen-frontend/src/messages/MessagesPage.tsx
```diff
@@ -0,0 +1,59 @@
...
+  const openBulletin = (bulletin: ReceivedBulletin) => {
+    markBulletinRead(bulletin.id)
+
+    setActiveBulletin({
+      ...bulletin,
+      isRead: true
+    })
+
+    if (bulletins.isSuccess) {
+      setBulletins(
```
jaahon, line 33: I would do this state update only after the `markBulletinRead` request succeeds.

frontend/packages/citizen-frontend/src/messages/const.ts
```diff
@@ -0,0 +1 @@
+export const messagesBreakpoint = '850px'
```
jaahon, line 1: Where does this breakpoint come from? I personally think it would make sense to have the breakpoint at phone widths. Tablets should be able to show both the message and the list simultaneously.

Joosakur: When I was resizing window this was about where the right side part was starting to get awkwardly narrow. To make this breakpoint lower the left side message list would need to somehow shrink (now it is 400px wide).

frontend/packages/citizen-frontend/src/messages/MessagesPage.tsx
Joosakur: computer says no
Screenshot 2021-02-24 at 13 08 07

Joosakur: after map b is still an array

frontend/packages/citizen-frontend/src/messages/MessageListItem.tsx
Joosakur: Hmm, good point.. button seems also kinda weird as a list row item though.. @massproduction got any wisdom on this?

frontend/packages/citizen-frontend/src/messages/MessageReadView.tsx
Joosakur: `flex-end` gives a wrong alignment. `center` is closer but still off

frontend/packages/citizen-frontend/src/messages/MessagesPage.tsx
jaahon: Yeah sorry I had a bit of a brain fart, but what I meant is that you can remove the whole `bulletins.isSuccess` check and manual `Success.of(...)` wrapping and just map `bulletins.`

frontend/packages/citizen-frontend/src/messages/MessageReadView.tsx
jaahon: It should be the same if you just remove the margin from `Title`.

frontend/packages/citizen-frontend/src/messages/const.ts
jaahon: Yeah that could be done by giving the left side a larger `flex-grow` value and setting a `max-width` and then setting a `min-width` for right side so that their sum is the breakpoint.

Joosakur: Now I think it works, but the css might not be the minimum and cleanest solution to achieve that.

frontend/packages/citizen-frontend/src/messages/MessageListItem.tsx
massproduction: Yeah if it behaves like a button then I would just use a button.

frontend/packages/citizen-frontend/src/messages/MessagesList.tsx
```diff
@@ -0,0 +1,104 @@
...
+  }
+`
+
+const Return = styled(IconButton)`
+  margin-left: ${defaultMargins.xs};
+`
+
+const Container = styled.div`
+  min-width: 250px;
+  width: 30%;
```
jaahon, line 86: You have to remove this, otherwise the component looks funny in mobile views that are over 250px wide 😄

jaahon: Apart from this, the layout looks very good now.

frontend/packages/citizen-frontend/src/messages/MessageListItem.tsx
Joosakur: changed to button, will consider once more when we are closer to getting to prod

jaahon (approved):

PR #512, opened 2021-02-23
Support both free and pro icon sets once again

Gekkio:
#### Summary

Install pro icons manually directly to node_modules, so yarn won't try to resolve them and fail


jaahon (commented):
Both free and pro icons worked with a cloned repo. You should probably add a mention of the new script to the README.

jaahon (approved):

PR #514, opened 2021-02-24
EVAKA-4122 tweak income from application

hruoho:
#### Summary

Let `income` reference `application` when it originates from one. 
Show employee that income is created from application. 
Postpone income creation to the point where enduser accepts decision.  

Screenshot 2021-02-24 at 12 20 12


hruoho:
Should we (somehow) map earlier income notes `created automatically from application` to application ids? o.O Or eg. just adjust frontend to handle those too?

reynders (approved):

PR #519, opened 2021-02-24
EVAKA 4030: Add child preferred name field

mapehe:
#### Summary

Add child preferred name field. I'm going slightly against Figma spec on this one: I think preferred name should be part of additional information instead of personal information for a couple of reasons:

1. Only children have preferred names. Hence it makes sense that this information is in the child table.
2. Updating personal information happens either through [an existing method](https://github.com/espoon-voltti/evaka/blob/3e836b1d0405c6e4c1d6f326c020142e3a99535a/service/src/main/kotlin/fi/espoo/evaka/pis/service/PersonService.kt#L202-L222) or a custom endpoint. The first option involves dancing around some existing SSN logic and leaves room for unnecessary blundering (ACL logic for example) and adds complexity. The second option seems like a lot of hassle for such a simple thing since it also requires a nontrivial amount of frontend work.

Btw, I don't know if the field should be called nickname instead. I decided to go for the more neutral sounding preferredName.


hruoho (commented):
Would it make sense to add an integration test for this? Or e2e-test for that matter. Otherwise looks good to me!

reynders (approved):

mapehe:
> Would it make sense to add an integration test for this? Or e2e-test for that matter. Otherwise looks good to me!

Thanks for the feedback @hruoho. Agreed that there should be tests. However, there will be bunch of other fields added in this update. It's probably easier to get all the new fields done and design the tests afterwards.

PR #522, opened 2021-02-25
Move Espoo client to profile-activated config

Gekkio:
#### Summary

* this is better for other cities, because they are explicitly required to provide an alternative implementation (vs. accidentally using a mock or Espoo client)
* also remove Spring stuff from AsyncJobRunnerTest since it's no longer needed. This makes it possible to put the mock client in `SharedIntegrationTestConfig`, which is now only used by bigger Spring tests

### DIFF
--- a/service/src/main/kotlin/fi/espoo/evaka/EspooConfig.kt
+++ b/service/src/main/kotlin/fi/espoo/evaka/EspooConfig.kt
@@ -2,16 +2,18 @@
 //
 // SPDX-License-Identifier: LGPL-2.1-or-later
 
-package fi.espoo.evaka.invoicing
+package fi.espoo.evaka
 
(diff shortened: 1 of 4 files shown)


service/src/main/kotlin/fi/espoo/evaka/EspooConfig.kt
```diff
@@ -1,17 +1,15 @@
-// SPDX-FileCopyrightText: 2017-2020 City of Espoo
```
reynders, line 1: Was dropping the copyright text out intentional?

reynders (approved):

service/src/main/kotlin/fi/espoo/evaka/EspooConfig.kt
Gekkio: No :grimacing:

PR #523, opened 2021-02-25
EVAKA-HOTFIX minor citizen improvements

jaahon:
#### Summary

Scroll to top of application verification view because smooth scrolling breaks if it happens during rendering.
Remove `flexWrap` prop from `FixedSpaceRow` since `FixedSpaceFlexWrap` should be used instead.
Fix `FixedSpaceFlexWrap` width, `width: 100%` actually makes the element too short.


kustaahuhtala (approved):
Tested locally, looks good to me 👍

PR #526, opened 2021-02-25
EVAKA-3952: unpolished version of employee sends messages UI

Joosakur:


hruoho (approved):
Went through this together and decided to merge it into master just to keep things rolling fast

PR #529, opened 2021-02-25
EVAKA-3952 units in employee's message view

hruoho:
#### Summary

Simple first draft of employee's unit list in the messaging section

Screenshot 2021-02-25 at 21 57 26

### DIFF
--- a/frontend/packages/employee-frontend/src/components/messages/types.ts
+++ b/frontend/packages/employee-frontend/src/components/messages/types.ts
@@ -1,3 +1,7 @@
+// SPDX-FileCopyrightText: 2017-2021 City of Espoo
+//
+// SPDX-License-Identifier: LGPL-2.1-or-later
+
 import { JsonOf } from '@evaka/lib-common/src/json'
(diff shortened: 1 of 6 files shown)


frontend/packages/employee-frontend/src/components/messages/types.ts
```diff
@@ -21,3 +25,18 @@ export type IdAndName = {
...
   name: string
 }
+
+export type SentBulletin = {
+  id: string
+  sentAt: Date
+  sender: string
+  title: string
+  content: string
+}
```
Joosakur, line 35: If I remember correctly this does not include field `sender`, that is only returned to citizens at the moment. The only difference is that `sentAt` cannot be null. I believe this could be written shorter and drier as:
```
export type SentBulletin = Bulletin & {
  sentAt: Date
}
```

Joosakur (approved):
Did not test this in action so I might have missed something, but we can return to that later. Approved with one minor comment - good job!

PR #531, opened 2021-02-26
EVAKA-3952 bulletin references daycare

hruoho:
Require param in draft creation

#### Summary

Add foreign key & required api param

### DIFF
--- a/service/src/main/kotlin/fi/espoo/evaka/messaging/bulletin/BulletinControllerEmployee.kt
+++ b/service/src/main/kotlin/fi/espoo/evaka/messaging/bulletin/BulletinControllerEmployee.kt
@@ -28,14 +28,16 @@ class BulletinControllerEmployee(
     @PostMapping
     fun createBulletin(
         db: Database.Connection,
-        user: AuthenticatedUser
+        user: AuthenticatedUser,
+        @RequestParam(required = true) unitId: UUID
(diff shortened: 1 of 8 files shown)


service/src/main/resources/db/migration/V39__add_unit_id_to_bulletin.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE bulletin ADD COLUMN unit_id uuid REFERENCES daycare(id);
```
Joosakur, line 1: add index

service/src/main/kotlin/fi/espoo/evaka/messaging/bulletin/BulletinControllerEmployee.kt
```diff
@@ -45,7 +47,8 @@ class BulletinControllerEmployee(
                     content = "",
                     createdByEmployee = user.id,
                     groupId = null,
-                    sentAt = null
+                    sentAt = null,
+                    unitId = unitId
```
Joosakur, line 51: getDraftBulletins endpoint should take and use `@RequestParam(required = true) unitId: UUID` in same way as getSentBulletins

Joosakur (changes requested):

Joosakur (approved):

PR #532, opened 2021-02-26
EVAKA-3952: employee sends and views bulletins - part 2

Joosakur:

### DIFF
--- a/frontend/packages/citizen-frontend/src/messages/MessageListItem.tsx
+++ b/frontend/packages/citizen-frontend/src/messages/MessageListItem.tsx
@@ -39,7 +39,7 @@ export default React.memo(function MessageListItem({
           </span>
         </Header>
         <Title>{bulletin.title}</Title>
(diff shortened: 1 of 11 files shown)


kustaahuhtala (approved):
Tested locally, works nicely. 💎 

Here's couple of minor notes, that could be improved at some point:

* In the employee message view, multiple line breaks are not shown. This works nicely on citizen view.
Screenshot 2021-02-27 at 12 40 00
Screenshot 2021-02-27 at 12 40 55

* Message date is not visible on employee side (time of the message is shown).
* There could be `cursor: pointer` style added when navigating on employee side. Units and sent and draft links don't seem clickable.

Joosakur:
Message date is visible instead of time when the date is not today, but maybe it should say "Tänään 12:37"

Joosakur:
Will fix line breaks and cursor on Monday 👍

PR #537, opened 2021-03-01
Use email from application if person email is missing

reynders:


reynders:
Person email originates currently from the application so this is redundant operation

PR #543, opened 2021-03-02
EVAKA-3952: send bulletins also to head of child

Joosakur:


jaahon (approved):

jaahon (commented):
By the way, maybe renaming `guardian_id` into something like `receiver_id` would be good at this point.

PR #548, opened 2021-03-02
Allow KeyCloak SAML requests

ilja-git:
#### Summary

We need to be able to allow form-action URLs for KeyCloak SAML authentication. Also cleared proxy parameters so there is no duplicates.

### DIFF
--- a/proxy/nginx.conf.template
+++ b/proxy/nginx.conf.template
@@ -124,8 +124,8 @@ server {
   add_header X-DNS-Prefetch-Control off;
   add_header Report-To '{"group","csp-endpoint","max_age":31536000,"endpoints":[{"url":"https://$host/api/csp"}]}';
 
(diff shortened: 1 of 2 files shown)


proxy/nginx.conf.template
```diff
@@ -124,8 +124,8 @@ server {
   add_header X-DNS-Prefetch-Control off;
   add_header Report-To '{"group","csp-endpoint","max_age":31536000,"endpoints":[{"url":"https://$host/api/csp"}]}';
 
-  set $contentSecurityPolicyBase "block-all-mixed-content; upgrade-insecure-requests; form-action 'self'; default-src 'self'; font-src 'self' data: https://fonts.gstatic.com; style-src 'self' 'unsafe-inline' https://fonts.googleapis.com; img-src 'self' data: https://maps.gstatic.com https://maps.googleapis.com https://a.tile.openstreetmap.org https://b.tile.openstreetmap.org https://c.tile.openstreetmap.org; connect-src 'self' https://o318158.ingest.sentry.io https://api.digitransit.fi; object-src 'none'; report-uri /api/csp; report-to csp-endpoint";
```
mikkopiu, line 127: Why'd you remove the OSM and digitransit entries?

mikkopiu: Sidenote: these also seem to be missing from the compose proxy config. Would you mind fixing that at the same time?

ilja-git: Accidentally went aways with merge/rebase. Fixed.

mikkopiu (approved):

PR #550, opened 2021-03-03
[EVAKA-HOTFIX] Licenses: auto-download missing licenses, add missing headers

mikkopiu:
#### Summary
- Add auto-downloading of missing license files to the license helper tool (`bin/add-license-headers.sh`)
- Add all missing license headers
    - For the new Yarn v2 plugins/releases it's not totally clear how the files should be marked but the safest assumption (selected here) is to license the whole `frontend/.yarn/` directory with the same license as Yarn V2 itself (BSD-2-Clause, 2016-present Yarn Contributors) and to describe the licensing situation of everything else included in the directory in the DEP5 comment section
        - Not everything in the directory is copyrighted by Yarn Contributors but everything else appears to have their own copyright notices included as prefixed comments in the minified source code
- Remove licenses from all city of Espoo logos (proper licensing still needs figuring out with these trademarked logos)
    - Add more hacky exclusions to helper tool
    - Also replace test image in E2E tests with a generated one

#### Testing

1. Delete some license file(s) from `LICENSES/`
1. Delete license header from some file(s)
1. Run `./bin/add-license-headers.sh` from repo root
1. Tool should download any missing licenses and add the missing license headers
1. Diff should only contain files you change


hruoho (approved):
Tested as instructed, seems to work as intended 👍

PR #552, opened 2021-03-04
Messages paging

jaahon:
#### Summary

Use paging with infinite scroll in both citizen and employee frontends.

Also create generic paging data classes and types and refactor all places with paging to use them.

### DIFF
--- a/frontend/packages/citizen-frontend/src/messages/MessagesList.tsx
+++ b/frontend/packages/citizen-frontend/src/messages/MessagesList.tsx
@@ -5,6 +5,7 @@
 import React from 'react'
 import styled from 'styled-components'
 import { Result } from '@evaka/lib-common/src/api'
+import useIntersectionObserver from '@evaka/lib-common/src/utils/useIntersectionObserver'
 import { faArrowLeft } from '@evaka/lib-icons'
 import { tabletMin } from '@evaka/lib-components/src/breakpoints'
 import colors from '@evaka/lib-components/src/colors'
@@ -18,17 +19,21 @@ import { ReceivedBulletin } from '../messages/types'
 import MessageListItem from '../messages/MessageListItem'
 
 type Props = {
-  bulletins: Result<ReceivedBulletin[]>
+  bulletins: ReceivedBulletin[]
+  nextPage: Result<void>
   activeBulletin: ReceivedBulletin | null
   onClickBulletin: (target: ReceivedBulletin) => void
   onReturn: () => void
+  loadNextPage: () => void
 }
 
 export default React.memo(function MessagesList({
   bulletins,
+  nextPage,
   activeBulletin,
   onClickBulletin,
-  onReturn
+  onReturn,
+  loadNextPage
 }: Props) {
   const t = useTranslation()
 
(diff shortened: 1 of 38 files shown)


frontend/packages/employee-frontend/src/components/messages/MessagesPage.tsx
```diff
@@ -54,22 +56,34 @@ export default React.memo(function MessagesPage() {
...
+  const loadMessageCounts = useRestApi(
+    (unitId: string) =>
+      Promise.all([
+        getDraftBulletins(unitId, 1, 1),
+        getSentBulletins(unitId, 1, 1)
+      ]).then(([drafts, sent]) =>
+        drafts.chain((ds) =>
+          sent.map((ss) => ({ SENT: ss.total, DRAFT: ds.total }))
+        )
+      ),
```
Joosakur, line 73: This feels somewhat hard to understand to me, but I don't really have better suggestions either. Part of it might be that Array.map vs Result.map always confuses me (maybe it should be e.g. Result.transform?) and the word sent being both singular and plural does not help either. :D

frontend/packages/citizen-frontend/src/messages/MessagesList.tsx
```diff
@@ -46,24 +53,22 @@ export default React.memo(function MessagesList({
       <Container className={activeBulletin ? 'desktop-only' : undefined}>
         <HeaderContainer>
           <H1 noMargin>{t.messages.inboxTitle}</H1>
-          {bulletins.isSuccess && bulletins.value.length === 0 && (
-            <span>{t.messages.noMessages}</span>
-          )}
+          {pages === 0 && <span>{t.messages.noMessages}</span>}
```
Joosakur, line 56: this is now visible while loading initial messages

frontend/packages/employee-frontend/src/components/messages/MessageList.tsx
```diff
@@ -2,85 +2,171 @@
...
+}
+
+const OnEnterView = React.memo(function IsInView({
+  onEnter
+}: {
+  onEnter: () => void
+}) {
+  const ref = useIntersectionObserver<HTMLDivElement>(onEnter)
+  return <div ref={ref} />
+})
```
Joosakur, line 168: Screenshot 2021-03-04 at 11 39 35
Sometimes this does not quite seem to work, there is hardly any scrollbar or it may be completely missing. Maybe adding some height for this div could help or maybe page size 15 is just too small for 1440p screen. Or maybe you could try to have it load next page when this element is visible (now it seems to trigger only when it becomes visible after initially being hidden).

frontend/packages/employee-frontend/src/components/messages/MessagesPage.tsx
```diff
@@ -83,8 +97,7 @@ export default React.memo(function MessagesPage() {
     setMessageUnderEdit(null)
     setMessageOpen(null)
     if (unit) {
-      loadDraftMessages(unit.id)
-      loadSentMessages(unit.id)
+      loadMessageCounts(unit.id)
```
Joosakur, line 100: now when you have message list open while creating a draft or sending a message, the list content is not updated

Joosakur (commented):
Great work! There's some merge conflicts

frontend/packages/employee-frontend/src/components/messages/MessagesPage.tsx
jaahon: Well `map` is really not specific to arrays or lists or other iterables, it's a property of [functors](https://en.wikipedia.org/wiki/Functor_(functional_programming)). Same thing with `chain` (sometimes known as `flatMap`).

frontend/packages/citizen-frontend/src/messages/MessagesList.tsx
jaahon: Ah yes, good catch.

frontend/packages/employee-frontend/src/components/messages/MessageList.tsx
jaahon: Whoops, I was just testing it with a page size of 15, it should have been 50 😅

frontend/packages/employee-frontend/src/components/messages/MessagesPage.tsx
jaahon: Yeah, I guess I'll have to refactor it all again 😆

jaahon:
This PR is now at a point that all the issues are fixed (I think). I don't feel good about the code though 🤣

Joosakur:
> This PR is now at a point that all the issues are fixed (I think). I don't feel good about the code though 🤣

Looks good to me!

Joosakur (approved):

PR #555, opened 2021-03-05
Increase proxy buffers for KeyCloak to avoid gateway issues

ilja-git:
#### Summary

KeyCloak need bigger Nginx proxy buffers.

### DIFF
--- a/proxy/nginx.conf.template
+++ b/proxy/nginx.conf.template
@@ -296,6 +296,10 @@ server {
     proxy_set_header X-Forwarded-Proto "https";
 
     proxy_pass $authUrl;
+
+    proxy_buffer_size 32k;
+    proxy_buffers 8 32k;
+    proxy_busy_buffers_size 64k;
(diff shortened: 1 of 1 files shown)


proxy/nginx.conf.template
```diff
@@ -296,6 +296,10 @@ server {
     proxy_set_header X-Forwarded-Proto "https";
 
     proxy_pass $authUrl;
+
+    proxy_buffer_size 128k;
```
mikkopiu, line 300: Does it _really_ need 128 kB for headers? How huge is that stuff?

ilja-git: Decreased and lets test if it works with it.

mikkopiu (approved):
32 kB sounds more reasonable but it would be nice to have concrete facts on this. Not that we're that memory constrained or have user volumes high enough to _really_ matter but still.

Won't block

PR #557, opened 2021-03-07
EVAKA-3952-unread-bulletin-notification-email

hruoho:
#### Summary

First draft - forked from `PendingDecisionEmailService`

### DIFF
--- a/service/src/main/kotlin/fi/espoo/evaka/messaging/bulletin/BulletinNotificationEmailService.kt
+++ b/service/src/main/kotlin/fi/espoo/evaka/messaging/bulletin/BulletinNotificationEmailService.kt
@@ -35,7 +35,7 @@ class BulletinNotificationEmailService(
         val id: UUID,
         val receiverId: UUID,
         val receiverEmail: String,
-        val language: String?
+        val language: Language
     )
 
     fun getBulletinNotificationReceivers(tx: Database.Transaction, bulletinId: UUID): List<BulletinReceiver> {
@@ -45,7 +45,7 @@ class BulletinNotificationEmailService(
                 bi.id,
                 bi.receiver_id,
                 g.email receiver_email,
-                g.language
+                coalesce(lower(g.language), 'fi') as language
             FROM bulletin_instance bi
             JOIN person g ON bi.receiver_id = g.id
             WHERE bi.bulletin_id = :bulletinId
@@ -85,15 +85,13 @@ class BulletinNotificationEmailService(
 
     fun sendBulletinNotification(db: Database, msg: SendUnreadBulletinNotificationEmail) {
         db.transaction { tx ->
-            val lang = getLanguage(msg.language)
-
             emailClient.sendEmail(
                 traceId = msg.id.toString(),
                 toAddress = msg.receiverEmail,
                 fromAddress = fromAddress,
-                getSubject(lang),
-                getHtml(lang),
-                getText(lang)
+                getSubject(msg.language),
+                getHtml(msg.language),
+                getText(msg.language)
             )
 
             // Mark as sent
@@ -110,19 +108,13 @@ class BulletinNotificationEmailService(
         }
     }
 
-    private fun getLanguage(languageStr: String?): Language = when (languageStr) {
-        "sv", "SV" -> Language.sv
-        "en", "EN" -> Language.en
-        else -> Language.fi
-    }
-
     private fun getSubject(language: Language): String {
         val postfix = if (System.getenv("VOLTTI_ENV") == "prod") "" else " [${System.getenv("VOLTTI_ENV")}]"
 
         return when (language) {
-            Language.en -> "Uusi tiedote eVakassa$postfix"
-            Language.sv -> "Uusi tiedote eVakassa$postfix"
-            else -> "New bulletin in eVaka$postfix"
+            Language.en -> "New bulletin in eVaka$postfix"
+            Language.sv -> "Ny$postfix"
+            else -> "Uusi tiedote eVakassa$postfix"
         }
     }
 
--- a/service/src/main/kotlin/fi/espoo/evaka/messaging/bulletin/BulletinNotificationEmailService.kt
+++ b/service/src/main/kotlin/fi/espoo/evaka/messaging/bulletin/BulletinNotificationEmailService.kt
@@ -1,0 +1,156 @@
+// SPDX-FileCopyrightText: 2017-2021 City of Espoo
+//
+// SPDX-License-Identifier: LGPL-2.1-or-later
+
+package fi.espoo.evaka.messaging.bulletin
+
+import fi.espoo.evaka.daycare.domain.Language
+import fi.espoo.evaka.emailclient.IEmailClient
+import fi.espoo.evaka.shared.async.AsyncJobRunner
+import fi.espoo.evaka.shared.async.SendUnreadBulletinNotificationEmail
+import fi.espoo.evaka.shared.db.Database
+import mu.KotlinLogging
+import org.jdbi.v3.core.kotlin.mapTo
+import org.springframework.core.env.Environment
+import org.springframework.stereotype.Service
+import java.time.Instant
+import java.time.OffsetDateTime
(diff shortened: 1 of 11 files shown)


service/src/main/kotlin/fi/espoo/evaka/messaging/bulletin/BulletinNotificationEmailService.kt
```diff
@@ -0,0 +1,169 @@
...
+                                receiverEmail = notification.receiverEmail,
+                                language = notification.language
+                            )
+                        ),
+                        runAt = Instant.now(),
+                        retryCount = 10
+                    )
+                    createdJobCount++
+                } else {
+                    logger.warn("Could not send bulletin notification email to guardian ${notification.receiverId}: invalid email")
```
mikkopiu, line 91: Nit: "missing email" would probably be clearer than "invalid email"

service/src/main/kotlin/fi/espoo/evaka/messaging/bulletin/BulletinNotificationEmailService.kt
```diff
@@ -0,0 +1,169 @@
...
+        }
+    }
+
+    private fun getSubject(language: Language): String {
+        val postfix = if (System.getenv("VOLTTI_ENV") == "staging") " [staging]" else ""
+
+        return when (language) {
+            Language.en -> "Uusi tiedote eVakassa$postfix"
+            Language.sv -> "Uusi tiedote eVakassa$postfix"
+            else -> "New bulletin in eVaka$postfix"
```
mikkopiu, line 138: These should probably come from properties as treVaka would obviously not want these to say "eVaka"

service/src/main/kotlin/fi/espoo/evaka/messaging/bulletin/BulletinNotificationEmailService.kt
```diff
@@ -0,0 +1,169 @@
...
+    private fun getLanguage(languageStr: String?): Language {
+        return when (languageStr) {
+            "sv", "SV" -> Language.sv
+            "en", "EN" -> Language.en
+            else -> Language.fi
+        }
+    }
+
+    private fun getSubject(language: Language): String {
+        val postfix = if (System.getenv("VOLTTI_ENV") == "staging") " [staging]" else ""
```
mikkopiu, line 133: Maybe something like this would be better (no reason to constrain this functionality to just staging):
```suggestion
        val postfix = if (System.getenv("VOLTTI_ENV") == "prod") "" else " [${System.getenv("VOLTTI_ENV")}]"
```

service/src/main/kotlin/fi/espoo/evaka/messaging/bulletin/BulletinNotificationEmailService.kt
```diff
@@ -0,0 +1,169 @@
...
+        asyncJobRunner.sendBulletinNotificationEmail = ::sendBulletinNotification
+    }
+
+    val fromAddress = env.getRequiredProperty("mail_reply_to_address")
+
+    data class BulletinNotification(
+        val bulletinId: UUID,
+        val receiverId: UUID,
+        val receiverEmail: String,
+        val language: String?
```
mikkopiu, line 37: Why not map directly to `Language`? IMO it's better to map earlier rather than later to avoid mistakes with forgotten mappings.

service/src/main/kotlin/fi/espoo/evaka/messaging/bulletin/BulletinNotificationEmailService.kt
```diff
@@ -0,0 +1,169 @@
...
+class BulletinNotificationEmailService(
+    private val asyncJobRunner: AsyncJobRunner,
+    private val emailClient: IEmailClient,
+    env: Environment
+) {
+    init {
+        asyncJobRunner.sendBulletinNotificationEmail = ::sendBulletinNotification
+    }
+
+    val fromAddress = env.getRequiredProperty("mail_reply_to_address")
```
mikkopiu, line 31: Not changed here so won't block but: it's better to use `fi_espoo_...` style names for eVaka specific configs to avoid any possible conflicts with other packages and to clarify which configs are meant for eVaka source code vs. other systems.

If you decided to change this, it will require a "migration" in evaka-infra: define both variations of the environment variable until only the new one is supported and deployed everywhere.

service/src/main/kotlin/fi/espoo/evaka/messaging/bulletin/BulletinNotificationEmailService.kt
```diff
@@ -0,0 +1,169 @@
...
+                                receiverId = notification.receiverId,
+                                bulletinId = notification.bulletinId,
+                                receiverEmail = notification.receiverEmail,
+                                language = notification.language
+                            )
+                        ),
+                        runAt = Instant.now(),
+                        retryCount = 10
+                    )
+                    createdJobCount++
```
mikkopiu, line 89: What the purpose here? Missing logging?

service/src/main/kotlin/fi/espoo/evaka/messaging/bulletin/BulletinNotificationEmailService.kt
```diff
@@ -0,0 +1,169 @@
...
+        }
+    }
+
+    private fun getLanguage(languageStr: String?): Language {
+        return when (languageStr) {
+            "sv", "SV" -> Language.sv
+            "en", "EN" -> Language.en
+            else -> Language.fi
+        }
+    }
```
mikkopiu, line 130: Purefly a stylistic preference but this can be simplified to:
```suggestion
    private fun getLanguage(languageStr: String?): Language = when (languageStr) {
        "sv", "SV" -> Language.sv
        "en", "EN" -> Language.en
        else -> Language.fi
    }
```

service/src/main/kotlin/fi/espoo/evaka/shared/controllers/ScheduledOperationController.kt
```diff
@@ -106,4 +108,10 @@ class ScheduledOperationController(
         pendingDecisionEmailService.scheduleSendPendingDecisionsEmails(db)
         return ResponseEntity.noContent().build()
     }
+
+    @PostMapping("/send-bulletin-notification-emails")
+    fun sendBulletinNotifications(db: Database.Connection): ResponseEntity<Unit> {
+        bulletinNotificationEmailService.scheduleSendingBulletinNotifications(db)
+        return ResponseEntity.noContent().build()
+    }
```
mikkopiu, line 116: More of a architectural Q: why are these notifications sent on a schedule instead of being sent as they're created (or rather: added to the async queue)? Or am I misunderstanding this?

service/src/main/kotlin/fi/espoo/evaka/messaging/bulletin/BulletinNotificationEmailService.kt
hruoho: Yes you are right, these are just placeholders since the actual content is still missing

service/src/main/kotlin/fi/espoo/evaka/shared/controllers/ScheduledOperationController.kt
hruoho: Yep no real reason for that - refactored this whole thing quite a bit after discussing with others

service/src/main/kotlin/fi/espoo/evaka/messaging/bulletin/BulletinNotificationEmailService.kt
hruoho: Actually it seems that this env variable is defined only locally on the app level, ie. is not in evaka-infra 🤔

mikkopiu: Even better

mikkopiu (approved):
LGTM

PR #558, opened 2021-03-08
EVAKA-HOTFIX hide ssn from log meta fields

jaahon:
#### Summary

Use a `ValueMasker` to hide ssns from all log fields. Uses the same regex used previously for masking `message` and `stackTrace` fields.


mikkopiu (approved):
Wonderful stuff :+1: Obviously value masking is relatively expensive but the endgame is anyway to move all masking to the log ingest pipelines -- and out of eVaka itself -- so doesn't matter too much IMO.

PR #573, opened 2021-03-09
[EVAKA-4159] Fix application received email links and update texts

mikkopiu:
#### Summary
- Invalid href value (missing protocol) caused links to be output as [link]link instead of being valid links
- Update "application received" email's Finnish and Swedish texts per new spec


massproduction (approved):

kustaahuhtala (changes requested):
There will be some additional changes for the texts. I suggest we put them into this PR, too (once we get them).

mikkopiu:
@kustaahuhtala Updated texts. We can add the English texts when they arrive, in a separate PR.

kustaahuhtala (approved):

kustaahuhtala:
It's good now. Maybe in the future we could link directly to decision page on the citizen frontend, but I think we should unify it at the same time everywhere (in decisions etc, too). Nice work 👌

PR #574, opened 2021-03-10
EVAKA-HOTFIX handle parsefi throws better

massproduction:
#### Summary
Change `parseFi` calls to `parseFiOrNull` and when using `parseFi` wrap them in a `try...catch`. This should help out with the sentry errors.

### DIFF
--- a/frontend/src/citizen-frontend/applications/editor/ApplicationFormData.ts
+++ b/frontend/src/citizen-frontend/applications/editor/ApplicationFormData.ts
@@ -264,7 +264,8 @@ export function formDataToApiData(
     otherGuardianId,
     otherGuardianLivesInSameAddress
   }: ApplicationDetails,
-  form: ApplicationFormData
+  form: ApplicationFormData,
+  isDraft = false
 ): ApplicationFormUpdate {
   const fullFamily =
     type === 'DAYCARE' ||
@@ -281,7 +282,9 @@ export function formDataToApiData(
             street: form.contactInfo.childFutureStreet,
             postalCode: form.contactInfo.childFuturePostalCode,
             postOffice: form.contactInfo.childFuturePostOffice,
-            movingDate: LocalDate.parseFiOrNull(form.contactInfo.childMoveDate)
+            movingDate: isDraft
+              ? LocalDate.parseFiOrNull(form.contactInfo.childMoveDate)
+              : LocalDate.parseFiOrThrow(form.contactInfo.childMoveDate)
           }
         : null,
       allergies: type !== 'CLUB' ? form.additionalDetails.allergies : '',
@@ -295,9 +298,9 @@ export function formDataToApiData(
             street: form.contactInfo.guardianFutureStreet,
             postalCode: form.contactInfo.guardianFuturePostalCode,
             postOffice: form.contactInfo.guardianFuturePostOffice,
-            movingDate: LocalDate.parseFiOrNull(
-              form.contactInfo.guardianMoveDate
-            )
+            movingDate: isDraft
+              ? LocalDate.parseFiOrNull(form.contactInfo.guardianMoveDate)
+              : LocalDate.parseFiOrThrow(form.contactInfo.guardianMoveDate)
           }
         : null,
       phoneNumber: form.contactInfo.guardianPhone,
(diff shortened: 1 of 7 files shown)


frontend/src/citizen-frontend/applications/editor/validations.ts
```diff
@@ -92,12 +92,16 @@ export const isValidDecisionStartDate = (
...
   type: DecisionType
 ): boolean => {
-  return date !== null
-    ? date.isEqualOrAfter(LocalDate.parseFi(startDate)) &&
-        date.isEqualOrBefore(
-          maxDecisionStartDate(LocalDate.parseFi(startDate), type)
-        )
-    : false
+  try {
+    return date !== null
```
Joosakur, line 96: signature says `date: LocalDate` so it cannot be null

frontend/src/citizen-frontend/applications/editor/validations.ts
```diff
@@ -92,12 +92,16 @@ export const isValidDecisionStartDate = (
...
+      ? date.isEqualOrAfter(LocalDate.parseFi(startDate)) &&
+          date.isEqualOrBefore(
+            maxDecisionStartDate(LocalDate.parseFi(startDate), type)
+          )
+      : false
+  } catch (e) {
+    return false
+  }
```
mikkopiu, line 104: Don't probably want to hide all possible exceptions from all the functions + just parsing `startDate` once would be preferable. My suggestion (comment-coding, cleanup needed):

```suggestion
  let parsedDate: LocalDate
  try {
    parsedDate = LocalDate.parseFi(startDate)
  } catch (e) {
    if (e instanceof RangeError) {
      return false
    }
    // Unexpected
    throw e
  }
  return date !== null
    ? date.isEqualOrAfter(parsedDate) &&
        date.isEqualOrBefore(maxDecisionStartDate(parsedDate, type))
    : false
```

mikkopiu (changes requested):

frontend/src/citizen-frontend/decisions/api.ts
```diff
@@ -37,9 +37,15 @@ export async function acceptDecision(
   decisionId: UUID,
   requestedStartDate: string
 ): Promise<Result<void>> {
+  let date
+  try {
+    date = LocalDate.parseFi(requestedStartDate)
+  } catch (e) {
+    return Failure.fromError(e)
+  }
```
Joosakur, line 45: wouldn't it make more sense to handle parsing and validation errors at the component. this api call should take LocalDate instead of string.

frontend/src/citizen-frontend/applications/editor/ApplicationFormData.ts
```diff
@@ -281,7 +281,7 @@ export function formDataToApiData(
             street: form.contactInfo.childFutureStreet,
             postalCode: form.contactInfo.childFuturePostalCode,
             postOffice: form.contactInfo.childFuturePostOffice,
-            movingDate: LocalDate.parseFi(form.contactInfo.childMoveDate)
+            movingDate: LocalDate.parseFiOrNull(form.contactInfo.childMoveDate)
```
Joosakur, line 284: these are also just hiding a problem. the validation function needs to ensure that if there is a future address then moving date has valid value

frontend/src/citizen-frontend/applications/editor/validations.ts
massproduction: Thanks, very nice suggestion 💎.

frontend/src/citizen-frontend/applications/editor/validations.ts
massproduction: True, removed the extra null check from `isValidPreferredStartDate` as well.

mikkopiu (approved):
To my eye :+1:. Not sure if @Joosakur wants to have another look.

frontend/src/citizen-frontend/decisions/decision-response-page/DecisionResponse.tsx
```diff
@@ -78,33 +79,42 @@ export default React.memo(function DecisionResponse({
   }
 
   const onSubmit = async () => {
+    if (parsedDate === null) throw new Error('Parsed date was null')
     setSubmitting(true)
     return (acceptChecked
-      ? acceptDecision(applicationId, decisionId, requestedStartDate)
+      ? acceptDecision(applicationId, decisionId, parsedDate)
       : rejectDecision(applicationId, decisionId)
```
Joosakur, line 86: actually I think there's a bug now, sorry missed it when we pair-coded. I guess the null-check on parsedDate should be done only when accepting, since rejection does not care about the date

Joosakur (approved):

PR #575, opened 2021-03-10
EVAKA-FIX prevent double submits in placement draft view

hruoho:
#### Summary

Quick fix to avoid accidental double placement plan submits


jaahon (commented):
I think a better idea would be to use `AsyncButton` since it has a better indicator of the request being in progress.

jaahon (approved):

PR #576, opened 2021-03-10
EVAKA-4016: voucher value freeze (sql edition)

Joosakur:


service/src/main/kotlin/fi/espoo/evaka/shared/controllers/ScheduledOperationController.kt
```diff
@@ -106,4 +107,14 @@ class ScheduledOperationController(
         pendingDecisionEmailService.scheduleSendPendingDecisionsEmails(db)
         return ResponseEntity.noContent().build()
     }
+
+    @PostMapping("/freeze-voucher-value-reports")
+    fun freezeVoucherValueReports(
+        db: Database.Connection,
+        @RequestParam year: Int,
+        @RequestParam month: Int
```
Joosakur, line 115: To be removed (makes testing easier)

jaahon (approved):

PR #578, opened 2021-03-12
EVAKA-4121 retroactive fee decision handler

hruoho:
#### Summary

For retroactive fee decisions, ensure approver is set as decision handler.

### DIFF
--- a/service/src/main/kotlin/fi/espoo/evaka/invoicing/data/FeeDecisionQueries.kt
+++ b/service/src/main/kotlin/fi/espoo/evaka/invoicing/data/FeeDecisionQueries.kt
@@ -509,7 +509,7 @@ fun findFeeDecisionsForHeadOfFamily(
         .let { it.merge() }
 }
 
-fun approveFeeDecisionDraftsForSending(h: Handle, ids: List<UUID>, approvedBy: UUID) {
+fun approveFeeDecisionDraftsForSending(h: Handle, ids: List<UUID>, approvedBy: UUID, approvedAt: LocalDate, isRetroactive: Boolean = false) {
     val sql =
         """
         WITH youngest_child AS (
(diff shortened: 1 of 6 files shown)


service/src/main/kotlin/fi/espoo/evaka/invoicing/data/FeeDecisionQueries.kt
```diff
@@ -509,6 +510,31 @@ fun findFeeDecisionsForHeadOfFamily(
...
+        batch
+            .bind("id", id)
+            .bind("status", FeeDecisionStatus.WAITING_FOR_SENDING.toString())
+            .bind("approvedBy", approvedBy)
+            .bind("now", LocalDateTime.now())
+            .add()
+    }
+    batch.execute()
+}
+
```
jaahon, line 537: Instead of creating a whole separate function for this I would have added a `useUnitHandlerOrSomething` parameter to `approveFeeDecisionDraftsForSending` and moved the
```
CASE
    WHEN youngest_child.finance_decision_handler_id IS NOT NULL AND fd.decision_type = 'NORMAL'
        THEN youngest_child.finance_decision_handler_id
    ELSE :approvedBy
END
```
part to kotlin code as well.

service/src/main/kotlin/fi/espoo/evaka/invoicing/service/FeeDecisionService.kt
```diff
@@ -86,7 +88,11 @@ class FeeDecisionService(
 
         val validIds = validDecisions.map { it.id }
 
-        approveFeeDecisionDraftsForSending(tx.handle, validIds, user.id)
+        val (retroactiveIds, otherValidIds) = validDecisions.partition { isRetroactive(it.validFrom, LocalDate.now()) }
```
jaahon, line 91: I guess it would have been polite to do this `LocalDate.now()` refactoring in this PR.

jaahon (approved):

PR #580, opened 2021-03-12
EVAKA-4168: hotfix child view role based layout

Joosakur:
#### Summary
- explicit layouts for each role, no default
- use role layout combination hotfix only when user has more than 1 role


mapehe:
Thanks for the help with this one 💎

mapehe (approved):

PR #585, opened 2021-03-12
Infer language from the browser settings

mapehe:
#### Summary

Get default language from the browser settings.


frontend/src/citizen-frontend/localization/state.tsx
```diff
@@ -11,13 +11,23 @@ export const langs = ['fi', 'sv', 'en'] as const
 
 export type Lang = typeof langs[number]
 
+const getDefaultLanguage: () => Lang = () => {
+  const language =
+    (window.navigator['userLanguage'] as string) || window.navigator.language
+  if (language === 'fi' || language === 'sv') {
```
massproduction, line 17: I think we need english as well. Also we should check strings like `en-US`, 'en-GB` etc. and not just `fi` and `sv`. I think this was a bug even in the old vue frontend.

Also wasn't there some query parameter check in the old enduser frontend? Something like you could use `espoonvarhaiskasvatus.fi?lang=sv`

mapehe: Thanks for the feedback @massproduction !

* Actually not including English was a decision I made after looking at the old vue code and some thought. The reason I proposed not including it here is that I, for example, use English settings on my devices even though I would like to see evaka in Finnish. I assume that there is a number of people like me and perhaps we should hesitate to assume the user actually wants to see the page in English. I can of course add English easily, but I'd like to hear your opinion on this first.
* Yeah I was unsure if we want to use the query parameter here. I'll add it.

kustaahuhtala (changes requested):
I tried this change locally on Firefox, but it seems the browser language was not picked up, for some reason:
Screenshot 2021-03-16 at 14 26 35

On firefox, `window.navigator.language` prints `sv-SE`.

mapehe:
Thanks for the feedback @kustaahuhtala. Seems that Firefox calls swedish `sv-SV` instead of `sv`. Fixed.

kustaahuhtala (approved):

PR #589, opened 2021-03-15
EVAKA-HOTFIX fee decision approved_at timestamp

jaahon:
#### Summary

Use an `Instant` instead of `LocalDate` to set a fee decision's `approved_at` value to get an accurate timestamp.


hruoho (approved):
🎖️

PR #590, opened 2021-03-15
Endpoint for scheduled removal of expired child daycare daily notes

reynders:
#### Summary
Scheduled endpoint for deleting expired daycare daily notes made for children
- notes modified at least 12h ago for children placed in non round the clock daycares
- round the clock daycare notes will be deleted manually
- group note deletion will be added later with the editor implementation

### DIFF
--- a/service/src/main/kotlin/fi/espoo/evaka/messaging/daycarydailynote/DaycareDailyNoteQueries.kt
+++ b/service/src/main/kotlin/fi/espoo/evaka/messaging/daycarydailynote/DaycareDailyNoteQueries.kt
@@ -2,6 +2,7 @@ package fi.espoo.evaka.messaging.daycarydailynote
 
 import fi.espoo.evaka.shared.db.Database
 import org.jdbi.v3.core.kotlin.mapTo
+import java.time.Instant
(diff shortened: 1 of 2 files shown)


service/src/main/kotlin/fi/espoo/evaka/messaging/daycarydailynote/DaycareDailyNoteQueries.kt
```diff
@@ -109,3 +109,25 @@ fun Database.Transaction.deleteChildDaycareDailyNotes(childId: UUID) {
...
 }
+
+fun Database.Transaction.deleteExpiredDaycareDailyNotes() {
+    createUpdate(
+        """
+WITH expired_child_notes AS (
+    SELECT id, child_id
+    FROM daycare_daily_note
+    WHERE child_id IS NOT null
+        AND modified_at < now() - INTERVAL '12 hours'
```
mikkopiu, line 120: Did we decide that we want to replace these `now()` calls with overridable arguments (for testing)?

mikkopiu (approved):

PR #596, opened 2021-03-16
Generalize AD SAML code by making espoo specific things configurable

reynders:
#### Summary
- Rename espoo-ad-saml.ts as ad-saml.ts
- Remove the word "espoo" from code
- Parameterize externalId, logout url, entrypoint url
Requires https://github.com/espoon-voltti/evaka-infra/pull/365

### DIFF
--- a/apigw/src/shared/config.ts
+++ b/apigw/src/shared/config.ts
@@ -136,12 +136,16 @@ const certificateNames = Object.keys(certificates) as ReadonlyArray<
 export const devLoginEnabled =
   env('DEV_LOGIN', parseBoolean) ?? ifNodeEnv(['local', 'test'], true) ?? false
 
-const eadCallbackUrl = process.env.EAD_SAML_CALLBACK_URL
+const adCallbackUrl = process.env.EAD_SAML_CALLBACK_URL
+const adEntryPointUrl = process.env.AD_SAML_ENTRYPOINT_URL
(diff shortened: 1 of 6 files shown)


apigw/src/shared/config.ts
```diff
@@ -150,6 +154,8 @@ export const eadConfig =
       }
     : undefined
 
+export const externalId = process.env.EXTERNAL_ID || 'espoo-ad'
```
Gekkio, line 157: Is there a reason to use `||` instead of `??` here? Or maybe we could remove the default value completely and use `required()`?
Also, the name isn't very clear (see the comment in the evaka-infra pull request)

Gekkio (approved):

apigw/src/shared/config.ts
```diff
@@ -136,12 +136,16 @@ const certificateNames = Object.keys(certificates) as ReadonlyArray<
 export const devLoginEnabled =
   env('DEV_LOGIN', parseBoolean) ?? ifNodeEnv(['local', 'test'], true) ?? false
 
-const eadCallbackUrl = process.env.EAD_SAML_CALLBACK_URL
+const adCallbackUrl = process.env.EAD_SAML_CALLBACK_URL
```
apeltola, line 139: Just naming conventions: noticed, that there are env variables starting with both EAD and AD.

apeltola:
Looks good! 👍🏻

PR #599, opened 2021-03-16
EVAKA-4016 more voucher report UX improvements

jaahon:
#### Summary

* Add an indicator for 'new' voucher report rows
* Store voucher report filters in url so that page refresh and navigation do not reset the filters
* Add unit name filter to area voucher report
* Enable estimated future voucher reports
* Add toggleable sorting by group name
* Remove the "coefficient" column from the report since it should show a coefficient that does not exist in our model yet


Joosakur (approved):
🏅

PR #602, opened 2021-03-17
EVAKA-HOTFIX planned absences and discounts

jaahon:
#### Summary

Do not give discounts for "planned absences". They are only used to indicate a child's schedule and are not actual absences.


hruoho (approved):
Seems great!

PR #603, opened 2021-03-17
Add locale sorting to a few API calls

mapehe:
#### Summary

There have been some issues with sorting absences, applications and units containing scandinavian letters. The real fix requires recreating the databases, but here's a quick frontend fix for the meantime.


jaahon (commented):
You should use an explicit locale and probably even `{ ignorePunctuation: true }`.

jaahon (approved):
There's quite a bit of repetition now, so you could create a utility function that does the comparison with the relevant configurations, but yeah everything works as it is.

mapehe:
Thanks for the feedback @jaahon. I suggest that we go ahead with this one since it's a hotfix anyway.

PR #604, opened 2021-03-18
EVAKA-3952 receiver selection

hruoho:
#### Summary

Add component to choose bulletin receivers in a flexible way
Screenshot 2021-03-18 at 15 08 58


service/src/main/resources/db/migration/V51__refactor_bulletin_reciever.sql
```diff
@@ -0,0 +1,2 @@
+ALTER TABLE bulletin_instance DROP COLUMN bulletin_receiver_id;
+ALTER TABLE bulletin_instance ADD COLUMN bulletin_id uuid REFERENCES bulletin(id);
```
Joosakur, line 2: add index

service/src/main/kotlin/fi/espoo/evaka/messaging/bulletin/BulletinQueries.kt
```diff
@@ -126,28 +124,53 @@ fun Database.Transaction.sendBulletin(
...
+            AND br.child_id IS NULL
+            
+            UNION DISTINCT
+            
+            SELECT pl.child_id, br.bulletin_id
+            FROM bulletin b
+            JOIN bulletin_receiver br ON br.bulletin_id = b.id
+            JOIN daycare_group dg ON br.unit_id = dg.daycare_id
+            JOIN daycare_group_placement gpl ON dg.id = gpl.daycare_group_id AND daterange(gpl.start_date, gpl.end_date, '[]') @> :date
+            JOIN placement pl ON gpl.daycare_placement_id = pl.id
```
Joosakur, line 154: we could forget groups and group placements here and just check that the child has placement in the unit.

FROM bulletin b
JOIN bulletin_receiver br ON br.bulletin_id = b.id
JOIN placement pl ON br.unit_id = pl.unit_id AND daterange(pl.start_date, pl.end_date, '[]') @> :date

service/src/main/kotlin/fi/espoo/evaka/messaging/bulletin/BulletinQueries.kt
```diff
@@ -126,28 +124,53 @@ fun Database.Transaction.sendBulletin(
...
+            WHERE b.id = :bulletinId 
+            AND br.unit_id IS NOT NULL 
+            AND br.group_id IS NULL 
+            AND br.child_id IS NULL
```
Joosakur, line 158: we are handling cases 
1. group not null, child not null
2. group not null, child is null
3. group is null, child is null
but what about the fourth possible permutation
4. group is null, child not null?

When does that happen, what does it mean, should it not be allowed? Depending on answers we should maybe either handle it here or prevent it with a database check constraint.

service/src/main/kotlin/fi/espoo/evaka/messaging/bulletin/BulletinQueries.kt
```diff
@@ -126,28 +124,53 @@ fun Database.Transaction.sendBulletin(
     // language=sql
     val insertBulletinInstancesSql = """
         WITH children AS (
-            SELECT pl.child_id, br.id AS bulletin_receiver_id
+            SELECT child_id, bulletin_id FROM bulletin_receiver
+            WHERE
+            bulletin_id = :bulletinId 
+            AND unit_id IS NOT NULL
```
Joosakur, line 130: Based on kotlin data class it looks to me like bulletin_receiver.unit_id should never be null. If that is the case then lets remove these checks from here and add the missing database not null constraint instead.

service/src/main/kotlin/fi/espoo/evaka/messaging/bulletin/BulletinQueries.kt
```diff
@@ -163,42 +186,148 @@ fun Database.Read.getSentBulletinsByUnit(
 ): Paged<Bulletin> {
     // language=sql
     val sql = """
-        SELECT COUNT(b.*) OVER (), b.*, br.group_id AS group_id, dg.name AS group_name, concat(e.first_name, ' ', e.last_name) AS created_by_employee_name
-        FROM bulletin b
+        WITH limited_bulletins AS (
```
Joosakur, line 189: not quite sure if this will now paginate correctly.. I think not?

Joosakur: I mean it looks like it's now paginating a set of bulletins but then filtering some out afterwards, which means that pages will have varying amounts of items and some may even be empty.

service/src/main/kotlin/fi/espoo/evaka/messaging/bulletin/BulletinQueries.kt
hruoho: let's discuss this through inbox

service/src/main/kotlin/fi/espoo/evaka/messaging/bulletin/BulletinQueries.kt
hruoho: I guess this depends on how we want to handle area- and citywise bulletins on the UI level - do we want to handle eg. Espoon keskus as single recipient or a collection of units 🤷

service/src/main/resources/db/migration/V51__refactor_bulletin_reciever.sql
hruoho: maybe rename `reciever` -> `receiver`

mapehe (approved):

PR #605, opened 2021-03-18
EVAKA-4030 e2e tests

massproduction:
#### Summary
Added tests for child attendances.

User should be able to do:
- Create a daily note for a child
- Delete a daily note for a child
- Mark a child as present at 08:30 and leaving at 16:00
- Mark a child as absent for the whole day
- Mark a child as present and returned back to coming list
- Undo the whole flow of marking present at 08:30 and leaving at 16:00
- Mark an absence if child arrives at 08:30 and leaves at 09:00

### DIFF
--- a/frontend/src/e2e-test/dev-api/fixtures.ts
+++ b/frontend/src/e2e-test/dev-api/fixtures.ts
@@ -737,6 +737,23 @@ export function createDaycarePlacementFixture(
   }
 }
 
+export function createPreschoolDaycarePlacementFixture(
+  id: string,
+  childId: string,
+  unitId: string,
+  startDate = '2020-05-01',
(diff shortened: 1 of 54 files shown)


frontend/src/e2e-test/dev-api/fixtures.ts
```diff
@@ -737,6 +737,23 @@ export function createDaycarePlacementFixture(
   }
 }
 
+export function createPreschoolDaycarePlacementFixture(
+  id: string,
+  childId: string,
+  unitId: string,
+  startDate = '2020-05-01',
```
reynders, line 744: We should not have fixed dates in tests

reynders (approved):

PR #610, opened 2021-03-18
Redirect sv-domain to main domain with lang parameter

ilja-git:
#### Summary

Optonal redirect from given SV domain to main domain on proxy. The redirect will add ?lang=sv to request.

### DIFF
--- a/proxy/nginx.conf.template
+++ b/proxy/nginx.conf.template
@@ -308,3 +308,19 @@ server {
   auth_basic_user_file  "/etc/nginx/.htpasswd";
   <% end %>
 }
+
+<% if ENV.key?("SERVICE_DOMAIN_SV") && ENV.key?("SERVICE_DOMAIN") %>
+server {
+    listen 8080;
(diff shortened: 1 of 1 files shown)


proxy/nginx.conf.template
```diff
@@ -308,3 +308,11 @@ server {
   auth_basic_user_file  "/etc/nginx/.htpasswd";
   <% end %>
 }
+
+<% if ENV.key?("SERVICE_DOMAIN_SV") && ENV.key?("SERVICE_DOMAIN") %>
+server {
+    listen 8080;
+    server_name <%= ENV["SERVICE_DOMAIN_SV"] %>;
+    rewrite ^ https://<%= ENV["SERVICE_DOMAIN"] %>$uri?lang=sv redirect;
+}
```
mikkopiu, line 317: IMO we should log correctly here, too. Also the security-headers like "Strict-Transport-Security".

mikkopiu (changes requested):

mikkopiu (approved):

PR #614, opened 2021-03-19
Minimal playwright setup + convert citizen map page tests to Playwright

Gekkio:


jaahon (commented):
I personally don't see the value in creating a separate class for every single "type" of a page element, e.g. `class MapMarker extends ElementSelector {}`. Just seems like unnecessary boilerplate.

In general I would prefer a more composable solution where you could combine different "features" together with a selector to get an object that has the desired behaviours and properties.

reynders (approved):

PR #616, opened 2021-03-19
EVAKA-TECH: mocked time PoC

Joosakur:
#### Summary
Thoughts?


Gekkio (commented):
Some comments:

- needs thread-safety (e.g. by using `AtomicReference`)
- it's probably simpler to store a `ZonedDateTime` instead of `Instant`. We can always easily get the latter from the former, and this should remove most of the extra back-and-forth conversions.
- you probably want to use `object` instead of `class` for the singletons. There's no point in using a class and allowing e.g. `new EvakaClock()`
- instead of duplicating modification functions from Java DateTime API in our API, we could just have one modify function that accepts a closure and is used like this: `EvakaClock.modify { it.plusHours(1) }`

PR #619, opened 2021-03-22
Most of messaging blocklist frontend

mapehe:
#### Summary

A commit that does most of the messaging blocklist frontend stuff.


mapehe:
Thought I'd create a pull request so it's easier to collaborate on this than simply pushing @Joosakur

PR #620, opened 2021-03-22
EVAKA-4027 send voucher decisions to Varda

jaahon:
#### Summary

Send voucher value decisions to Varda. In essence the fee data integration stays the same, voucher value decision data is just included in the data sets with `UNION`.


reynders (approved):
Only question is about the performance of the varda feedata query - should it be analyzed? I'm willing to trust on the gut feeling of the author.

jaahon:
> Only question is about the performance of the varda feedata query - should it be analyzed? I'm willing to trust on the gut feeling of the author.

This PR shouldn't make the performance much worse and it's adequate at the moment. We also don't really have any voucher value decision data to test this properly.

PR #623, opened 2021-03-23
New frontend project structure and build - part 4

Gekkio:
#### Summary

- remove `@evaka/foo` path mapping. It works, but we don't need three different ways of importing
- use a single consistent `baseUrl` configuration. This is needed for eslint, which absolutely requires all code to be in one TS project and all imports to be resolvable in this project. Previously `from 'state/i18n'` was ambiguous since two frontends had the same path internally. Now the full path from `src` is included: `from 'employee-frontend/state/i18n'` and it's no longer ambiguous
- compile with ES2020 by default and transpile frontend bundles with babel-loader with a separate browserslist configuration. TS project references require us to compile modules like lib-common using one configuration only, so we need to either be pessimistic (compile with ES6 and suffer the consequences in non-browser code too), or optimistic (compile with ES2020 and somehow make sure the code works in supported browsers)
- upgrade TestCafe and work around new internal TestCafe errors. Apparently TestCafe does custom transformations to the our code and breaks it in various ways. In other words, I might be testing using Chrome 89 locally and our code works perfectly in it, but TestCafe doesn't fully support optional chaining (`?.`) operator so it mangles the code into something that doesn't work


jaahon (approved):
🎖️ 💪

PR #627, opened 2021-03-24
Allow round the clock unit supervisors to open application attachments

jaahon:
#### Summary

Round the clock unit supervisors should be able to open "extended care" attachments for applications with a placement proposal to their units.


hruoho (approved):
🏅

PR #630, opened 2021-03-25
EVAKA-FIX Parse bulletin notification language from string

hruoho:
#### Summary

Revert having simple way to handle unsupported languages, ie. to avoid eg.

```
Unexpected error (no Language value could be matched to the name et)
```


service/src/main/kotlin/fi/espoo/evaka/messaging/bulletin/BulletinNotificationEmailService.kt
```diff
@@ -90,6 +90,14 @@ class BulletinNotificationEmailService(
         }
     }
 
+    private fun getLanguage(languageStr: String?): Language {
+        return when (languageStr) {
+            "sv", "SV" -> Language.sv
+            "en", "EN" -> Language.en
```
jaahon, line 96: You could do `language.toLowerCase()` so you wouldn't have to always give the matches in different cases.

jaahon (approved):

PR #633, opened 2021-03-26
EVAKA-4031 backup pickups

massproduction:
#### Summary
Added backup pickups to child information page.


service/src/main/resources/db/migration/V51__backup_pickup.sql
```diff
@@ -0,0 +1,9 @@
+CREATE TABLE public.backup_pickup(
+    id uuid DEFAULT ext.uuid_generate_v1mc() NOT NULL,
+    child_id uuid NOT NULL,
+    name text NOT NULL,
+    phone text NOT NULL
+);
+
+ALTER TABLE ONLY public.backup_pickup
+    ADD CONSTRAINT backup_pickup_child_fkey FOREIGN KEY (child_id) REFERENCES public.person(id);
```
Joosakur, line 9: index would be good

service/src/main/kotlin/fi/espoo/evaka/backuppickup/BackupPickupController.kt.kt
```diff
@@ -0,0 +1,157 @@
...
+    backupPickup: ChildBackupPickup
+): Int {
+    // language=sql
+    val sql = """
+        UPDATE backup_pickup
+        SET
+            child_id = :childId,
+            name = :name,
+            phone = :phone
+        WHERE id  = :id
```
Joosakur, line 136: nitpick: is there any case where the child id should be updated or would it be safer to add `child_id = :childId` as part of the where-clause?

service/src/main/kotlin/fi/espoo/evaka/backuppickup/BackupPickupController.kt.kt
```diff
@@ -0,0 +1,157 @@
...
+    """.trimIndent()
+    return this.createUpdate(sql)
+        .bind("id", backupPickup.id)
+        .bind("childId", backupPickup.childId)
+        .bind("name", backupPickup.name)
+        .bind("phone", backupPickup.phone)
+        .execute()
```
Joosakur, line 143: this returns Int but the value is not checked in controller. Could also already throw NotFound here as you did with delete

service/src/main/kotlin/fi/espoo/evaka/backuppickup/BackupPickupController.kt.kt
massproduction: Yeah we don't need childId at all in the update function. Removed it. 👍

reynders (approved):

PR #634, opened 2021-03-29
Make sure that decisions are sent only to current guardians

reynders:
#### Summary
Because there might be a long time between submitting an application and creating the decisions for the guardians, make sure that the parents are still custodians.


kustaahuhtala (commented):
This looks good to me, although there will probably be further issues if decision does not get sent to any guardian. I did not yet test this locally.

kustaahuhtala (approved):

PR #637, opened 2021-03-30
Refresh user roles on page load + refactor role data model in apigw slightly

Gekkio:
#### Summary

* pass roles in two arrays to apigw, separating "global" and "scoped" roles. This separation will later be applied to frontend and service code as well to avoid mixing the two types of roles accidentally
* remove old leftover scoped roles from employee table `roles` column, which is supposed to only contain global roles
* add database constraint to prevent scoped roles from being added to the employee table
* refresh apigw session roles in /auth/status endpoint if they have been changed in the database. In practice this means a page reload is enough for changes in user roles to be applied in frontend (no longer needs logout + login!)

### DIFF
--- a/apigw/src/internal/routes/auth-status.ts
+++ b/apigw/src/internal/routes/auth-status.ts
@@ -16,15 +16,16 @@ export default toRequestHandler(async (req, res) => {
   if (user) {
     if (user.userType === 'MOBILE') {
       const device = await getMobileDevice(req, user.id)
+      const globalRoles = user.globalRoles ?? []
+      const allScopedRoles = user.allScopedRoles ?? ['MOBILE']
       if (device) {
         const { id, name, unitId } = device
         res.status(200).json({
           loggedIn: true,
           user: { id, name, unitId },
-          roles: user.roles ?? [
-            ...(user.globalRoles ?? []),
-            ...(user.allScopedRoles ?? [])
-          ]
+          globalRoles,
+          allScopedRoles,
+          roles: [...globalRoles, ...allScopedRoles]
         })
       } else {
         // device has been removed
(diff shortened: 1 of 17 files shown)


service/src/main/kotlin/fi/espoo/evaka/pis/EmployeeQueries.kt
```diff
@@ -59,20 +67,26 @@ WHERE (:id::uuid IS NULL OR e.id = :id)
...
+"""
+)
+fun Database.Read.getEmployeeUser(id: UUID): EmployeeUser? = createEmployeeUserQuery("WHERE id = :id")
+    .bind("id", id)
+    .mapTo<EmployeeUser>()
+    .singleOrNull()
+fun Database.Read.getEmployeeUserByExternalId(externalId: ExternalId): EmployeeUser? = createEmployeeUserQuery("WHERE external_id = :externalId")
+    .bind("externalId", externalId)
+    .mapTo<EmployeeUser>()
+    .singleOrNull()
```
jaahon, line 89: Line breaks between the functions would be nice for readability.

apigw/src/internal/routes/auth-status.ts
```diff
@@ -19,20 +21,39 @@ export default toRequestHandler(async (req, res) => {
         res.status(200).json({
           loggedIn: true,
           user: { id, name, unitId },
-          roles: user.roles
+          roles:
+            user.roles ??
+            concat(user.globalRoles ?? [], user.allScopedRoles ?? [])
```
jaahon, line 26: Wouldn't array spread do the job without concat?

jaahon, line 26: Why not add `globalRoles` to the response as a separate property? Or is the idea to keep using this old `roles` property in the frontend?

Gekkio: I didn't want to include frontend changes here at this point, so it still uses the old property.

But I think I'll add the two new properties to the response, so we can already return the data in the correct format even though the frontend doesn't yet use it -> less chance for breakage during upgrade

jaahon (approved):

PR #638, opened 2021-03-30
[EVAKA] CI workflow for forked PRs

mikkopiu:
#### Summary

- CI: Don't run jobs for fork branches, only origin
   - All jobs currently require secrets due to Slack notifications -> need to stop already at checkout
   - **NOTE:** This does not prevent our actual workflow for fork pull requests where we push the references to origin as then `CIRCLE_BRANCH` will be our name for it, instead of `pull/[0-9]+` and the CircleCI setting "Build forked pull requests" will enable the builds (a bit confusingly, I admit)
- Docs: CI for forked PRs

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -941,6 +941,11 @@ workflows:
     jobs:
       - checkout_repo:
           <<: *default_contexts
+          filters:
+            branches:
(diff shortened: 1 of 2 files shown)


kustaahuhtala:
The documentation looks good to me, but I will leave merging to @mikkopiu .

PR #639, opened 2021-03-31
EVAKA-4150 bulletin notification email content

hruoho:
#### Summary

Add notification email content

https://voltti.atlassian.net/browse/EVAKA-4150

NB: depends on https://github.com/espoon-voltti/evaka-infra/pull/381

### DIFF
--- a/service/src/main/kotlin/fi/espoo/evaka/messaging/bulletin/BulletinNotificationEmailService.kt
+++ b/service/src/main/kotlin/fi/espoo/evaka/messaging/bulletin/BulletinNotificationEmailService.kt
@@ -29,6 +29,8 @@ class BulletinNotificationEmailService(
         asyncJobRunner.sendBulletinNotificationEmail = ::sendBulletinNotification
     }
 
+    val baseUrl: String = env.getRequiredProperty("application.frontend.baseurl")
+    val baseUrlSv: String = env.getRequiredProperty("application.frontend.baseurl.sv")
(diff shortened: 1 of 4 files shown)


service/src/main/kotlin/fi/espoo/evaka/messaging/bulletin/BulletinNotificationEmailService.kt
```diff
@@ -129,35 +129,49 @@ class BulletinNotificationEmailService(
 
         return when (language) {
             Language.en -> "New bulletin in eVaka$postfix"
-            Language.sv -> "Ny$postfix"
+            Language.sv -> "Ny meddelande i eVaka$postfix"
             else -> "Uusi tiedote eVakassa$postfix"
         }
     }
 
+    private fun getCitizenMessagesUrl() = if (System.getenv("VOLTTI_ENV") == "prod") "https://espoonvarhaiskasvatus.fi/messages"
```
kustaahuhtala, line 137: Couple of comments regarding this:
- How does this work in Tampere?
- How about swedish domain; should the URL be https://esbosmabarnspedagogik.fi/messages if certain condition is met (i.e. swedish speaking citizen etc.?). Maybe pass language as a parameter for this function?

We can probably proceed with this implementation but these will bite back later :)

hruoho: Yes I wished to take the easy way out but let's do this properly now 🤡  

Just opened an PR to read base url from env.

kustaahuhtala (approved):

PR #644, opened 2021-04-01
EVAKA-4031 second phone and medication

massproduction:
#### Summary
Added second phone for users and medication info for children


jaahon (commented):
You accidentally included an empty `EmployeePinCode.tsx` file in the PR.

Also I'm not really a fan of the name `phone2`, I think something like `backupPhone` would be better, but it's not a major issue.

service/src/main/resources/db/migration/V54__add_backup_number_to_person.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE person ADD COLUMN phone2 text DEFAULT NULL::character varying;
```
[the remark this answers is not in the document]
massproduction: Yeah, I don't remember why I did it that way when I just could have used an empty string just like with medication. Fixed it to an empty string. Also renamed the phone2 to backupPhone.

service/src/main/kotlin/fi/espoo/evaka/pis/service/PersonService.kt
```diff
@@ -357,6 +359,7 @@ data class PersonDTO(
     val lastName: String?,
     val email: String?,
     val phone: String?,
+    val backupPhone: String? = "",
```
jaahon, line 362: I don't think this field should be nullable since the data can't be null in backend.

service/src/main/kotlin/fi/espoo/evaka/pis/service/PersonService.kt
```diff
@@ -383,6 +386,7 @@ data class PersonJSON(
     val lastName: String? = null,
     val email: String? = null,
     val phone: String? = null,
+    val backupPhone: String? = "",
```
jaahon, line 389: Same here.

service/src/main/kotlin/fi/espoo/evaka/pis/FamilyContact.kt
```diff
@@ -20,6 +20,7 @@ data class FamilyContact(
     val lastName: String?,
     val email: String?,
     val phone: String?,
+    val backupPhone: String?,
```
jaahon, line 23: And here.

frontend/src/employee-frontend/types/person.ts
```diff
@@ -19,6 +19,7 @@ export interface PersonDetails {
   lastName: string | null
   email: string | null
   phone: string | null
+  backupPhone: string | null
```
jaahon, line 22: And here.

jaahon (approved):
Good job!

PR #645, opened 2021-04-01
Employee pin login and child sensitive info view

reynders:
#### Summary

- Allow seeing child's sensitive info after correctly entering PIN on employee mobile app 
- Logout functionality will be part of the next PR
- Call to number -links will be part of the next PR
- - - 

https://www.figma.com/file/8FXipaHxJcHwwfCl273PWS/Kasvattajan-mobiili?node-id=968%3A1210


jaahon:
Most children don't have any family contacts set explicitly, they just use the default values, which are set by [this function](https://github.com/espoon-voltti/evaka/blob/bbb8a879850c781065db45bd6f1b819af81157db/service/src/main/kotlin/fi/espoo/evaka/pis/controllers/FamilyController.kt#L202).

service/src/main/kotlin/fi/espoo/evaka/pis/EmployeeQueries.kt
```diff
@@ -45,7 +45,7 @@ RETURNING id, first_name, last_name, email, external_id, created, updated, roles
 private fun Handle.searchEmployees(id: UUID? = null) = createQuery(
     // language=SQL
     """
-SELECT e.id, first_name, last_name, email, external_id, e.created, e.updated, roles
+SELECT e.id, first_name, last_name, email, external_id, e.created, e.updated, roles, pin
```
Gekkio, line 48: This query change + adding the field to the common `Employee` type leaks employee pins to frontend in **all JSON responses** that use this query.

My recommendation: instead of trying to reuse `Handle.getEmployee` / `Handle.searchEmployees`, answer the actual question you're trying to answer with a separate query: `Handle.isEmployeePinCorrect(employeeId: UUID, pin: String): Boolean`.

service/src/main/kotlin/fi/espoo/evaka/attendance/ChildAttendanceController.kt
```diff
@@ -60,6 +64,34 @@ class ChildAttendanceController(
...
+    fun getChildSensitiveInfo(
+        db: Database.Connection,
+        user: AuthenticatedUser,
+        @PathVariable childId: UUID,
+        @RequestParam pin: String,
+        @RequestParam staffId: UUID
+    ): ResponseEntity<ChildResult> {
+        Audit.ChildSensitiveInfoRead.log(targetId = staffId)
+
+        db.read { it.handle.getEmployee(staffId) }?.let {
```
Gekkio, line 77: - if the `staffId` is not found in the db, the check is silently passed so the pin check can be bypassed trivially by just passing a random UUID as staffId
- there are no role checks for the mobile device, so anybody (even citizens) can call this endpoint
- there are no role checks for the staff ID, so any employee can access information for any child
- one of the requirements was that audit logs show who is the actual employee accessing the data. This doesn't happen, because audit logs show the generic mobile device id

service/src/main/kotlin/fi/espoo/evaka/attendance/ChildAttendance.kt
```diff
@@ -12,6 +12,40 @@ import java.time.Instant
...
+    val diet: String?,
+    val medication: String?,
+    val contact1: ContactInfo?,
+    val contact2: ContactInfo?,
+    val backupPickup1: ContactInfo?,
+    val backupPickup2: ContactInfo?
+)
+
+data class ContactInfo(
+    val firstName: String?,
```
Gekkio, line 42: Do we really need to handle NULLs and empty strings separately? IMHO it's generally simpler to use empty strings for values that haven't been set

service/src/main/kotlin/fi/espoo/evaka/attendance/ChildAttendanceController.kt
reynders: Revised the whole implementation and added the staffId authorization

frontend/src/employee-mobile-frontend/components/attendances/child-info/ChildSensitiveInfo.tsx
```diff
@@ -0,0 +1,238 @@
...
+            paddingHorizontal={'s'}
+            paddingVertical={'m'}
+          >
+            <FixedSpaceColumn alignItems={'center'} spacing={'m'}>
+              <Title>{i18n.attendances.childInfo.header}</Title>
+            </FixedSpaceColumn>
+          </ContentAreaWithShadow>
+
+          <ContentAreaWithShadow opaque={true}>
+            <CollapsibleSection
```
jaahon, line 55: I don't know if you used this component intentionally, but it doesn't match the one used in spec.

frontend/src/employee-mobile-frontend/components/attendances/child-info/ChildSensitiveInfo.tsx
```diff
@@ -0,0 +1,238 @@
...
+  child: Child | null
+}
+
+const KeyValue = styled.div``
+
+const Divider = styled.div`
+  border-bottom: 1px solid #d8d8d8;
+`
+
+const render = (label: string, value: string | null, dataQa: string) => {
```
jaahon, line 22: I think a more descriptive name would be better.

frontend/src/employee-mobile-frontend/components/attendances/child-info/ChildSensitiveInfo.tsx
```diff
@@ -0,0 +1,238 @@
...
+  return value ? (
+    <KeyValue>
+      <Label>{label}</Label>
+      <P fitted data-qa={dataQa}>
```
jaahon, line 26: Why not just use a `<span>`?

frontend/src/employee-mobile-frontend/components/attendances/child-info/ChildSensitiveInfo.tsx
```diff
@@ -0,0 +1,238 @@
...
+                      <P fitted data-qa={'child-info-contact2-name'}>{`${
+                        child.contact2.firstName || ''
+                      } ${child.contact2.lastName || ''}`}</P>
+                    </KeyValue>
+
+                    {render(
+                      i18n.attendances.childInfo.phone,
+                      child.contact2.phone,
+                      'child-info-contact2-phone'
+                    )}
```
jaahon, line 168: The spec has icon telephone links for phone numbers.

service/src/main/kotlin/fi/espoo/evaka/attendance/ChildAttendance.kt
```diff
@@ -12,6 +12,40 @@ import java.time.Instant
...
+    val ssn: String,
+    val childAddress: String,
+    val placementTypes: List<PlacementType>,
+    val allergies: String,
+    val diet: String,
+    val medication: String,
+    val contact1: ContactInfo?,
+    val contact2: ContactInfo?,
+    val backupPickup1: ContactInfo?,
+    val backupPickup2: ContactInfo?
```
jaahon, line 38: Is there a reason why these are a maximum of two people each?

service/src/main/kotlin/fi/espoo/evaka/attendance/ChildAttendanceController.kt
```diff
@@ -60,6 +66,46 @@ class ChildAttendanceController(
...
 
+    @GetMapping("/child/{childId}")
+    fun getChildSensitiveInfo(
+        db: Database.Connection,
+        user: AuthenticatedUser,
+        @PathVariable childId: UUID,
+        @RequestParam pin: String,
+        @RequestParam staffId: UUID
+    ): ResponseEntity<ChildResult> {
+        Audit.ChildSensitiveInfoRead.log(targetId = staffId)
```
jaahon, line 77: Shouldn't the `targetId` be `childId` and the `staffId` could be `objectId`?

service/src/main/kotlin/fi/espoo/evaka/attendance/ChildAttendanceQueries.kt
```diff
@@ -339,3 +357,63 @@ fun Database.Transaction.deleteCurrentDayAbsences(childId: UUID) {
...
+            },
+            contact2 = familyContacts.getOrNull(1)?.let {
+                ContactInfo(
+                    firstName = it.firstName ?: "",
+                    lastName = it.lastName ?: "",
+                    phone = it.phone ?: "",
+                    backupPhone = it.backupPhone ?: "",
+                    email = it.email ?: ""
+                )
+            },
```
jaahon, line 397: The contacts that should be shown in mobile view are those that have a "priority" set, so it's not always a fixed 1-2 people, it could be more. The `fetchFamilyContacts(childId)` query also doesn't set default values.

jaahon (commented):
You should change `pin-login.ts` file name to `pin-login.spec.ts` so that `yarn e2e-mobile` runs it.

frontend/src/employee-mobile-frontend/components/attendances/child-info/ChildSensitiveInfo.tsx
reynders: yes - sorry forgot this from the done - in - next - pr list, added

frontend/src/employee-mobile-frontend/components/attendances/child-info/ChildSensitiveInfo.tsx
reynders: Checking this from UX

reynders: There is an ongoing discussion how this should be solved. Will be fixed in next PR if needed

service/src/main/kotlin/fi/espoo/evaka/attendance/ChildAttendanceQueries.kt
reynders: Revised after discussion

service/src/main/kotlin/fi/espoo/evaka/attendance/ChildAttendance.kt
reynders: Supports now any number of contacts or backup pickups

jaahon (approved):

PR #661, opened 2021-04-08
Split AuthenticatedUser to separate variants - part 1

Gekkio:
#### Summary

* it's now possible to do different things depending on what kind of session the user is using without relying on fragile role checks
* we can add extra fields to only some of the session types (e.g. mobile device session could include id of the actual user)
* this change makes several invalid combinations impossible


jaahon (approved):

jaahon:
🏆

PR #665, opened 2021-04-08
Support running add-license-headers.sh within a submodule

Gekkio:
#### Summary

The reuse tool assumes it can see the git repository root, so we need to
ensure we mount the whole git repository as a docker volume and not just
the working directory, which might be a git submodule and doesn't have a
.git directory.


mikkopiu (approved):
LGTM :+1:

PR #670, opened 2021-04-09
EVAKA-4157 Daily service times (mobile)

akheron:

### DIFF
--- a/frontend/src/employee-mobile-frontend/api/attendances.ts
+++ b/frontend/src/employee-mobile-frontend/api/attendances.ts
@@ -41,6 +41,7 @@ export interface AttendanceChild {
   placementType: PlacementType
   attendance: Attendance | null
   absences: Absence[]
+  dailyServiceTimes: DailyServiceTimes | null
   entitledToFreeFiveYearsOldDaycare: boolean
   dailyNote: DailyNote | null
 }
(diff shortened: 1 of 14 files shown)


frontend/src/employee-mobile-frontend/api/attendances.ts
```diff
@@ -57,6 +58,29 @@ interface Absence {
...
+  monday: TimeRange | null
+  tuesday: TimeRange | null
+  wednesday: TimeRange | null
+  thursday: TimeRange | null
+  friday: TimeRange | null
+}
+
+export type DailyServiceTimes =
+  | DailyServiceTimesRegular
+  | DailyServiceTimesIrregular
```
Joosakur, line 82: Recently we've started using `lib-common` for typings that are shared between multiple frontends. These could potentially also be extracted there instead of duplicating, although since these don't exactly use the same endpoint one could argue that the typings are also separate and just happen to be identical. So just FYI.

Joosakur: https://github.com/espoon-voltti/evaka/tree/master/frontend/src/lib-common/api-types

frontend/src/employee-mobile-frontend/components/attendances/AttendanceDailyServiceTimes.tsx
```diff
@@ -0,0 +1,51 @@
...
+  'wednesday',
+  'thursday',
+  'friday'
+] as const
+
+type DayName = typeof dayNames[number]
+
+function getToday(): DayName {
+  return dayNames[new Date().getDay()]
+}
```
Joosakur, line 23: This might have a bug or two. First I'm wondering if it was so that getDay has 0 for Sunday, so there might be an off-by-one error. Secondly if the current day is saturday/sunday this can return undefined I guess. In those cases we should also just show that noServiceToday text.

( I see this PR is still draft, but already commenting as I don't have much of anything better to do right now. )

Joosakur:
looks great 💎

frontend/src/employee-mobile-frontend/components/attendances/AttendanceDailyServiceTimes.tsx
akheron: Fixed!

frontend/src/employee-mobile-frontend/api/attendances.ts
akheron: Moved to somewehere under `lib-common/api-types`, but not sure about file naming.

Joosakur (approved):

PR #673, opened 2021-04-09
EVAKA-4160 Restrict end user attachments

kimifoud:
#### Summary

Do not return attachments uploaded by employees to end users.

### DIFF
--- a/service/src/main/kotlin/fi/espoo/evaka/application/ApplicationQueries.kt
+++ b/service/src/main/kotlin/fi/espoo/evaka/application/ApplicationQueries.kt
@@ -535,7 +535,8 @@ private val toPersonApplicationSummary: (ResultSet, StatementContext) -> PersonA
     )
 }
 
-fun fetchApplicationDetails(h: Handle, applicationId: UUID): ApplicationDetails? {
+fun fetchApplicationDetails(h: Handle, applicationId: UUID, includeCitizenAttachmentsOnly: Boolean = false): ApplicationDetails? {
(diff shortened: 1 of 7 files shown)


service/src/main/kotlin/fi/espoo/evaka/application/ApplicationQueries.kt
```diff
@@ -566,7 +567,8 @@ fun fetchApplicationDetails(h: Handle, applicationId: UUID): ApplicationDetails?
         LEFT JOIN person g1 ON g1.id = a.guardian_id
         LEFT JOIN (
             SELECT application_id, jsonb_agg(jsonb_build_object('id', id, 'name', name, 'contentType', content_type, 'updated', updated, 'receivedAt', received_at, 'type', type)) json
-            FROM attachment GROUP BY application_id
+            FROM attachment $attachmentWhereClause
```
kimifoud, line 570: This feels hacky. Is there a better way to do this?
We could filter non-citizen attachments in the CitizenController instead?

hruoho: I think this is somewhat related to the upcoming task of indicating which attachments are uploaded by who. There we must expand the current data class so that it has some indication of the uploader and then it might be reasonable to do the filtering on the code level, instead of sql. Of course that data class augmentation could be done here already but I dunno... This is fine anyway, to me at least.

hruoho (approved):

hruoho:
Ah please fix that commit message before merging :)

PR #674, opened 2021-04-11
Fix maybe-null urgent flag in application summary query

kimifoud:
#### Summary

`urgent` might be null, default to false to fix non-nullable field.


Gekkio (approved):
:+1:

PR #683, opened 2021-04-13
EVAKA-4160 Service worker attachments

kimifoud:
#### Summary

* Implement attachment upload for employees
* Extract FileUpload component to lib-components for re-use, add story


hruoho (commented):
Seems good to me! 🏅 Manually tested that palveluohjaaja can add and remove attachments both initially and on later editing stage.

reynders (approved):

PR #686, opened 2021-04-13
Misc. receiver selection UI fixes

mapehe:
#### Summary

Minor messaging UI fixes.


hruoho (commented):
Otherwise seems good but the e2e-tests are broken now that the unit is not selected as a receiver by default 🤡

mapehe:
Thanks for the feedback @hruoho. Updated!

reynders (approved):

PR #688, opened 2021-04-13
EVAKA-4160 Show receivedAt in attachment list

hruoho:
#### Summary

Show `receivedAt` in attachment list by mapping received timestamp string to LocalDate while deserializing (wonder if we should have `LocalDateTime` too).

### DIFF
--- a/frontend/src/employee-frontend/components/common/Attachment.tsx
+++ b/frontend/src/employee-frontend/components/common/Attachment.tsx
@@ -18,10 +18,18 @@ import FileDownloadButton from 'lib-components/molecules/FileDownloadButton'
 import { getAttachmentBlob } from '../../api/applications'
 import { useTranslation } from '../../state/i18n'
 import { UIContext } from '../../state/ui'
+import { defaultMargins } from 'lib-components/white-space'
+import { Dimmed } from 'lib-components/typography'
+import LocalDate from 'lib-common/local-date'
 
 const AttachmentContainer = styled.div`
   display: flex;
   justify-content: flex-start;
+`
+
+const ReceivedAtText = styled(Dimmed)`
+  font-style: italic;
+  margin-left: ${defaultMargins.s};
 `
 
 interface Props {
@@ -68,6 +76,10 @@ function Attachment({ attachment, dataQa }: Props) {
           }
           dataQa={'attachment-download'}
         />
+        <ReceivedAtText>
+          {i18n.application.attachments.receivedAt}{' '}
+          {LocalDate.fromSystemTzDate(attachment.receivedAt).format()}
+        </ReceivedAtText>
       </FixedSpaceRow>
     </AttachmentContainer>
   )
--- a/frontend/src/employee-frontend/components/common/Attachment.tsx
+++ b/frontend/src/employee-frontend/components/common/Attachment.tsx
@@ -76,7 +76,7 @@ function Attachment({ attachment, dataQa }: Props) {
           }
           dataQa={'attachment-download'}
         />
-        <ReceivedAtText>
+        <ReceivedAtText data-qa={`attachment-${attachment.name}-received-at`}>
           {i18n.application.attachments.receivedAt}{' '}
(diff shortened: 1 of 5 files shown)


frontend/src/citizen-frontend/applications/editor/service-need/PreferredStartSubSection.tsx
```diff
@@ -52,8 +52,8 @@ export default React.memo(function PreferredStartSubSection({
                 id: result.value,
                 name: file.name,
                 contentType: file.type,
-                updated: new Date(),
-                receivedAt: new Date(),
+                updated: LocalDate.fromSystemTzDate(new Date()),
```
kimifoud, line 55: `LocalDate.today()` could work here too?

frontend/src/employee-frontend/api/applications.ts
```diff
@@ -33,6 +33,10 @@ export async function getApplication(
   return client
     .get<JsonOf<ApplicationResponse>>(`/v2/applications/${id}`)
     .then((res) => res.data)
+    .then((data) => {
+      console.log(data)
+      return data
+    })
```
kimifoud, line 39:
```suggestion
```

frontend/src/employee-frontend/api/applications.ts
```diff
@@ -49,8 +53,8 @@ export async function getApplication(
       })),
       attachments: data.attachments.map((attachment) => ({
         ...attachment,
-        updated: new Date(attachment.updated),
-        receivedAt: new Date(attachment.receivedAt)
+        updated: LocalDate.fromSystemTzDate(new Date(attachment.updated)),
+        receivedAt: LocalDate.fromSystemTzDate(new Date(attachment.receivedAt))
```
kimifoud, line 57: `LocalDate.parseIso` seems to be the used for ISO string -> LocalDate conversion. Should we use that too?

frontend/src/employee-frontend/components/common/Attachment.tsx
```diff
@@ -68,6 +75,10 @@ function Attachment({ attachment, dataQa }: Props) {
           }
           dataQa={'attachment-download'}
         />
+        <ReceivedAtText>
+          {i18n.application.serviceNeed.receivedAt}{' '}
```
kimifoud, line 79: `application.serviceNeed.receivedAt` seems a little odd place for this. 🤷

frontend/src/employee-frontend/api/applications.ts
hruoho: That was my first guess but resulted to a hard-to-catch bug - the reason being that the function can't parse timestamps, only iso dates

frontend/src/employee-frontend/api/applications.ts
hruoho: Oh dear...

frontend/src/employee-frontend/api/applications.ts
Gekkio: Shouldn't we return dates from the backend then? It seems confusing to me that we're explicitly asking for extra information (the time components in the timestamp) and then immediately throwing it away in the API call deserialization code.

Or, if we can't change the backend payload and just don't want to show the time components to users, it sounds to me like a formatting concern that belongs to the components showing the data.

kimifoud: Imho all dates should be serialized as iso strings in response payloads. The time component might be relevant in another use case so I would not get rid of it.

> just don't want to show the time components to users, it sounds to me like a formatting concern that belongs to the components showing the data.

I think this is the right answer; pass the value around as a Date and render it without the time component. Using LocalDate _everywhere_ in the frontend might not be a viable long-term solution anyway because there will most likely be use cases where we need the time too.

hruoho: Yes I think you guys are absolutely correct here 👍 I was just blindly stuck in the idea of _always_ using `LocalDate` over `Date`.

kimifoud (approved):

PR #689, opened 2021-04-14
EVAKA-4160 urgent due date based on attachments

kimifoud:
#### Summary

Set application due date based on attachments for urgent applications

If the application is urgent:
- the due date should not be set at all if attachments are missing
- the due date is the day application is sent or the first attachment received date, whichever is later

### DIFF
--- a/service/src/integrationTest/kotlin/fi/espoo/evaka/application/ApplicationStateServiceIntegrationTests.kt
+++ b/service/src/integrationTest/kotlin/fi/espoo/evaka/application/ApplicationStateServiceIntegrationTests.kt
@@ -41,6 +41,7 @@ import fi.espoo.evaka.shared.domain.BadRequest
 import fi.espoo.evaka.shared.domain.FiniteDateRange
 import fi.espoo.evaka.shared.domain.Forbidden
 import fi.espoo.evaka.shared.message.MockEvakaMessageClient
+import fi.espoo.evaka.shared.utils.zoneId
 import fi.espoo.evaka.testAdult_1
 import fi.espoo.evaka.testAdult_4
 import fi.espoo.evaka.testAdult_5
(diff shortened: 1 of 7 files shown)


service/src/integrationTest/kotlin/fi/espoo/evaka/application/ApplicationStateServiceIntegrationTests.kt
```diff
@@ -161,11 +177,58 @@ class ApplicationStateServiceIntegrationTests : FullApplicationTest() {
...
-            // then
-            val application = fetchApplicationDetails(it.handle, applicationId)!!
-            assertEquals(LocalDate.now().plusWeeks(2), application.dueDate)
+        // then
+        assertDueDate(applicationId, null) // missing attachment
+
+        // when
+        assertTrue(uploadAttachment(applicationId, AuthenticatedUser.Citizen(testAdult_1.id)))
+        db.transaction { tx ->
+            tx.createUpdate("UPDATE attachment set received_at = :receivedAt")
```
reynders, line 186: This updates all the attachments. Tests should not trust on existing data, but still it would be nice to work only on the specific data the tests actually uses

reynders (approved):

service/src/integrationTest/kotlin/fi/espoo/evaka/application/ApplicationStateServiceIntegrationTests.kt
kimifoud: True that. We are also resetting the database in `@BeforeEach`, so there should not be any other rows that get updated here. I had a plan to support `received_at` as a parameter in the attachment upload endpoint, but it wouldn't be used in production code (yet), so I refrained from doing so.

PR #691, opened 2021-04-14
Set preferredStartDate in test data to 5 months from now

kimifoud:


jaahon (commented):
This approach has the problem that the date might not hit a club or preschool term, which will break the tests again later 😅

jaahon:
I would just remove that one failing assertion since it's not strictly related to the test anyway. And fix this mess later when the tests are migrated to playwright.

kimifoud:
Superseded by #692

PR #694, opened 2021-04-14
Add reset database functionality for E2E tests

akheron:
#### Summary

- Add `/dev-api/reset-db` for e2e tests. It
    - removes all data from the database, and
    -  re-inserts `preschool_term`s, because there's no endpoint to add them.
- Use the db reset in some e2e tests.

In addition:
- Insert dev data in bulk. This saves a few seconds in local server startup.
- Run `docker-compose build` in CI to rebuild local images if they have changed.


service/src/main/resources/dev-data/reset-database.sql
```diff
@@ -13,21 +11,13 @@ BEGIN
...
-    )
+    AND table_name <> 'flyway_schema_history'
   );
   FOR sequence IN
     SELECT sequence_name
     FROM information_schema.sequences
     WHERE sequence_schema = 'public'
   LOOP
-    EXECUTE format('ALTER SEQUENCE %I RESTART', sequence);
+    EXECUTE format('SELECT setval(%L, 1, false)', sequence);
```
Gekkio, line 21: Doesn't this restart all sequences to 1? We've got some sequences which don't start from 1 by default, so this might cause problems

Gekkio (commented):
:heart_eyes:

Gekkio (approved):

Joosakur (approved):
great stuff!

service/src/main/resources/dev-data/reset-database.sql
akheron: Changed the `setval` calls to restart from the correct sequence start value.

PR #697, opened 2021-04-14
EVAKA-4176: new service needs data model

Joosakur:


service/src/main/resources/db/migration/V65__new_service_needs.sql
```diff
@@ -0,0 +1,35 @@
+CREATE TABLE service_need_option (
+    id uuid PRIMARY KEY DEFAULT ext.uuid_generate_v1mc(),
+    created timestamp with time zone DEFAULT now() NOT NULL,
+    updated timestamp with time zone DEFAULT now() NOT NULL,
+
+    name text NOT NULL,
+    valid_placement_types placement_type NOT NULL,
```
jaahon, line 7: This column should be renamed now that it's not an array type.

PR #699, opened 2021-04-14
EVAKA 4152: Require guardian email when an application is sent

mapehe:

### DIFF
--- a/service/src/main/kotlin/fi/espoo/evaka/application/ApplicationForm.kt
+++ b/service/src/main/kotlin/fi/espoo/evaka/application/ApplicationForm.kt
@@ -255,8 +255,8 @@ data class ApplicationForm(
                         postOffice = guardian.postOffice ?: ""
                     ),
                     futureAddress = null,
-                    phoneNumber = "",
-                    email = ""
+                    phoneNumber = guardian.phone ?: "",
+                    email = guardian.email ?: ""
                 ),
                 secondGuardian = null,
                 otherPartner = null,
--- a/frontend/src/citizen-frontend/applications/editor/ApplicationFormData.ts
+++ b/frontend/src/citizen-frontend/applications/editor/ApplicationFormData.ts
@@ -53,6 +53,8 @@ export type ContactInfoFormData = {
   guardianHomeAddress: string
   guardianPhone: string
   guardianEmail: string
+  noGuardianEmail: boolean
+  guardianEmailVerification: string
   guardianFutureAddressExists: boolean
   guardianFutureAddressEqualsChild: boolean
   guardianMoveDate: string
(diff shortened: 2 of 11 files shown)


service/src/main/kotlin/fi/espoo/evaka/application/ApplicationForm.kt
```diff
@@ -255,8 +255,8 @@ data class ApplicationForm(
                         postOffice = guardian.postOffice ?: ""
                     ),
                     futureAddress = null,
-                    phoneNumber = "",
-                    email = ""
+                    phoneNumber = guardian.phone ?: "",
```
mapehe, line 258: Is it ok to default to a known phone number here?

service/src/main/kotlin/fi/espoo/evaka/application/ApplicationForm.kt
```diff
@@ -255,8 +255,8 @@ data class ApplicationForm(
                         postOffice = guardian.postOffice ?: ""
                     ),
                     futureAddress = null,
-                    phoneNumber = "",
-                    email = ""
```
mapehe, line 259: Noticed that I'm not currently storing whether the `noEmail` checkbox is checked. I'll fix that.

mapehe: On the other hand. This may be something we don't event want to store.

service/src/main/kotlin/fi/espoo/evaka/application/ApplicationForm.kt
Joosakur: yeah I think that makes sense

service/src/main/kotlin/fi/espoo/evaka/application/ApplicationForm.kt
Joosakur: I guess when we map backend data to form data we can interpret missing email so that the checkbox is on

Joosakur: no need for a separate flag in backend/db imho

mapehe: Wouldn't that default to the checkbox being checked on creating a new application. We probably don't want that?

The current implementation would require an user that does not have an email address to re-check the checkbox on each update of the application. It sounds like an edge case to me: maybe we should just proceed with this one?

Joosakur: Ah yeah, true that.. Ok then, I guess we can go with "it's not a bug it's a feature". :D

Joosakur (approved):

PR #700, opened 2021-04-15
WCAG contrast compliance tweaks

kimifoud:
#### Summary

- All dimmed text on white background should use #6e6e6e instead of #b1b1b1 to be contrast compliant.
- Same goes for disabled buttons


mikkopiu (approved):
Just to check: UX peeps know about these changes? LGTM but just checking

kimifoud:
> Just to check: UX peeps know about these changes? LGTM but just checking

Yup, the request came from them while we were reviewing a related task 👍

PR #701, opened 2021-04-15
Add experimental HelsinkiDateTime type

Gekkio:
#### Summary

- basically a wrapper for `Instant` with an implicitly defined but guaranteed Europe/Helsinki timezone
- consistent serialization with Jackson and JDBI, which takes care of interpreting the timestamp in the correct time zone. Serialization to/from a zoneless timestamp throws an error, so "local date time" type of values are not accidentally used
- `from()` functions always re-interpret the given timestamp in Europe/Helsinki
- `to*()` functions and serialization always guarantee the resulting value is in Europe/Helsinki
- switch to `HelsinkiDateTime` in pairing functionality as a demo

### DIFF
--- a/service/src/main/kotlin/fi/espoo/evaka/shared/domain/HelsinkiDateTime.kt
+++ b/service/src/main/kotlin/fi/espoo/evaka/shared/domain/HelsinkiDateTime.kt
@@ -1,0 +1,50 @@
+// SPDX-FileCopyrightText: 2017-2021 City of Espoo
+//
+// SPDX-License-Identifier: LGPL-2.1-or-later
+
+package fi.espoo.evaka.shared.domain
+
+import com.fasterxml.jackson.databind.annotation.JsonDeserialize
+import com.fasterxml.jackson.databind.annotation.JsonSerialize
+import com.fasterxml.jackson.databind.util.StdConverter
+import fi.espoo.evaka.shared.utils.zoneId
+import java.time.Clock
+import java.time.Instant
+import java.time.LocalDate
+import java.time.LocalDateTime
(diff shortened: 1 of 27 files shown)


kimifoud (commented):
Looks good to me! The PR summary could be added to `HelsinkiDateTime` javadocs.

service/src/main/kotlin/fi/espoo/evaka/shared/domain/HelsinkiDateTime.kt
```diff
@@ -0,0 +1,100 @@
...
+
+    override fun compareTo(other: HelsinkiDateTime): Int = this.instant.compareTo(other.instant)
+    override fun toString(): String = toZonedDateTime().toString()
+
+    companion object {
+        fun of(date: LocalDate, time: LocalTime): HelsinkiDateTime = from(ZonedDateTime.of(date, time, europeHelsinki))
+        fun of(dateTime: LocalDateTime): HelsinkiDateTime = from(ZonedDateTime.of(dateTime, europeHelsinki))
+        fun now(clock: Clock? = Clock.systemUTC()): HelsinkiDateTime = HelsinkiDateTime(Instant.now(clock))
+        fun from(value: Instant): HelsinkiDateTime = HelsinkiDateTime(value)
+        fun from(value: ZonedDateTime): HelsinkiDateTime = HelsinkiDateTime(value.toInstant())
```
Joosakur, line 90: Is there a difference between `of` and `from`? I guess this mimics java api?

Joosakur (approved):
great stuff!

PR #704, opened 2021-04-15
PIN logout

reynders:
#### Summary
- Make collapsible section match spec (https://voltti.slack.com/archives/C015HMMRRFB/p1618376253025200)
- Logout button
- Logout modal with cancel and accept buttons
- Limited amount of PIN login tries
- Focus on PIN input after user selection
- E2E test2


jaahon (commented):
Like I mentioned in the previous PR, you need to rename `pin-login.ts` to `pin-login.spec.ts` so that our `yarn e2e-x` commands run the tests. I didn't realize you never changed it.

jaahon (commented):
Also, I don't think it makes sense to lock an individual user when there have been multiple failed attempts. In the case someone tries to hack into child's sensitive information it doesn't matter which employee's PIN they try to guess, the whole phone should just lock up. If only a specific employee's PIN gets locked the amount of guesses an attacker gets is 4 * number of employees.

reynders:
> Also, I don't think it makes sense to lock an individual user when there have been multiple failed attempts. In the case someone tries to hack into child's sensitive information it doesn't matter which employee's PIN they try to guess, the whole phone should just lock up. If only a specific employee's PIN gets locked the amount of guesses an attacker gets is 4 * number of employees.

Good points! After a discussion with product management it was decided to go with suggested locking of employee -approach, though. Main reasoning being that locking up the device will cause too much confusion compared to the security risk of locking just the user exposes.

reynders:
> Also, I don't think it makes sense to lock an individual user when there have been multiple failed attempts. In the case someone tries to hack into child's sensitive information it doesn't matter which employee's PIN they try to guess, the whole phone should just lock up. If only a specific employee's PIN gets locked the amount of guesses an attacker gets is 4 * number of employees.

Good points! After a discussion with product management it was decided to go with suggested locking of employee -approach, though. Main reasoning being that locking up the device will cause too much confusion compared to the security risk of locking just the user exposes.

jaahon (commented):
I just realized that the PIN getting locked doesn't actually stop an attacker in any way. They can still continue to try to guess the PIN and if it's correct they get in.

reynders:
> I just realized that the PIN getting locked doesn't actually stop an attacker in any way. They can still continue to try to guess the PIN and if it's correct they get in.

Good catch! A fix and a E2E test for it has been added

service/src/main/kotlin/fi/espoo/evaka/pis/EmployeeQueries.kt
```diff
@@ -192,35 +192,67 @@ WHERE external_id = :externalId
...
     """.trimIndent()
 ).bind("employeeId", employeeId)
     .bind("pin", pin)
     .mapTo<Boolean>()
     .first()
+
+fun Handle.resetEmployeePinFailureCount(employeeId: UUID) = createUpdate(
+    """
+UPDATE employee_pin
+SET failure_count = 0, locked = false
```
jaahon, line 236: I think you should remove the `locked = false` part here as well just in case since the locked state shouldn't be reset here.

jaahon: Apart from this thing this PR should be good to merge.

jaahon (approved):
Good job 🎖️

reynders:
> Good job 🎖️

...eventually :D

PR #705, opened 2021-04-15
EVAKA-4046: Voucher value decision templates

kustaahuhtala:
#### Summary

- Generate voucher value decision based on template
- Add swedish translations to voucher value decision
- Handle different income options for service voucher decision

How to test:
- Run `fi.espoo.evaka.invoicing.service.PdfServiceTest` and uncomment PDF generation. Decision
  PDF gets then created in temp folder.
- Try different languages (`fi` and `sv`) in the test
- Try creating voucher value decision through user interface.


mapehe (approved):
LGTM, tried locally and seems to work as intended.

Small nit: perhaps this could be squashed to a single commit?

PR #706, opened 2021-04-15
New service need options

jaahon:
#### Summary

Add first set of service need options and sync the new data every time service need is updated.


service/src/main/kotlin/fi/espoo/evaka/invoicing/messaging/FeeDecisionGenerationJobProcessor.kt
```diff
@@ -51,8 +52,72 @@ class FeeDecisionGenerationJobProcessor(
...
+        least(p.end_date, sn.end_date) AS end_date,
+        sn.shift_care
+    FROM placement p
+    JOIN service_need sn ON p.child_id = sn.child_id AND daterange(p.start_date, p.end_date, '[]') && daterange(sn.start_date, sn.end_date, '[]')
+    WHERE p.child_id = :childId
+)
+INSERT INTO new_service_need (option_id, placement_id, start_date, end_date, shift_care)
+SELECT (SELECT id FROM service_need_option WHERE name = option_name), placement_id, start_date, end_date, shift_care
+FROM data
+WHERE option_name != 'undefined'
```
Joosakur, line 115: should we temporarily add an option with name illegal or something? after running this to all children in prod we could easily search where those have been used tio find the cases that need fixing. could even create a quick dynamic report of them for supervisors

Joosakur (approved):
Did not validate every line but generally looks good!

service/src/main/kotlin/fi/espoo/evaka/invoicing/messaging/FeeDecisionGenerationJobProcessor.kt
jaahon: We could do that, but it's also trivial to just query the children that get an undefined service need option (using a modified version of this query), so I personally don't see much value in it.

PR #709, opened 2021-04-16
Refactor sfi messages to allow customization

patari:
#### Summary

Messages sent by the app includes words like "Espoo" and "eVaka" which needs to customized for other installations.

### DIFF
--- a/service/src/main/kotlin/fi/espoo/evaka/shared/message/MessageProvider.kt
+++ b/service/src/main/kotlin/fi/espoo/evaka/shared/message/MessageProvider.kt
@@ -5,16 +5,12 @@
 package fi.espoo.evaka.shared.message
 
 interface IMessageProvider {
-    fun get(type: MessageType, lang: MessageLanguage): String
-}
-
-enum class MessageType {
-    DECISION_HEADER,
-    DECISION_CONTENT,
-    FEE_DECISION_HEADER,
-    FEE_DECISION_CONTENT,
-    VOUCHER_VALUE_DECISION_HEADER,
-    VOUCHER_VALUE_DECISION_CONTENT,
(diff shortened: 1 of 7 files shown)


Gekkio:
OWASP errors have been fixed in our master -> this needs rebase

patari:
Done, thanks!

service/src/main/kotlin/fi/espoo/evaka/shared/message/MessageProvider.kt
```diff
@@ -0,0 +1,26 @@
+// SPDX-FileCopyrightText: 2017-2021 City of Espoo
+//
+// SPDX-License-Identifier: LGPL-2.1-or-later
+
+package fi.espoo.evaka.shared.message
+
+interface IMessageProvider {
+    fun get(type: MessageType, lang: MessageLanguage): String
```
Gekkio, line 8: Is there some advantage to using one function + enum with N variants + when checks, vs just using N functions? The enum solution seems more complicated to me.

In other words:

```kotlin
enum class MessageType {
  A,
  B
}

fun get(type: MessageType, lang: MessageLanguage) {
  return when(type) {
    MessageType.A -> // ...
    MessageType.B -> // ...    
  }
}
```

vs

```kotlin
fun getA(lang: MessageLanguage) {
  // ...
}
fun getB(lang: MessageLanguage) {
  // ...
}
```

patari: Yeah, I think you're right. I'll change this

patari: Done

Gekkio (approved):

PR #710, opened 2021-04-16
EVAKA-4160 show attachment uploader 

hruoho:
#### Summary

Show that attachment is (or is not) uploaded by employee. For now this is indicated only in the application read view for the FileUpload component is too general for context dependant details like this (discussed & approved with UX).

Screenshot 2021-04-16 at 10 46 28


kimifoud (approved):
👍

PR #726, opened 2021-04-20
Duplicate fee decision model with new service need

jaahon:
#### Summary

Basically duplicate all fee decision generation code and run it after generating the regular fee decisions. Five year old daycare is still unsolved and there is no migration for old fee decision data yet. The migration is a bit tricky so it needs more planning.


service/src/main/resources/db/migration/V68__fee_decisions_with_new_service_needs.sql
```diff
@@ -0,0 +1,66 @@
...
+    created timestamp with time zone DEFAULT now() NOT NULL,
+    updated timestamp with time zone DEFAULT now() NOT NULL,
+
+    fee_decision_id uuid NOT NULL REFERENCES new_fee_decision (id) ON DELETE CASCADE,
+    child_id uuid NOT NULL REFERENCES person (id),
+    child_date_of_birth date NOT NULL,
+    sibling_discount integer NOT NULL,
+    placement_unit_id uuid NOT NULL REFERENCES daycare (id),
+    placement_type placement_type NOT NULL,
+    service_need_id uuid NOT NULL REFERENCES service_need_option (id),
```
Joosakur, line 52: might rename this as `service_need_option_id`, now it could be very confusing

Joosakur (approved):

PR #728, opened 2021-04-20
EVAKA-3938 Employees page

kimifoud:
#### Summary

- Add Employees page for searching employees and their rights
- The page and search endpoint can be accessed by admins

#### TODO

- [x] e2e test


Joosakur (commented):
Should we exclude mobile users? They aren't really employees/users at all but devices which just happen to be in the same table for tech reasons.

Joosakur (approved):

PR #729, opened 2021-04-20
Map search area rect coordinates customizable

apeltola:


frontend/src/lib-customizations/types.d.ts
```diff
@@ -12,6 +12,12 @@ export interface CitizenCustomizations {
     src: string
     alt: string
   }
+  mapSearchAreaRect: {
+    minLongitude: number
+    maxLongitude: number
+    minLatitude: number
+    maxLatitude: number
+  }
```
mikkopiu, line 20: I'd rather put these into `MapConfig` as that is just our own config type for everything related to the map.

mikkopiu:
Triggered CI

mikkopiu (approved):

mikkopiu:
Service & message-service builds are failing due to an OWASP/NVD issue (will be fixed in master / externally) -> looking into it already.

PR #731, opened 2021-04-21
EVAKA-UX: use async button quick mode on commonly used mobile actions

Joosakur:

### DIFF
--- a/frontend/src/employee-mobile-frontend/components/attendances/actions/MarkAbsent.tsx
+++ b/frontend/src/employee-mobile-frontend/components/attendances/actions/MarkAbsent.tsx
@@ -122,6 +122,7 @@ export default React.memo(function MarkAbsent() {
                 {selectedAbsenceType ? (
                   <AsyncButton
                     primary
+                    quick
                     text={i18n.common.confirm}
                     onClick={() => postAbsence(selectedAbsenceType)}
                     onSuccess={() => history.goBack()}
--- a/frontend/src/employee-mobile-frontend/components/attendances/actions/MarkAbsent.tsx
+++ b/frontend/src/employee-mobile-frontend/components/attendances/actions/MarkAbsent.tsx
@@ -122,7 +122,6 @@ export default React.memo(function MarkAbsent() {
                 {selectedAbsenceType ? (
                   <AsyncButton
                     primary
-                    quick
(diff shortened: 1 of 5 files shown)


jaahon:
Instead of having a `quick` prop would it make more sense to just make `onSuccess` optional and call the callback immediately when the promise resolves? And maybe rename `onSuccess` to something like `afterSuccessAnimation` and reduce the animation duration.

Joosakur:
> Instead of having a `quick` prop would it make more sense to just make `onSuccess` optional and call the callback immediately when the promise resolves? And maybe rename `onSuccess` to something like `afterSuccessAnimation` and reduce the animation duration.

Not sure I understand. Do you propose that there would be two optional callbacks, `onSuccess` and `afterSuccessAnimation`, where the first one is called immediately after promise is resolved and the latter after animation? So if in this case I would want to skip the animation, I would put the `history.goBack()` to `onSuccess` instead of `afterSuccessAnimation`? Although I'm not sure I even want to skip the animation entirely.

jaahon:
> > Instead of having a `quick` prop would it make more sense to just make `onSuccess` optional and call the callback immediately when the promise resolves? And maybe rename `onSuccess` to something like `afterSuccessAnimation` and reduce the animation duration.
> 
> Not sure I understand. Do you propose that there would be two optional callbacks, `onSuccess` and `afterSuccessAnimation`, where the first one is called immediately after promise is resolved and the latter after animation? So if in this case I would want to skip the animation, I would put the `history.goBack()` to `onSuccess` instead of `afterSuccessAnimation`? Although I'm not sure I even want to skip the animation entirely.

What I mean is that the `onClick` callback can just do the navigation when the promise resolves, there is no need to do it in a separate callback.

But, another idea I got is that the `onSuccess` callback would be fired after about 500ms, but the success state would just be shown for a longer time. That way users see the success state for longer, but are not slowed down in cases where you are navigated out anyway and the successful state is implied anyway.

Joosakur:
> But, another idea I got is that the `onSuccess` callback would be fired after about 500ms, but the success state would just be shown for a longer time. That way users see the success state for longer, but are not slowed down in cases where you are navigated out anyway and the successful state is implied anyway.

I decided to try this approach now

jaahon (approved):

PR #738, opened 2021-04-22
[EVAKA-HOTFIX] Support Single Logout w/o 3rd party cookies

mikkopiu:
#### Summary of SLO issue / vulnerability

- Many browsers are starting to disable 3rd party cookies by default (even SameSite None) which breaks Single Logout with passport-saml that currently only supports logging out with a cookie
    - Browser details: https://webkit.org/blog/10218/full-third-party-cookie-blocking-and-more/ and https://blog.google/products/chrome/more-intuitive-privacy-and-security-controls-chrome/
    - `passport-saml` issues: https://github.com/node-saml/passport-saml/issues/419 and https://github.com/node-saml/passport-saml/issues/221
        - This also causes an usability issue when the user tries to initate SLO from another application, fails to end the eVaka session, attempts a logout from eVaka and gets an ugly JSON error message as a response when APIGW attempts to make a LogoutRequest to the IdP that already ended the session -> will be fixed separately

1. User logs into eVaka via IdP
2. User logs into service B via same IdP SSO session
3. User starts Single Logout from service B and IdP initiates SAML `LogoutRequest`s via the User's browser
4. User's browser blocks 3rd party cookies, i.e. eVaka's logout cookie/token
5. eVaka **incorrectly** responds with SAML success message and User **thinks they've logged out of eVaka but haven't** <- the major security issue
6. Another entity/person with access to the same computer can use User's eVaka session

#### Summary
- apigw: support SLO without cookies, drop logout cookie
    - Other systems like Shibboleth get around the 3rd party cookie issue with Single Logout by not utilizing cookies at all for this but instead use the SAML nameID (and sessionIndex) properties presented in all SAML LogoutRequest messages
        - Source: https://wiki.shibboleth.net/confluence/display/DEV/IdP+SameSite+Testing#IdPSameSiteTesting-SameSiteandSingleLogout
    - When logouts are always only done through SAML there's no need for the logout cookie itself but the idea is actually useful as an effective "secondary index" for Redis:
        - By storing a nameID + sessionIndex keyed item pointing to the session ID, we effectively create a second index that can be used with just the SAML LogoutRequest's nameID and sessionIndex properties
- apigw: clarify logout start error logging
    - No SAML error will ever be present when starting the logout flow from the SP (eVaka itself) -> don't waste time attempting to parse SAML errors and clarify the error description
- apigw: split SAML config creation from `Strategy` creation
    - Allows separate instation of the `SAML` class (from `passport-saml`) used for parsing SAML messages for SLO
    - **NOTE:** This would be entirely unnecessary if `passport-saml` either supported hooking into `passport.authenticate()` _before_ it automatically redirects on `SAMLRequest` or if it had separate methods for first parsing the request and _then_ doing everything else (redirects etc.)
- apigw: don't use deprecated `body-parser` directly
    - Express still actually uses the same body-parser package/methods behind the scenes but as the library user we should use the official wrapper methods `express.json()` and `express.urlencoded()`
    - Options stay as-is as the implementations stay as-is
- apigw: Extend `GatewayTester` to allow further cookie manipulation and non-dummy authentication
- apigw: add test certificates for unit testing
- apigw: add dependencies and features to test tooling for testing the new SAML functionality & mocking required parts of app setup and authentication
- Add `.editorconfig` for `apigw/` (matches `frontend/.editorconfig`)
- apigw: enable `@typescript/no-floating-promises` rule for ESLint (can't forget to `await` a `Promise`, already enabled in frontend projects)

##### Testing instructions

On **master** version (e.g. staging):

1. Go to https://staging.espoonvarhaiskasvatus.fi/ -> login with Suomi.fi (any user)
1. Go to https://staging.espoositoavustukset.fi/nuoriso/ **in the same browser session** -> login with Suomi.fi -> will use the same IdP session
1. Open dev tools for request logs
1. Attempt to log out in https://staging.espoositoavustukset.fi/nuoriso/
1. -> Suomi.fi will trigger Single Logout (SLO) for https://staging.espoonvarhaiskasvatus.fi/api/application/auth/saml/logout/callback (with a `SAMLRequest` query parameter)
1. -> eVaka will respond with 302 **with a `SAMLResponse` indicating success**
1. -> Suomi.fi shows "Sinut on jo kirjattu ulos seuraavista palveluista		Espoon kaupungin varhaiskasvatus 	Kirjauduttu ulos"
1. Go to eVaka tab and attempt to log out
1. -> You will get an (ugly) JSON error response page with `{"message":"Internal server error"}` + an alert

On **dev** (once this branch is deployed):

1. Repeat process
1. -> you will be successfully logged out without any errors

- [ ] Manually tested in staging against Suomi.fi
- [ ] Manually tested in staging against Azure AD
- [ ] Manually tested in staging against Keycloak
- [ ] Manually tested in staging with mobile employee authentication


apigw/src/shared/routes/auth/saml/index.ts
```diff
@@ -177,7 +183,13 @@ export default function createSamlRouter(config: SamlEndpointConfig): Router {
       req,
       'Logout callback called'
     )
-    await logoutExpress(req, res, config.sessionType)
+    const profile = await tryParseProfile(req, saml)
+    await logoutExpress(
+      req,
+      res,
+      config.sessionType,
+      profile?.nameID && createLogoutToken(profile.nameID, profile.sessionIndex)
```
mikkopiu, line 191: Replacing this with `undefined` should effectively simulate the situation in `master` when running the regression tests: if nothing is parsed from the SAML message, only cookies are used (and the tests don't currently care about `SameSite` in this context, so `evaka.eugw.session` will be available in the "IdP-initiated logout" test case).

apigw/src/shared/__tests__/saml-slo.ts
```diff
@@ -0,0 +1,360 @@
...
+  afterEach(async () => {
+    await tester.afterEach()
+    await fromCallback((cb) => redisClient.flushall(cb))
+  })
+  afterAll(async () => await tester?.stop())
+
+  test('reference case (3rd party cookies available)', async () => {
+    await tester.client.get(SECURED_ENDPOINT, {
+      validateStatus: (status) => status === 401
+    })
```
mikkopiu, line 136: validateStatus always true -> expect -> stack traces

apigw/src/shared/auth/index.ts
```diff
@@ -68,3 +70,50 @@ function createJwtToken(user: SamlUser): string {
...
+ * @param saml Config must match active strategy's config
+ */
+export async function tryParseProfile(
+  req: Request,
+  saml?: SAML
+): Promise<Profile | undefined> {
+  if (!saml) {
+    logDebug('No SAML parser provided, skipping profile parsing from request')
+    return
+  }
```
mikkopiu, line 92: No longer relevant

Gekkio (approved):

PR #747, opened 2021-04-23
Automagically log out after 120 seconds of inactivity

reynders:
After logging in with PIN, 
- log out after 120 seconds of inactivity OR
- log out on blur (for example screen lock)
Tested to work on IOS 14 device

### DIFF
--- a/frontend/src/employee-mobile-frontend/components/attendances/child-info/AutoPinLogout.tsx
+++ b/frontend/src/employee-mobile-frontend/components/attendances/child-info/AutoPinLogout.tsx
@@ -1,0 +1,52 @@
+import React, { useCallback, useEffect } from 'react'
+
+interface Props {
+  timeout: number
+  onTimeOut: () => void
+}
+
+export default function AutoLogout({ timeout, onTimeOut }: Props) {
+  const logout = useCallback(() => {
+    clearTimeout()
+    onTimeOut && onTimeOut()
+  }, [onTimeOut])
+
+  useEffect(() => {
+    let logoutTimeout: number
+
+    const setTimeouts = () => {
+      logoutTimeout = setTimeout(logout, timeout)
+    }
+
+    const clearTimeouts = () => {
+      if (logoutTimeout) clearTimeout(logoutTimeout)
+    }
+    const events = [
+      'load',
+      'mousemove',
+      'mousedown',
+      'click',
+      'scroll',
+      'keypress'
+    ]
+
+    const resetTimeout = () => {
+      clearTimeouts()
+      setTimeouts()
+    }
+
+    for (const event of events) {
+      window.addEventListener(event, resetTimeout)
+    }
+
+    setTimeouts()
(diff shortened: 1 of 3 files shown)


frontend/src/employee-mobile-frontend/components/attendances/child-info/AutoPinLogout.tsx
```diff
@@ -0,0 +1,55 @@
+import { useCallback, useEffect } from 'react'
+
+export default function useAutoLogout(timeout: number, onTimeOut: () => void) {
+  const logout = useCallback(() => {
+    clearTimeout()
+    onTimeOut && onTimeOut()
+  }, [onTimeOut])
```
Joosakur, line 7: Abstraction level seems inconsistent. This takes onTimeOut callback which could be anything, but part of the naming implies that this hook is logout specific. I guess you can go either direction, but if the logout logic has to be passed in anyway and nothing inside this is really logout specific, I would probably remove mentions of logout and make this a generic useTimeOut or useInactivity hook.

frontend/src/employee-mobile-frontend/components/attendances/child-info/AutoPinLogout.tsx
```diff
@@ -0,0 +1,55 @@
...
+    }
+
+    const logoutOnResume = () => {
+      clearTimeouts()
+      logout()
+    }
+
+    for (const event of events) {
+      window.addEventListener(event, resetTimeout, { passive: true })
+      window.addEventListener('blur', logoutOnResume, { passive: true })
```
Joosakur, line 41: should this be outside for loop? now blur handler gets added 7 times

frontend/src/employee-mobile-frontend/components/attendances/child-info/AutoPinLogout.tsx
reynders: Renamed to useInactivityTimeout

frontend/src/employee-mobile-frontend/components/attendances/child-info/AutoPinLogout.tsx
reynders: Good catch, fixed!

frontend/src/employee-mobile-frontend/components/attendances/child-info/InactivityTimeout.tsx
```diff
@@ -0,0 +1,58 @@
...
+  timeout: number,
+  onTimeOut: () => void
+) {
+  const logout = useCallback(() => {
+    clearTimeout()
+    onTimeOut && onTimeOut()
+  }, [onTimeOut])
+
+  useEffect(() => {
+    let logoutTimeout: number
```
Joosakur, line 13: vars on lines 7 and 13 could still be renamed

Joosakur (approved):

PR #756, opened 2021-04-27
EVAKA-4190 new messaging data model and initial implementation

kimifoud:
#### Summary

Initial database model for bidirectional messaging. Mock data for frontend development.

### DIFF
--- a/service/src/main/kotlin/fi/espoo/evaka/messaging/message/MessageController.kt
+++ b/service/src/main/kotlin/fi/espoo/evaka/messaging/message/MessageController.kt
@@ -1,0 +1,153 @@
+// SPDX-FileCopyrightText: 2017-2021 City of Espoo
+//
+// SPDX-License-Identifier: LGPL-2.1-or-later
+
+package fi.espoo.evaka.messaging.message
+
+import fi.espoo.evaka.shared.Paged
+import fi.espoo.evaka.shared.auth.AuthenticatedUser
+import fi.espoo.evaka.shared.auth.UserRole
+import fi.espoo.evaka.shared.db.Database
+import fi.espoo.evaka.shared.domain.Forbidden
+import fi.espoo.evaka.shared.domain.HelsinkiDateTime
+import org.springframework.web.bind.annotation.GetMapping
+import org.springframework.web.bind.annotation.PostMapping
+import org.springframework.web.bind.annotation.RequestBody
+import org.springframework.web.bind.annotation.RequestMapping
+import org.springframework.web.bind.annotation.RequestParam
+import org.springframework.web.bind.annotation.RestController
+import java.util.UUID
+
+data class UnreadMessagesResponse(val count: Int)
+
+@RestController
+@RequestMapping("/messages")
+class MessageController {
+
+    @GetMapping("/my-accounts")
+    fun getAccountsByUser(db: Database.Connection, user: AuthenticatedUser): List<MessageAccount> {
+        return db.read { it.getMessageAccountsForUser(user) }
+    }
+
+    @GetMapping("/threads")
+    fun getThreads(
+        db: Database.Connection,
+        user: AuthenticatedUser,
+        @RequestParam accountId: UUID,
+        @RequestParam pageSize: Int,
+        @RequestParam page: Int,
+    ): Paged<MessageThread> {
+        if (!db.read { it.getMessageAccountsForUser(user) }.map { it.id }.contains(accountId))
+            throw Forbidden("User is not authorized to access the account")
+        return db.read { it.getMessagesByAccount(accountId, pageSize, page) }
+    }
+
+    @GetMapping("/unread")
+    fun getUnreadMessages(
+        db: Database.Connection,
+        user: AuthenticatedUser
+    ): UnreadMessagesResponse {
+        val accountIds = db.read { it.getMessageAccountsForUser(user) }.map { it.id }
+        val count = if (accountIds.isEmpty()) 0 else db.read { it.getUnreadMessages(accountIds) }
+        return UnreadMessagesResponse(count)
+    }
+
+    data class PostMessageBody(
+        val title: String,
+        val content: String,
+        val type: MessageType,
+        val senderAccountId: UUID,
+        val recipientAccountIds: List<UUID>
+    )
+
+    @PostMapping
+    fun createMessage(
+        db: Database.Connection,
+        user: AuthenticatedUser,
+        @RequestBody body: PostMessageBody
+    ): UUID {
+        user.requireOneOfRoles(UserRole.UNIT_SUPERVISOR, UserRole.STAFF, UserRole.SPECIAL_EDUCATION_TEACHER)
+        val sender = db.read { it.getMessageAccountsForUser(user) }.find { it.id == body.senderAccountId }
+            ?: throw Forbidden("User is not authorized to access the account")
+
+        // TODO recipient account authorization
+
+        return db.transaction {
+            it.createMessageThread(
+                title = body.title,
+                content = body.content,
+                sender = sender,
+                type = body.type,
+                recipientAccountIds = body.recipientAccountIds
+            )
+        }
+    }
+
+    val supervisorAccount: MessageAccount = MessageAccount(id = UUID.randomUUID(), name = "Esimies Ella")
+    val aatusMomAccount: MessageAccount = MessageAccount(id = UUID.randomUUID(), name = "Aatun äiti")
+    val aatusDadAccount: MessageAccount = MessageAccount(id = UUID.randomUUID(), name = "Aatun isä")
+
+    @GetMapping
+    fun getThreadsMock(): Paged<MessageThread> {
+        val aatuMeritahtiMessage = Message(
+            id = UUID.randomUUID(),
+            senderId = supervisorAccount.id,
+            senderName = supervisorAccount.name,
+            title = "Aatun siirtyminen Meritähtiin",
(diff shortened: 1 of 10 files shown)


service/src/main/resources/db/migration/R__message_account_name_view.sql
```diff
@@ -0,0 +1,17 @@
+CREATE VIEW message_account_name_view(
```
Joosakur, line 1: I guess this could already start with 'drop view if exists' statement to make future changes easier

service/src/main/resources/db/migration/V72__messages.sql
```diff
@@ -0,0 +1,63 @@
...
+    created timestamp with time zone DEFAULT now() NOT NULL,
+    updated timestamp with time zone DEFAULT now() NOT NULL,
+    daycare_group_id uuid references daycare_group(id),
+    employee_id uuid references employee(id),
+    person_id uuid references person(id)
+);
+CREATE TRIGGER set_timestamp BEFORE UPDATE ON message_account FOR EACH ROW EXECUTE PROCEDURE trigger_refresh_updated();
+
+ALTER TABLE message_account ADD CONSTRAINT message_account_created_for_fk CHECK (
+    num_nonnulls(daycare_group_id, employee_id, person_id) = 1
```
Joosakur, line 12: wow, didn't know of this function nor would have thought to use it here, nice!

service/src/main/resources/db/migration/R__message_account_name_view.sql
kimifoud: Ach, of course. 👍

service/src/main/resources/db/migration/V72__messages.sql
```diff
@@ -0,0 +1,63 @@
...
+
+CREATE INDEX idx$message_account_daycare_group_id ON message_account (daycare_group_id);
+CREATE INDEX idx$message_account_employee_id ON message_account (employee_id);
+CREATE INDEX idx$message_account_person_id ON message_account (person_id);
+
+CREATE TABLE message (
+    id uuid PRIMARY KEY DEFAULT ext.uuid_generate_v1mc(),
+    created timestamp with time zone DEFAULT now() NOT NULL,
+    updated timestamp with time zone DEFAULT now() NOT NULL,
+    title text,
```
Joosakur, line 30: should this be not null? empty string is still possible

kimifoud: The UX design indicated that thread replies do not have titles so I figured null would be semantically more correct than empty string.

service/src/main/resources/migrations.txt
```diff
@@ -70,3 +70,4 @@ V69__five_year_old_daycare_placement_types.sql
 V70__employee_pin.sql
 V71__club_term_table.sql
 V72__staff_attendance_other.sql
+V5000__temp__messages.sql
```
mapehe, line 73: Just adding a reminder to check this migration number before merge.

service/src/main/kotlin/fi/espoo/evaka/messaging/message/MessageQueries.kt
```diff
@@ -0,0 +1,163 @@
...
+                m.content,
+                m.title,
+                m.created,
+                sender.account_id AS sender_id,
+                name_view.account_name as sender_name
+         FROM message m
+         JOIN thread_participation sender ON sender.message_id = m.id AND sender.sender = TRUE
+         JOIN message_account_name_view name_view ON sender.account_id = name_view.id
+         ORDER BY m.created ASC
+     ),
```
mapehe, line 65: I don't personally like using a large number of sub-query aliases like this. However, it is certainly in line with how things are done in this project.

I would love it if we could reduce SQL complexity in favour of Kotlin complexity, but that would probably result in a performance hit. 🤷

service/src/main/kotlin/fi/espoo/evaka/messaging/message/MessageQueries.kt
```diff
@@ -0,0 +1,163 @@
...
+    val batch = this.prepareBatch(insertParticipantsSql)
+
+    val participants = listOf(Pair(sender.id, true)) + recipientAccountIds.map { Pair(it, false) }
+    for (p in participants) {
+        batch
+            .bind("messageId", messageId)
+            .bind("accountId", p.first)
+            .bind("sender", p.second)
+            .add()
+    }
```
mapehe, line 158: Nit: I would probably not define `participants` at all and just use `forEach` on the anonymous list on the right hand side of the definition on line 151.

service/src/main/kotlin/fi/espoo/evaka/messaging/message/MessageController.kt
```diff
@@ -0,0 +1,153 @@
...
+import fi.espoo.evaka.shared.domain.HelsinkiDateTime
+import org.springframework.web.bind.annotation.GetMapping
+import org.springframework.web.bind.annotation.PostMapping
+import org.springframework.web.bind.annotation.RequestBody
+import org.springframework.web.bind.annotation.RequestMapping
+import org.springframework.web.bind.annotation.RequestParam
+import org.springframework.web.bind.annotation.RestController
+import java.util.UUID
+
+data class UnreadResponse(val unreadCount: Int)
```
mapehe, line 21: Nit: Something like `UnreadCountResponse` would probably be slightly more idiomatic. I would also consider dropping this entirely and just returning `ResponseEntity<Int>` on the corresponding endpoints.

service/src/main/kotlin/fi/espoo/evaka/messaging/message/MessageControllerCitizen.kt
```diff
@@ -0,0 +1,34 @@
...
+@RequestMapping("/citizen/messages")
+class MessageControllerCitizen {
+
+    @GetMapping("/my-accounts")
+    fun getAccountsByUser(db: Database.Connection, user: AuthenticatedUser): List<MessageAccount> {
+        user.requireOneOfRoles(UserRole.END_USER)
+        return db.read { it.getMessageAccountsForUser(user) }
+    }
+
+    @GetMapping("/unread")
```
mapehe, line 24: Nit: Something like `/unread-count` would probably be slightly more idiomatic.

service/src/main/kotlin/fi/espoo/evaka/messaging/message/MessageController.kt
```diff
@@ -0,0 +1,153 @@
...
+        @RequestParam accountId: UUID,
+        @RequestParam pageSize: Int,
+        @RequestParam page: Int,
+    ): Paged<MessageThread> {
+        if (!db.read { it.getMessageAccountsForUser(user) }.map { it.id }.contains(accountId))
+            throw Forbidden("User is not authorized to access the account")
+        return db.read { it.getMessagesByAccount(accountId, pageSize, page) }
+    }
+
+    @GetMapping("/unread")
```
mapehe, line 45: Nit: Something like `/unread-count` would probably be slightly more idiomatic.

service/src/main/kotlin/fi/espoo/evaka/messaging/message/MessageController.kt
```diff
@@ -0,0 +1,153 @@
...
+import org.springframework.web.bind.annotation.RequestMapping
+import org.springframework.web.bind.annotation.RequestParam
+import org.springframework.web.bind.annotation.RestController
+import java.util.UUID
+
+data class UnreadResponse(val unreadCount: Int)
+
+@RestController
+@RequestMapping("/messages")
+class MessageController {
```
mapehe, line 25: None of these endpoints require any specific ACL at the moment. Should they?

This is almost certainly a non-issue, but it seems that not all roles are relevant here. It might not be a terrible idea to limit these endpoints to the relevant roles.

mapehe (commented):
LGTM. Just some nitty-gritty stuff. Bonus points for bothering with the tests in such an early stage 💎

We probably don't want this on production yet, I'm. not sure how to make sure things are behind a feature flag. I'd be happy to join a call where we discuss that.

EDIT: One more thing. The commit history is probably not worth storing. I'd personally just squash that.

mapehe (approved):

service/src/main/kotlin/fi/espoo/evaka/messaging/message/MessageController.kt
kimifoud: Changed as discussed.

service/src/main/kotlin/fi/espoo/evaka/messaging/message/MessageControllerCitizen.kt
kimifoud: Changed as discussed.

service/src/main/kotlin/fi/espoo/evaka/messaging/message/MessageController.kt
kimifoud: Changed as discussed.

service/src/main/kotlin/fi/espoo/evaka/messaging/message/MessageController.kt
kimifoud: Good point. Created a separate task for authorization.

service/src/main/kotlin/fi/espoo/evaka/messaging/message/MessageQueries.kt
kimifoud: Yeah, this approach gets messy fast.
What we could do here is break this up two separate queries: fetch threads with paging first and their messages in another query. What do you think?

PR #759, opened 2021-04-27
Add script for manual initial club term data insert

kimifoud:
#### Summary

This should be run in staging and production after deployment.


mikkopiu (changes requested):
I'd like some documentation on this directory's/script's existence. Also, if these are _only_ relevant for Espoo -> maybe https://github.com/espoon-voltti/evaka-infra/tree/master/sql-scripts would be the more correct place?

kimifoud:
Irrelevant now. Should be handled with an admin UI in the near future.

PR #761, opened 2021-04-27
Evaka 3914 pin frozen info desktop

reynders:
https://www.figma.com/file/UyFTAPt5kFhErtP56e7Ze8/eVaka-employee?node-id=3536%3A0

### DIFF
--- a/frontend/src/e2e-playwright/specs/5_employee/employee-pin.spec.ts
+++ b/frontend/src/e2e-playwright/specs/5_employee/employee-pin.spec.ts
@@ -1,0 +1,60 @@
+// SPDX-FileCopyrightText: 2017-2021 City of Espoo
+//
+// SPDX-License-Identifier: LGPL-2.1-or-later
+
+import { newBrowserContext } from 'e2e-playwright/browser'
+import EmployeeNav from 'e2e-playwright/pages/employee/employee-nav'
+import { waitUntilEqual } from 'e2e-playwright/utils'
+import config from 'e2e-test-common/config'
+import { resetDatabase } from 'e2e-test-common/dev-api'
+import { Page } from 'playwright'
+import { EmployeePinPage } from '../../pages/employee/employee-pin'
+import { Fixture } from '../../../e2e-test-common/dev-api/fixtures'
+
+let page: Page
+let nav: EmployeeNav
+let pinPage: EmployeePinPage
+
+beforeEach(async () => {
+  await resetDatabase()
+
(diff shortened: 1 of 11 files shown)


frontend/src/e2e-playwright/specs/5_employee/employee-pin.spec.ts
```diff
@@ -0,0 +1,60 @@
...
+  await nav.login('ADMIN')
+  pinPage = new EmployeePinPage(page)
+})
+
+describe('Employees PIN', () => {
+  beforeEach(async () => {
+    await nav.openAndClickDropdownMenuItem('pinCode')
+  })
+
+  test('shows a warning if PIN is too easy, and warning disappears once PIN is valid', async () => {
```
mikkopiu, line 33: This isn't exactly what was changed here but still: what should happen when the user inputs four letters or other non-numeric symbols? Currently it looks like the input only checks that input length == 4:

```ts
  function isValidNumber(pin: string) {
    const int = parseInt(pin)
    const str = int.toString()
    return str.length === 4
  }
```

-> `0x1000` would be accepted as `parseInt` (as no `radix` is defined) would interpret is as hexadecimal 4096 -> 4 characters as a string. Currently the PIN would still be stored as the string "0x1000" but that would be confusing when prompted for a 4-number token.

Also, looks like anything beginning with 0 wouldn't be a valid PIN. Is that desired?

mikkopiu (commented):
The actual change LGTM but I had one question about the actual PIN input logic (see above).

frontend/src/employee-frontend/components/employee/EmployeePinCodePage.tsx
```diff
@@ -2,31 +2,38 @@
...
+  useEffect(() => {
+    void isPinCodeLocked().then(setPinLocked)
+  }, [setPinLocked])
 
   function isValidNumber(pin: string) {
     const int = parseInt(pin)
-    const str = int.toString()
-    return str.length === 4
+    const str = int.toString().padStart(4, '0')
+    return pin == str && pin.length === 4
```
mikkopiu, line 36: It might just be my preference for regex but this could be slightly more readable/robust:

```ts
function isValidNumber(pin: string) {
  return /^\d{4}$/.test(pin)
}
```

That would be exactly 4 characters, all of which must be explicitly digits 0-9.

mikkopiu (approved):
LGTM

PR #763, opened 2021-04-27
Add Tampere AD staging certificate

patari:
#### Summary

It was agreed some time ago that AD certificates will reside inside the app.


mikkopiu (approved):
LGTM with the current setup but I'll create a task to figure out how these could be removed from evaka upstream (it would be really inconvenient to have to go through this whole upstreaming & downstream updating workflow just to update an expired certificate).

PR #766, opened 2021-04-28
Fix flaky value decision e2e test

jaahon:
#### Summary

The test failed because `AsyncButton`'s delayed `onSuccess` callback was run after the component had already unmounted causing a wrong set of data to be displayed in the list. Do a proper clean up of all the `setTimeout`s when the component is unmounted.

### DIFF
--- a/frontend/src/lib-components/atoms/buttons/AsyncButton.tsx
+++ b/frontend/src/lib-components/atoms/buttons/AsyncButton.tsx
@@ -6,7 +6,7 @@ SPDX-License-Identifier: LGPL-2.1-or-later
 */
 }
 
-import React, { useEffect, useState } from 'react'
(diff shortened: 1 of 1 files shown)


akheron:
It seems to me that `onSuccessRef` is not needed. If the `onSuccess` prop changes (or the component is unmounted), the timeouts are cancelled and new ones started.

jaahon:
> It seems to me that `onSuccessRef` is not needed. If the `onSuccess` prop changes (or the component is unmounted), the timeouts are cancelled and new ones started.

I thought about that too, but then if the callback is changed a lot the delay grows every time. I don't know if that is a relevant consern. Maybe we should just go with the behaviour that is less surprising, but I'm not sure what that would be.

reynders (approved):

akheron (approved):

PR #769, opened 2021-04-28
EVAKA-4176: CRUD for new service needs

Joosakur:
#### Summary
Also UI improvements to child-information page


frontend/src/employee-frontend/components/child-information/Assistance.tsx
```diff
@@ -2,34 +2,39 @@
...
+  const [open, setOpen] = useState(startOpen)
+
   return (
     <div>
-      <CollapsibleSection
-        icon={faHandHolding}
-        title={i18n.childInformation.assistance.title}
-        startCollapsed={!open}
+      <CollapsibleContentArea
+        //icon={faHandHolding}
```
jaahon, line 26: I would just remove all these commented out icons.

frontend/src/lib-components/atoms/buttons/AddButton.tsx
```diff
@@ -73,6 +73,7 @@ const StyledButton = styled.button`
 
   &.flipped {
     flex-direction: row-reverse;
+    margin-right: -12px;
```
jaahon, line 76: This kind of messes with layout in an unexpected way if the button is used in any different context since it's not even symmetrical. I would just remove the padding from the button, which should align it correctly on every browser.

frontend/src/employee-frontend/components/child-information/placements/NewServiceNeeds.tsx
```diff
@@ -36,28 +83,237 @@ function NewServiceNeeds({ serviceNeeds }: Props) {
...
+              setErrorMessage({
+                type: 'error',
+                title: i18n.common.error.unknown,
+                text: i18n.common.error.saveFailed,
+                resolveLabel: i18n.common.ok
+              })
+            }
+          })
+          .finally(() => setSubmitting(false))
+      }
```
jaahon, line 256: Everything apart from the initial request is identical between the branches, so I would do this instead:
```
const request = editingId ? updateNewServiceNeed(...) : createNewServiceNeed(...)
void request.then(...).finally(...)
```

frontend/src/employee-frontend/components/child-information/placements/NewServiceNeeds.tsx
```diff
@@ -1,29 +1,76 @@
...
   const { i18n } = useTranslation()
   const t = i18n.childInformation.placements.serviceNeeds
 
+  const [creatingNew, setCreatingNew] = useState(false)
+  const [editingId, setEditingId] = useState<string | null>(null)
+
+  const { serviceNeedOptions } = useContext(ChildContext)
+  const options = serviceNeedOptions.isSuccess
+    ? serviceNeedOptions.value
+        .filter((opt) => opt.validPlacementType === placementType)
```
jaahon, line 51: You should also filter out the "default" options, they shouldn't be picked manually. Or at least that was my intention, maybe it was a bad idea.

frontend/src/employee-frontend/components/child-information/placements/NewServiceNeeds.tsx
```diff
@@ -36,28 +83,237 @@ function NewServiceNeeds({ serviceNeeds }: Props) {
...
+            onClick={onCancel}
+            text={i18n.common.cancel}
+            disabled={submitting}
+          />
+          <InlineButton
+            onClick={onSubmit}
+            text={i18n.common.save}
+            disabled={submitting || !formIsValid}
+          />
+        </FixedSpaceRow>
```
jaahon, line 307: I would add a slightly larger space between the buttons, it didn't feel good when I used the editor. The ui spec also has a `24px` margin between them.

jaahon (commented):
Good stuff, just some pretty minor comments.

By the way, should the editor be hidden behind `useNewServiceNeeds` feature flag?

Joosakur:
Thanks for comments! I'll work on those and discuss some more.

frontend/src/employee-frontend/components/child-information/placements/NewServiceNeeds.tsx
Joosakur: Hmm I think this needs more thought. If there is only one option which is also default (e.g. case club) I agree that even the whole service need section could be hidden. However, if we think about daycare placement type, my understanding was that the default option is used as a fallback when the service need is missing but we need some coefficients for e.g. invoicing. Nevertheless, the explicit selection is still missing, so we need to notify the unit supervisor that he needs to set the service need, which might end up being the same as the default or not. But yeah lets maybe discuss this in a call so we understand the defaults the same way.

frontend/src/employee-frontend/components/child-information/service-need/ServiceNeedRow.tsx
```diff
@@ -100,6 +100,7 @@ function ServiceNeedRow({ serviceNeed, onReload }: Props) {
         ) : (
           <LabelValueList
             spacing="large"
+            labelWidth="240px"
```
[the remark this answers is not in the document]
Joosakur: Hmm I'll try that. Does fit-content(40%) mean fit content but max width 40%? I think 40% might be too much here now since the container width has increased.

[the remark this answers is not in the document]
jaahon: Yeah it's the maximum. 40% can be a lot, but it's not that much with smaller screens. The left side shouldn't be very long in any case anyway.

frontend/src/lib-components/atoms/buttons/AddButton.tsx
Joosakur: Screenshot 2021-04-30 at 10 16 39
Hmm yeah, that is an awkward hax, but if I just remove the padding it kind of bothers me that the center of the plus icon and caret are not aligned. But maybe that still looks fine enough? 🤷‍♂️

jaahon: Well that's always going to be a problem with buttons that are not the same size 😄 But at least in figma the icons themselves are not aligned.

jaahon (approved):

PR #772, opened 2021-04-29
EVAKA-4199 add private units to map

massproduction:
#### Summary
Show private units that do not have applyPeriods on the citizen map with a text that tells that you can't apply to them through eVaka and you need to contact the unit yourself.


mikkopiu (changes requested):
Could this have an E2E test to avoid regressions?

mikkopiu (approved):

PR #778, opened 2021-04-30
Remove `jdbi.handle` and `jdbi.transaction` calls from integration tests

akheron:
#### Summary

Use `db.transaction` and `db.read` instead of `jdbi.handle` and `jdbi.transaction`. Quite a few tests had to be refactored to avoid nested transactions or changes not visible outside of transactions.

This paves way for removing `jdbi.Handle` usage elsewhere.


Joosakur (approved):
Wow, that was a lot of effort, great stuff!

PR #786, opened 2021-05-04
EVAKA-4176 occupancy and raw report with new service needs

jaahon:
#### Summary

Add toggles for using new service needs to occupancy and raw reports. Implement the occupancy report with most of the computations done in kotlin code to make the implementation easier to read. If the performance hit is too big, revert back to the single database query.

The group occupancy reports are not implemented yet because they take a bit more work to implement and it's good to test the performance part first.

### DIFF
--- a/frontend/src/employee-frontend/components/reports/Occupancies.tsx
+++ b/frontend/src/employee-frontend/components/reports/Occupancies.tsx
@@ -11,6 +11,7 @@ import { Container, ContentArea } from 'lib-components/layout/Container'
 import Loader from 'lib-components/atoms/Loader'
 import Title from 'lib-components/atoms/Title'
 import { Th, Tr, Td, Thead, Tbody } from 'lib-components/layout/Table'
+import Checkbox from 'lib-components/atoms/form/Checkbox'
 import { reactSelectStyles } from '../../components/common/Select'
 import { Translations, useTranslation } from '../../state/i18n'
 import { Loading, Result, Success } from 'lib-common/api'
@@ -36,6 +37,7 @@ import {
   TableScrollable
 } from '../../components/reports/common'
 import { FlexRow } from '../../components/common/styled/containers'
+import { RequireRole } from '../../utils/roles'
 
 const StyledTd = styled(Td)`
   white-space: nowrap;
@@ -207,6 +209,7 @@ function Occupancies() {
     type: 'UNIT_CONFIRMED'
   })
   const [usedValues, setUsedValues] = useState<ValueOnReport>('percentage')
+  const [useNewServiceNeeds, setUseNewServiceNeeds] = useState(false)
 
   useEffect(() => {
     void getAreas().then((res) => res.isSuccess && setAreas(res.value))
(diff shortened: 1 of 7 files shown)


frontend/src/employee-frontend/components/reports/Occupancies.tsx
```diff
@@ -369,6 +372,15 @@ function Occupancies() {
             />
           </Wrapper>
         </FilterRow>
+        <RequireRole oneOf={['ADMIN']}>
+          <FilterRow>
+            <Checkbox
+              label="Käytä uusia palveluntarpeita"
```
reynders, line 378: I18N

service/src/main/kotlin/fi/espoo/evaka/reports/OccupancyReport.kt
```diff
@@ -273,3 +287,223 @@ private fun Database.Read.calculateOccupancyGroupReport(
         .filter { (isOperationDay, _) -> isOperationDay }
         .map { (_, reportRow) -> reportRow }
 }
+
+private data class Caretakers(
+    @Nested
+    val unit: UnitKey,
+    val date: LocalDate,
+    val caretakers: BigDecimal
```
reynders, line 295: caretakerCount?

service/src/main/kotlin/fi/espoo/evaka/reports/OccupancyReport.kt
```diff
@@ -273,3 +287,223 @@ private fun Database.Read.calculateOccupancyGroupReport(
...
+    today: LocalDate,
+    areaId: UUID,
+    queryPeriod: FiniteDateRange,
+    type: OccupancyType
+): List<OccupancyUnitReportResultRow> {
+    if (queryPeriod.start.plusDays(50) < queryPeriod.end) {
+        throw BadRequest("Date range ${queryPeriod.start} - ${queryPeriod.end} is too long. Maximum range is 50 days.")
+    }
+
+    val period =
```
reynders, line 339: Just something to check during testing: if the report is really slow, maybe we want to assert max period length here.

service/src/main/kotlin/fi/espoo/evaka/reports/RawReport.kt
```diff
@@ -197,6 +203,112 @@ private fun Database.Read.getRawRows(from: LocalDate, to: LocalDate): List<RawRe
...
+    sn IS NOT NULL AS has_service_need,
+    sno.part_day,
+    sno.part_week,
+    sn.shift_care,
+    sno.daycare_hours_per_week AS hours_per_week,
+
+    an IS NOT NULL as has_assistance_need,
+    coalesce(an.capacity_factor, 1.0) as capacity_factor,
+    coalesce(capacity_factor, 1.0) * (CASE
+        WHEN u.type && array['FAMILY', 'GROUP_FAMILY']::care_types[] THEN 1.75
```
reynders, line 244: Is this the same "1.75" as in the code above? If so, should this be a constant?

reynders (approved):
Very readable style!

service/src/main/kotlin/fi/espoo/evaka/reports/OccupancyReport.kt
jaahon: There is an assertion right above? 😄

But, it does make more sense to do the assertion for the actual period used in calculations.

service/src/main/kotlin/fi/espoo/evaka/reports/OccupancyReport.kt
jaahon: Sure.

service/src/main/kotlin/fi/espoo/evaka/reports/RawReport.kt
jaahon: Yeah, it probably should.

PR #793, opened 2021-05-04
[EVAKA-HOTFIX] Further prevent SAML replay attacks with InResponseTo checks, custom Redis cache provider

mikkopiu:
#### Background
- When `validateInResponseTo` is set to `false` (default), `passport-saml` won't check that the SAML responses it receives are in response to anything it actually initiated -> the responses can be used in replay attacks and `passport-saml` won't complain
- `passport-saml` only provides an `InMemoryCacheProvider` which simply stores the requires checking information (request IDs) in local memory -> this won't work environments with multiple instances of API GW where one instance might create the initial SAML request and another instance receive the matching SAML response, and not know about it as the ID would be only in the first instance's cache
- `passport-saml` does however provide a config option to use a custom cache provider, i.e. Redis in our case
- Part 3 of the features/fixes recommended by a white hat

#### Summary
- APIGW: Create Redis passport-saml cache provider
   - Redis is the single source-of-truth for all APIGW instances in a single environment
   - For now, use the same Redis config as is used for session management -- just separate key prefixes to avoid key conflicts
   - Add `redis-mock` for testing the cache provider
- APIGW: Use the new cache provider with all SAML Strategies to further prevent replay attacks (responses can only be re-used while the matching request IDs are in the cache instead of forever, excluding `notOnOrAfter` timestamping)
   - Instantiate the Redis clients at the app level, to simplify future testing/stubbing with redis-mock etc.
       - **NOTE:** Not sure if this is the best solution but it's the first, and felt reasonable
   - Also remove unnecessary `createAuthEndpoints` wrapper function for Suomi.fi SAML Strategy (does nothing but cause confusion)
- APIGW: Replace `@types/passport-saml` with local implementation
   - More specifically: include the latest version of `@types/passport-saml` types in APIGW source code and modify types for `CacheItem` and `CacheProvider` to match the actual `passport-saml@1.3.5` implementation
   - Original license is MIT, i.e. compatible with our LGPL-2.1-or-later license
- Add all missing license headers

#### Testing

This has been manually deployed and tested in staging and we should mostly be able to rely on `passport-saml`'s own tests as this is just a "plug-in" implementation + configuration.


apigw/src/enduser/app.ts
```diff
@@ -43,10 +53,38 @@ app.use(refreshLogoutToken('enduser'))
...
+      : undefined
+
+  if (redisClient) {
+    redisClient.on('error', (err) =>
+      logError('Redis error', undefined, undefined, err)
+    )
+
+    // don't prevent the app from exiting if a redis connection is alive
+    redisClient.unref()
+  }
```
mikkopiu, line 75: I'm not totally sure this is the best place for these but it felt better than instantiating the clients inside the cache providers (enables unit testing).

Gekkio: Seems ok, although now there's two clients: one for the sessions and one for the SAML stuff. Having two clients is fine, but they also point to the same db (database 0 I think?) so it seems a bit pointless right now. Would it be better to put SAML things in database 1 and keep the two separate clients?

mikkopiu: I'm not sure what the current status is but at least in the past Redis developers have indicated that having multiple databases was a "mistake" (e.g. comments from Salvatore: https://groups.google.com/g/redis-db/c/vS5wX8X4Cjg/discussion). It also looks like Redis Cluster doesn't support using anything else than database 0, so I think that's a no-go.

It seems the recommended way is to either have multiple Redis instances or by using "namespaces" (i.e. key prefixes as done here), and AWS obviously doesn't provide very small instances -> unnecessary costs for splitting by instance. It does appear, however, that `node-redis` has a [`prefix` option in `createClient()`](https://github.com/NodeRedis/node-redis/#options-object-properties) which would make the namespacing transparent to the cache implementation.

So now I think there are two ways to go about this:
1. Single client for everything, instantiated once and **manual** namespaces
    - Comes with the risk of key conflicts (I clearly already made the mistake once, naming everything "keycloak...")
    - Removes duplicate configs & error logging
    - Logical to have one client per one instance
1. Client per context (sessions and for each active strategy) and **pre-defined** namespaces per client (upon instantiation)
    - With this, I'd also move the session client init to the same place as the others
    - Namespaces are transparent to implementation -> risk of key conflicts is centralized in the one place where all clients are created
    - Requires duplicating configs and error logging (though this can be alleviated with shared callbacks, base configs etc.)
    - _Feels_ silly, as everything is targeting a single instance but is it?

mikkopiu: I think I'll try with the single client approach first and see how that goes.

mikkopiu:
Rebased (and merged the new Keycloak strategy with the new changes here) and implemented the "shared Redis client" idea.

apigw/src/shared/auth/customer-saml.ts
```diff
@@ -28,16 +33,20 @@ function createKeycloakSamlStrategy(
   })
   return new SamlStrategy(
     {
-      issuer: 'evaka-customer',
-      callbackUrl: samlConfig.callbackUrl,
-      entryPoint: samlConfig.entryPoint,
-      logoutUrl: samlConfig.entryPoint,
       acceptedClockSkewMs: -1,
+      cacheProvider: redisClient
+        ? redisCacheProvider(redisClient, { keyPrefix: 'suomifi-saml-resp:' })
```
Gekkio, line 38: Citizen SAML should probably use a different prefix than actual suomi.fi SAML

Gekkio (approved):

apigw/src/shared/auth/customer-saml.ts
mikkopiu: Goddamit :D

PR #796, opened 2021-05-05
EVAKA-4176 group occupancy reports

jaahon:
#### Summary

Implement group occupancy reports using new service needs.


service/src/main/kotlin/fi/espoo/evaka/reports/OccupancyReport.kt
```diff
@@ -470,41 +590,41 @@ WHERE sn.placement_id = ANY(:placementIds)
...
-        .map { (unit, values) ->
+    return caretakerCounts
+        .map { (key, values) ->
             val occupancies = values
-                .associate { (_, date, caretakers) ->
-                    val unitPlacementsOnDate = (placementsAndPlans[unit.unitId] ?: listOf())
-                        .filter { it.period.includes(date) }
-                        .filterNot { absences[it.childId]?.any { absence -> absence.date == date } ?: false }
+                .associate { caretakers ->
+                    val unitPlacementsOnDate = (placementsAndPlans[key.groupingId] ?: listOf())
```
Joosakur, line 599: is the correct name still `unitPlacementsOnDate` or can it also be `groupPlacementsOnDate` depending on type of `K`?

service/src/main/kotlin/fi/espoo/evaka/reports/OccupancyReport.kt
```diff
@@ -470,41 +590,41 @@ WHERE sn.placement_id = ANY(:placementIds)
...
+                        .filter { it.period.includes(caretakers.date) }
+                        .filterNot { absences[it.childId]?.any { absence -> absence.date == caretakers.date } ?: false }
 
                     val sum = unitPlacementsOnDate
                         .groupBy { it.childId }
-                        .mapNotNull { (_, placements) -> placements.map { getCoefficient(date, it) }.maxOrNull() }
+                        .mapNotNull { (_, placements) ->
```
Joosakur, line 605: nitpick: `sum` could be renamed to be more specific - now need to read ahead to figure out sum of what.

service/src/main/kotlin/fi/espoo/evaka/reports/OccupancyReport.kt
```diff
@@ -459,7 +579,7 @@ WHERE sn.placement_id = ANY(:placementIds)
         val serviceNeedCoefficient = when {
             placement.familyUnitPlacement -> BigDecimal(youngChildOccupancyCoefficient)
             date < dateOfBirth.plusYears(3) -> BigDecimal(youngChildOccupancyCoefficient)
-            else -> serviceNeedCoefficients[placement.id]
+            else -> serviceNeedCoefficients[placement.placementId]
```
Joosakur, line 582: I'm a bit uncertain whether a nested function inside a huge function is a good idea. Moving it outside could be worth considering even if the parameter list will get somewhat long.

service/src/main/kotlin/fi/espoo/evaka/reports/OccupancyReport.kt
```diff
@@ -470,41 +590,41 @@ WHERE sn.placement_id = ANY(:placementIds)
...
+                            placements.map { getCoefficient(caretakers.date, it) }.maxOrNull()
+                        }
                         .fold(BigDecimal.ZERO) { sum, value -> sum + value }
 
                     val percentage =
-                        if (caretakers.compareTo(BigDecimal.ZERO) == 0) null
+                        if (caretakers.caretakerCount.compareTo(BigDecimal.ZERO) == 0) null
                         else sum
-                            .divide(caretakers * BigDecimal(7), 10, RoundingMode.HALF_UP)
+                            .divide(caretakers.caretakerCount * BigDecimal(7), 10, RoundingMode.HALF_UP)
```
Joosakur, line 613: Not sure if using BigDecimal api is a bit of an overkill here, but the readability isn't too bad so maybe thats good 🤷‍♂️

Joosakur (approved):
Nice stuff! 🏅 Some random thoughts included but nothing that requires changes

service/src/main/kotlin/fi/espoo/evaka/reports/OccupancyReport.kt
```diff
@@ -289,15 +305,20 @@ private fun Database.Read.calculateOccupancyGroupReport(
         .map { (_, reportRow) -> reportRow }
 }
 
-private data class Caretakers(
-    @Nested
-    val unit: UnitKey,
+private interface Key {
+    val groupingId: UUID
+    val unitId: UUID
+}
```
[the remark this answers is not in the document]
jaahon: A sealed class doesn't really offer anything over using an interface since we don't do anything conditionally depending on values, just reified type parameters. I can explain this if you want.

service/src/main/kotlin/fi/espoo/evaka/reports/OccupancyReport.kt
```diff
@@ -423,7 +543,7 @@ JOIN service_need_option sno ON sn.option_id = sno.id
 WHERE sn.placement_id = ANY(:placementIds)
 """
     )
-        .bind("placementIds", placements.map { it.id }.toList().toTypedArray())
+        .bind("placementIds", placements.map { it.placementId }.toList().toTypedArray())
```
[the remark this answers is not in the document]
jaahon: > ^ on line 531 should we remove duplicate childIds?
Sure

service/src/main/kotlin/fi/espoo/evaka/reports/OccupancyReport.kt
jaahon: I'm not completely sure what you mean, but I think using kotlin's `/` operator with `BigDecimal` doesn't make it very clear what the end result is so I dislike using it.

service/src/main/kotlin/fi/espoo/evaka/reports/OccupancyReport.kt
jaahon: What do you think is the problem with it?

service/src/main/kotlin/fi/espoo/evaka/reports/OccupancyReport.kt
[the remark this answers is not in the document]
Joosakur: I think there was some when-block where the else-branch would not be needed if this was a sealed class, because compiler would then know that all options have been handled. But yeah thats a very minor benefit.

service/src/main/kotlin/fi/espoo/evaka/reports/OccupancyReport.kt
Joosakur: yeah i mean i was wondering whether a normal Double would be enough here since this is not about money and we dont care that much about rounding errors

service/src/main/kotlin/fi/espoo/evaka/reports/OccupancyReport.kt
Joosakur: just readability

service/src/main/kotlin/fi/espoo/evaka/reports/OccupancyReport.kt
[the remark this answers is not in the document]
jaahon: It doesn't seem to work with type parameters, only with actual values.

PR #797, opened 2021-05-05
Confirm when navigating away from unsaved PIN

reynders:
#### Summary
Confirm when navigating away from unsaved PIN:
- If tab is closed or back button pressed, use browser beforeunload support (https://developer.mozilla.org/en-US/docs/Web/API/Window/beforeunload_event)
- If clicking another link, use react routed DOM support

Tested on latest chrome and firefox. On playwright chromium the popup disappears immediately.

The two cases shown on chrome and firefox:

Screenshot 2021-05-06 at 8 53 28
Screenshot 2021-05-06 at 9 02 10
Screenshot 2021-05-06 at 13 22 50
Screenshot 2021-05-06 at 13 23 18

### DIFF
--- a/frontend/src/employee-frontend/components/employee/EmployeePinCodePage.tsx
+++ b/frontend/src/employee-frontend/components/employee/EmployeePinCodePage.tsx
@@ -25,6 +25,7 @@ export default React.memo(function EmployeePinCodePage() {
   const [pin, setPin] = useState<string>('')
   const [error, setError] = useState<boolean>(false)
   const [pinLocked, setPinLocked] = useState<Result<boolean>>()
+  const [dirty, setDirty] = useState<boolean>(false)
 
   useEffect(() => {
     void isPinCodeLocked().then(setPinLocked)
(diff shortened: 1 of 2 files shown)


frontend/src/employee-frontend/components/employee/EmployeePinCodePage.tsx
```diff
@@ -74,8 +80,27 @@ export default React.memo(function EmployeePinCodePage() {
...
   }
 
+  const beforeUnloadHandler = (e: BeforeUnloadEvent) => {
+    if (dirty) {
+      // Support different browsers: https://developer.mozilla.org/en-US/docs/Web/API/Window/beforeunload_event
+      e.preventDefault()
+      e.returnValue = i18n.pinCode.unsavedDataWarning
+      return i18n.pinCode.unsavedDataWarning
+    }
+    return
```
akheron, line 90: Useless `return`

akheron (approved):

frontend/src/employee-frontend/components/employee/EmployeePinCodePage.tsx
reynders: lint complaints without it: "TS7030: Not all code paths return a value."

PR #800, opened 2021-05-06
Injectable proxy configuration and proxy Dockerfile cleanup

ilja-git:
#### Summary

- We want to inject configurations into proxy so we can do more dynamic configurations, so here we can use environment to inject configuration and also we can download configurations from S3 buckets
- Cleaned up the docker build


proxy/files/bin/proxy-start.sh
```diff
@@ -21,7 +22,18 @@ if [ "${INTERNAL_GW_URL:-X}" = 'X' ]; then
   exit 1
 fi
 
-erb /etc/nginx/conf.d/evaka-nginx.conf.template > /etc/nginx/conf.d/evaka-nginx.conf
+if test -n "$DEPLOYMENT_BUCKET"; then
+  s3download "$DEPLOYMENT_BUCKET" "nginx-proxy" /etc/nginx/
```
mikkopiu, line 26: Nit: Naming the S3 "directory" just "proxy" would be more consistent with the other names

proxy/files/bin/proxy-start.sh
```diff
@@ -21,7 +22,18 @@ if [ "${INTERNAL_GW_URL:-X}" = 'X' ]; then
...
+fi
+
+for template in `ls -v /etc/nginx/conf.d/*.template`; do
```
mikkopiu, line 29: Supernit: I'd prefer `$(...)` over backticks just to keep a consistent style (and avoid unnecessary complaints from shellcheck, https://github.com/koalaman/shellcheck/wiki/SC2006).

Also, prefer globs over `ls` output, see [rationale](https://github.com/koalaman/shellcheck/wiki/SC2045).

proxy/files/etc/nginx/conf.d/nginx.conf.template
```diff
@@ -266,7 +266,9 @@ server {
     proxy_pass  $internalUrl/api/csp/csp-report;
   }
 
-  <% if ENV.key?("KEYCLOAK_URL") && ENV["KEYCLOAK_URL"] != "" %>
+  <% if ENV.key?("KEYCLOAK_CONFIGURATION") && ENV["KEYCLOAK_CONFIGURATION"] != "" %>
+  <%= ENV["KEYCLOAK_CONFIGURATION"] %>
+  <% elsif ENV.key?("KEYCLOAK_URL") && ENV["KEYCLOAK_URL"] != "" %>
```
mikkopiu, line 271: I'd maybe add a comment here to describe the rationale behind allowing a complete override vs. the pre-defined config for posterity.

proxy/.dockerignore
```diff
@@ -2,7 +2,5 @@
 #
 # SPDX-License-Identifier: LGPL-2.1-or-later
 
-*
-!entrypoint.sh
-!nginx.conf.template
-!proxy_params
+README.md
+build.sh
```
mikkopiu, line 6: I really prefer the whitelist method to ensure we don't accidentally add the wrong files to any images, and as you've moved everything under `files/` anyway, couldn't you just change this to:

```
*
!files/
```

mikkopiu (changes requested):

proxy/files/etc/nginx/conf.d/nginx.conf.template
ilja-git: The pre-defined was left there for backwards combability so it supports both configurations without failing.

mikkopiu: OK, so was the idea to remove it at some later point completely (that's OK, just affects whether I'd like the comment or not)?

ilja-git: We could leave it as example there, but if we go this way, we will not be using it anymore when this is installed on all systems and becomes unnessary and can be removed then. This way we do not have to worry synching the infra with this commit.

I can leave deprecation comment there.

mikkopiu (approved):

PR #804, opened 2021-05-06
EVAKA-4176: splitting placements splits service needs

Joosakur:
#### Summary
Service needs are split the same way as group placements. If a placement is created inside another, both service needs and group placements are shortened. Later it would be nice to copy them to the new placement inserted at the end, but this is a rare use-case and not trivial to implement so I left that out.

Unit supervisors may also now delete placements from their units which have not yet started.


service/src/integrationTest/kotlin/fi/espoo/evaka/placement/PlacementServiceIntegrationTest.kt
```diff
@@ -546,56 +570,17 @@ class PlacementServiceIntegrationTest : FullApplicationTest() {
         assertEquals(1, groupPlacements.size)
         assertEquals(LocalDate.of(year, month, 1), groupPlacements.first().startDate)
         assertEquals(LocalDate.of(year, month, 14), groupPlacements.first().endDate)
-    }
 
-    @Test
-    fun `changing placement type to preparatory by creating new placement preserves group placement history`() {
```
Joosakur, line 552: this test was basically a duplicate

jaahon (approved):

PR #808, opened 2021-05-07
EVAKA-4190 Citizen reply endpoint

kimifoud:
#### Summary

Implement citizen thread reply functionality.


kimifoud:
Will be superseded.

PR #814, opened 2021-05-10
EVAKA-4176 fix service need migration query

jaahon:


Joosakur (approved):
oops, seems I missed that, thanks for fixing!

reynders (approved):

PR #818, opened 2021-05-10
Embedded fonts

mapehe:
#### Summary

Posti pointed out that our PDF:s don't come with embedded fonts. Include font files in resources and embed those to fix this.


mikkopiu (changes requested):
The font files need their original licenses as `*.license` sidecar files:
- Montserrat:
    - https://fonts.google.com/specimen/Montserrat?selection.family=Montserrat#license -> unclear which version of SIL Open Font License (but could _assume_ latest, 1.1)
    - https://www.fontsquirrel.com/license/montserrat  -> 1.1, so pretty safe to assume
    - SPDX code: https://spdx.org/licenses/OFL-1.1.html
    - Based on this: https://tldrlegal.com/license/open-font-license-(ofl)-explained -> safe to include in eVaka
- OpenSans:
    - https://github.com/googlefonts/opensans/blob/master/LICENSE.txt -> Apache 2.0 (compatible with our license)
    - https://fonts.google.com/specimen/Open+Sans#license

Use the `./bin/add-license-headers.sh` tool

mapehe:
Thanks for the feedback @mikkopiu. Added licences.

service/src/main/resources/ttf/Montserrat-Bold.ttf.license
```diff
@@ -0,0 +1,93 @@
+Copyright 2011 The Montserrat Project Authors (https://github.com/JulietaUla/Montserrat)
```
mikkopiu, line 1: What you actually for these files, is like:

```txt
SPDX-FileCopyrightText: 2011 The Montserrat Project Authors (https://github.com/JulietaUla/Montserrat)

SPDX-License-Identifier: OFL-1.1
```

and let the tool download these actual license files under `LICENSES/`. Looks like currently the tool doesn't really care about the contents of `*.license` files but that's what they should be (see existing files).

mikkopiu (changes requested):
Almost there!

service/src/main/resources/ttf/Montserrat-Bold.ttf.license
mapehe: Ah Thanks!

service/src/main/resources/ttf/OpenSans-Regular.ttf.license
```diff
@@ -0,0 +1,3 @@
+SPDX-FileCopyrightText: The Open Sans Project Authors
```
mapehe, line 1: Couldn't find copyright holders. Suggesting to writing it like this.

mikkopiu: Looks like it should be:
```suggestion
SPDX-FileCopyrightText: Digitized data copyright  2010-2011, Google Corporation
```

Source: https://fonts.google.com/attribution

mapehe: Thanks, updated

mikkopiu (approved):
Good job :+1: , now at least one more person is familiar with the license tool :sweat_smile:

mapehe:
> Good job 👍 , now at least one more person is familiar with the license tool 😅

😆 Thanks for your patience!

PR #821, opened 2021-05-10
Upgrade owasp dependencycheck version

akheron:


mikkopiu (changes requested):
`message-service/buildSrc/src/main/kotlin/Version.kt` needs the same update

akheron:
> `message-service/buildSrc/src/main/kotlin/Version.kt` needs the same update

Changed, seems to pass now 🎉

mikkopiu (approved):

PR #824, opened 2021-05-11
EVAKA-4190 Split new messages to threads by guardianship

kimifoud:
#### Summary

Split new messages to threads by guardianship.

The idea is to isolate each "household" to a single thread so that people participating in each thread are allowed see all the participants and reply to them.

### DIFF
--- a/service/src/main/kotlin/fi/espoo/evaka/messaging/message/MessageAccountQueries.kt
+++ b/service/src/main/kotlin/fi/espoo/evaka/messaging/message/MessageAccountQueries.kt
@@ -91,13 +91,15 @@ SELECT * FROM common_guardians
         .mapTo<AccountToChild>()
         .list()
 
-    val singleRecipients = (accountIds - accountsWithCommonChildren.map { it.id }).map { setOf(it) }
-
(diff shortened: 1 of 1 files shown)


akheron (approved):

service/src/main/kotlin/fi/espoo/evaka/messaging/message/MessageAccountQueries.kt
```diff
@@ -63,3 +63,41 @@ fun Database.Transaction.deleteDaycareGroupMessageAccount(daycareGroupId: UUID)
...
+
+    val singleRecipients = (accountIds - accountsWithCommonChildren.map { it.id }).map { setOf(it) }
+
+    val distinctSetsOfAccountsWithCommonChildren = accountsWithCommonChildren
+        .groupBy { it.childId }
+        .values
+        .map { row -> row.map { it.id }.toSet() }
+        .toSet()
+
+    return distinctSetsOfAccountsWithCommonChildren + singleRecipients
```
akheron, line 102: The logic above is a bit hairy 😄

PR #829, opened 2021-05-11
EVAKA-4190 Add messaging api-types to employee frontend

kimifoud:
#### Summary

Initial api types for employee frontend messaging

### DIFF
--- a/frontend/src/employee-frontend/components/messages/api.ts
+++ b/frontend/src/employee-frontend/components/messages/api.ts
@@ -7,13 +7,17 @@ import { client } from '../../api/client'
 import { UUID } from '../../types'
 import {
   Bulletin,
-  IdAndName,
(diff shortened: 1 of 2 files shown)


frontend/src/employee-frontend/components/messages/api.ts
```diff
@@ -127,3 +131,63 @@ export async function getSenderOptions(
     .then((res) => Success.of(res.data))
     .catch((e) => Failure.fromError(e))
 }
+
+export async function getAccounts(): Promise<Result<MessageAccount[]>> {
```
mapehe, line 135: We updated the return type to `EnrichedMessageAccount`, so this will be outdated soon. But it probably makes sense to make the update it in the branch that I'm working on atm.

PR #830, opened 2021-05-11
Make invoice account configurable

reynders:


reynders:
It was agreed that for now the account change will be done on the integration platform side

PR #833, opened 2021-05-12
EVAKA-4190 Supervisor message account

akheron:
#### Summary
- Add support for marking message accounts as active/inactive
- Add message account for unit supervisor when giving them a supervisor role
- Deactivate the message account when removing the last supervisor role

### DIFF
--- a/service/src/integrationTest/kotlin/fi/espoo/evaka/daycare/controllers/UnitAclControllerIntegrationTest.kt
+++ b/service/src/integrationTest/kotlin/fi/espoo/evaka/daycare/controllers/UnitAclControllerIntegrationTest.kt
@@ -22,6 +22,8 @@ import fi.espoo.evaka.shared.dev.insertTestEmployee
 import fi.espoo.evaka.testAreaCode
 import fi.espoo.evaka.testAreaId
 import fi.espoo.evaka.testDaycare
+import fi.espoo.evaka.testDaycare2
+import org.jdbi.v3.core.kotlin.mapTo
 import org.junit.jupiter.api.Assertions.assertEquals
 import org.junit.jupiter.api.Assertions.assertTrue
 import org.junit.jupiter.api.BeforeEach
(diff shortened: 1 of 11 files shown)


service/src/integrationTest/kotlin/fi/espoo/evaka/daycare/controllers/UnitAclControllerIntegrationTest.kt
```diff
@@ -77,6 +90,19 @@ class UnitAclControllerIntegrationTest : FullApplicationTest() {
...
+    fun `supervisor message account`() {
+        assertEquals(employeeMessageAccountState(), MessageAccountState.NO_ACCOUNT)
+        insertSupervisor(testDaycare.id)
+        assertEquals(employeeMessageAccountState(), MessageAccountState.ACTIVE_ACCOUNT)
+        insertSupervisor(testDaycare2.id)
+        assertEquals(employeeMessageAccountState(), MessageAccountState.ACTIVE_ACCOUNT)
+        deleteSupervisor(testDaycare.id)
+        assertEquals(employeeMessageAccountState(), MessageAccountState.ACTIVE_ACCOUNT)
+        deleteSupervisor(testDaycare2.id)
+        assertEquals(employeeMessageAccountState(), MessageAccountState.INACTIVE_ACCOUNT)
```
kimifoud, line 103: These should be flipped `(expected, actual)`

service/src/main/kotlin/fi/espoo/evaka/shared/auth/AclQueries.kt
```diff
@@ -30,6 +30,20 @@ WHERE daycare_id = :daycareId
...
+            SELECT * FROM daycare_acl
+            WHERE employee_id = :employeeId AND role = :role
+        )
+    """.trimIndent()
+)
+    .bind("employeeId", employeeId)
+    .bind("role", role)
+    .mapTo<Boolean>()
+    .first()
+    .let { it == true }
```
kimifoud, line 45: Couldn't this be simplified to 
```kt
    .mapTo<Boolean>()
    .one()
```
?

kimifoud (commented):
Do we implement `where active = TRUE` filtering in a separate PR?

service/src/integrationTest/kotlin/fi/espoo/evaka/daycare/controllers/UnitAclControllerIntegrationTest.kt
akheron: > Do we implement `where active = TRUE` filtering in a separate PR?

Yes, added a ticket to miro

service/src/main/kotlin/fi/espoo/evaka/shared/auth/AclQueries.kt
akheron: The idea is that `null` and `false` both map to `false`. Does `.one()` throw if the value is `null`?

akheron: Ah, `Boolean!` seems to mean "`Boolean` or `Boolean?`", so `.one()` will do!

service/src/main/kotlin/fi/espoo/evaka/shared/dev/DevApi.kt
```diff
@@ -954,11 +953,22 @@ fun Database.Transaction.deleteApplication(id: UUID) {
 }
 
 fun Database.Transaction.deleteAndCascadeEmployee(id: UUID) {
+    execute("DELETE FROM message_account WHERE employee_id = ?", id)
```
kimifoud, line 956: This will fail if messages are created, but it can be ignored for now and handled when/if this code is ever used in a messaging related test code.

kimifoud (approved):

PR #836, opened 2021-05-12
Enable Evaka Employee login on Production

ilja-git:
#### Summary

This enables Employee login on production.


mikkopiu (approved):
:+1: and checked with @ilja-git that Keycloak configs are correct & safe (flow and SAML-wise)

PR #848, opened 2021-05-14
EVAKA-4176: updated occupancy calculation in unit view

Joosakur:

### DIFF
--- a/service/src/integrationTest/kotlin/fi/espoo/evaka/ChildBuilder.kt
+++ b/service/src/integrationTest/kotlin/fi/espoo/evaka/ChildBuilder.kt
@@ -1,13 +1,17 @@
 package fi.espoo.evaka
 
 import fi.espoo.evaka.application.ApplicationStatus
+import fi.espoo.evaka.daycare.service.AbsenceType
+import fi.espoo.evaka.daycare.service.CareType
 import fi.espoo.evaka.placement.PlacementType
 import fi.espoo.evaka.shared.db.Database
 import fi.espoo.evaka.shared.dev.DevAssistanceNeed
 import fi.espoo.evaka.shared.dev.DevChild
 import fi.espoo.evaka.shared.dev.DevPerson
+import fi.espoo.evaka.shared.dev.insertTestAbsence
 import fi.espoo.evaka.shared.dev.insertTestApplication
 import fi.espoo.evaka.shared.dev.insertTestAssistanceNeed
+import fi.espoo.evaka.shared.dev.insertTestBackUpCare
 import fi.espoo.evaka.shared.dev.insertTestChild
 import fi.espoo.evaka.shared.dev.insertTestDaycareGroupPlacement
 import fi.espoo.evaka.shared.dev.insertTestNewServiceNeed
@@ -37,7 +41,9 @@ class ChildBuilder(
     ) {
         fun hasAssistanceNeed() = AssistanceNeedBuilder(tx, today, this)
         fun hasPlacementPlan() = PlacementPlanBuilder(tx, today, this)
+        fun hasBackupCare() = BackupCareBuilder(tx, today, this)
         fun hasPlacement() = PlacementBuilder(tx, today, this)
+        fun hasAbsence() = AbsenceBuilder(tx, today, this)
     }
 
     class AssistanceNeedBuilder(
@@ -55,7 +61,7 @@ class ChildBuilder(
         fun toDay(day: Int) = this.apply { this.to = day }
         fun createdBy(employeeId: UUID) = this.apply { this.employeeId = employeeId }
 
-        fun execAndDone(): ChildFixture {
+        fun execAndReturn(): ChildFixture {
             tx.insertTestAssistanceNeed(
                 DevAssistanceNeed(
                     childId = childFixture.childId,
@@ -69,6 +75,41 @@ class ChildBuilder(
         }
     }
 
+    class AbsenceBuilder(
+        private val tx: Database.Transaction,
+        private val today: LocalDate,
+        private val childFixture: ChildFixture
+    ) {
+        private var day: Int = 0
+        private var type: AbsenceType? = null
+        private var careTypes: List<CareType>? = null
+
+        fun onDay(day: Int) = this.apply { this.day = day }
+        fun ofType(type: AbsenceType) = this.apply { this.type = type }
+        fun forCareTypes(vararg careTypes: CareType) = this.apply { this.careTypes = careTypes.asList() }
+
+        private fun exec() {
+            careTypes?.forEach { careType ->
+                tx.insertTestAbsence(
+                    childId = childFixture.childId,
+                    date = today.plusDays(day.toLong()),
+                    absenceType = type ?: throw IllegalStateException("absence type not set"),
+                    careType = careType
+                )
+            } ?: throw IllegalStateException("care types not set")
+        }
+
+        fun execAndReturn(): ChildFixture {
+            exec()
+            return childFixture
+        }
+
+        fun andAnother(): AbsenceBuilder {
+            exec()
+            return AbsenceBuilder(tx, today, childFixture)
+        }
+    }
+
     class PlacementPlanBuilder(
         private val tx: Database.Transaction,
         private val today: LocalDate,
@@ -78,13 +119,15 @@ class ChildBuilder(
         private var type: PlacementType? = null
         private var from: Int = 0
         private var to: Int = 0
+        private var deleted = false
 
         fun toUnit(id: UUID) = this.apply { this.unitId = id }
         fun ofType(type: PlacementType) = this.apply { this.type = type }
         fun fromDay(day: Int) = this.apply { this.from = day }
         fun toDay(day: Int) = this.apply { this.to = day }
+        fun asDeleted() = this.apply { this.deleted = true }
 
-        fun execAndDone(): ChildFixture {
+        fun execAndReturn(): ChildFixture {
             val applicationGuardianId = tx.insertTestPerson(DevPerson())
             val applicationId = tx.insertTestApplication(
                 guardianId = applicationGuardianId,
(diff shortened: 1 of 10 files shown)


service/src/integrationTest/kotlin/fi/espoo/evaka/ChildBuilder.kt
```diff
@@ -0,0 +1,280 @@
...
+    private val today: LocalDate
+) {
+    fun childOfAge(years: Int, months: Int = 0, days: Int = 0): ChildFixture {
+        val dob = today.minusYears(years.toLong()).minusMonths(months.toLong()).minusDays(days.toLong())
+        val childId = tx.insertTestPerson(DevPerson(dateOfBirth = dob))
+        tx.insertTestChild(DevChild(childId))
+        return ChildFixture(tx, today, childId)
+    }
+
+    class ChildFixture(
```
jaahon, line 37: If the idea is to not use these classes directly (i.e. calling `ChildBuilder.ChildFixture(...)`) then the constructors shouldn't be public in my opinion.

service/src/integrationTest/kotlin/fi/espoo/evaka/ChildBuilder.kt
```diff
@@ -0,0 +1,280 @@
...
+        fun withFactor(factor: Double) = this.apply { this.factor = factor }
+        fun fromDay(day: Int) = this.apply { this.from = day }
+        fun toDay(day: Int) = this.apply { this.to = day }
+        fun createdBy(employeeId: UUID) = this.apply { this.employeeId = employeeId }
+
+        fun execAndReturn(): ChildFixture {
+            tx.insertTestAssistanceNeed(
+                DevAssistanceNeed(
+                    childId = childFixture.childId,
+                    updatedBy = employeeId ?: throw IllegalStateException("createdBy not set"),
```
jaahon, line 68: Instead of throwing here, why not pass the `employeeId` as a parameter to this function?

service/src/integrationTest/kotlin/fi/espoo/evaka/ChildBuilder.kt
```diff
@@ -0,0 +1,280 @@
...
+        tx.insertTestChild(DevChild(childId))
+        return ChildFixture(tx, today, childId)
+    }
+
+    class ChildFixture(
+        private val tx: Database.Transaction,
+        private val today: LocalDate,
+        val childId: UUID
+    ) {
+        fun hasAssistanceNeed() = AssistanceNeedBuilder(tx, today, this)
```
jaahon, line 42: Maybe some other name would be better? Like `withAssistanceNeed` or `addAssistanceNeed`.

service/src/integrationTest/kotlin/fi/espoo/evaka/ChildBuilder.kt
```diff
@@ -0,0 +1,280 @@
...
+        private var factor: Double = 1.0
+        private var from: Int = 0
+        private var to: Int = 0
+
+        fun withFactor(factor: Double) = this.apply { this.factor = factor }
+        fun fromDay(day: Int) = this.apply { this.from = day }
+        fun toDay(day: Int) = this.apply { this.to = day }
+        fun createdBy(employeeId: UUID) = this.apply { this.employeeId = employeeId }
+
+        fun execAndReturn(): ChildFixture {
```
jaahon, line 64: I think `save`, `persist` or something like that would describe the intent better then `exec`. I also think that the `AndReturn` part is kind of redundant as well.

service/src/integrationTest/kotlin/fi/espoo/evaka/ChildBuilder.kt
```diff
@@ -0,0 +1,280 @@
...
+        private var unitId: UUID? = null
+        private var type: PlacementType? = null
+        private var from: Int = 0
+        private var to: Int = 0
+        private var deleted = false
+
+        fun toUnit(id: UUID) = this.apply { this.unitId = id }
+        fun ofType(type: PlacementType) = this.apply { this.type = type }
+        fun fromDay(day: Int) = this.apply { this.from = day }
+        fun toDay(day: Int) = this.apply { this.to = day }
```
jaahon, line 127: I don't think this is a very obvious API.

service/src/integrationTest/kotlin/fi/espoo/evaka/ChildBuilder.kt
```diff
@@ -0,0 +1,280 @@
...
+        private val today: LocalDate,
+        val childFixture: ChildFixture,
+        val placementId: UUID,
+        val placementPeriod: FiniteDateRange
+    ) {
+        fun withGroupPlacement() = GroupPlacementBuilder(tx, today, this)
+
+        fun withServiceNeed() = ServiceNeedBuilder(tx, today, this)
+
+        fun returnFromPlacement() = childFixture
```
jaahon, line 225: Seems unnecessarily low level having to manually jump between "fixtures".

jaahon (commented):
Here's my two cents on the builders and their API. Overall I like the idea! I'll look through the occupancy stuff separately.

service/src/integrationTest/kotlin/fi/espoo/evaka/ChildBuilder.kt
Joosakur: Yeah `add` could work. I reserved `with` for hierarchical chaining, e.g. `add` placement `with` service need.

service/src/integrationTest/kotlin/fi/espoo/evaka/ChildBuilder.kt
Joosakur: Maybe change to LocalDate and add these relative ones as overloads? Renaming to fromRelativeDay would be more explicit but the line lengths would become a problem.

service/src/integrationTest/kotlin/fi/espoo/evaka/ChildBuilder.kt
jaahon: Well, there _is_ a hierarchy between children and their placements 😄

service/src/integrationTest/kotlin/fi/espoo/evaka/ChildBuilder.kt
Joosakur: Yeah +1 for `save` over `exec`. 

For placements the return part is important because you may want to add some stuff related to the placement before returning
```
.addPlacement() ... .save()
   .withServiceNeed() ... .saveAndReturn()
   .withGroupPlacement() ... .saveAndReturn()
  .return()
.addPlacement() ...
```
In most cases there is no next level though so saving without returning is not necessary. My thought was that this way the API is nevertheless more consistent, but maybe there's some better option.. Maybe `save` should return by default since that is most common and then have some other function name for the special case of `save and then add some child entities to this`.

service/src/integrationTest/kotlin/fi/espoo/evaka/ChildBuilder.kt
Joosakur: maybe my other comment related to this also answers this one

service/src/integrationTest/kotlin/fi/espoo/evaka/ChildBuilder.kt
Joosakur: hmm this one I did not understand

service/src/integrationTest/kotlin/fi/espoo/evaka/ChildBuilder.kt
Joosakur: Hah well thats true.. But if it was 
```
child
   .withAbsence() ...
   .withPlacement() ...
      .withServiceNeed() ...
```
then without indentation it would not be clear that service need is child of placement but placement is not child of absence.

Hmm how about this?
```
child
   .addAbsence() ... .save()
   .addPlacement() ... .save()
   .addPlacement() ... .save { placement ->
      placement.addServiceNeed() ... save()
      placement.addGroupPlacement() ... save()
   }
   .addAssistanceNeed() ... .save()
```
Then for the sake of consistency the child save maybe should also take a block function, but maybe it's ok to handle root level differently.

service/src/integrationTest/kotlin/fi/espoo/evaka/occupancy/OccupancyTest.kt
```diff
@@ -0,0 +1,542 @@
...
+            assertEquals(daycareInArea1, occupancyValues[0].key.unitId)
+            assertEquals(2, occupancyValues[0].occupancies.size)
+            assertEquals(1.75, occupancyValues[0].occupancies[period.start]!!.sum)
+            assertEquals(1.0, occupancyValues[0].occupancies[period.end]!!.sum)
+            assertEquals(1, occupancyValues[0].occupancies[period.start]!!.headcount)
+            assertEquals(1, occupancyValues[0].occupancies[period.end]!!.headcount)
+            assertEquals(6.0, occupancyValues[0].occupancies[period.start]!!.caretakers)
+            assertEquals(6.0, occupancyValues[0].occupancies[period.end]!!.caretakers)
+            assertEquals(4.2, occupancyValues[0].occupancies[period.start]!!.percentage)
+            assertEquals(2.4, occupancyValues[0].occupancies[period.end]!!.percentage)
```
jaahon, line 113: With all these assertions I feel it would be simpler to just do `assertEquals(listOf(...), occupancyValues)` 😄

jaahon (approved):
Apart from some of the assertations being a bit hard to read, everything looks good. I'll just approve this so you can merge when you're happy with everything.

service/src/integrationTest/kotlin/fi/espoo/evaka/ChildBuilder.kt
jaahon: I think that's a good idea. Another possible solution would be to make `addServiceNeed` a part of `PlacementBuilder` and "saving" the service need would just store a callback that persists the service need in `PlacementBuilder` that would be called after the placement is persisted. So things would be written to the database only when the whole "building process" is done.

Joosakur:
Redesigned the API based on discussions

PR #858, opened 2021-05-17
Keep external link icon and text on the same row

apeltola:


mikkopiu (commented):
@apeltola Where is this currently broken? Didn't manage to find an example case.

apeltola:
> 
> 
> @apeltola Where is this currently broken? Didn't manage to find an example case.
> EDIT: Before / after screenshots would be nice (unless I'm missing something totally obvious).

The problem occurred in Tampere customization

Before:
image

After:
image

mikkopiu (approved):

PR #861, opened 2021-05-17
Add font weight configurations to theme

patari:
#### Summary

treVaka design requires slightly different font weights for headings. This adds support for it.


akheron:
I think a much cleaner solution would be to make the typography options required, and add the "default values" to the Espoo theme.

patari:
Done.

akheron (approved):

PR #863, opened 2021-05-18
Make CI fail on lint warnings

reynders:

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -666,7 +666,7 @@ jobs:
             - frontend/node_modules
       - run:
           working_directory: *workspace_frontend
-          command: yarn lint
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -666,7 +666,7 @@ jobs:
             - frontend/node_modules
       - run:
           working_directory: *workspace_frontend
-          command: yarn lint
+          command: yarn lint-strict
```
mikkopiu, line 669: You can actually just do:
```suggestion
          command: yarn lint --max-warnings 0
```

as ESLint parses options even after the target files list but either way is fine with me.

mikkopiu (approved):
See comments

PR #877, opened 2021-05-19
EVAKA-4176 occupancy calculation fixes

jaahon:
#### Summary

* Count planned absence as an absence in occupancy calculations
* Use half even rounding for consistency with the old implementation

### DIFF
--- a/service/src/main/kotlin/fi/espoo/evaka/occupancy/Occupancy.kt
+++ b/service/src/main/kotlin/fi/espoo/evaka/occupancy/Occupancy.kt
@@ -5,6 +5,8 @@
 package fi.espoo.evaka.occupancy
 
 import fi.espoo.evaka.daycare.service.AbsenceType
+import fi.espoo.evaka.daycare.service.CareType
+import fi.espoo.evaka.daycare.service.getAbsenceCareTypes
 import fi.espoo.evaka.placement.PlacementType
 import fi.espoo.evaka.shared.db.Database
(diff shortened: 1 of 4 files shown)


Joosakur (approved):

service/src/main/kotlin/fi/espoo/evaka/occupancy/Occupancy.kt
```diff
@@ -750,10 +752,10 @@ WHERE sn.placement_id = ANY(:placementIds)
 
     val absences =
         if (type == OccupancyType.REALIZED)
-            this.createQuery("SELECT child_id, date FROM absence WHERE child_id = ANY(:childIds) AND :period @> date AND NOT absence_type = ANY(:nonAbsences)")
+            this.createQuery("SELECT child_id, date, care_type FROM absence WHERE child_id = ANY(:childIds) AND :period @> date AND NOT absence_type = :presence")
                 .bind("childIds", childIds)
                 .bind("period", period)
-                .bind("nonAbsences", AbsenceType.nonAbsences.toTypedArray())
```
Joosakur, line 756: don't know if some renaming would help here, `financeNonAbsences`? 🤷‍♂️

jaahon: Maybe, but since it's not currently used anyway, I just wanted to delete it. I'll let the absences rewrite solve it.

PR #878, opened 2021-05-19
Add new combobox based on downshift library

Gekkio:
#### Summary

Combobox = input field + dropdown combo for selecting *one* thing. Let's handle multi select and other use cases separately

* implement `Combobox` with downshift `useCombobox` hook
* replace poorly typed react-select component in unit ACL editor
* apply correct colors from theme


akheron:
The new combobox looks a bit different than the old one:

OLD:
Näyttökuva 2021-5-20 kello 8 34 19

NEW:
Näyttökuva 2021-5-20 kello 8 36 02

Not sure about the rounded corners, but I think at least the dropdown should match the width of the input, and the highlighted item should have a light blue background.

Gekkio:
It's *mostly* intentional, because the old component just used react-select default styles and didn't follow our style conventions. But I think you're right about the menu width and the highlight color!

reynders (approved):

PR #882, opened 2021-05-19
EVAKA-4176: missing service needs in unit groups view

Joosakur:

### DIFF
--- a/service/src/main/kotlin/fi/espoo/evaka/backupcare/BackupCareQueries.kt
+++ b/service/src/main/kotlin/fi/espoo/evaka/backupcare/BackupCareQueries.kt
@@ -64,16 +64,10 @@ USING (id)
 JOIN (
   SELECT
     bc.id,
-    coalesce(sum(days_in_range(daterange(bc.start_date, bc.end_date, '[]') * daterange('2020-03-01', NULL) * sn.period)), 0) AS days_with_new_service_need
+    coalesce(sum(days_in_range(daterange(bc.start_date, bc.end_date, '[]') * daterange('2020-03-01', NULL) * daterange(sn.start_date, sn.end_date, '[]'))), 0) AS days_with_new_service_need
   FROM backup_care bc
-  LEFT JOIN (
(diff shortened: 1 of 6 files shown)


service/src/main/kotlin/fi/espoo/evaka/backupcare/BackupCareQueries.kt
```diff
@@ -60,6 +61,22 @@ JOIN (
...
+  LEFT JOIN (
+    SELECT id, child_id, daterange(start_date, end_date, '[]') AS period
+    FROM placement
+  ) AS pl ON bc.child_id = pl.child_id AND daterange(bc.start_date, bc.end_date, '[]') && pl.period
+  LEFT JOIN (
+    SELECT placement_id, daterange(start_date, end_date, '[]') AS period
+    FROM new_service_need
+  ) AS sn ON pl.id = sn.placement_id
+  GROUP BY bc.id
+) AS new_service_need_stats
```
jaahon, line 78: I know this was just basically copied from the `service_need_stats` part above, but I think keeping this simple would be better, ie.
```
  SELECT
    bc.id,
    coalesce(sum(days_in_range(daterange(bc.start_date, bc.end_date, '[]') * daterange('2020-03-01', NULL) * daterange(sn.start_date, sn.end_date, '[]'))), 0) AS days_with_new_service_need
  FROM backup_care bc
  LEFT JOIN placement pl ON bc.child_id = pl.child_id AND daterange(bc.start_date, bc.end_date, '[]') && daterange(pl.start_date, pl.end_date, '[]')
  LEFT JOIN new_service_need sn ON pl.id = sn.placement_id
  GROUP BY bc.id
```

jaahon (approved):

service/src/main/kotlin/fi/espoo/evaka/backupcare/BackupCareQueries.kt
Joosakur: good point, much more readable that way!

PR #887, opened 2021-05-20
EVAKA-4161 absences list

massproduction:
#### Summary
Adds future absences list to the `MarkAbsentBeforehand` page and adds modals with mobile styles to the page.

The designs can be found from [figma](https://www.figma.com/file/8FXipaHxJcHwwfCl273PWS/Kasvattajan-mobiili?node-id=3158%3A3398).


frontend/src/employee-mobile-frontend/components/attendances/actions/MarkAbsentBeforehand.tsx
```diff
@@ -74,6 +86,12 @@ export default React.memo(function MarkAbsentBeforehand() {
     loadFutureAbsences(childId)
   }, [loadFutureAbsences, childId])
 
+  useEffect(() => {
+    if (absences.isSuccess && absences.value.length > 1) {
+      setAbsenceRanges(groupAbsencesByDateRange(absences))
+    }
+  }, [absences])
+
```
jaahon, line 94: In general I don't think storing `absenceRanges` in a separate `useState` hook is a good idea. In reality it's just data computed from `absences` and could just be stored in a variable. If the idea is to minimize unnecessary computations, use `useMemo` instead.

frontend/src/employee-mobile-frontend/components/attendances/actions/MarkAbsentBeforehand.tsx
```diff
@@ -49,7 +56,12 @@ export default React.memo(function MarkAbsentBeforehand() {
   const [selectedAbsenceType, setSelectedAbsenceType] = useState<
     AbsenceType | undefined
   >(undefined)
-  const [, setAbsences] = useState<Result<Absence[]>>(Loading.of())
+  const [absences, setAbsences] = useState<Result<Absence[]>>(Loading.of())
+  const [absenceRanges, setAbsenceRanges] = useState<LocalDate[][]>([])
+  const [uiMode, setUiMode] = useState<
+    'default' | 'confirmDelete' | 'confirmExit'
+  >('default')
+  const [deleteRange, setDeleteRange] = useState<LocalDate[]>([])
```
jaahon, line 64: Doesn't seem like there is any reason to store this data in a `useState` hook, it should just be passed to the `deleteAbsences` function.

service/src/main/kotlin/fi/espoo/evaka/attendance/ChildAttendanceController.kt
```diff
@@ -380,6 +376,40 @@ class ChildAttendanceController(
...
+        user: AuthenticatedUser,
+        @PathVariable childId: UUID,
+        @RequestBody body: AbsenceDateDeleteRequest
+    ): ResponseEntity<Unit> {
+        Audit.AbsenceDelete.log(targetId = childId)
+        acl.getRolesForChild(user, childId).requireOneOfRoles(UserRole.MOBILE)
+        return db.transaction { tx ->
+            tx.deleteAbsencesByDate(childId, body.date)
+        }.let { ResponseEntity.ok(it) }
+    }
```
Joosakur, line 412: The latter endpoint does not even mention delete in path so it's very surprising that such an endpoint deletes something. I would change these from post to delete and pass date/range as request param to keep it more restfull.

frontend/src/lib-components/molecules/modals/InfoModal.tsx
```diff
@@ -165,19 +165,18 @@ function InfoModal({
               </ModalTitle>
               {children}
               {resolve && (
-                <ModalButtons>
+                <ModalButtons $singleButton={!reject}>
```
Joosakur, line 168: what's that `$`?

frontend/src/employee-mobile-frontend/components/attendances/actions/MarkAbsentBeforehand.tsx
```diff
@@ -209,3 +351,35 @@ const TimeInputWrapper = styled.span`
 const Separator = styled.span`
   padding-bottom: 5px;
 `
+
+function groupAbsencesByDateRange(absences: Success<Absence[]>): LocalDate[][] {
```
Joosakur, line 355: `Success` wrapper seems weird here, just unpack it on caller side already. Shouldn't this return `FiniteDateRange[]` instead of `LocalDate[][]`?

frontend/src/employee-mobile-frontend/components/attendances/actions/MarkAbsentBeforehand.tsx
```diff
@@ -209,3 +351,35 @@ const TimeInputWrapper = styled.span`
...
+  const sorted = orderBy(absences.value, ['date'])
+  const absencesGroupedByDate: Absence[][] = [[sorted[0]]]
+  let previousDate: LocalDate = absencesGroupedByDate[0][0].date
+  for (const absence of sorted) {
+    if (
+      differenceInDays(
```
Joosakur, line 361: Lets not use date-fns here. LocalDate should have all the necessary methods. It may use date-fns internally.

frontend/src/employee-mobile-frontend/components/attendances/actions/MarkAbsentBeforehand.tsx
Joosakur: Actually this function does not need to be absence specific at all but
`function groupDatesToRanges(dates: LocalDate[]): FiniteDateRange[]` 
which then should be unit-tested.

frontend/src/employee-mobile-frontend/components/attendances/actions/MarkAbsentBeforehand.tsx
```diff
@@ -74,6 +86,12 @@ export default React.memo(function MarkAbsentBeforehand() {
     loadFutureAbsences(childId)
   }, [loadFutureAbsences, childId])
 
+  useEffect(() => {
+    if (absences.isSuccess && absences.value.length > 1) {
+      setAbsenceRanges(groupAbsencesByDateRange(absences))
```
Joosakur, line 91: Do we need the original absences? We could pass a set function into useRestApi which already does the mapping.

Joosakur (changes requested):

frontend/src/lib-components/molecules/modals/InfoModal.tsx
massproduction: It's a [transient props](https://styled-components.com/docs/api#transient-props) that does not show up in the dom.

frontend/src/employee-mobile-frontend/components/attendances/actions/MarkAbsentBeforehand.tsx
massproduction: We only delete the absences after we have clicked the confirm button in the modal. I added the deleteRange to the state so we know what absence range the user clicked in the list and what are we confirming in the modal. Can't figure out how to pass the information to the modal without storing it in the component state.

jaahon: Yes, sorry, you are absolutely right. I was thinking of a completely different way to do the confirmation modal, but it's not relevant here.

frontend/src/employee-mobile-frontend/components/attendances/actions/MarkAbsentBeforehand.tsx
massproduction: Didn't realise (or I had forgotten that we had FiniteDateRange in the frontend as well. 👍

frontend/src/lib-components/molecules/modals/FormModal.tsx
```diff
@@ -117,12 +125,17 @@ export const ModalBackground = styled.div`
   background: rgba(15, 15, 15, 0.86);
 `
 
-export const ModalButtons = styled.div`
+export const ModalButtons = styled.div<{ $singleButton?: boolean }>`
```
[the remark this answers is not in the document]
massproduction: It's a [transient props](https://styled-components.com/docs/api#transient-props) that does not show up in the dom.

frontend/src/lib-common/utils/absences-date-range.spec.ts
```diff
@@ -0,0 +1,112 @@
...
+      }
+      const absence2: Absence = {
+        id: uniqueId(),
+        childId: uniqueId(),
+        date: LocalDate.today().addDays(2),
+        absenceType: 'SICKLEAVE',
+        careType: 'PRESCHOOL'
+      }
+      const absences = [absence, absence2]
+      expect(groupAbsencesByDateRange(absences)[0].durationInDays()).toBe(-2)
```
Joosakur, line 41: why are these negative?

massproduction: FiniteDateRange.durationInDays() returns the duration as negative values. Maybe it would return positive numbers if instead of start.differenceInDays(end) we would use end.differenceInDays(start).

frontend/src/lib-common/utils/absences-date-range.spec.ts
```diff
@@ -0,0 +1,112 @@
...
+        careType: 'PRESCHOOL'
+      }
+      const absence4: Absence = {
+        id: uniqueId(),
+        childId: uniqueId(),
+        date: LocalDate.today().addDays(8),
+        absenceType: 'SICKLEAVE',
+        careType: 'PRESCHOOL'
+      }
+      const absences = [absence, absence2, absence3, absence4]
```
Joosakur, line 74: I suspect there may be a bug that lodash orderBy does not work with our LocalDate. Lets shuffle this array out of order so then we are also testing whether sorting works or not.

frontend/src/lib-common/utils/absences.ts
```diff
@@ -0,0 +1,41 @@
+import { orderBy } from 'lodash'
+
+import { Absence } from 'lib-common/api-types/child/Absences'
+import FiniteDateRange from 'lib-common/finite-date-range'
+import LocalDate from 'lib-common/local-date'
+
+export function groupAbsencesByDateRange(
+  absences: Absence[]
+): FiniteDateRange[] {
```
Joosakur, line 9: just a thought:
```
export function groupAbsencesByDateRange(
  absences: Absence[]
): FiniteDateRange[] = groupDatesToRanges(absences.map(absence => absence.date))
```

```
export function groupDatesToRanges(
  dates: LocalDate[]
): FiniteDateRange[] {
  // this is now more readable, more reusable and easier to test
}
```

frontend/src/lib-common/utils/absences.ts
```diff
@@ -0,0 +1,41 @@
...
+  if (absences.length > 0) {
+    const sorted = orderBy(absences, ['date'])
+    const absencesGroupedByDate: Absence[][] = [[sorted[0]]]
+    let previousDate: LocalDate = absencesGroupedByDate[0][0].date
+    for (const absence of sorted) {
+      if (absence.date.differenceInDays(previousDate) > 1) {
+        previousDate = absence.date
+        absencesGroupedByDate.push([absence])
+      } else {
+        previousDate = absence.date
```
Joosakur, line 19: this is same on both if and else -> can be moved outside after the block

frontend/src/lib-common/utils/absences.ts
```diff
@@ -0,0 +1,41 @@
...
+      absence.length > 1
+        ? absenceDateRanges.push(
+            new FiniteDateRange(
+              absence[0].date,
+              absence[absence.length - 1].date
+            )
+          )
+        : absenceDateRanges.push(
+            new FiniteDateRange(absence[0].date, absence[0].date)
+          )
```
Joosakur, line 35: doesn't the first part of ternary work for both cases?

frontend/src/lib-common/utils/absences-date-range.spec.ts
Joosakur: lets change it that way

Joosakur (approved):
looks good to me now!

frontend/src/lib-common/utils/local-date.ts
```diff
@@ -10,3 +11,33 @@ import LocalDate from 'lib-common/local-date'
...
+    for (const date of sorted) {
+      if (date.differenceInDays(previousDate) > 1) {
+        groupedDates.push([date])
+      } else {
+        groupedDates[groupedDates.length - 1].push(date)
+      }
+      previousDate = date
+    }
+
+    const absenceDateRanges: FiniteDateRange[] = []
```
Joosakur, line 33: nitpick: variable naming has absences left

frontend/src/lib-common/utils/local-date.ts
```diff
@@ -10,3 +11,33 @@ import LocalDate from 'lib-common/local-date'
...
   return LocalDate.today().differenceInYears(dateOfBirth)
 }
+
+/**
+ *
+ * Groups a list of dates to list of consecutive date ranges
+ */
+export function groupDatesToRanges(dates: LocalDate[]): FiniteDateRange[] {
+  if (dates.length > 0) {
+    const sorted = dates.sort()
```
Gekkio, line 21: `.sort()` mutates the input array, which will at some point lead to problems that are difficult to debug

PR #888, opened 2021-05-20
EVAKA-4148 varda child reset

hruoho:
#### Summary

First draft of functions to delete varda records.

Also add a dev controller to test these against their QA env.

### DIFF
--- a/service/src/main/kotlin/fi/espoo/evaka/varda/VardaDevController.kt
+++ b/service/src/main/kotlin/fi/espoo/evaka/varda/VardaDevController.kt
@@ -1,0 +1,29 @@
+package fi.espoo.evaka.varda
+
+import fi.espoo.evaka.shared.db.Database
+import org.springframework.http.ResponseEntity
+import org.springframework.web.bind.annotation.DeleteMapping
+import org.springframework.web.bind.annotation.PathVariable
+import org.springframework.web.bind.annotation.RequestMapping
+import org.springframework.web.bind.annotation.RestController
+
+@RestController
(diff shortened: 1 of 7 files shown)


service/src/main/kotlin/fi/espoo/evaka/varda/VardaDevController.kt
```diff
@@ -0,0 +1,29 @@
...
+) {
+    @DeleteMapping("/child/{vardaChildId}")
+    fun deleteChild(
+        db: Database.Connection,
+        @PathVariable vardaChildId: Long
+    ): ResponseEntity<Unit> {
+        if (System.getenv("VOLTTI_ENV") != "staging") {
+            return ResponseEntity.notFound().build()
+        }
+        vardaUpdateService.deleteFeeDataByChild(vardaChildId, db)
```
reynders, line 23: Should we wrap all the deletes in a try catch and report any errors with info which child deletion failed (varda id, person id)? The situation requires manual work so it would be helpful to log necessary information for that. Of course an error code will also have to be returned to the caller so that the calling end will know that the child cannot be processed further.

reynders (commented):
Maybe add tests with the help of MockVardaIntegrationEndpoint?

reynders (approved):

service/src/main/kotlin/fi/espoo/evaka/varda/VardaDevController.kt
hruoho: Added some logging, let's see if it suffices

PR #893, opened 2021-05-21
EVAKA-4190 Unify message endpoints

kimifoud:
#### Summary

Move account id from body to request path in POST /messages endpoint to unify paths


mapehe (approved):
👍 better consistency

PR #895, opened 2021-05-21
Fix list-migrations.sh instructions

akheron:
#### Summary

The old instructions just didn't work correctly

### DIFF
--- a/service/README.md
+++ b/service/README.md
@@ -56,10 +56,12 @@ Run integration tests (needs to have docker daemon running):
 ```
 
 To start dev server connected to integration test database (suitable for e.g. running E2E tests without cluttering
- development data):
+development data):
+
 ```sh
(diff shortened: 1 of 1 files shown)


mikkopiu (approved):
The change is good but what exactly didn't work? Should be a straightforward symlink

akheron:
The instruction was `ln -s ../../list-migrations.sh .git/hooks/pre-commit`. This doesn't work because the script to run is `service/list-migrations.sh` (not in the root directory).

If I symlink it with the correct path of `ln -s ../../service/list-migrations.sh .git/hooks/pre-commit`, then I get this error when committing:

```
.git/hooks/pre-commit: line 11: /Users/akheron/work/evaka/.git/hooks/src/main/resources/migrations.txt: No such file or directory
```

PR #897, opened 2021-05-21
Take screenshots after failed Playwright tests

Gekkio:
#### Summary

* split global jest config into separate "projects" so we can customize the configuration just for tests under src/e2e-playwright
* use jest-circus test runner, which has support for various lifecycle hooks
* implement a simple custom jest environment, which allows us to run code after a failing test
* expose screenshot function from `browser.ts` to the jest environment through a nodejs global object
* take screenshots after failed Playwright tests from all browser contexts and pages

### DIFF
--- a/frontend/src/e2e-playwright/browser.ts
+++ b/frontend/src/e2e-playwright/browser.ts
@@ -12,6 +12,17 @@ import playwright, {
 } from 'playwright'
 import config from 'e2e-test-common/config'
 
+declare global {
+  // eslint-disable-next-line @typescript-eslint/no-namespace
+  namespace NodeJS {
+    interface Global {
+      evaka?: {
+        takeScreenshots: (namePrefix: string) => Promise<void>
+      }
+    }
+  }
+}
+
 let browser: Browser
 
 const DISABLE_JEST_TIMEOUT = 1_000_000_000 // 0 or Infinity unfortunately don't work
@@ -21,6 +32,7 @@ beforeAll(async () => {
   browser = await playwright[config.playwright.browser].launch({
     headless: config.playwright.headless
   })
+  global.evaka = { takeScreenshots }
 })
 
 afterEach(async () => {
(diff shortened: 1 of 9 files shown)


jaahon:
Does not work on my machine at the moment.

frontend/src/e2e-playwright/browser.ts
```diff
@@ -21,23 +32,42 @@ beforeAll(async () => {
...
 })
 
 const injected = fs.readFileSync(
   path.resolve(__dirname, '../e2e-test-common/injected.js'),
   'utf-8'
 )
 
+async function takeScreenshots(namePrefix: string): Promise<void> {
+  if (!browser) return
+  const pages = browser.contexts().flatMap((ctx, ctxIndex) =>
```
kimifoud, line 56: `pages.length` seems to always be 0 at this point. However, changing the event handler to listen to `test_fn_failure` seems to do the trick.

frontend/src/e2e-playwright/jest-environment.js
```diff
@@ -0,0 +1,24 @@
...
+  @typescript-eslint/no-unsafe-assignment,
+  @typescript-eslint/no-unsafe-call,
+  @typescript-eslint/no-var-requires,
+  no-undef
+*/
+const NodeEnvironment = require('jest-environment-node')
+
+class PlaywrightEnvironment extends NodeEnvironment {
+  async handleTestEvent(event, _state) {
+    if (event.name === 'test_done' && event.test.errors.length > 0) {
```
kimifoud, line 12:
```suggestion
    if (event.name === 'test_fn_failure') {
```

frontend/src/e2e-playwright/browser.ts
Gekkio: Yeah it looks like some but not all of our tests do `afterEach(() => page?.close())` which means the page is closed before the `test_done` event. Those `page.close()` calls are unnecessary so we could remove them, but maybe `test_fn_failure` is in any case a better event for this use case.

kimifoud (approved):

PR #898, opened 2021-05-21
Refactor messaging

mapehe:
#### Summary

* Close the message editor on send.
* Clear the receiver selection and message content on editor close.
* Reduce the number of /receivers/ requests.

### DIFF
--- a/frontend/src/employee-frontend/components/messages/MessagesPage.tsx
+++ b/frontend/src/employee-frontend/components/messages/MessagesPage.tsx
@@ -4,13 +4,9 @@
 
 import ReceiverSelection from 'employee-frontend/components/messages/ReceiverSelection'
 import {
+  deselectAll,
   getReceiverOptions,
-  getSelected,
-  getSelectedBottomElements,
-  SelectorChange,
-  SelectorNode,
-  unitAsSelectorNode,
-  updateSelector
+  SelectorNode
 } from 'employee-frontend/components/messages/SelectorNode'
 import { Loading, Result } from 'lib-common/api'
 import { useRestApi } from 'lib-common/utils/useRestApi'
@@ -18,11 +14,11 @@ import Container from 'lib-components/layout/Container'
 import { Gap } from 'lib-components/white-space'
 import React, { useEffect, useState } from 'react'
 import styled from 'styled-components'
-import { getMessagingAccounts, getReceivers, postMessage } from './api'
+import { getMessagingAccounts } from './api'
 import MessageEditor from './MessageEditor'
 import MessageList from './MessageList'
 import Sidebar from './Sidebar'
-import { Message, MessageAccount, ReceiverGroup } from './types'
+import { MessageAccount } from './types'
 import { AccountView } from './types-view'
 
 const PanelContainer = styled.div`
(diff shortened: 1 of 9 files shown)


frontend/src/employee-frontend/components/messages/types.ts
```diff
@@ -74,13 +74,13 @@ export const isGroupMessageAccount = (
...
   id: UUID
-  senderId: UUID
   senderName: string
   sentAt: Date
   readAt: Date | null
   title: string
   content: string
   receivers: UUID[]
+  type: MessageType
 }
```
kimifoud, line 84: This interface should reflect the data we get from the backend (`thread.messages`). Type and title are properties of thread. It seems like `Message` and `MessageBody` types have melded. We should be using `MessageBody` or a derivative of it when creating a new message.

kimifoud: Message should only include properties that are found in the backend. https://github.com/espoon-voltti/evaka/blob/52a8f0faca9d6dcca4c2201c389daa86cdef5954/service/src/main/kotlin/fi/espoo/evaka/messaging/message/Message.kt#L17

frontend/src/employee-frontend/components/messages/Sidebar.tsx
```diff
@@ -64,6 +64,13 @@ function Accounts(props: AccountsParams) {
   const [selectedUnit, setSelectedUnit] = useState<SelectOptionProps>(
     unitOptions[0]
   )
+  useEffect(() => {
+    props.view &&
+      props.setView({
+        ...props.view,
+        selectedUnitId: props.view.account.daycareGroup?.unitId
+      })
+  }, [props.view?.account.daycareGroup?.unitId])
```
kimifoud, line 73: What is happening here? 🤔 If we get a view that has an account, immediately set the view again only converting the view to include unit? Shouldn't this happen elsewhere?

Moreover, props will change when any prop change so this will trigger on all renders.

frontend/src/employee-frontend/components/messages/types-view.d.ts
```diff
@@ -9,4 +9,5 @@ export type View = 'RECEIVED' | 'SENT' | 'RECEIVERS' | 'DRAFTS'
 export interface AccountView {
   account: MessageAccount
   view: View
+  selectedUnitId: UUID | undefined
```
kimifoud, line 12: I'm not entirely sure that the `selectedUnitId` belongs here because, the `AccountView` interface is used to  control the selected message box.

frontend/src/employee-frontend/components/messages/MessagesPage.tsx
```diff
@@ -133,14 +137,15 @@ export default React.memo(function MessagesPage() {
...
-              postMessage(accounts.value[0].id, {
+              view?.account &&
+              postMessage(view.account.id, {
                 title: message.title,
                 content: message.content,
-                type: 'MESSAGE',
+                type: message.type,
                 recipientAccountIds: message.receivers
-              })
+              }).then(closeEditor)
```
kimifoud, line 148: `onSend` could be refactored outside of JSX.

frontend/src/employee-frontend/components/messages/MessagesPage.tsx
```diff
@@ -133,14 +137,15 @@ export default React.memo(function MessagesPage() {
           <MessageEditor
             message={message}
             onChange={(message: Message) => setMessage(message)}
-            onClose={() => setShowEditor(false)}
+            onClose={closeEditor}
             onSend={() =>
-              postMessage(accounts.value[0].id, {
+              view?.account &&
+              postMessage(view.account.id, {
```
kimifoud, line 143: Shouldn't the `MessageEditor` give the selected sender account in the callback instead of deducing it from the selected account in that page state?
It would feel more natural if the editor was initialized with the selected account. It's needed there anyway because we want to at least show the sender (if not even select it).

newmsg

frontend/src/employee-frontend/components/messages/MessagesPage.tsx
```diff
@@ -39,56 +51,48 @@ export default React.memo(function MessagesPage() {
...
-    title: '',
-    receivers: [],
-    content: '',
-    id: '',
-    senderId: '',
-    senderName: '',
-    sentAt: new Date(),
-    readAt: new Date()
-  })
+  const [message, setMessage] = useState<Message>(emptyMessage)
```
kimifoud, line 60: `MessageBody` or some derivative of that should be used instead of `Message`

kimifoud (approved):
Okay!

PR #904, opened 2021-05-24
EVAKA-4148 Allow traffic to varda-dev controller

hruoho:
#### Summary

Add missing config to enable dev controller unless in prod

### DIFF
--- a/service/src/main/kotlin/fi/espoo/evaka/varda/VardaDevController.kt
+++ b/service/src/main/kotlin/fi/espoo/evaka/varda/VardaDevController.kt
@@ -17,13 +17,13 @@ class VardaDevController(
         db: Database.Connection,
         @PathVariable vardaChildId: Long
     ): ResponseEntity<Unit> {
-        if (System.getenv("VOLTTI_ENV") == "prod") {
-            return ResponseEntity.notFound().build()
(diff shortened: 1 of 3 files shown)


service/src/main/kotlin/fi/espoo/evaka/varda/VardaDevController.kt
```diff
@@ -17,7 +17,7 @@ class VardaDevController(
         db: Database.Connection,
         @PathVariable vardaChildId: Long
     ): ResponseEntity<Unit> {
-        if (System.getenv("VOLTTI_ENV") != "staging") {
+        if (System.getenv("VOLTTI_ENV") == "prod") {
```
reynders, line 20: I'd rather have it the other way round - check if we are in environments where the delete is allowed (shortlist: dev and staging). This way we never ever accidentally delete data by some misconfiguration

reynders (approved):

service/src/main/kotlin/fi/espoo/evaka/varda/VardaDevController.kt
hruoho: Agreed, that's somewhat safer way

PR #906, opened 2021-05-24
EVAKA-4190 Save message draft state automatically

kimifoud:
#### Summary

- Save message content as draft
- Open draft from the draft section
- Discard draft after message is sent or when manually discarded
- Move message state to context as it is needed to be refreshed from multiple places
- Fix pre-selection of sender and recipients
- Refactor message editor logic
  - Handle receiver selection internally within the editor. Previously the receiver selection was passed from the outside and updated from within the editor creating a two-way data flow. Now the editor just gets the initial state and handles it from there by itself.
- Make all fields of draft content required.

### DIFF
--- a/frontend/src/employee-frontend/components/messages/MessageEditor.tsx
+++ b/frontend/src/employee-frontend/components/messages/MessageEditor.tsx
@@ -1,4 +1,5 @@
 import { Result } from 'lib-common/api'
+import { UpdateStateFn } from 'lib-common/form-state'
 import { UUID } from 'lib-common/types'
 import { useDebounce } from 'lib-common/utils/useDebounce'
 import { useRestApi } from 'lib-common/utils/useRestApi'
@@ -13,38 +14,47 @@ import { Label } from 'lib-components/typography'
 import { defaultMargins, Gap } from 'lib-components/white-space'
 import colors from 'lib-customizations/common'
 import { faTimes, faTrash } from 'lib-icons'
-import React, { useCallback, useEffect, useState } from 'react'
+import React, { useCallback, useEffect, useMemo, useState } from 'react'
 import styled from 'styled-components'
 import Select from '../../components/common/Select'
 import { useTranslation } from '../../state/i18n'
 import * as api from './api'
 import {
+  deselectAll,
+  getReceiverOptions,
   getSelected,
   getSelectedBottomElements,
-  SelectorChange,
   SelectorNode,
   updateSelector
 } from './SelectorNode'
-import { DraftContent, MessageBody } from './types'
+import { DraftContent, MessageBody, UpsertableDraftContent } from './types'
 
-const emptyMessageBody: MessageBody = {
-  title: '',
-  content: '',
-  type: 'MESSAGE',
-  recipientAccountIds: []
+type Message = UpsertableDraftContent & {
+  senderId: UUID
+  recipientAccountIds: UUID[]
 }
 
-const draftContentToInitialState = (
-  draft: DraftContent | undefined
-): MessageBody =>
-  draft
-    ? {
-        content: draft.content ?? '',
-        recipientAccountIds: draft.recipientAccountIds ?? [],
-        title: draft.title ?? '',
-        type: draft.type ?? 'MESSAGE'
-      }
-    : emptyMessageBody
+const getInitialMessage = (
+  draft: DraftContent | undefined,
+  senderId: UUID
+): Message => ({
+  senderId,
+  content: draft?.content ?? '',
+  recipientIds: draft?.recipientIds ?? [],
+  recipientNames: draft?.recipientNames ?? [],
+  recipientAccountIds: [],
+  title: draft?.title ?? '',
+  type: draft?.type ?? 'MESSAGE'
+})
+
+const areRequiredFieldsFilledForMessage = (msg: Message): boolean =>
+  !!(msg.recipientAccountIds?.length && msg.type && msg.content && msg.title)
+
+const createReceiverTree = (tree: SelectorNode, selectedIds: UUID[]) =>
+  selectedIds.reduce(
+    (acc, next) => updateSelector(acc, { selectorId: next, selected: true }),
+    deselectAll(tree)
+  )
 
 type Option = {
   label: string
@@ -54,11 +64,7 @@ type Option = {
 interface Props {
   defaultSender: Option
   senderOptions: Option[]
-  selectedReceivers: SelectorNode
-  receiverOptions: Option[]
-  setSelectedReceivers: React.Dispatch<
-    React.SetStateAction<SelectorNode | undefined>
-  >
+  availableReceivers: SelectorNode
   onSend: (
     accountId: UUID,
     messageBody: MessageBody,
@@ -72,9 +78,7 @@ interface Props {
 export default React.memo(function MessageEditor({
   defaultSender,
   senderOptions,
-  selectedReceivers,
-  receiverOptions,
-  setSelectedReceivers,
+  availableReceivers,
   onSend,
   onDiscard,
   onClose,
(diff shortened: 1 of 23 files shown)


frontend/src/employee-frontend/components/messages/MessageEditor.tsx
```diff
@@ -84,11 +151,39 @@ export default React.memo(function MessageEditor({
...
+    const timeoutHandle = setTimeout(clearStatus, 1500)
+    return () => clearTimeout(timeoutHandle)
+  }, [i18n, draftSaveResult])
+
+  const title =
+    debouncedTitleStatus ||
+    messageBody.title ||
+    i18n.messages.messageEditor.newMessage
+
+  const onCloseHandler = useCallback(() => onClose(!!draftSaveResult), [
```
mapehe, line 177: As a user I'd expect this to save the result on close, unless I'd choose to discard the draft. At the moment it doesn't to seem to work that way locally.

frontend/src/employee-frontend/components/messages/MessageEditor.tsx
```diff
@@ -54,17 +76,62 @@ export default React.memo(function MessageEditor({
...
+
+  // initialize draft when requested
+  useEffect(() => {
+    if (newDraftRequested && !draftId) {
+      initDraft(selectedAccount.value)
+    }
+  }, [draftId, newDraftRequested, initDraft, selectedAccount.value])
+
+  // save draft on input change or ask for a new draft to be initialized
+  const debouncedMessageBody = useDebounce(messageBody, 2000)
```
mapehe, line 115: There's something weird going on with the UI at my end:

The only way for me to update the draft seems to be to change the `sender`. Editing any other fields doesn't seem to send any XHRs.

frontend/src/employee-frontend/components/messages/MessagesPageContext.tsx
```diff
@@ -0,0 +1,171 @@
...
+import { useRestApi } from '../../../lib-common/utils/useRestApi'
+import { UUID } from '../../types'
+import { getMessageDrafts, getReceivedMessages, getSentMessages } from './api'
+import { DraftContent, MessageThread, SentMessage } from './types'
+import { AccountView } from './types-view'
+
+const PAGE_SIZE = 20
+
+export interface MessagesPageState {
+  selectedDraft: DraftContent | undefined
```
mapehe, line 18: It seems that at the moment, while everything else works, the receivers are not preserved correctly when a draft is reopened.

Perhaps the reason is that instead of the actual messaging account ids we should store the `SelectorNode` ids after extracting them with `getSelected(selectedReceivers).map(({value}) => value)`. Reselecting those nodes would then be done by something like
```
recordedIds
  .map(id => ({selectorId: id, selected: true}))
  .reduce((selection, selectorChange) => updateSelector(selection, selectorChange), emptySelection)
```
The downside is that there is no guarantee that the messaging account ids are rendered the same way when the draft is reopened (there could have been changes to a group, for example). This has been discussed before and I think that the consensus was that it still is the way it this is supposed to work.

mapehe (changes requested):

frontend/src/employee-frontend/components/messages/MessagesPageContext.tsx
kimifoud: Changed the architecture so that the selected `SelectorNode`s are saved instead of message accounts.

frontend/src/employee-frontend/components/messages/MessageEditor.tsx
kimifoud: Revamped the whole draft save logic. All fields use the same debounced save logic now.

frontend/src/employee-frontend/components/messages/MessageEditor.tsx
kimifoud: Close is now disabled while there are unsaved changes (mvp).

service/src/main/resources/db/migration/V91__message_draft_non_nullable.sql
```diff
@@ -0,0 +1,20 @@
...
+ALTER TABLE message_draft ALTER COLUMN title SET DEFAULT '';
+ALTER TABLE message_draft ALTER COLUMN title SET NOT NULL;
+
+UPDATE message_draft SET content = '' WHERE content IS NULL;
+ALTER TABLE message_draft ALTER COLUMN content SET DEFAULT '';
+ALTER TABLE message_draft ALTER COLUMN content SET NOT NULL;
+
+UPDATE message_draft SET type = 'MESSAGE' WHERE type IS NULL;
+ALTER TABLE message_draft ALTER COLUMN type SET DEFAULT 'MESSAGE';
+ALTER TABLE message_draft ALTER COLUMN type SET NOT NULL;
```
akheron, line 20: 💎

frontend/src/employee-frontend/components/messages/MessagesPageContext.tsx
```diff
@@ -0,0 +1,171 @@
...
+  useEffect,
+  useMemo,
+  useState
+} from 'react'
+import { Loading, Paged, Result } from '../../../lib-common/api'
+import { useRestApi } from '../../../lib-common/utils/useRestApi'
+import { UUID } from '../../types'
+import { getMessageDrafts, getReceivedMessages, getSentMessages } from './api'
+import { DraftContent, MessageThread, SentMessage } from './types'
+import { AccountView } from './types-view'
```
akheron, line 13: Stuff from `types-view.d.ts` could be moved to `types.ts`, because they cross-import each other.

akheron:
```
export interface BaseMessageAccount {
  id: UUID
  name: string
}
export interface MessageAccount extends BaseMessageAccount {
  personal: boolean
  daycareGroup?: {
    id: UUID
    name: string
    unitId: UUID
    unitName: string
  }
  unreadCount: number
}
export type GroupMessageAccount = Required<MessageAccount>
```

This is a confusing construction. What happens if `personal === true` and `daycareGroup !== undefined`?

frontend/src/employee-frontend/components/messages/MessageEditor.tsx
```diff
@@ -1,105 +1,232 @@
...
+  senderId: UUID
+): Message => ({
+  senderId,
+  content: draft?.content ?? '',
+  recipientIds: draft?.recipientIds ?? [],
+  recipientNames: draft?.recipientNames ?? [],
+  recipientAccountIds: [],
+  title: draft?.title ?? '',
+  type: draft?.type ?? 'MESSAGE'
+})
```
akheron, line 48:
```suggestion
const emptyMessage = {
  content: '',
  recipientIds: [],
  recipientNames: [],
  recipientAccountIds: [],
  title: '',
  type: 'MESSAGE' as const
}

const getInitialMessage = (
  draft: DraftContent | undefined,
  senderId: UUID
): Message =>
  draft
    ? { ...draft, senderId, recipientAccountIds: [] }
    : { senderId, ...emptyMessage }
```

Something like this would be cleaner.

kimifoud: Sure would! 👍

frontend/src/employee-frontend/components/messages/MessagesPageContext.tsx
kimifoud: It's 2021 and there's no built-in tool for checking circular dependencies ☹️  Moved the view-related message boxes to `types-view`.

frontend/src/employee-frontend/components/messages/MessageEditor.tsx
kimifoud: Draft is now saved on close.

mapehe (changes requested):
image

Great fixes on my previous points!

Still noticed what seems like a small bug: When a draft is saved, I get these empty extra drafts. Each new draft seems to create one.

frontend/src/employee-frontend/components/messages/MessagesPage.tsx
```diff
@@ -2,52 +2,96 @@
...
+    selectedDraft,
+    setSelectedDraft,
+    view,
+    setView,
+    refreshMessages
+  } = useContext(MessagesPageContext)
+
   const [accounts, setResult] = useState<Result<MessageAccount[]>>(Loading.of())
   const loadAccounts = useRestApi(getMessagingAccounts, setResult)
   useEffect(() => loadAccounts(), [loadAccounts])
```
akheron, line 39:
```suggestion
  const [accounts, setResult] = useState<Result<MessageAccount[]>>(Loading.of())
  const loadAccounts = useRestApi(
    getMessagingAccounts,
    (result: Result<MessageAccount[]>) => {
      setResult(result)
      if (result.isSuccess && result.value[0]) {
        setView({
          view: 'RECEIVED',
          account: result.value[0]
        })
      }
    }
  )
  useEffect(() => loadAccounts(), [loadAccounts])
```
One less `useEffect` if you will. Would be even cooler if account handling was fully moved to the context.

akheron: This would replace the `useEffect` around line 96

frontend/src/employee-frontend/components/messages/useDraft.ts
```diff
@@ -0,0 +1,93 @@
...
+  }, [])
+
+  return {
+    draftId: id,
+    state: saveState,
+    setDraft: setDraftCallback,
+    saveDraft: saveDraftCallback,
+    wasModified
+  }
+}
```
akheron, line 93: 💎

mapehe (approved):
Great work 👏

frontend/src/employee-frontend/components/messages/MessagesPage.tsx
kimifoud: Moved account handling to context.

PR #907, opened 2021-05-24
Reimplement messaging e2e tests

mapehe:


kimifoud (approved):
The test file could be renamed, too.

PR #910, opened 2021-05-25
Update README based on new dev experience

huhtari:
- Clarified (apparently optional) additional configuration need
- Document problem with another database server running


mikkopiu (approved):
:+1:

PR #919, opened 2021-05-26
EVAKA-4176: fix and refactor gap calculation

Joosakur:

### DIFF
--- a/frontend/src/lib-common/finite-date-range.ts
+++ b/frontend/src/lib-common/finite-date-range.ts
@@ -5,6 +5,7 @@
 import LocalDate from './local-date'
 import DateRange from './date-range'
 import { JsonOf } from './json'
+import { groupDatesToRanges } from './utils/local-date'
 
 function maxOf(a: LocalDate, b: LocalDate) {
   return a.isAfter(b) ? a : b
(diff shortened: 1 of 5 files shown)


frontend/src/lib-common/finite-date-range.ts
```diff
@@ -123,6 +124,13 @@ export default class FiniteDateRange {
     return this.start.isEqual(other.start) && this.end.isEqual(other.end)
   }
 
+  getGaps(childRanges: FiniteDateRange[]): FiniteDateRange[] {
+    const gapDates = [...this.dates()].filter(
+      (date) => childRanges.find((range) => range.includes(date)) === undefined
```
jaahon, line 129: I think using `!childRanges.some((range) => range.includes(date))` or `childRanges.every((range) => !range.includes(date))` makes the intent clearer here.

jaahon (approved):
Good idea to reuse `groupDatesToRanges` here 👍

Joosakur:
> Good idea to reuse `groupDatesToRanges` here 👍

@massproduction did not actually expect this to become useful so fast, but good thing it was written to be reusable! :D

PR #924, opened 2021-05-27
EVAKA-4190 Page layout fixes

kimifoud:
#### Summary

- Keep message title in its container
- Make messages page occupy the full page height. Currently it is implemented by setting a fixed height to the page container, which is sub-optimal.
- Make accounts panel and message list scrollable.
- "Sticky" new message button
- Add a gap between accounts and new message button

### DIFF
--- a/frontend/src/employee-frontend/components/messages/MessagesPage.tsx
+++ b/frontend/src/employee-frontend/components/messages/MessagesPage.tsx
@@ -4,9 +4,9 @@
 
 import { UUID } from 'lib-common/types'
 import Container from 'lib-components/layout/Container'
-import { Gap } from 'lib-components/white-space'
 import React, { useContext, useEffect, useState } from 'react'
 import styled from 'styled-components'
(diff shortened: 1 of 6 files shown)


frontend/src/employee-frontend/components/messages/MessagesPage.tsx
```diff
@@ -19,7 +19,19 @@ import { deselectAll, SelectorNode } from './SelectorNode'
 import Sidebar from './Sidebar'
 import { MessageBody } from './types'
 
+// TODO is fixed header height possible?
+// If not, replace with a stretching flex container with scrollable children
+const approximatedHeaderHeight = `164px`
+const MessagesPageContainer = styled(Container)`
+  height: calc(100vh - ${approximatedHeaderHeight});
+`
```
kimifoud, line 27: I played around with this for an hour so but couldn't come up with a solution that would retain the correct page width while always occupying all the available height.

mapehe (approved):
LGTM 💎

PR #927, opened 2021-05-27
EVAKA-4190 Finalize receiver selection layout

kimifoud:
#### Summary

- Add missing date of birth and parents to recipient selector
- Make childNodes a required property in SelectorNode
- Unify layout by wrapping receiver selection in ContentArea

### DIFF
--- a/frontend/src/employee-frontend/components/messages/SelectorNode.ts
+++ b/frontend/src/employee-frontend/components/messages/SelectorNode.ts
@@ -4,6 +4,7 @@
 
 import { ReceiverGroup } from 'employee-frontend/components/messages/types'
 import { UUID } from 'employee-frontend/types'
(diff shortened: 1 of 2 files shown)


frontend/src/employee-frontend/components/messages/SelectorNode.ts
```diff
@@ -18,7 +19,7 @@ export interface SelectorNode {
   selectorId: UUID
   selected: boolean
   name: string
-  childNodes?: SelectorNode[]
+  childNodes: SelectorNode[]
```
mapehe, line 22: Yeah, tbh `childNodes = []` is better than `childNodes = undefined`. 👍

mapehe (approved):
Nice work 👍

PR #928, opened 2021-05-27
Additional e2e tests

mapehe:
#### Summary

* Test that a draft is saved correctly.
* Test that a draft is not saved when a message is sent.
* Test that a draft is not saved when it's discarded.

### DIFF
--- a/frontend/src/e2e-test/pages/employee/messaging/messages-page.ts
+++ b/frontend/src/e2e-test/pages/employee/messaging/messages-page.ts
@@ -6,17 +6,40 @@ import { t, Selector } from 'testcafe'
 
 export default class MessagesPage {
   readonly unitsListUnit = (id: string) => Selector(`[data-qa="unit-${id}"]`)
-
+  readonly draftBoxes = Selector(`[data-qa="message-box-row-DRAFTS"]`)
+  readonly draftBox = (index: number) => this.draftBoxes.nth(index)
+  readonly draftMessageRows = Selector(`[data-qa="draft-message-row"]`)
+  readonly draftMessageRow = (index: number) => this.draftMessageRows.nth(index)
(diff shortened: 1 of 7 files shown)


frontend/src/e2e-test/pages/employee/messaging/messages-page.ts
```diff
@@ -6,17 +6,40 @@ import { t, Selector } from 'testcafe'
...
+  async draftNewMessage(title: string, content: string) {
+    await t.click(this.newMessageBtn)
+    await t.click(this.messageEditorSender)
+    await t.pressKey('enter')
+    await t.click(this.messageEditorReceiver)
+    await t.pressKey('enter')
+    await t.typeText(this.messageEditorTitle, title)
+    await t.typeText(this.messageEditorContent, content)
+    // Draft is not saved immediately so we wait for a bit.
+    await t.wait(5000)
```
kimifoud, line 32: Could we wait until 'Tallennettu' appears instead? It would also be nice to be able to detect that we're in CI mode and set the delay to a much lower value.

service/src/main/kotlin/fi/espoo/evaka/shared/dev/DevApi.kt
```diff
@@ -554,6 +554,7 @@ RETURNING id
 
     @DeleteMapping("/message/delete-all")
     fun deleteMessages(db: Database) {
+        db.transaction { it.execute("DELETE FROM message_draft") }
```
kimifoud, line 557: Doesn't really matter but the same transaction could be used for all these operations.

mapehe:
Thanks for the feedback @kimifoud, made some improvements accordingly 🙏

kimifoud (approved):
👍

PR #931, opened 2021-05-28
EVAKA-4190 Citizen replies to message thread

kimifoud:
#### Summary

- Implement citizen's reply to message feature with an inline editor
  - pixel tweaks and e2e tests can be done separately
- Move citizen message state to context
- Add recipient accounts to message threads schema in order to display thread participants


frontend/src/citizen-frontend/messages/state.tsx
```diff
@@ -2,43 +2,276 @@
...
+            ...state,
+            threads: [
+              {
+                ...old,
+                messages: [...old.messages, message]
+              },
+              ...rest
+            ]
+          }
+        })
```
akheron, line 170: This could be given a name. After reading 5 times and checking the types, I now _think_ I understand what it does 😅

frontend/src/lib-components/atoms/form/InputField.tsx
```diff
@@ -283,9 +283,8 @@ export const TextArea = styled(TextareaAutosize)`
   resize: none;
 
   background-color: transparent;
-  border-style: solid;
-  border-width: 0 0 1px;
-  border-color: ${({ theme: { colors } }) => colors.greyscale.medium};
+  border: 2px solid transparent;
+  border-bottom: 1px solid ${({ theme: { colors } }) => colors.greyscale.medium};
```
akheron, line 287: Do these changes affect many parts of the project?

mapehe:
image

Nice! The messages seem to move back and forth as intended. 👏 

Noticed that, at the employee end, the received message count seems to count messages even though it should probably (?) count threads. For example, in the above image, I'd expect the count to be 1, not 3.

Also, this is what I see at the sent view:

image

I'd expect that clicking on the sent message on the right would do something, but at the moment it doesn't. Is this intended?

frontend/src/lib-components/atoms/form/InputField.tsx
kimifoud: Yes, and I actually checked that the new implementation fixes the flickering effect in FF/Safari/Chrome.

kimifoud:
@mapehe unread counts are out of scope, there's a separate ticket on the Miro board. Same goes for navigating from sent messages to single thread view.

frontend/src/citizen-frontend/messages/state.tsx
kimifoud: Better? I ditched lodash too, because the performance impact is negligible.

akheron (approved):

PR #934, opened 2021-05-28
EVAKA-4190 Refactor message account queries

akheron:
#### Summary
- Use person/employee ids instead of `AuthenticatedUser` when querying message accounts, because only the `id` was used.
- Use account id to reference message accounts, instead of `MessageAccount` that also contains the rarely needed account name.
- Rename `getMessageAccountFor*()` to `get*MessageAccount()`
- Rename `getAuthorizedMessageAccountsForEmployee()` to `getEmployeeDetailedMessageAccount()`


kimifoud (approved):
🥇

PR #936, opened 2021-05-28
EVAKA-4184 Replace pricing with feeThresholds

hruoho:
#### Summary

Replace pricing with feeThresholds


service/src/main/kotlin/fi/espoo/evaka/invoicing/domain/FeeThresholdsWithValidity.kt
```diff
@@ -144,6 +108,14 @@ data class FeeThresholds(
             4 -> maxIncomeThreshold2
             5 -> maxIncomeThreshold5
             6 -> maxIncomeThreshold6
-            else -> maxIncomeThreshold6 + ((familySize - 6) * thresholdIncrease6Plus)
+            else -> maxIncomeThreshold6 + ((familySize - 6) * incomeThresholdIncrease6Plus)
+        }
+
+    fun siblingDiscountPercent(siblingOrdinal: Int): Int =
```
hruoho, line 114: I guess this should be the coefficient not percentage

service/src/main/kotlin/fi/espoo/evaka/shared/dev/DataInitializers.kt
```diff
@@ -524,7 +525,7 @@ fun Database.Transaction.insertTestFeeAlteration(
     return id
 }
 
-fun Database.Transaction.insertTestPricing(pricing: DevPricing) = insertTestDataRow(
+fun Database.Transaction.insertTestPricingOld(pricing: DevPricing) = insertTestDataRow(
```
jaahon, line 528: This should just be deleted.

service/src/main/kotlin/fi/espoo/evaka/shared/dev/DataInitializers.kt
```diff
@@ -533,6 +534,15 @@ RETURNING id
     """.trimIndent()
 )
 
+fun Database.Transaction.insertTestPricing(pricing: FeeThresholdsWithValidity) = insertTestDataRow(
```
jaahon, line 537: I think it would be a good idea to get rid of the old "pricing" name.

service/src/main/kotlin/fi/espoo/evaka/invoicing/domain/FeeThresholdsWithValidity.kt
jaahon: I agree. Why not refactor it now?

jaahon (commented):
Everything looks good in general. Maybe it would be a good idea to do as much renaming as is convenient.

Also, we'll need to do a fee decision migration if we don't want to generate new fee decisions unnecessarily after this change. We can plan this together.

service/src/main/kotlin/fi/espoo/evaka/shared/dev/DataInitializers.kt
hruoho: Let's do this exhaustively later

service/src/main/kotlin/fi/espoo/evaka/shared/dev/DevApi.kt
```diff
@@ -361,27 +362,35 @@ class DevApi(
...
+INSERT INTO fee_thresholds (id, valid_during,
+    income_multiplier_2, income_multiplier_3, income_multiplier_4, income_multiplier_5, income_multiplier_6,
+    min_income_threshold_2, min_income_threshold_3, min_income_threshold_4, min_income_threshold_5, min_income_threshold_6,
+    max_income_threshold_2, max_income_threshold_3, max_income_threshold_4, max_income_threshold_5, max_income_threshold_6,
+    income_threshold_increase_6_plus, min_fee, max_fee, sibling_discount_2, sibling_discount_2_plus)
+VALUES (:id, :validDuring,
+    :incomeMultiplier2, :incomeMultiplier3, :incomeMultiplier4, :incomeMultiplier5, :incomeMultiplier6,
+    :minIncomeThreshold2, :minIncomeThreshold3, :minIncomeThreshold4, :minIncomeThreshold5, :minIncomeThreshold6,
+    :maxIncomeThreshold2, :maxIncomeThreshold3, :maxIncomeThreshold4, :maxIncomeThreshold5, :maxIncomeThreshold6,
+    :incomeThresholdIncrease6Plus, :minFee, :maxFee, :siblingDiscount2, :siblingDiscount2Plus)
```
jaahon, line 378: Couldn't you just use `insertTestPricing` from `DataInitializers.kt` here?

service/src/main/resources/db/migration/V98__update_fee_decision_pricing.sql
```diff
@@ -0,0 +1,30 @@
...
+            WHEN family_size = 2 THEN (pricing->'minThreshold2')::int
+            WHEN family_size = 3 THEN (pricing->'minThreshold3')::int
+            WHEN family_size = 4 THEN (pricing->'minThreshold4')::int
+            WHEN family_size = 5 THEN (pricing->'minThreshold5')::int
+            WHEN family_size = 6 THEN (pricing->'minThreshold6')::int
+            WHEN family_size > 6 THEN (pricing->'minThreshold6')::int + (pricing->'thresholdIncrease6Plus')::int * (family_size - 6)
+        END,
+        'maxIncomeThreshold',
+        CASE
+            WHEN family_size = 2 THEN 0
```
jaahon, line 16: Should be `WHEN family_size = 1 THEN 0`.

service/src/main/resources/db/migration/V98__update_fee_decision_pricing.sql
```diff
@@ -0,0 +1,30 @@
...
+            WHEN family_size = 3 THEN (pricing->'minThreshold3')::int
+            WHEN family_size = 4 THEN (pricing->'minThreshold4')::int
+            WHEN family_size = 5 THEN (pricing->'minThreshold5')::int
+            WHEN family_size = 6 THEN (pricing->'minThreshold6')::int
+            WHEN family_size > 6 THEN (pricing->'minThreshold6')::int + (pricing->'thresholdIncrease6Plus')::int * (family_size - 6)
+        END + (pricing->'maxThresholdDifference')::int,
+        'incomeMultiplier',
+        pricing->'multiplier',
+        'maxFee',
+        (pricing->'maxThresholdDifference')::int * (pricing->'multiplier')::numeric,
```
jaahon, line 27: This needs to be rounded back to whole euros, so something like `((pricing->'maxThresholdDifference')::int * (pricing->'multiplier')::numeric / 100.0)::int * 100`.

jaahon (approved):

PR #937, opened 2021-05-28
EVAKA-4043: child image on mobile

Joosakur:


frontend/src/employee-mobile-frontend/components/attendances/child-info/ImageEditor.tsx
```diff
@@ -0,0 +1,170 @@
...
+      onReturn()
+    }
+
+    window.addEventListener('popstate', popStateHandler)
+
+    return () => {
+      window.removeEventListener('popstate', popStateHandler)
+    }
+  }, [onReturn, h])
+
```
Joosakur, line 54: this feels like a hack and causes warning in console. is there a better way to prevent back navigation and call `onReturn` instead?

massproduction (approved):
💎

PR #941, opened 2021-05-31
EVAKA-4184 basic fee thresholds editor

jaahon:
#### Summary

Add a basic fee thresholds editor for creating new fee thresholds.

Also changes InputField's wrapper div to be inline so that the component behaves more like an input by default.

### DIFF
--- a/frontend/src/employee-frontend/components/finance-basics/FeeThresholdsItemEditor.tsx
+++ b/frontend/src/employee-frontend/components/finance-basics/FeeThresholdsItemEditor.tsx
@@ -1,0 +1,342 @@
+import React, { useState } from 'react'
+import styled from 'styled-components'
+import DateRange from 'lib-common/date-range'
+import LocalDate from 'lib-common/local-date'
+import { H3LikeLabel, Label } from 'lib-components/typography'
+import { defaultMargins } from 'lib-components/white-space'
+import {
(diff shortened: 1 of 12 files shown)


frontend/src/employee-frontend/components/finance-basics/FeeThresholdsItemEditor.tsx
```diff
@@ -0,0 +1,393 @@
...
+              }))
+            }
+            placeholder="0"
+            symbol={'€'}
+            info={validationErrorInfo('minFee')}
+            hideErrorsBeforeTouched
+          />
+        </FixedSpaceColumn>
+      </RowWithMargin>
+      {showMaxFeeErrorsInfo ? <div>ERROR</div> : null}
```
hruoho, line 247: Is this intentional?

jaahon: 😂 No

hruoho:
What about editing, should we implement the edit mode later in a separate task?

hruoho (approved):
Looks good to me despite that error thing 👍

jaahon:
> What about editing, should we implement the edit mode later in a separate task?

Yeah there's a task for that. I wanted to get this first part out of the way and do the editing and copying later.

PR #942, opened 2021-05-31
EVAKA-4190 Delete message account when daycare group is deleted

akheron:


service/src/main/kotlin/fi/espoo/evaka/daycare/service/DaycareService.kt
```diff
@@ -56,11 +57,11 @@ class DaycareService {
             groupId = groupId,
             startDate = null,
             endDate = null
-        ).isEmpty()
+        ).isEmpty() && !daycareGroupHasMessages(tx, groupId)
 
         if (!isEmpty) throw Conflict("Cannot delete group which has children placed in it")
```
kimifoud, line 62: Error message should be updated or even better, have separate messages.
```kt
tx.getPlacements(..).isEmpty() || throw ..
daycareGroupHasMessages(..) || throw ..
```

But. Shouldn't the catch clause already handle the existing fk reference even without an explicit "has messages" query?

service/src/main/resources/db/migration/V95__message_account_foreign_keys.sql
```diff
@@ -0,0 +1,8 @@
+ALTER TABLE message_account DROP CONSTRAINT message_account_created_for_fk;
+
+-- CASCADE also drops the message_account_name_view, but it's recreated in an R__ migration
+ALTER TABLE message_account DROP COLUMN deleted_owner_name CASCADE;
+
+ALTER TABLE message_account ADD CONSTRAINT message_account_created_for_fk CHECK (
+    num_nonnulls(daycare_group_id, employee_id, person_id) = 1
+);
```
kimifoud, line 8: Doesn't this now have a slight chance of failing if someone has already deleted a message group with messages?
It shouldn't be the case, but let's keep eyes open when deploying.

akheron: Yes, has to be monitored

kimifoud (approved):
👍

PR #944, opened 2021-06-01
EVAKA-4190 Authorize daycare group message account access

akheron:
#### Summary

- Supervisors see all group accounts in their daycares
- Staff members only see their own groups' accounts (based on `daycare_group_acl`)
- Admins do **not** see all message accounts. Not sure if this is desirable.

### DIFF
--- a/service/src/main/kotlin/fi/espoo/evaka/messaging/message/MessageAccountQueries.kt
+++ b/service/src/main/kotlin/fi/espoo/evaka/messaging/message/MessageAccountQueries.kt
@@ -36,8 +36,9 @@ SELECT acc.id
 FROM message_account acc
 LEFT JOIN daycare_group dg ON acc.daycare_group_id = dg.id
     LEFT JOIN daycare dc ON dc.id = dg.daycare_id
-    LEFT JOIN daycare_acl acl ON acl.daycare_id = dg.daycare_id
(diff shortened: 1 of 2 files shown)


service/src/main/kotlin/fi/espoo/evaka/messaging/message/MessageAccountQueries.kt
```diff
@@ -67,6 +68,7 @@ FROM message_account acc
     LEFT JOIN daycare_group dg ON acc.daycare_group_id = dg.id
     LEFT JOIN daycare dc ON dc.id = dg.daycare_id
     LEFT JOIN daycare_acl acl ON acl.daycare_id = dg.daycare_id
+    LEFT JOIN daycare_group_acl gacl ON gacl.daycare_group_id = dg.id
```
kimifoud, line 71: Redundant join? Actually, it seems like both of these acl joins are redundant now.

kimifoud (changes requested):

service/src/main/kotlin/fi/espoo/evaka/messaging/message/MessageAccountQueries.kt
akheron: Good catch, fixed!

kimifoud (approved):

PR #947, opened 2021-06-01
EVAKA-4190 Fix sender selection

kimifoud:
#### Summary

- Move sender to message state instead of separate useState
- Use named functions for effects instead of comments


mapehe (approved):
Seems to fix the issue locally 👍

PR #948, opened 2021-06-01
EVAKA-HOTFIX use first name and last name on fee decision approver in…

massproduction:
…stead just name

#### Summary
Show full last name and only first first name for approver in fee decisions.


jaahon (commented):
You need to update frontend as well. Also, I think you should drop the name splitting in `decision.financeDecisionHandlerFirstName?.split(" ")?.get(0)`.

jaahon (approved):

PR #951, opened 2021-06-01
EVAKA-4190: Prevent sender-receiver conflicts

mapehe:
#### Summary

* Prevent sender-receiver conflicts in frontend: If sending message as `daycare X group Y`, disable selecting receivers from daycares other than `X` and groups other than `Y`.
* Use sidebar unit selection to filter available accounts upon opening the editor (also specify which daycare is selected in the title).
* Misc improvements:
  * Minor refactoring
  * Prevent choosing empty groups as the selected sender account (this is to rule out some edge cases this update would introduce, but could make sense otherwise too).
  * Hide receiver selection from the sidebar.

Suggest doing everything in the frontend, as there doesn't seem to be a need to involve the backend in any of this.

### DIFF
--- a/frontend/src/employee-frontend/components/messages/MessageContext.tsx
+++ b/frontend/src/employee-frontend/components/messages/MessageContext.tsx
@@ -36,7 +36,7 @@ const PAGE_SIZE = 20
 type RepliesByThread = Record<UUID, string>
 
 export interface MessagesState {
-  accountsRequest: Result<MessageAccount[]>
+  accounts: Result<MessageAccount[]>
   loadAccounts: () => void
   selectedDraft: DraftContent | undefined
   setSelectedDraft: (draft: DraftContent | undefined) => void
@@ -61,7 +61,7 @@ export interface MessagesState {
 }
 
 const defaultState: MessagesState = {
-  accountsRequest: Loading.of(),
+  accounts: Loading.of(),
   loadAccounts: () => undefined,
   selectedDraft: undefined,
   setSelectedDraft: () => undefined,
@@ -116,10 +116,10 @@ export const MessageContextProvider = React.memo(
       SelectOptionProps | undefined
     >()
 
-    const [accountsRequest, setAccountsRequest] = useState<
-      Result<MessageAccount[]>
-    >(Loading.of())
-    const getAccounts = useRestApi(getMessagingAccounts, setAccountsRequest)
+    const [accounts, setAccounts] = useState<Result<MessageAccount[]>>(
+      Loading.of()
+    )
+    const getAccounts = useRestApi(getMessagingAccounts, setAccounts)
     const loadAccounts = useDebouncedCallback(getAccounts, 100)
 
     useEffect(() => {
@@ -247,7 +247,7 @@ export const MessageContextProvider = React.memo(
           0
         )
         if (unreadCount > 0) {
-          setAccountsRequest((state) =>
+          setAccounts((state) =>
             state.map((accounts) =>
               accounts.map((acc) =>
                 acc.id === accountId
@@ -266,7 +266,7 @@ export const MessageContextProvider = React.memo(
 
     const value = useMemo(
       () => ({
-        accountsRequest,
+        accounts,
         loadAccounts,
         selectedDraft,
         setSelectedDraft,
@@ -290,7 +290,7 @@ export const MessageContextProvider = React.memo(
         refreshMessages
       }),
       [
-        accountsRequest,
+        accounts,
         loadAccounts,
         selectedDraft,
         selectedAccount,
--- a/frontend/src/employee-frontend/components/messages/MessageContext.tsx
+++ b/frontend/src/employee-frontend/components/messages/MessageContext.tsx
@@ -1,3 +1,4 @@
+import { SelectOptionProps } from 'employee-frontend/components/common/Select'
 import React, {
   createContext,
   useCallback,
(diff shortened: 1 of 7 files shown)


frontend/src/employee-frontend/components/messages/MessageEditor.tsx
```diff
@@ -169,12 +176,27 @@ export default React.memo(function MessageEditor({
...
+    function updateReceiversOnSenderChange() {
+      const account = accounts.find(
+        (account) => account.id === message.sender.value
+      )
+      if (account?.type === 'PERSONAL') {
+        setReceiverTree(availableReceivers)
+      } else if (account?.type === 'GROUP') {
+        const groupId = account.daycareGroup.id
+        const selection = getSubTree(availableReceivers, groupId)
+        selection && setReceiverTree(selection)
```
akheron, line 189: Just a style thing, but I prefer `if (selection) setReceiverTree(selection)` style when the code is about side effects and not about returning a value.

frontend/src/employee-frontend/components/messages/MessageEditor.tsx
```diff
@@ -222,7 +244,22 @@ export default React.memo(function MessageEditor({
...
+              if (account.type === 'PERSONAL') {
+                return true
+              } else if (account.type === 'GROUP') {
+                return !!getSelectorName(
+                  account.daycareGroup.id,
+                  availableReceivers
+                )
+              }
+              return false
+            })
```
akheron, line 258: And on the contrary, when returning a value like here, I would prefer an expression like 
```typescript
(account: MessageAccount) =>
  account.type === 'PERSONAL' ||
  (account.type === 'GROUP' && !!getSelectorName(account.daycareGroup.id, availableReceivers))
```

akheron (approved):
Only style issues, good to go without doing anything :)

frontend/src/employee-frontend/components/messages/SelectorNode.ts
```diff
@@ -162,3 +159,20 @@ export const getSelected = (selector: SelectorNode): ReactSelectOption[] => {
...
+): SelectorNode | undefined => {
+  if (selector.selectorId === selectorId) {
+    return selector
+  } else {
+    for (const child of selector.childNodes) {
+      const out = getSubTree(child, selectorId)
+      if (out) {
+        return out
+      }
+    }
```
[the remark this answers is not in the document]
mapehe: Yeah, that would be exactly what we need here 💘

frontend/src/employee-frontend/components/messages/MessageEditor.tsx
kimifoud: I would like to have this logic to be extracted outside of JSX.  `const senderOptions = accounts.filter(..)`

frontend/src/employee-frontend/components/messages/Sidebar.tsx
```diff
@@ -81,9 +81,16 @@ interface AccountsParams {
...
+  selectedUnit: SelectOptionProps | undefined
+  setSelectedUnit: (unit: SelectOptionProps) => void
 }
 
-function Accounts({ accounts, setSelectedReceivers }: AccountsParams) {
+function Accounts({
+  accounts,
+  setSelectedReceivers,
+  selectedUnit,
+  setSelectedUnit
```
kimifoud, line 92: These could be taken straight from the context instead of passing them through the whole component tree.

frontend/src/employee-frontend/components/messages/MessagesPageContext.tsx
```diff
@@ -103,10 +110,28 @@ export const MessagesPageContextProvider = React.memo(
...
-    const loadAccounts = useRestApi(getMessagingAccounts, setAccounts)
+    const [selectedUnit, setSelectedUnit] = useState<
+      SelectOptionProps | undefined
+    >()
+
+    const [accountsRequest, setAccountsRequest] = useState<
+      Result<MessageAccount[]>
+    >(Loading.of())
+    const loadAccounts = useRestApi(getMessagingAccounts, setAccountsRequest)
+    const [accounts, setAccounts] = useState<MessageAccount[]>()
```
kimifoud, line 121: I don't really like the naming here as it doesn't indicate that the result set is filtered.

kimifoud: Moreover, the whole filtering could be done in the component where it's needed. You could just `.map()` the original `accounts: Result<MessageAccount[]>` and filter from there.

frontend/src/employee-frontend/components/messages/MessageContext.tsx
```diff
@@ -28,16 +29,18 @@ import {
...
 
 const PAGE_SIZE = 20
-
 type RepliesByThread = Record<UUID, string>
 
 export interface MessagesState {
-  accounts: Result<MessageAccount[]>
+  accountsRequest: Result<MessageAccount[]>
   loadAccounts: () => void
+  accounts: MessageAccount[] | undefined
```
kimifoud, line 37: I don't think this is needed

kimifoud (changes requested):
Request changes to prevent auto-merge while discussing.

mapehe:
Thanks for the feedback @kimifoud and @akheron. Refactored 90% accordingly, except for the small suggestion about local types.

mapehe:
I won't be around for the rest of the week so I'll enable auto merge.

kimifoud (approved):

PR #952, opened 2021-06-02
EVAKA-HOTFIX do not fetch units on every edit of application

massproduction:
#### Summary
Do not fetch units on every edit of application.


jaahon (commented):
Instead of adding `// eslint-disable-next-line react-hooks/exhaustive-deps` you should just drop `editedApplication` from the if-clause and use `editedApplication?.type` instead.

jaahon (approved):

PR #955, opened 2021-06-02
More Combobox

Gekkio:
#### Summary

:boxing_glove:

### DIFF
--- a/frontend/src/e2e-playwright/pages/employee/reports.ts
+++ b/frontend/src/e2e-playwright/pages/employee/reports.ts
@@ -3,16 +3,16 @@
 // SPDX-License-Identifier: LGPL-2.1-or-later
 
 import { Page } from 'playwright'
-import { RawElement } from 'e2e-playwright/utils/element'
+import { Combobox, RawElement } from 'e2e-playwright/utils/element'
 import { waitUntilEqual } from 'e2e-playwright/utils'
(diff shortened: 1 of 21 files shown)


jaahon:
I have not looked through the PR yet, but why would we ever want to use an "uncontrolled" version of Combobox? I'm referring to this commit `Add uncontrolled support for Combobox`.

jaahon:
> I have not looked through the PR yet, but why would we ever want to use an "uncontrolled" version of Combobox? I'm referring to this commit `Add uncontrolled support for Combobox`.

It doesn't seem like the uncontrolled version is used anywhere anyway. Feels like just another way to make a mistake.

Gekkio:
> > I have not looked through the PR yet, but why would we ever want to use an "uncontrolled" version of Combobox? I'm referring to this commit `Add uncontrolled support for Combobox`.
> 
> It doesn't seem like the uncontrolled version is used anywhere anyway. Feels like just another way to make a mistake.

True, I added it only because the unit editor originally used uncontrolled inputs. But this turned out to be a real bug, so they were changed to controlled. We also have more uncontrolled ReactSelect/Select components in some places that were not updated in this pull request, but it's probably better to make them controlled inputs at the same time when they are switched to use Combobox.

jaahon (approved):

PR #956, opened 2021-06-02
Add placement type customization

patari:
#### Summary

1. treVaka doesn't use all placement types -> add support to hide options 0b6e8955c385a8c594aa2e1f5bfff3d68e5fcc8d
2. treVaka requires one additional placement type 4f25ed7fcc07df6b1dad99cb3ced446b9cbdb99e


Joosakur:
Based on a quick look this looks good. 👍 I think we should maybe postpone the actual review a few days until the service needs v2 feature is finished and the old code has been cleaned away and renamings done. That will decrease the amount of places where the new placement type may affect. It will anyway need some effort to make sure it has been handled in all the places since I don't think all of them will cause compilation errors (e.g. sql code, some of i18n etc). Luckily though most of the possible bugs won't cause regression for us as long as we don't use this placement type.

Joosakur:
ready for rebase!

service/src/main/kotlin/fi/espoo/evaka/decision/DecisionDraftService.kt
```diff
@@ -28,6 +28,7 @@ class DecisionDraftService {
             PlacementType.DAYCARE, PlacementType.DAYCARE_PART_TIME -> planDaycareDecisionDrafts(placementPlan)
             PlacementType.PRESCHOOL, PlacementType.PRESCHOOL_DAYCARE, PlacementType.PREPARATORY, PlacementType.PREPARATORY_DAYCARE -> planPreschoolDecisionDrafts(placementPlan, application)
             PlacementType.TEMPORARY_DAYCARE, PlacementType.TEMPORARY_DAYCARE_PART_DAY,
+            PlacementType.SCHOOL_SHIFT_CARE -> listOf()
```
Gekkio, line 31: This probably doesn't matter, but doesn't this change the semantics of temporary daycares?

Previously `TEMPORARY_DAYCARE`, `TEMPORARY_DAYCARE_PART_DAY`, `DAYCARE_FIVE_YEAR_OLDS`, `DAYCARE_PART_TIME_FIVE_YEAR_OLDS` all ended up throwing errors.

Now `TEMPORARY_DAYCARE`, `TEMPORARY_DAYCARE_PART_DAY` return an empty list

patari: Fixed

Joosakur (approved):

PR #958, opened 2021-06-02
EVAKA-TECH: faster missing group placements?

Joosakur:
#### Summary
Test how calculating missing group placements in kotlin affects performance

### DIFF
--- a/service/src/main/kotlin/fi/espoo/evaka/placement/PlacementService.kt
+++ b/service/src/main/kotlin/fi/espoo/evaka/placement/PlacementService.kt
@@ -366,6 +366,8 @@ fun getMissingGroupPlacements(
         val gapRange: FiniteDateRange
     )
 
+    val evakaLaunch = LocalDate.of(2020, 3, 1)
+
     data class PlacementResult(
         val id: UUID,
         val type: PlacementType,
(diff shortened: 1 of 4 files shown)


service/src/main/kotlin/fi/espoo/evaka/placement/PlacementService.kt
```diff
@@ -352,6 +353,101 @@ private fun handleFiveYearOldDaycare(tx: Database.Transaction, childId: UUID, ty
...
+        """
+        SELECT 
+            FALSE AS backup,
+            pl.id, 
+            pl.type,
+            daterange(pl.start_date, pl.end_date, '[]') AS range,
+            pl.child_id,
+            array_remove(array_agg(
+                case when dgp.id is null then null else daterange(dgp.start_date, dgp.end_date, '[]') end
+            ), null) AS group_placement_ranges
```
Joosakur, line 386: is there a better way to avoid zero group placements resulting in an array which contains daterange(null, null)?

Gekkio: One option is to use aggregate function filering (`FILTER (WHERE)`):

```
array_agg(...) FILTER (WHERE dgp is NOT NULL)
```

Joosakur: excellent!

Joosakur:
Added optimization to fetch about 90% less rows from db.

reynders (approved):

PR #968, opened 2021-06-04
EVAKA-4190 Implement toggling of reply recipients

kimifoud:
#### Summary

- Possible recipients are those who have participated in the thread (except self)
- Default recipients are those who participated in the last message
- The original thread author cannot be removed from recipients
- Disable reply button when the user has not entered any text and when the state is loading


akheron (approved):
LGTM

PR #970, opened 2021-06-04
EVAKA-4184 sibling discounts in fee thresholds editor

jaahon:
#### Summary

* Add sibling discounts to fee thresholds editor.
* Insert dev data into `fee_thresholds` table instead of `pricing`.
* Add overlap constraint and `created` & `updated` columns to `fee_thresholds`


Joosakur (approved):
looks good to me!

PR #971, opened 2021-06-04
Capture Playwright page errors and logs

Gekkio:

### DIFF
--- a/frontend/src/e2e-playwright/browser.ts
+++ b/frontend/src/e2e-playwright/browser.ts
@@ -8,7 +8,8 @@ import playwright, {
   Browser,
   BrowserContext,
   BrowserContextOptions,
-  Download
+  Download,
+  Page
 } from 'playwright'
 import config from 'e2e-test-common/config'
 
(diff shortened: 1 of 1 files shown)


frontend/src/e2e-playwright/browser.ts
```diff
@@ -77,10 +78,33 @@ async function takeScreenshots(namePrefix: string): Promise<void> {
...
+function configurePage(page: Page) {
+  page.on(
+    'console',
+    (msg) =>
+      void (async () => {
+        const args: unknown[] = []
+        for (const arg of msg.args()) {
+          const value = (await arg.jsonValue()) as unknown
+          args.push(value)
+          await arg.dispose()
```
akheron, line 90: Is the dispose call needed? It's not used in the example: https://playwright.dev/docs/api/class-page#pageonconsole

akheron (approved):
Seems to work fine 👍

frontend/src/e2e-playwright/browser.ts
Gekkio: Hmm, maybe not...I added it because I remembered that JSHandle values retain memory in the browser unless disposed manually.
But it's probably not a big problem since all memory is reclaimed once the page/context/browser is closed

PR #973, opened 2021-06-04
EVAKA-4190 Enable messaging in prod for pilot units

kimifoud:
#### Summary

Prerequisites:
- [x] Suomi.fi has confirmed that keycloak certificate request is approved
- [x] Suomi.fi authentication works
- [x] VPN restriction is lifted https://github.com/espoon-voltti/evaka-infra/pull/428

### DIFF
--- a/frontend/src/citizen-frontend/config.ts
+++ b/frontend/src/citizen-frontend/config.ts
@@ -34,7 +34,7 @@ const features: Features = {
     messaging: true
   },
   prod: {
-    messaging: false
+    messaging: true
   }
 }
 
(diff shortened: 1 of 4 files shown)


Joosakur:
Should we filter account list to only pilot unit group accounts?

Joosakur:
Citizen side comes in separate PR? Maybe both should be enabled at the same time or else some citizen will end up getting emails about unread messages without actually being able to read them.

kimifoud:
> Citizen side comes in separate PR? Maybe both should be enabled at the same time or else some citizen will end up getting emails about unread messages without actually being able to read them.

Ah, good catch. I didn't realize there was a separate feature flag config for citizen.

kimifoud:
> Should we filter account list to only pilot unit group accounts?

Hmmm, maybe..? Should be pretty easy anyway.

reynders (approved):

mikkopiu:
Also: don't merge this before the Keycloak realm's whitelisting is removed from the deployment (otherwise users will end up on a 403 nginx page).

PR #977, opened 2021-06-07
Update proxy OS packages on build

ilja-git:
#### Summary

We want to upgrade operating system packages on build so we will fix possible issues before upstream container.

Includes the fixed versions of packages with the following vulnerabilities:

- CVE-2021-3517: HIGH
- CVE-2021-3518: MEDIUM
- CVE-2021-3537: MEDIUM
- CVE-2021-22901: UNDEFINED
- CVE-2021-22898: UNDEFINED

### DIFF
--- a/proxy/Dockerfile
+++ b/proxy/Dockerfile
@@ -9,7 +9,10 @@ LABEL maintainer="https://github.com/espoon-voltti/evaka"
 ENV NGINX_ENV=local \
     TZ=UTC
 
+ARG CACHE_BUST=2021-06-07
+
 RUN apk add --no-cache 'ruby=~2.7' \
+ && apk --no-cache upgrade \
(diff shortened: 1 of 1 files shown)


proxy/Dockerfile
```diff
@@ -10,6 +10,7 @@ ENV NGINX_ENV=local \
     TZ=UTC
 
 RUN apk add --no-cache 'ruby=~2.7' \
+ && apk --no-cache upgrade \
```
mikkopiu, line 13: Improvement suggestion: As we use a Docker layer cache in CI, we could preemptively create a cache busting layer like: `ARG CACHE_BUST=2021-06-07` that can then just be updated if we need to explicitly do a similar forced update.

mikkopiu (approved):
LGTM but see suggestion (can ignore or implement, either is fine for this occasion by me)

mikkopiu (approved):

PR #985, opened 2021-06-08
Add template provider for customization

patari:
#### Summary

Refactor PDF templates to support customization.

Also moves hard-coded default address used for decisions into message provider.


Joosakur:
Good stuff! I'll test it quickly a bit later and approve then.

Joosakur (approved):
Haven't tested changing templates but at least I can't find anything broken on our version

PR #995, opened 2021-06-09
EVAKA added welcome page for users without roles

massproduction:
#### Summary
Added welcome page for users without roles.


frontend/src/employee-frontend/App.tsx
```diff
@@ -416,7 +423,7 @@ function RedirectToMainPage() {
   } else if (hasRole(roles, 'DIRECTOR')) {
     return <Redirect to={'/reports'} />
   } else {
-    return <Redirect to={'/search'} />
+    return <Redirect to={'/welcome'} />
```
Joosakur, line 426: Now it forwards some of the roles to welcome page because this if-else chain is not exhaustive. 

You need to add something like
`if(roles.length === 0) return <Redirect to={'/welcome'} />`
on line 415 instead

Joosakur (changes requested):

Joosakur (approved):

PR #1000, opened 2021-06-10
Speed up freetext searches

akheron:
#### Summary

Precompute the tsvector used for person free text search search using a generated column `freetext_vec`, and index that column. Use the new column for person, application, invoice, fee decision and voucher value decision searches.

With a set of 100,000 generated persons, this change speeds up the person search request from 300-400 ms to about 50-80 ms on my machine (~75-85 % faster). I didn't test the the performance effect for other search endpoints, because they would have needed much more complicated test data.

This also fixes the value decision free text search that was broken. Add some simple regression tests for it, too.

### DIFF
--- a/service/src/main/kotlin/fi/espoo/evaka/invoicing/data/VoucherValueDecisionQueries.kt
+++ b/service/src/main/kotlin/fi/espoo/evaka/invoicing/data/VoucherValueDecisionQueries.kt
@@ -209,7 +209,7 @@ fun Database.Read.searchValueDecisions(
         "financeDecisionHandlerId" to financeDecisionHandlerId
     )
 
-    val (freeTextQuery, freeTextParams) = freeTextSearchQuery(listOf("head", "partner", "child"), searchTerms)
+    val (freeTextQuery, freeTextParams) = freeTextSearchQuery(listOf("head", "child"), searchTerms)
 
     val conditions = listOfNotNull(
         if (areas.isNotEmpty()) "area.short_name = ANY(:areas)" else null,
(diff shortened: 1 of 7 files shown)


service/src/main/kotlin/fi/espoo/evaka/invoicing/data/VoucherValueDecisionQueries.kt
```diff
@@ -209,7 +209,7 @@ fun Database.Read.searchValueDecisions(
         "financeDecisionHandlerId" to financeDecisionHandlerId
     )
 
-    val (freeTextQuery, freeTextParams) = freeTextSearchQuery(listOf("head", "partner", "child"), searchTerms)
+    val (freeTextQuery, freeTextParams) = freeTextSearchQuery(listOf("head", "child"), searchTerms)
```
akheron, line 212: There was no `partner` table in the SQL query anymore, that's why the search was broken.

jaahon: Why not include partner in the query instead of taking it off the free text search altogether?

akheron: Not sure what's the correct thing to do 🤔 Because the search was broken, I assume that `partner` was there but has been removed since.

jaahon: I remember wondering why `partner` wasn't in that list and adding it at some point, but I guess I didn't realize that it wasn't in the actual query. I think `partner` should be there.

akheron: Added partner and a test that checks that partner's details can be used for searching.

jaahon (approved):
Works 😎

PR #1005, opened 2021-06-10
Capture videos from e2e test failures in CI

akheron:
#### Summary

Capture videos from failed e2e tests in CI. Individual e2e tests run a bit slower, but the total build time seems not to be affected noticeably.

Other stuff:
- Attempt to fix a few flaky tests
- Avoid extra navigation in `employeeLogin`
- Run `VACUUM` after database reset in an attempt to avoid automatic postgres cleanups taking place in the middle of the tests
- Remove screenshots and videos in `yarn clean`


frontend/src/e2e-playwright/jest-environment.js
```diff
@@ -20,7 +20,9 @@ class PlaywrightEnvironment extends NodeEnvironment {
       const specName = event.test.name.replace(/\W/g, '-')
 
       const namePrefix = `${parentName}_${specName}`
-      await this.global.evaka?.takeScreenshots(namePrefix)
+      await this.global.evaka?.saveScreenshotsAndVideos(namePrefix)
+    } else if (event.name === 'test_fn_success') {
```
mikkopiu, line 24: Where this `test_fn_success` event name from?

akheron: From `jest-circus` test runner, like the `test_fn_failure` above. The library's documentation is pretty sparse, but seems to work 🤷

mikkopiu (changes requested):
Would you mind adding an `rm -rf videos` to the npm `clean` script? I at least run E2E tests a lot locally with the CI setup and (if I'm not mistaken) now the Testcafe videos would never be deleted.

compose/db/init_evaka_local.sh
```diff
@@ -1,30 +0,0 @@
...
-    -- App login user
-    CREATE ROLE "evaka_application_local" WITH LOGIN PASSWORD 'app'
-      IN ROLE "evaka_application_role_local";
-EOSQL
-
-PGPASSWORD=flyway psql -v ON_ERROR_STOP=1 --username evaka_migration_local --dbname evaka_local <<EOSQL
-    -- The reset_database function, used in e2e tests, truncates tables and resets sequences
-    ALTER DEFAULT PRIVILEGES IN SCHEMA public GRANT TRUNCATE ON TABLES TO "evaka_application_local";
-    ALTER DEFAULT PRIVILEGES IN SCHEMA public GRANT UPDATE ON SEQUENCES TO "evaka_application_local";
-EOSQL
```
mikkopiu, line 30: Are these no longer unnecessary?

akheron: The `reset_database` function is now run by the `evaka_migration_local` (the flyway user), which is the owner of all tables and sequences, so the `evaka_application_local` user no longer needs these permissions.

The change was made because `VACUUM` can only be run by the owner of the tables. There's no permission that could be given to enable vacuuming by other users.

akheron:
> Would you mind adding an `rm -rf videos` to the npm `clean` script? I at least run E2E tests a lot locally with the CI setup and (if I'm not mistaken) now the Testcafe videos would never be deleted.

Done! The `clean` script now also removes screenshots.

PR #1007, opened 2021-06-10
EVAKA-4203: first draft of vasu backend

Joosakur:

### DIFF
--- a/service/src/main/kotlin/fi/espoo/evaka/vasu/VasuController.kt
+++ b/service/src/main/kotlin/fi/espoo/evaka/vasu/VasuController.kt
@@ -4,6 +4,7 @@ import fi.espoo.evaka.shared.auth.AuthenticatedUser
 import fi.espoo.evaka.shared.auth.UserRole
 import fi.espoo.evaka.shared.db.Database
 import fi.espoo.evaka.shared.domain.BadRequest
+import fi.espoo.evaka.shared.domain.NotFound
 import org.springframework.http.ResponseEntity
 import org.springframework.web.bind.annotation.GetMapping
 import org.springframework.web.bind.annotation.PathVariable
@@ -33,13 +34,14 @@ class VasuController {
         user.requireOneOfRoles(UserRole.ADMIN)
 
         val document = db.transaction { tx ->
-            tx.getVasuTemplate(body.templateId).let { template ->
+            tx.getVasuTemplate(body.templateId)?.let { template ->
                 if (!template.valid.includes(LocalDate.now())) {
                     throw BadRequest("Template is not currently valid")
                 }
-            }
+            } ?: throw NotFound("template ${body.templateId} not found")
+
             val id = tx.insertVasuDocument(body.childId, body.templateId)
-            tx.getVasuDocumentResponse(id)
+            tx.getVasuDocumentResponse(id)!!
         }
 
         return ResponseEntity.created(URI.create("/vasu/${document.id}")).body(document)
@@ -50,12 +52,12 @@ class VasuController {
         db: Database.Connection,
         user: AuthenticatedUser,
         @PathVariable id: UUID
-    ): ResponseEntity<VasuDocumentResponse> {
+    ): VasuDocumentResponse {
         user.requireOneOfRoles(UserRole.ADMIN)
 
         return db.read { tx ->
-            tx.getVasuDocumentResponse(id)
-        }.let { ResponseEntity.ok(it) }
+            tx.getVasuDocumentResponse(id) ?: throw NotFound("template $id not found")
+        }
     }
 
     data class UpdateDocumentRequest(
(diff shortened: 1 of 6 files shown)


service/src/main/kotlin/fi/espoo/evaka/vasu/VasuController.kt
```diff
@@ -0,0 +1,91 @@
...
+    fun getDocument(
+        db: Database.Connection,
+        user: AuthenticatedUser,
+        @PathVariable id: UUID
+    ): ResponseEntity<VasuDocumentResponse> {
+        user.requireOneOfRoles(UserRole.ADMIN)
+
+        return db.read { tx ->
+            tx.getVasuDocumentResponse(id)
+        }.let { ResponseEntity.ok(it) }
```
kimifoud, line 58: You can just use `VasuDocumentResponse` as return type and return the transaction result `return db.read { it -> getVasuDocumentResponse(id) }`. Same goes for other `ResponseEntity` usages.

service/src/main/kotlin/fi/espoo/evaka/vasu/VasuController.kt
```diff
@@ -0,0 +1,91 @@
...
+    data class UpdateDocumentRequest(
+        val content: VasuContent
+    )
+    @PutMapping("/{id}")
+    fun putDocument(
+        db: Database.Connection,
+        user: AuthenticatedUser,
+        @PathVariable id: UUID,
+        @RequestBody body: UpdateDocumentRequest
+    ): ResponseEntity<Unit> {
```
kimifoud, line 70: `: ResponseEntity<Unit>` can be omitted if you don't return anything from the controller.

service/src/main/kotlin/fi/espoo/evaka/vasu/VasuQueries.kt
```diff
@@ -0,0 +1,105 @@
...
+    val sql = """
+        SELECT *
+        FROM vasu_template
+        WHERE id =:id
+    """.trimIndent()
+
+    return createQuery(sql)
+        .bind("id", id)
+        .mapTo<VasuTemplate>()
+        .firstOrNull() ?: throw NotFound("template $id not found")
```
kimifoud, line 40:
```suggestion
        .one()
```

service/src/main/kotlin/fi/espoo/evaka/vasu/VasuQueries.kt
```diff
@@ -0,0 +1,105 @@
...
+        JOIN vasu_content vc on vd.content_id = vc.id
+        JOIN vasu_template vt on vd.template_id = vt.id
+        JOIN person p on p.id = vd.child_id
+        WHERE vd.id =:id
+    """.trimIndent()
+
+    return createQuery(sql)
+        .bind("id", id)
+        .mapTo<VasuDocumentResponse>()
+        .firstOrNull() ?: throw NotFound("template $id not found")
```
kimifoud, line 89:
```suggestion
        .one()
```

service/src/main/kotlin/fi/espoo/evaka/vasu/VasuQueries.kt
```diff
@@ -0,0 +1,105 @@
...
+        UPDATE vasu_content
+        SET content = :content
+        WHERE id IN (SELECT vd.content_id FROM vasu_document vd WHERE vd.id = :id)
+    """.trimIndent()
+
+    createUpdate(sql)
+        .bind("id", id)
+        .bind("content", content)
+        .execute()
+        .also { if (it < 1) throw NotFound("document content to update not found") }
```
kimifoud, line 104: 👍 This made me realize that there must be a couple of places in messaging code where `createUpdate` no-op has not been handled.

Is it idiomatic in this project to `throw NotFound` from the query instead of returning the count and reacting to it on the caller side?

service/src/main/resources/db/migration/R__dev_temp.sql
```diff
@@ -0,0 +1,38 @@
...
+    content jsonb NOT NULL
+);
+
+CREATE TRIGGER set_timestamp BEFORE UPDATE ON vasu_content FOR EACH ROW EXECUTE PROCEDURE trigger_refresh_updated();
+
+CREATE TABLE vasu_document(
+    id uuid PRIMARY KEY DEFAULT ext.uuid_generate_v1mc(),
+    created timestamp with time zone NOT NULL DEFAULT now() NOT NULL,
+    updated timestamp with time zone DEFAULT now() NOT NULL,
+    child_id uuid REFERENCES person(id),
```
kimifoud, line 29: Should we reference `child` instead?

service/src/main/resources/db/migration/R__dev_temp.sql
```diff
@@ -0,0 +1,38 @@
...
+    template_id uuid REFERENCES vasu_template(id),
+    content_id uuid REFERENCES vasu_content(id)
```
kimifoud, line 31: Should all three foreign keys be `NOT NULL`?

service/src/main/kotlin/fi/espoo/evaka/vasu/VasuQueries.kt
Joosakur: Found some extension function we have and used that. I think 0 updates is basically always an error so no need to handle it on caller side.

kimifoud (approved):

mikkopiu (commented):
Should these endpoints have audit events logged? I'm not familiar with the context but there seem to be child IDs going around so that would indicate yes.

Joosakur:
> Should these endpoints have audit events logged? I'm not familiar with the context but there seem to be child IDs going around so that would indicate yes.

at some point yes, for now this won't be used in prod, but I'll add them already in next pr so they wont be forgotten 👍

PR #1010, opened 2021-06-10
Capture top-level clicks in Combobox

Gekkio:
#### Summary

- E2E tests always click the Root element which had no click listener, but the clicks *usually* hit the input wrapper
- let clicks propagate from children to Root instead
- clicks from the menu are not propagated, so menu clicks don't trigger the Root listener


mikkopiu (approved):
LGTM, though stories for this component would be nice. Are we still doing those?

akheron (approved):

PR #1011, opened 2021-06-11
EVAKA-HOTFIX application received email texts

hruoho:
#### Summary

Update bad urls & phone numbers


mikkopiu (changes requested):
Looks like we have the same issue in:
- `service/src/main/kotlin/fi/espoo/evaka/application/PendingDecisionEmailService.kt`
- `service/src/main/kotlin/fi/espoo/evaka/shared/message/EvakaMessageProvider.kt`

(look for 'på adressen')

PR #1014, opened 2021-06-11
EVAKA-4176: remove 'new' prefixes from service needs and fee decisions

Joosakur:
#### Summary
... and remove some dead code


service/src/main/resources/db/migration/V109__new_service_needs_rename.sql
```diff
@@ -0,0 +1,8 @@
+-- drop these backups later if not needed
+ALTER TABLE service_need RENAME TO old_service_need;
+ALTER TABLE fee_decision RENAME TO old_fee_decision;
+ALTER TABLE fee_decision_part RENAME TO old_fee_decision_part;
+
+ALTER TABLE new_service_need RENAME TO service_need;
+ALTER TABLE new_fee_decision RENAME TO fee_decision;
+ALTER TABLE new_fee_decision_child RENAME TO fee_decision_child;
```
mikkopiu, line 8: Should we also have a temporary view that points e.g. `new_service_need -> service_need` so that during the deployment, other task instances won't fail their queries? The view could then be deleted once this version has been deployed to all environments (incl. Tampere).

mikkopiu (changes requested):

service/src/main/resources/db/migration/V109__new_service_needs_rename.sql
Joosakur: Hmm good idea!

Yeah `old_**` tables should be fully unused, although I'm a bit worried that I missed some place. Those tables can be dropped later but I thought I'd temporarily back up the data in case we still find out that the migration was somehow incomplete.

mikkopiu (approved):
LGTM but I'm not an expert on the business context here.

mikkopiu (approved):
Re-affirming what was discussed: the worst outcome of not creating a temporary view is that users get a few errors while they have the old frontend (with old paths) cached/loaded and the frontend should anyway be deployed at the same time as the service -> good to go without the view.

PR #1015, opened 2021-06-11
CI: Add shell script linting

mikkopiu:
#### Summary
- CI: add shell script linting with shellcheck
   - Lint all shell scripts with the excellent shellcheck tool to avoid the most common mistakes (variable expansion, incorrect whitespace, unused variables etc)
   - Fix minor issues in existing scripts found by the new script
   - Ran in parallel to multiple slower jobs -> no increase in total workflow time
- Improve AWS metadata error logging in all image entrypoints
   - Don't suppress error messages from `curl` when fetching host IP from AWS metadata during image startups, while still falling back to a default value
   - Also ignore SC2155: we already handle the error case with the fallback value "UNAVAILABLE" and output error messages from `curl` to stderr (without crashing as that's not desirable here)
- Fix minor existing issues with all the shell scripts

### DIFF
--- a/service/entrypoint.sh
+++ b/service/entrypoint.sh
@@ -6,9 +6,9 @@
 
 set -euo pipefail
 
-# For logback tagging (allow for local fallback)
-HOST_IP=$(curl --fail -s http://169.254.169.254/latest/meta-data/local-ipv4 || printf 'UNAVAILABLE')
-export HOST_IP
+# For log tagging (with a default value and error logging without crashing)
+# shellcheck disable=SC2155
+export HOST_IP=$(curl --silent --fail --show-error http://169.254.169.254/latest/meta-data/local-ipv4 || printf 'UNAVAILABLE')
 
 # Download deployment specific files from S3 if in a non-local environment
 if [ "${VOLTTI_ENV:-X}" != "local" ]; then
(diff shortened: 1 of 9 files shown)


service/entrypoint.sh
```diff
@@ -7,7 +7,8 @@
 set -euo pipefail
 
 # For logback tagging (allow for local fallback)
-export HOST_IP=$(wget -qO- http://169.254.169.254/latest/meta-data/local-ipv4 || printf 'UNAVAILABLE')
+HOST_IP=$(curl --fail -s http://169.254.169.254/latest/meta-data/local-ipv4 || printf 'UNAVAILABLE')
+export HOST_IP
```
ilja-git, line 11: Is separate export required as we use "||"? We might want to use -sSf also o get the error message to stderr for debugging purposes.

ilja-git: If the separate export is because linting then ignore me.

ilja-git (approved):

service/entrypoint.sh
mikkopiu: It was indeed because of the linter but you make a good point, I'll fix these.

ilja-git (approved):
LGTM

PR #1016, opened 2021-06-13
EVAKA-4203: vasu template editor v1

Joosakur:


frontend/src/employee-frontend/components/Header.tsx
```diff
@@ -263,6 +263,17 @@ const Header = React.memo(function Header({ location }: RouteComponentProps) {
                     </Link>
                   </RequireRole>
                 ) : null}
+                {featureFlags.vasu ? (
```
kimifoud, line 266: `featureFlags.vasu && <..` would be more idiomatic, but I see that ternary has been used in this file already.

frontend/src/employee-frontend/components/vasu/templates/CreateQuestionModal.tsx
```diff
@@ -0,0 +1,141 @@
...
+          type: 'MULTISELECT',
+          name: name,
+          optionNames: options,
+          minSelections: 0,
+          maxSelections: null,
+          value: []
+        }
+      default:
+        throw Error('Unexpected type')
+    }
```
kimifoud, line 72: I don't know how our typescript compiler / linter is configured, but I prefer switch without default case and handling all cases explicitly and throwing after the switch. This way the compiler might warn about an unhandled case when a new type is added.

kimifoud (approved):
Magnificent work! IMO it's good to merge now to avoid further conflicts and smooth the rough edges afterwards.

frontend/src/employee-frontend/components/vasu/templates/CreateQuestionModal.tsx
Joosakur: ah, didn't work without default before but adding `as const` to the enum helped

PR #1017, opened 2021-06-14
Add service need name to application lists

patari:
#### Summary

Service need should be in application form data only if frontend feature flag `serviceNeedOptionsEnabled` is true -> this change should not affect Espoo.


mikkopiu (approved):
LGTM, pushed for CI

PR #1018, opened 2021-06-14
Await playwright video promises properly

akheron:

### DIFF
--- a/frontend/src/e2e-playwright/jest-environment.js
+++ b/frontend/src/e2e-playwright/jest-environment.js
@@ -13,6 +13,14 @@ const NodeEnvironment = require('jest-environment-node')
 
 class PlaywrightEnvironment extends NodeEnvironment {
   async handleTestEvent(event, _state) {
+    if (!this.global.evaka) return
+    const {
+      captureScreenshots,
+      saveVideos,
+      deleteTemporaryVideos,
+      promises
+    } = this.global.evaka
+
     if (event.name === 'test_fn_failure') {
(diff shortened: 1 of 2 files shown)


frontend/src/e2e-playwright/jest-environment.js
```diff
@@ -13,16 +13,22 @@ const NodeEnvironment = require('jest-environment-node')
...
+    let promise = undefined
     if (event.name === 'test_fn_failure') {
       /** @type {string} **/
       const parentName = event.test.parent.name.replace(/\W/g, '-')
       /** @type {string} **/
       const specName = event.test.name.replace(/\W/g, '-')
 
       const namePrefix = `${parentName}_${specName}`
-      await this.global.evaka?.saveScreenshotsAndVideos(namePrefix)
+      promise = this.global.evaka?.saveScreenshotsAndVideos(namePrefix)
```
Gekkio, line 24: Doesn't this cause a possible race condition? Now that we don't await anything, there's no guarantee the screenshots have been saved before the browser contexts are closed in `beforeEach`. The main problem seems to be that `saveScreenshotsAndVideos` involves multiple steps and we want to defer only step to be awaited later.

akheron: Oh, I didn't realize that screenshots have to be catured _before_ closing the context, and videos are ready only _after_ closing the context 🤦

Gekkio (approved):

PR #1021, opened 2021-06-15
EVAKA-4184 decision generation on fee thresholds change

jaahon:
#### Summary

Generate new finance decisions on fee thresholds change. Since I would have had to add yet another async job type for this I decided to remove the existing async job types related to finance decision generation and create a new one that serves all of the cases. Async job runner can still run jobs of the old types so that we don't have to convert any of the existing jobs. We can remove the code in a couple of months when the old jobs have all been removed from the database.

Add some e2e tests for the new fee thresholds editor.

Drop the unused `pricing` table and remove the term "pricing" from code.


service/src/main/kotlin/fi/espoo/evaka/invoicing/domain/FeeDecisions.kt
```diff
@@ -183,11 +183,13 @@ data class FeeDecisionDetailed(
         return isRetroactive(this.validDuring.start, LocalDate.from(sentAtLocalDate ?: LocalDate.now()))
     }
 
+    // TODO: remove
```
reynders, line 186: Either add more descriptive comment when to remove this or remote this and the next todo below

service/src/main/kotlin/fi/espoo/evaka/pis/controllers/PartnershipsController.kt
```diff
@@ -98,7 +102,12 @@ class PartnershipsController(private val asyncJobRunner: AsyncJobRunner, private
                 val partnership = partnershipService.updatePartnershipDuration(tx, partnershipId, body.startDate, body.endDate)
                 asyncJobRunner.plan(
                     tx,
-                    listOf(NotifyFamilyUpdated(partnership.partners.last().id, partnership.startDate, partnership.endDate))
+                    listOf(
+                        GenerateFinanceDecisions.forAdult(
+                            partnership.partners.last().id,
```
reynders, line 107: Could there be another way to ensure we generate the decision for the correct partner than the partners.last()? I know it is obvious which partner it is when looking at the implementation of partnershipService.createPartnership, but there is no guarantee in the API that the order is maintained when creating the parnership. Anyway, left for author's consideration whether this can be improved or not.

reynders (approved):

service/src/main/kotlin/fi/espoo/evaka/invoicing/domain/FeeDecisions.kt
jaahon: My idea was to remove the fields immediately after this PR is in prod. I can make a draft PR to make sure that I won't forget to do it.

service/src/main/kotlin/fi/espoo/evaka/pis/controllers/PartnershipsController.kt
jaahon: Well to be honest it doesn't matter which adult is picked, the end result should be the same. I'm not sure why it was `last()` originally. I'll look into this.

service/src/main/kotlin/fi/espoo/evaka/invoicing/domain/FeeDecisions.kt
jaahon: https://github.com/espoon-voltti/evaka/pull/1038

PR #1022, opened 2021-06-15
EVAKA-4243 Updated access

massproduction:
#### Summary
Updated following accesses:

Child additional info should be editable by `STAFF` and `SPECIAL_EDUCATION_TEACHER`
Child daily notes should be editable by `SPECIAL_EDUCATION_TEACHER`
Absences should be editable by `SPECIAL_EDUCATION_TEACHER`
Application notes should be hidden from `UNIT_SUPERVISOR`
Hide partners and children in different addresses reports from `UNIT_SUPERVISORS`
Person contact info should be editable by `STAFF`

### DIFF
--- a/frontend/src/employee-frontend/api/person.ts
+++ b/frontend/src/employee-frontend/api/person.ts
@@ -220,6 +220,7 @@ export interface PatchPersonRequest {
   streetAddress: string | undefined
   postalCode: string | undefined
   postOffice: string | undefined
+  backupPhone: string | undefined
 }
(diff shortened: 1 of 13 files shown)


Joosakur:
When accessing unit page group tabs as SERVICE_WORKER, the daily notes renders an error segment. I assume it responds 403. Probably it shouldn't even try to load it and just display disabled icon or whatever.

reynders (approved):

PR #1023, opened 2021-06-15
EVAKA-4203 Glue current WIP vasu implementations together

kimifoud:
#### Summary

- Return child's actual vasu documents instead of mock data
- Navigate to vasu edit view from vasu list
- Initialize new vasu document and navigate to edit view if successful

Initialization has to be reworked to select the currently valid template and restrict to only one document per template.

### DIFF
--- a/service/src/main/kotlin/fi/espoo/evaka/vasu/VasuQueries.kt
+++ b/service/src/main/kotlin/fi/espoo/evaka/vasu/VasuQueries.kt
@@ -61,3 +61,25 @@ fun Database.Transaction.updateVasuDocument(id: UUID, content: VasuContent) {
         .bind("content", content)
         .updateExactlyOne()
 }
+
+fun Database.Read.getVasuDocumentSummaries(childId: UUID): List<VasuDocumentSummary> {
+    // language=sql
+    val sql = """
+        SELECT 
+            vd.id,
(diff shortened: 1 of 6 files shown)


service/src/main/kotlin/fi/espoo/evaka/vasu/VasuQueries.kt
```diff
@@ -61,3 +61,25 @@ fun Database.Transaction.updateVasuDocument(id: UUID, content: VasuContent) {
...
         .updateExactlyOne()
 }
+
+fun Database.Read.getVasuDocumentSummaries(childId: UUID): List<VasuDocumentSummary> {
+    // language=sql
+    val sql = """
+        SELECT 
+            vd.id,
+            vt.name AS name,
+            vc.updated AS modified_at,
```
Joosakur, line 71: at some point we need to add a new db column for this since `updated` is automatically changed e.g. by migrations

Joosakur (approved):

service/src/main/kotlin/fi/espoo/evaka/vasu/VasuQueries.kt
kimifoud: Added a todo to board

PR #1025, opened 2021-06-15
Prevent marking unit closed before it was opened

reynders:
#### Summary

Check that opening date is on or before closing date, and vice versa. Default min and max values are 1.1.1960 - 1.1.2100


jaahon (commented):
Should the validation function check the dates as well? In case the date picker component has a way to bypass the min and max dates.

jaahon (approved):

PR #1029, opened 2021-06-16
EVAKA-4203 Select vasu template when creating new vasu document

kimifoud:
#### Summary

 Select vasu template when creating new vasu document


massproduction (approved):

Joosakur (approved):
Not sure if this is any less code than a proper modal, but as an MVP this should work :D We can polish it later

PR #1031, opened 2021-06-16
EVAKA-TECH optional email whitelist

jaahon:
#### Summary

Add optional email whitelisting to service. Depends on https://github.com/espoon-voltti/evaka-infra/pull/432.

Tested in staging with both whitelisted and non-whitelisted emails.

### DIFF
--- a/service/src/main/kotlin/fi/espoo/evaka/emailclient/EmailClient.kt
+++ b/service/src/main/kotlin/fi/espoo/evaka/emailclient/EmailClient.kt
@@ -18,13 +18,13 @@ import mu.KotlinLogging
 
 private val logger = KotlinLogging.logger {}
 
-class EmailClient(private val client: AmazonSimpleEmailService) : IEmailClient {
+class EmailClient(private val client: AmazonSimpleEmailService, private val whitelist: List<Regex>?) : IEmailClient {
     private val charset = "UTF-8"
 
     override fun sendEmail(traceId: String, toAddress: String, fromAddress: String, subject: String, htmlBody: String, textBody: String) {
         logger.info { "Sending email (traceId: $traceId)" }
 
-        if (validateToAddress(traceId, toAddress)) {
+        if (validateToAddress(traceId, toAddress) && checkWhitelist(toAddress)) {
             try {
                 val request = SendEmailRequest()
                     .withDestination(Destination(listOf(toAddress)))
@@ -54,5 +54,15 @@ class EmailClient(private val client: AmazonSimpleEmailService) : IEmailClient {
         } else {
             logger.warn("Will not send email due to invalid toAddress: (traceId: $traceId)")
         }
+    }
+
+    private fun checkWhitelist(address: String): Boolean {
+        val isWhitelisted = whitelist?.any { it.matches(address) } ?: true
+
+        if (!isWhitelisted) logger.info {
+            "Not sending email to $address because it does not match any of the entries in whitelist"
+        }
+
+        return isWhitelisted
     }
 }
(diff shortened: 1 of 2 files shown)


service/src/main/kotlin/fi/espoo/evaka/emailclient/EmailConfig.kt
```diff
@@ -8,13 +8,25 @@ import com.amazonaws.services.simpleemail.AmazonSimpleEmailService
...
+                client = client,
+                whitelist = env.getRequiredProperty<List<String>>("application.email.whitelist").map { pattern ->
+                    // Assumes the patterns are of form "*", "*@test.com", etc.
+                    pattern
+                        .split("*")
+                        .joinToString(".*", "^", "$") {
+                            if (it.isNotBlank()) Regex.escape(it) else it
+                        }
+                        .let { Regex(it) }
+                }
```
mikkopiu, line 28: Looking at this, I feel this is quite brittle and IMO it would be better to just move the regex-reformatting into Terraform (i.e. the env vars should be expected to be valid regex as-is). In Terraform it wouldn't bee to difficult either and making assumptions about the variable format would make more sense closer to where the variables are actually defined:

```hcl
APPLICATION_EMAIL_WHITELIST = join(",", [for r in local.allowed_recipients: format("^%s$", replace(r, "*", ".*"))])
```

jaahon: Sure, that does make sense.

service/src/main/kotlin/fi/espoo/evaka/emailclient/EmailClient.kt
```diff
@@ -55,4 +55,14 @@ class EmailClient(private val client: AmazonSimpleEmailService) : IEmailClient {
             logger.warn("Will not send email due to invalid toAddress: (traceId: $traceId)")
         }
     }
+
+    private fun checkWhitelist(address: String): Boolean {
+        val isWhitelisted = whitelist == null || whitelist.any { it.matches(address) }
```
mikkopiu, line 60: Q: wouldn't `whitelist?.any { it.matches(address) } ?: true` do the same and be more idiomatic than explicit null checking?

service/src/main/kotlin/fi/espoo/evaka/emailclient/EmailConfig.kt
```diff
@@ -8,13 +8,17 @@ import com.amazonaws.services.simpleemail.AmazonSimpleEmailService
...
 
 @Configuration
 class EmailConfig {
     @Bean
     fun emailClient(client: AmazonSimpleEmailService, env: Environment): IEmailClient =
         when (env.getProperty("application.email.enabled", Boolean::class.java, false)) {
-            true -> EmailClient(client = client)
+            true -> EmailClient(
+                client = client,
+                whitelist = env.getRequiredProperty<List<String>>("application.email.whitelist").map { Regex(it) }
```
mikkopiu, line 20: As this is now a required property (when mail is enabled), should we have a default for it in `application.yaml`? Or if the idea is to force an active decision for the person deploying this service -> we should inform Trevaka about this one.

mikkopiu (changes requested):

service/src/main/kotlin/fi/espoo/evaka/emailclient/EmailConfig.kt
jaahon: It shouldn't be a required property, I just forgot to change that method. Good catch 👍

service/src/main/kotlin/fi/espoo/evaka/emailclient/EmailClient.kt
jaahon: Yeah, I guess, I just wanted to be very explicit that if there is not whitelist it's ok.

mikkopiu: Yeah, totally up to you. More "wondering" than "questioning" on my part.

jaahon: But I guess it's pretty obvious anyway.

mikkopiu (approved):
LGTM, tests would be nice but I think I've bullied you enough :D

PR #1035, opened 2021-06-16
EVAKA-HOTFIX add application info to incomes in frontend

massproduction:
#### Summary
If income has been generated from application then show it on the income list

### DIFF
--- a/frontend/src/employee-frontend/components/person-profile/income/IncomeItemBody.tsx
+++ b/frontend/src/employee-frontend/components/person-profile/income/IncomeItemBody.tsx
@@ -2,7 +2,9 @@
 //
 // SPDX-License-Identifier: LGPL-2.1-or-later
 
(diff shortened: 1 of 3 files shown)


frontend/src/employee-frontend/components/person-profile/income/IncomeItemBody.tsx
```diff
@@ -43,6 +49,32 @@ const IncomeItemBody = React.memo(function IncomeItemBody({ income }: Props) {
             ? i18n.personProfile.income.details.originApplication
             : income.updatedBy}
         </span>
+        {income.notes === 'created automatically from application' &&
```
reynders, line 52: It would be better have the backend signal this with a flag rather than this string deduction, but ok for a hotfix if it proves to be to big of a change

reynders (approved):

PR #1040, opened 2021-06-16
EVAKA-4190: Messaging Playwright Tests

mapehe:
#### Summary

Add an e2e to test sending and replying to a message.

### DIFF
--- a/frontend/README.md
+++ b/frontend/README.md
@@ -213,8 +213,34 @@ test('Application main page is visible', async (t) => {
 
 #### Run all tests locally
 
+E2E-tests are implemented using testcafe (legacy) and playwright.
+Testcafe tests can be run via the following command:
+
 ```sh
 yarn e2e
+```
+
+If you don't want to wipe data from your dev environment,
+then start the dev server by running the following commands in
+the `service` directory before running the playwright tests:
+
+```sh
+pm2 stop service
+./gradlew bootRunTest
+```
+
+Playwright tests can be started with:
+
+```sh
+yarn e2e-playwright
+```
+
+You can run a single spec by specifying the corresponding subdirectory. For example,
+in order to only run tests related to messaging, use the following command:
+
+```sh
+yarn jest --testTimeout 600000 --runInBand src/e2e-playwright/specs/7_messaging
+```
 
 #### Run tests by fixture or test meta data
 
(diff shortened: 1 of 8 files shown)


frontend/src/e2e-playwright/pages/citizen/citizen-messages.ts
```diff
@@ -0,0 +1,20 @@
...
+  constructor(private readonly page: Page) {}
+
+  #messageReplyContent = new RawTextInput(
+    this.page,
+    '[data-qa="message-reply-content"]'
+  )
+  #threadListItem = new RawElement(this.page, '[data-qa="thread-list-item"]')
+  #sendReplyButton = new RawElement(this.page, '[data-qa="message-send-btn"]')
+
+  async replyToThread(index: number, content: string) {
```
jaahon, line 14: `index` is an unused parameter.

frontend/src/e2e-playwright/pages/citizen/citizen-messages.ts
```diff
@@ -0,0 +1,20 @@
...
+    await this.#threadListItem.click()
+    await this.#messageReplyContent.fill(content)
+    await this.#sendReplyButton.click()
+    await this.page.waitForTimeout(1000)
```
jaahon, line 18: In general it's not a good idea to just do a fixed time timeout. If you need to wait for some async action to complete, read the UI for an indication that it completed.

frontend/README.md
```diff
@@ -213,8 +213,39 @@ test('Application main page is visible', async (t) => {
...
+pm2 stop service
+```
+
+Then start the dev server by running the following command in the `service`
+directory:
+
+```sh
+./gradlew bootRunTest
+```
+
```
jaahon, line 236: This is actually not required, so you might want to reword this bit.

mapehe: Thanks for the feedback, could you explain this in a bit more detail?

mapehe:
Thanks for the feedback @jaahon, made some changes accordingly.

frontend/README.md
jaahon: I mean that running `service` with `./gradlew bootRunTest` is not required to run playwright tests. It's a useful thing sometimes when you don't want to wipe data from your dev environment, but it's not required.

mapehe: Ok, thanks. Reworded README a bit.

jaahon (approved):

PR #1042, opened 2021-06-17
EVAKA-HOTFIX axios connection error handling

jaahon:
#### Summary

The axios error's `response` property was incorrectly cast into a type that is never undefined although the property might sometimes be missing, like when there was a connection error, so remove the cast.


mikkopiu (approved):
:+1:

PR #1051, opened 2021-06-18
EVAKA better handling for dead people

jaahon:
#### Summary

* End family relation on person death
* Handle person death only with definitive date of death
* Date of death validations to fridge family modals
* Manually send application decisions if applicant has died
* Add indication of applicant death on application


service/src/main/kotlin/fi/espoo/evaka/dvv/DvvModificationsService.kt
```diff
@@ -95,14 +104,46 @@ class DvvModificationsService(
...
+
+                endFamilyRelations(tx, person.id, dateOfDeath)
+                asyncJobRunner.plan(tx, listOf(GenerateFinanceDecisions.forAdult(person.id, DateRange(dateOfDeath, null))))
+                asyncJobRunner.plan(tx, listOf(GenerateFinanceDecisions.forChild(person.id, DateRange(dateOfDeath, null))))
+            }
         }
     }
 
+    private fun endFamilyRelations(tx: Database.Transaction, personId: UUID, dateOfDeath: LocalDate) {
+        tx
```
reynders, line 124: What about the guardian table?

reynders (approved):

service/src/main/kotlin/fi/espoo/evaka/dvv/DvvModificationsService.kt
jaahon: Isn't mutpa supposed to handle that? If the guardianship ends, mutpa should send an event about that. At least that is my assumption. My opinion is that since it's not something our users have control over anyway, I don't think it should be touched here.

PR #1052, opened 2021-06-21
EVAKA-4203 Vasu header section

kimifoud:
#### Summary

Implement vasu header section.

header section


jaahon (approved):
By the way, not related to this PR, but if I try to create a new vasu for a child without a row in `child` table, the frontend bombards the backend with create requests. It happens if there is any kind of backend error.

kimifoud:
> By the way, not related to this PR, but if I try to create a new vasu for a child without a row in `child` table, the frontend bombards the backend with create requests. It happens if there is any kind of backend error.

Yikes, good catch, will handle that in a separate PR!

PR #1056, opened 2021-06-21
EVAKA-FIX exclude varda person if vtj timestamp is null

hruoho:
#### Summary

Aim is to avoid a class of errors by creating person only if data originates from dvv


reynders (approved):

mikkopiu:
Is this pending on purpose?

hruoho:
> Is this pending on purpose?

In a sense yes but I guess it's better to close this for now.

PR #1063, opened 2021-06-22
Use new job scheduler for all jobs

Gekkio:
#### Summary

- not all jobs are daily jobs -> refactor and rename new scheduler functionality
- move all scheduled jobs to use the new system
- scheduled times are now in Europe/Helsinki timezone. Existing schedules in evaka-infra were in UTC time, and were converted by assuming a UTC+2 offset
- ~`ScheduledOperationController` will be removed in a later pull request~ removed now

**Don't merge yet! Some jobs are disabled by default in the code, and need to be enabled in a separate infra pull request**


service/src/integrationTest/kotlin/fi/espoo/evaka/shared/job/ScheduledJobsTest.kt
```diff
@@ -1,8 +1,4 @@
-// SPDX-FileCopyrightText: 2017-2020 City of Espoo
-//
-// SPDX-License-Identifier: LGPL-2.1-or-later
-
```
mikkopiu, line 4: Mistakenly dropped?

service/src/main/kotlin/fi/espoo/evaka/koski/KoskiUpdateService.kt
```diff
@@ -29,7 +29,10 @@ class KoskiUpdateService(
         asyncJobRunner.scheduleKoskiUploads = { db, msg -> scheduleKoskiUploads(db, msg.params) }
     }
 
-    fun scheduleKoskiUploads(db: Database, params: KoskiSearchParams) {
+    fun scheduleKoskiUploads(db: Database, params: KoskiSearchParams) = db.connect {
+        scheduleKoskiUploads(it, params)
+    }
```
mikkopiu, line 34: It's probably just too hot in here and I'm being stupid but why doesn't this care about "fi.espoo.integration.koski.enabled"?

service/src/main/kotlin/fi/espoo/evaka/shared/controllers/ScheduledOperationController.kt
```diff
@@ -4,52 +4,20 @@
...
-    private val asyncJobRunner: AsyncJobRunner
-) {
+class ScheduledOperationController {
 
     @PostMapping("/dvv/update")
-    fun dvvUpdate(db: Database.Connection): ResponseEntity<Int> {
-        Audit.VtjBatchSchedule.log()
-        return ResponseEntity.ok(dvvModificationsBatchRefreshService.scheduleBatch(db))
+    fun dvvUpdate(db: Database.Connection) {
+        // executed by DailyJobRunner instead
```
mikkopiu, line 20: Does this respond with "no content" implicitly (comparing this and e.g. `/koski/update`)?

service/src/main/kotlin/fi/espoo/evaka/shared/job/JobSchedule.kt
```diff
@@ -0,0 +1,37 @@
...
+        ScheduledJob.VardaUpdate -> JobSchedule.daily(LocalTime.of(23, 0))
+        ScheduledJob.EndOfDayAttendanceUpkeep -> JobSchedule.daily(LocalTime.of(0, 0))
+        ScheduledJob.KoskiUpdate -> JobSchedule.daily(LocalTime.of(0, 0))
+        ScheduledJob.RemoveOldDraftApplications -> JobSchedule.daily(LocalTime.of(0, 30))
+        ScheduledJob.CancelOutdatedTransferApplications -> JobSchedule.daily(LocalTime.of(0, 35))
+        ScheduledJob.EndOutdatedVoucherValueDecisions -> JobSchedule.daily(LocalTime.of(1, 0))
+        ScheduledJob.RemoveOldAsyncJobs -> JobSchedule.daily(LocalTime.of(3, 0))
+        ScheduledJob.DvvUpdate -> JobSchedule.daily(LocalTime.of(6, 0))
+        ScheduledJob.RemoveOldDaycareDailyNotes -> JobSchedule.daily(LocalTime.of(6, 30))
+        ScheduledJob.SendPendingDecisionReminderEmails -> JobSchedule.daily(LocalTime.of(7, 0))
```
mikkopiu, line 33: Would these schedules be overridable in other implementations (i.e. trevaka)? These times can be highly dependent on the deployment environment (e.g. in Espoo, we have a specific downtime for VTJ access which forces `DvvUpdate` to be at a specific time).

If not as-is (I haven't yet familiarized myself with how the fork-system works for evaka-service), then maybe these should be env vars with defaults?

service/src/main/kotlin/fi/espoo/evaka/shared/job/ScheduledJobs.kt
```diff
@@ -0,0 +1,135 @@
...
+
+    fun vardaUpdate(db: Database.Connection) {
+        vardaUpdateService.scheduleVardaUpdate(db)
+    }
+
+    fun removeOldDraftApplications(db: Database.Connection) {
+        db.transaction { it.removeOldDrafts(attachmentsController::deleteAttachment) }
+    }
+
+    fun cancelOutdatedTransferApplications(db: Database.Connection) {
```
mikkopiu, line 95: Should the audit logging be placed here now that it won't be in the controller? Same for all of the jobs where auditing is relevant.

And just for posterity: audit logs are (in Voltti's case at least but also commonly) stored for much longer than other "technical" logs.

mikkopiu (changes requested):

service/src/integrationTest/kotlin/fi/espoo/evaka/shared/job/ScheduledJobsTest.kt
Gekkio: Oops :sweat_smile: 
The file has a lot of other unintended changes too...I'll redo this file

service/src/main/kotlin/fi/espoo/evaka/koski/KoskiUpdateService.kt
Gekkio: The check is done in the inner `scheduleKoskiUploads(it, params)` call

mikkopiu: :facepalm:

service/src/main/kotlin/fi/espoo/evaka/shared/job/ScheduledJobs.kt
Gekkio: I thought about this and decided to drop the audit logs, because

a) the jobs are nightly "batch" jobs, and not something that are started by humans. If we add a REST endpoint for manual triggering, we can of course add audit logging there
b) the old jobs were inconsistent: some had audit logging and some didn't

service/src/main/kotlin/fi/espoo/evaka/shared/job/JobSchedule.kt
Gekkio: This is just the default configuration, and in practice a `JobSchedule` implementation needs in any case to be provided before the service can start (in our case it's in `EspooConfig.kt`). It's up to the implementation to decide whether it uses this default configuration, uses some mix of default + overrides, or pulls values from env vars.

service/src/main/kotlin/fi/espoo/evaka/shared/job/ScheduledJobs.kt
jaahon: By the way, I don't know if you noticed, but canceling outdated transfer applications is not on in production. I don't know if this PR enables it in prod, but that's something that should be taken into account.

Gekkio: Good catch :cold_sweat: I didn't notice, because I just looked at the "schedule" variables in terraform and didn't notice many jobs had separate "enabled" variables.

I'll check all the jobs and will probably have to look at env variables in EspooConfig.kt

mikkopiu: Regarding audit logging: I think the only questionable one is `dvvUpdate` **but** I think we fulfill the VTJ requirements by having audit logging when users make requests for data that originated from VTJ:

> Luvan saaja vastaa siitä, että jälkikäteen on selvitettävissä henkilötasolla, kuka väestötietojärjestelmästä haettuja tietoja on saanut asiakasjärjestelmän kautta nähtäväkseen. Organisaatio, jolle oikeus sovelluskyselyyn on annettu, on velvollinen säilyttämään lokitiedot viideltä (5) vuodelta.

So on that part -> :heavy_check_mark: 

But good catch on the jobs :+1:

service/src/main/kotlin/fi/espoo/evaka/shared/controllers/ScheduledOperationController.kt
mikkopiu: What about this one ^

mikkopiu (commented):
One Q, otherwise looks good

service/src/main/kotlin/fi/espoo/evaka/shared/controllers/ScheduledOperationController.kt
Gekkio: A void function returns status 200 with empty body (`Content-Length: 0`), so this behaves differently than earlier but it shouldn't matter.

But I think I'll add one more commit and remove the REST controller completely in this pull request...it means the nightly lambdas will fail, but it's just a temporary problem until they're removed as well

mikkopiu (approved):
LGTM but make sure this is communicated to Tampere before merging and that we have PR to drop the scheduled jobs from evaka-infra, too (mainly to avoid noise).

PR #1067, opened 2021-06-22
Add placement reject reason customization support

patari:
#### Summary

Adds support to customize placement reject reason selection list. Also adds third reason for Tampere.


mikkopiu (approved):
LGTM

PR #1069, opened 2021-06-23
EVAKA-4203 Auto-save vasu document on changes

kimifoud:
#### Summary

Auto-save vasu document on changes. Display latest save time, show spinner while saving.
Also fix infinite loop if create request fails.

### DIFF
--- a/frontend/src/employee-frontend/components/vasu/use-vasu.ts
+++ b/frontend/src/employee-frontend/components/vasu/use-vasu.ts
@@ -47,27 +47,31 @@ export function useVasu(id: string): Vasu {
   const [vasu, setVasu] = useState<VasuMetadata>()
   const [content, setContent] = useState<VasuContent>({ sections: [] })
 
-  useEffect(() => {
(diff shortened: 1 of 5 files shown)


frontend/src/employee-frontend/components/vasu/use-vasu.ts
```diff
@@ -0,0 +1,103 @@
...
+      success: () => setStatus({ state: 'clean', savedAt: new Date() })
+    })
+  )
+  const saveNow = useCallback(
+    (params: PutVasuDocumentParams) => {
+      setStatus((prev) => ({ ...prev, state: 'saving' }))
+      save(params)
+    },
+    [save]
+  )
```
jaahon, line 81: Just a heads-up, this `save` callback's identity changes on every render since the `setState` parameter passed to it is recreated on every render. That's why the `useCallback` used in `saveNow` does practically nothing.

jaahon (approved):
Just a minor comment, no real need to do anything about it.

Also, if the vasu creation fails, the button gets stuck in disabled state. Feel free to add it to the backlog, though.

PR #1073, opened 2021-06-23
EVAKA-FIX: unique people for invoice children cell

hruoho:
#### Summary

Use unique before passing children to component


frontend/src/employee-frontend/components/invoices/Invoices.tsx
```diff
@@ -287,7 +288,9 @@ const InvoiceTableBody = React.memo(function InvoiceTableBody({
             <NameWithSsn {...item.headOfFamily} i18n={i18n} />
           </Td>
           <Td>
-            <ChildrenCell people={item.rows.map(({ child }) => child)} />
+            <ChildrenCell
+              people={unique(item.rows.map(({ child }) => child))}
+            />
```
jaahon, line 293: `child` is an object so `unique` won't work with it. Lodash has `uniqBy` which you can use here instead.

jaahon (approved):

PR #1074, opened 2021-06-23
Upgrade backend dependencies

Gekkio:
#### Summary

Highlights:

- Kotlin 1.5
- Spring Boot 2.5.1

### DIFF
--- a/message-service/build.gradle.kts
+++ b/message-service/build.gradle.kts
@@ -4,6 +4,7 @@
 
 import org.jetbrains.kotlin.gradle.plugin.KotlinSourceSet
 import org.jetbrains.kotlin.gradle.tasks.KotlinCompile
+import org.springframework.boot.gradle.tasks.bundling.BootJar
 
 buildscript {
     repositories {
@@ -150,6 +151,10 @@ tasks.withType<KotlinCompile> {
         jvmTarget = Version.java
         allWarningsAsErrors = true
     }
+}
+
+tasks.getByName<Jar>("jar") {
+    enabled = false
 }
 
 project.wsdl2javaExt {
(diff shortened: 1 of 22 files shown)


service/vtjclient/build.gradle.kts
```diff
@@ -30,7 +30,7 @@ dependencies {
 }
 
 project.wsdl2javaExt {
-    cxfVersion = "3.3.7"
+    cxfVersion = "3.4.4"
```
mikkopiu, line 33: Doesn't this mean the WSDL stuff should be re-built/-compiled/-whatever-it-is? I just remember trying to update this at some point and everything going wonky with those classes. Maybe I'm misremembering...

Gekkio: Yeah this has no practical effect other than removing cxf from the list of "potential updates" in gradle versions plugin, and fixing a CVE.

WSDL generation is indeed broken at the moment and the wsdl2java plugin we use for it has been abandoned, but I've got a fix for this and will submit it in a follow-up pull request

mikkopiu (approved):

message-service/build.gradle.kts
```diff
@@ -150,6 +153,10 @@ tasks.withType<KotlinCompile> {
     }
 }
 
+tasks.getByName<Jar>("jar") {
+    enabled = false
+}
+
 project.wsdl2javaExt {
     cxfVersion = "3.4.2"
```
mikkopiu, line 161: Actually, update this one, too

mikkopiu (changes requested):

mikkopiu (approved):

PR #1079, opened 2021-06-24
Use clearable selector for finance unit and handler

reynders:

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -335,6 +335,10 @@ commands:
             yarn --version
       - *restore_frontend_deps
       - run:
+          name: Install Playwright dependencies
+          command: |
+            sudo apt-get update && sudo apt-get --yes install --no-install-recommends libgbm1
+      - run:
           name: Run e2e tests against compose
           working_directory: *workspace_frontend
           command: |
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -30,7 +30,7 @@ aliases:
   - &shellcheck_image_version v0.7.2
   - &nodejs_image cimg/node:14.15
   - &openjdk_image cimg/openjdk:11.0
-  - &ubuntu_machine_image ubuntu-2004:202010-01
+  - &ubuntu_machine_image ubuntu-2004:202104-01
   - &builder_aws_image 307238562370.dkr.ecr.eu-west-1.amazonaws.com/voltti/builder-aws:18.04-809ad64eaee48225fba439142026079a49d0f1d5
 
   - &default_contexts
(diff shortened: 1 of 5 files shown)


frontend/src/employee-frontend/components/common/Filters.tsx
```diff
@@ -273,13 +274,16 @@ export const UnitFilter = React.memo(function UnitFilter({
...
+          items={options}
           placeholder={i18n.filters.unitPlaceholder}
-          options={options}
-          onChange={(option) =>
-            option && 'id' in option ? select(option.id) : undefined
-          }
-          value={selected}
+          selectedItem={selected ?? null}
+          onChange={(option) => {
+            option && 'id' in option ? select(option.id) : select(null)
```
jaahon, line 282: There is no need to do the `'id' in option` check with `Combobox`, it was only mandatory with react select since it had crappy types.

Also, if you typed the `select` prop like this `select: (unit?: string) => void`, this could be simplified to 
```
onChange={(option) => {
  select(option.id)
}
```
and you wouldn't have to do `unit: id || undefined` in the `selectUnit` and `selectFinanceDecisionHandler` functions.

frontend/src/employee-frontend/components/fee-decisions/FeeDecisionFilters.tsx
```diff
@@ -90,13 +90,13 @@ function FeeDecisionFilters() {
...
-    setSearchFilters((filters) => ({ ...filters, unit: id }))
+  const selectUnit = (id?: string) =>
+    setSearchFilters((filters) => ({ ...filters, unit: id || undefined }))
 
-  const selectFinanceDecisionHandler = (id: string) =>
+  const selectFinanceDecisionHandler = (id?: string) =>
     setSearchFilters((filters) => ({
       ...filters,
-      financeDecisionHandlerId: id
+      financeDecisionHandlerId: id || undefined
```
jaahon, line 99: These `id || undefined`s can just be `id`

jaahon (approved):
Just fix the `InvoiceFilters.tsx` file and this is good to merge

.circleci/config.yml
```diff
@@ -274,6 +274,10 @@ commands:
               && sudo mv chromedriver /usr/local/bin/chromedriver \
               && sudo chmod +x /usr/local/bin/chromedriver \
               && chromedriver --version
+      - run:
+          name: Install Playwright dependencies
+          command: |
+            sudo apt-get update && sudo apt-get --yes install --no-install-recommends libgbm1
```
mikkopiu, line 280: This should be in the `e2e_test_playwright` command (they're separate commands for some reason for now, will be fixed in my future PR)

mikkopiu (changes requested):

mikkopiu (approved):

.circleci/config.yml
```diff
@@ -334,6 +334,10 @@ commands:
             sudo apt-get update && sudo apt-get --yes install --no-install-recommends yarn=<< parameters.yarn_version >>
             yarn --version
       - *restore_frontend_deps
+      - run:
+          name: Install Playwright dependencies
+          command: |
+            sudo apt-get update && sudo apt-get --yes install --no-install-recommends libgbm
```
mikkopiu, line 340: Actually, this should be:
```suggestion
            sudo apt-get update && sudo apt-get --yes install --no-install-recommends libgbm1
```

mikkopiu (changes requested):

mikkopiu (approved):

PR #1084, opened 2021-06-28
Update db schema diagram

reynders:
#### Summary
Update db schema diagram to reflect the current db schema


huhtari:
Staging database has 81 tables, this around 40? How about for example messaging tables?

huhtari (approved):
LGTM

PR #1086, opened 2021-06-29
Refactor default mail address to support customization

patari:
#### Summary

Refactor hard coded values to support customization.

Also fixes voucher value fee decision po box visibility.


service/src/main/kotlin/fi/espoo/evaka/invoicing/service/PdfService.kt
```diff
@@ -155,6 +157,7 @@ class PDFService(
             "showTotalIncome" to !hideTotalIncome,
             "coPayment" to formatCents(decision.finalCoPayment),
             "decisionNumber" to decision.decisionNumber,
+            "hasPoBox" to (sendAddress.poBox != null),
```
patari, line 160: Fixes voucher value fee decision po box visibility (see voucher-value-decision.html)

reynders (approved):

PR #1088, opened 2021-06-29
TREV-237 enable/disable urgent application's attachments on employee …

apeltola:
…views. As already done in citizen views.


mikkopiu (approved):
Pushing for CI

PR #1090, opened 2021-06-29
[EVAKA] CI: refactor workflow for speed and parallelization

mikkopiu:
#### Summary

| Thing | Current | New |
|-|-|-|
| Branch workflow | ~23 min | **~17 min** :tada: |
| Service tests | ~13 min | **~3 min** :tada: |
| Frontend deploys | ~3 min | **~30 s** :rocket: |
| E2E tests | ~12 min (slowest) | ~9 min :+1: |
| Workspace attaching | 30 s - 2.5 min | **~2 s** :muscle: |
| Developers | :cry: | :smiley: |

- Parallelize E2E tests and service integration tests based on historical test timing data
    - Total actual time (when all parallel tests are finished) is minimized
    - Automatic test splitting means that we don't need to manually select which tests are run together for an optimal set (timings wise)
    - New tests are randomly split until timings for them exist
    - For E2E tests this has the additional benefit of somewhat ensuring that fixtures are independent as any two fixtures aren't guaranteed to be ran on the same node every time or in a specific order
- Move most files out of the CircleCI workspace into workflow-run-specific caches to avoid downloading unnecessary files in all downstream jobs
    - **Major** speed up for frontend deploys: ~2.5 min is immediately shaved, bringing total down to ~15 s
    - This is totally abusing CircleCI's caches but currently there isn't any sort of "sub-workspace" solution (that I know of) that could do partial downloads
- Cache Gradle builds within a single workflow execution and pre-compile [integration] test classes for test parallelization
    - Also avoid caching unnecessary Gradle files
- Split service & message-service test and integration test runs into parallel jobs
- Drop layer caching for E2E jobs
    - This feature is really costly (:dollar:) without _that_ big of a benefit, especially with parallelization (200 credits * 8 parallel jobs for every single workflow execution adds up quickly)
    - Lost performance is gained back with the other optimizations in this PR
- Merge build and push image jobs
    - Pushing doesn't take _that_ long in the context of the whole workflow and just complicates the config (and workflow visualization)
    - Starting another executor just for pushing takes ~20s, meaning the time to push an image roughly doubles (depending on image size etc. of course)
- Update Docker to 20.10.6
- Update `voltti/builder-aws` to the latest version & use a more optimized (read: smaller) variant where some tools aren't needed
    - Image sizes drop down from ~1.13GB -> ~300MB in the best case and even in the worst case (Terraform deploys) roughly 200MB can be saved, leading to a small improvement over time
- Use Yarn cache for `node_modules` persistence
    - Instead of "caching" the `frontend/node_modules` directory in the workspace, use a real cache
    - This should speed up all jobs attaching workspaces (pretty much all jobs) as they don't need to fetch the ~500MB of `node_modules` contents just to start
- Use the official `JAVA_TOOL_OPTIONS` instead of the unofficial (but pretty widely supported) `_JAVA_OPTIONS` for setting JVM configs
- General refactoring: aliases to commands & parameters, dedupe jobs, clearer names for steps
- Adjust JVM memory configs to better utilize resources
    - E.g. prefer the modern `MaxRAMPercentage` over `Xmx`
    - No real performance benefit right now, just not leaving resources explicitly unused (2 GB out of 4 GB vs. 80 % of 4 GB)
- Fix Kotlin linter warnings and deprecated Gradle features
    - Couldn't yet figure out some warnings related to Gradle 8.x deprecations
- Ensure unique E2E fixture names
    - As `testcafe-reporter-junit` doesn't output the `filename` attribute in its reports, we have to use fixture names -> they need to be unique to ensure correct test splitting
- Cache commercial dependencies (minor optimization, couple of seconds)
- Use new TS config config for Testcafe (`tsConfigPath is deprecated)
- Update CircleCI orbs (nothing worth noting, just keeping up-to-date)

##### Current branch

Total branch workflow execution (~23 min):

image

##### New workflow

Total branch workflow execution (~16 min):

image

Parallel executions in UI:

image

#### Notes

- For the cache abuse trick, there's the danger of the cache getting randomly cleared _just_ when its needed but IME CircleCI pretty consistenly keeps the caches available for the documented 14 days and so far I haven't faced this issue in a few days of testing
- Frontend (TypeScript mainly) builds are still slow and slow down the entire workflow as they block E2E test jobs from starting (they need the built frontends)
- Decided to _not_ cache Gradle _builds_ between workflow executions to avoid poisoning other builds, especially with Spring included
    - This slows down the build by ~1.5 min but IMO clean builds are worth it
- Service unit tests and message service unit & integration tests are fast enough to not benefit from parallelization (executor start and Gradle boot take enough time to take away any benefits gained from parallel test execution)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -829,6 +829,11 @@ jobs:
       - run:
           working_directory: *workspace_service
           command: ./gradlew assemble compileIntegrationTestKotlin
+      - run:
+          working_directory: *workspace_service
+          command: |
+            cd custom-ktlint-rules
+            ./gradlew assemble
       - *store_service_gradle
       - persist_to_workspace:
           root: *workspace_root
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -233,7 +233,7 @@ commands:
   store_service_gradle:
     steps:
       - save_cache:
-          key: gradle-home-service-v4-{{ checksum ".circleci/config.yml" }}-{{ checksum "service/build.gradle.kts" }}-{{ checksum "service/gradle.properties" }}-{{ checksum "service/settings.gradle.kts" }}-{{ checksum "service/buildSrc/build.gradle.kts" }}-{{ checksum "service/buildSrc/src/main/kotlin/Version.kt" }}
+          key: gradle-home-service-v5-{{ checksum ".circleci/config.yml" }}-{{ checksum "service/build.gradle.kts" }}-{{ checksum "service/gradle.properties" }}-{{ checksum "service/settings.gradle.kts" }}-{{ checksum "service/buildSrc/build.gradle.kts" }}-{{ checksum "service/buildSrc/src/main/kotlin/Version.kt" }}
           paths:
             - service/buildSrc/.gradle
             - service/.gradle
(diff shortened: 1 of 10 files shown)


Gekkio (approved):

mikkopiu:
Verified that new tests (and tests that don't have timings) are randomly placed: the output is just a bit confusing:

```
No timing found for "fi.espoo.evaka.pis.AbstractIntegrationTest"
```

comes from the splitting going through _all_ the tests -> the same log will appear on all nodes **and one of the nodes will actually  receive those classes** ([example](https://app.circleci.com/pipelines/github/espoon-voltti/evaka/8032/workflows/581e916c-111d-40a6-b930-15921ff4ac06/jobs/148202/parallel-runs/3?filterBy=ALL&invite=true#step-105-32)). And just for further proof, here's a run with a totally new test class: https://app.circleci.com/pipelines/github/espoon-voltti/evaka/8032/workflows/581e916c-111d-40a6-b930-15921ff4ac06/jobs/148202/parallel-runs/0?filterBy=ALL&invite=true#step-105-30

mikkopiu:
Improved test splitting output for debugging purposes, split Playwright tests (I think I just forgot about this after I started off the path of running E2E tests in a container...) and updated to the latest `voltti/builder-aws` images.

mikkopiu:
Fixed caching between retried workflow runs

Gekkio (approved):

PR #1091, opened 2021-06-29
EVAKA-4203 Copy vasu template

kimifoud:
#### Summary

Add support for copying vasu templates.

- Copy custom vasu sections from a template to another
- Add 1 year to both start and end date by default

### DIFF
--- a/service/src/main/kotlin/fi/espoo/evaka/vasu/OphQuestion.kt
+++ b/service/src/main/kotlin/fi/espoo/evaka/vasu/OphQuestion.kt
@@ -59,3 +59,9 @@ private val ophQuestionMap = mapOf(
         )
     )
 )
+
+fun copyTemplateContentWithCurrentlyValidOphSections(template: VasuTemplate): VasuContent {
+    val copyableSections = template.content.sections.filter { section -> section.questions.all { question -> question.ophKey == null } }
+    val defaultSections = getDefaultTemplateContent(template.language).sections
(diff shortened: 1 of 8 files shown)


service/src/main/kotlin/fi/espoo/evaka/vasu/OphQuestion.kt
```diff
@@ -59,3 +59,9 @@ private val ophQuestionMap = mapOf(
         )
     )
 )
+
+fun copyTemplateContentWithCurrentlyValidOphSections(template: VasuTemplate): VasuContent {
+    val copyableSections = template.content.sections.filter { section -> section.questions.all { question -> question.ophKey == null } }
+    val defaultSections = getDefaultTemplateContent(template.language).sections
+    return VasuContent(sections = copyableSections + defaultSections)
```
Joosakur, line 66: Kinda unfortunate that both ordering and some questions get lost even if there have been no chages to default questions. I wonder if we could somehow check if the default sections haven't been changed since and in that case just copy the whole template as is.. Oh well, maybe not worth spending much time and money on thinking about that.

Joosakur (approved):

service/src/main/kotlin/fi/espoo/evaka/vasu/OphQuestion.kt
kimifoud: Yup, these concerns can be properly addressed once the default template is actually implemented (with versioning).

PR #1099, opened 2021-07-01
TREV-254 employee application creation: show/hide preschool selection…

apeltola:
… with featureFlag


mikkopiu (commented):
Should preschool be hidden from here as well: `frontend/src/employee-frontend/components/common/Filters.tsx#L750`?

apeltola:
> 
> 
> Should preschool be hidden from here as well: `frontend/src/employee-frontend/components/common/Filters.tsx#L750`?

@mikkopiu it already is (line 769), but in a different way. Maybe it should be made same way as in this PR. I kind of would like that these options would come through customization files so that one should not every file having preschool-option would need a change.

mikkopiu (approved):
Pushing for CI

PR #1102, opened 2021-07-01
EVAKA-4203: vasu and evaluation discussion sections

Joosakur:

### DIFF
--- a/frontend/src/employee-frontend/components/vasu/sections/VasuEvents.tsx
+++ b/frontend/src/employee-frontend/components/vasu/sections/VasuEvents.tsx
@@ -4,7 +4,6 @@
 
 import React from 'react'
 import styled from 'styled-components'
-import { formatDate } from '../../../../lib-common/date'
 import HorizontalLine from '../../../../lib-components/atoms/HorizontalLine'
 import { ContentArea } from '../../../../lib-components/layout/Container'
 import ListGrid from '../../../../lib-components/layout/ListGrid'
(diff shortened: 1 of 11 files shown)


frontend/src/employee-frontend/components/vasu/sections/VasuEvents.tsx
```diff
@@ -74,8 +65,16 @@ export function VasuEvents({
...
-        <EventRow label={i18n.vasu.lastModified} date={modifiedAt} />
-        <EventRow label={i18n.vasu.lastPublished} date={lastPublished} />
+        <EventRow
+          label={i18n.vasu.lastModified}
+          date={LocalDate.fromSystemTzDate(modifiedAt)}
+        />
+        <EventRow
+          label={i18n.vasu.lastPublished}
+          date={
+            lastPublished ? LocalDate.fromSystemTzDate(lastPublished) : null
```
kimifoud, line 75: These conversions are now done on each render. Eg. on every keystroke in a text input. It might have a (maybe negligible) performance impact when there are multiple events.
Should we just convert all event dates to `LocalDate` in response mapping? I cannot think of a feature where timestamps are required.

Joosakur: Hmm yeah, in figma it only shows dates at the moment, but I was actually wondering if the publishing should also include time. Sometimes you might publish multiple times within a day, e.g. fix some typo in between. But maybe your idea is best for now.

kimifoud: Or just disregard the potential perf issue for now 😄

kimifoud (approved):

PR #1109, opened 2021-07-02
Fix static chip color and add story

kimifoud:
#### Summary

- Add storybook story for `StaticChip`
- Fix `readableColor` parameter order in `StaticChip`
- Remove color override from `VasuStateChip`

image


Joosakur (approved):
💎

PR #1112, opened 2021-07-05
EVAKA-4203: publishing backend

Joosakur:

### DIFF
--- a/service/src/main/resources/db/migration/R__dev_temp.sql
+++ b/service/src/main/resources/db/migration/R__dev_temp.sql
@@ -38,7 +38,8 @@ CREATE TABLE vasu_content(
     created timestamp with time zone DEFAULT now() NOT NULL,
     updated timestamp with time zone DEFAULT now() NOT NULL,
     document_id uuid NOT NULL REFERENCES vasu_document(id),
-    published_revision bigint DEFAULT NULL,
+    published_at timestamp with time zone DEFAULT NULL,
+    master bool GENERATED ALWAYS AS (published_at IS NULL) STORED,
     content jsonb NOT NULL,
     authors_content jsonb NOT NULL,
     vasu_discussion_content jsonb NOT NULL,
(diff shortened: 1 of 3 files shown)


service/src/main/resources/db/migration/R__dev_temp.sql
```diff
@@ -19,33 +19,36 @@ CREATE TABLE vasu_template(
...
+    evaluation_discussion_content jsonb NOT NULL
 );
 
-CREATE TRIGGER set_timestamp BEFORE UPDATE ON vasu_document FOR EACH ROW EXECUTE PROCEDURE trigger_refresh_updated();
+CREATE TRIGGER set_timestamp BEFORE UPDATE ON vasu_content FOR EACH ROW EXECUTE PROCEDURE trigger_refresh_updated();
 
-CREATE INDEX fk$vasu_document_child_id ON vasu_document(child_id);
-CREATE INDEX fk$vasu_document_template_id ON vasu_document(template_id);
-CREATE INDEX fk$vasu_document_content_id ON vasu_document(content_id);
+CREATE UNIQUE INDEX uniq$vasu_content_document_id_master ON vasu_content(document_id, (published_revision IS NULL)) WHERE published_revision IS NULL;
```
Joosakur, line 50: is this the best way to ensure only one row per document with null value? borrowed this from internet but maybe it could be even simpler?

Joosakur: would this work as well?
`CREATE UNIQUE INDEX uniq$vasu_content_document_id_master ON vasu_content(document_id) WHERE published_revision IS NULL;`

Gekkio: Yeah I think that's enough and there's no need to duplicate the condition in the index itself...a very similar kind of partial index is used with `application_form` latest flag

kimifoud (approved):

PR #1113, opened 2021-07-05
Group end date validation

Gekkio:
#### Summary

Group end date should never be before start date

### DIFF
--- a/service/src/main/resources/db/migration/V119__daycare_group_start_before_end.sql
+++ b/service/src/main/resources/db/migration/V119__daycare_group_start_before_end.sql
@@ -1,0 +1,1 @@
+ALTER TABLE daycare_group ADD CONSTRAINT start_before_end CHECK ((start_date <= end_date));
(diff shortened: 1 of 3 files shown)


Joosakur (approved):

service/src/main/resources/db/migration/V119__daycare_group_start_before_end.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE daycare_group ADD CONSTRAINT start_before_end CHECK ((start_date <= end_date));
```
kimifoud, line 1: Do we want to allow `start_date == end_date`?

Joosakur: yes, technically it is ok (end date is inclusive)

Gekkio: It's probably not a very meaningful case, but semantically it's a valid date range so there's no reason to forbid it

reynders (approved):

PR #1115, opened 2021-07-05
Init child table when a child person is created

reynders:
#### Summary
Init child when a child person is created so that for example editing the child profile info works before an application is created


mikkopiu (approved):
Maybe something worth testing, though?

PR #1118, opened 2021-07-05
Only send SENT fee data to Varda

reynders:

### DIFF
--- a/service/src/main/kotlin/fi/espoo/evaka/varda/VardaUpdateServiceV2.kt
+++ b/service/src/main/kotlin/fi/espoo/evaka/varda/VardaUpdateServiceV2.kt
@@ -7,8 +7,11 @@ package fi.espoo.evaka.varda
 import com.fasterxml.jackson.databind.ObjectMapper
 import com.github.kittinunf.fuel.core.FuelManager
 import fi.espoo.evaka.daycare.domain.ProviderType
+import fi.espoo.evaka.invoicing.data.getFeeDecision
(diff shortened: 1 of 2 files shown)


service/src/main/kotlin/fi/espoo/evaka/varda/VardaUpdateServiceV2.kt
```diff
@@ -624,7 +645,8 @@ WITH child_fees AS (
     fdc.child_id,
     fd.valid_during
   FROM fee_decision fd JOIN fee_decision_child fdc ON fd.id = fdc.fee_decision_id 
-  ${if (startingFrom != null) " WHERE fd.sent_at >= :startingFrom" else ""}  
+  ${if (startingFrom != null) " WHERE fd.sent_at >= :startingFrom" else ""}
+  ${if (feeDecisionStatus != null) " WHERE fd.status >= :feeDecisionStatus" else ""}
```
hruoho, line 649:
```suggestion
  ${if (feeDecisionStatus != null) " WHERE fd.status = :feeDecisionStatus" else ""}  
```

hruoho: Maybe we should also add a check that not both params `feeDecisionStatus` and `voucherValueDecisionStatus` are null? Eg something like `check(feeDecisionStatus != null || voucherValueDecisionStatus != null)`

hruoho (approved):

PR #1119, opened 2021-07-05
EVAKA-4203 Edit vasu template

kimifoud:
#### Summary

- Allow editing name, language and date range of a template
- Validate changes


service/src/main/kotlin/fi/espoo/evaka/vasu/VasuTemplateController.kt
```diff
@@ -44,10 +45,28 @@ class VasuTemplateController {
...
+        @PathVariable id: UUID,
+        @RequestBody body: VasuTemplateUpdate
+    ) {
+        Audit.VasuTemplateEdit.log()
+        user.requireOneOfRoles(UserRole.ADMIN)
+
+        db.transaction { tx ->
+            val template = tx.getVasuTemplateForUpdate(id) ?: throw NotFound("Template not found")
+            validateTemplateUpdate(template, body)
+            tx.updateVasuTemplate(id, body)
```
Joosakur, line 61: I'm a bit unsure if editing name/language is fine. Should that be allowed only when document count is zero?

service/src/integrationTest/kotlin/fi/espoo/evaka/vasu/VasuTemplateTest.kt
```diff
@@ -0,0 +1,134 @@
...
+// SPDX-License-Identifier: LGPL-2.1-or-later
+
+package fi.espoo.evaka.vasu
+
+import fi.espoo.evaka.shared.domain.BadRequest
+import fi.espoo.evaka.shared.domain.FiniteDateRange
+import fi.espoo.evaka.shared.domain.HelsinkiDateTime
+import org.junit.jupiter.api.Test
+import org.junit.jupiter.api.assertDoesNotThrow
+import org.junit.jupiter.api.assertThrows
```
Joosakur, line 12: These will fail after rebase. Maybe we should also put my integration test PR to master first and then merge these tests in.

service/src/main/kotlin/fi/espoo/evaka/vasu/VasuTemplateController.kt
kimifoud: Yes, maybe. We should probably also add some kind of a "publish state" enum which would indicate that the template is ready for use. Otherwise we might end up with bogus templates when someone starts using a template that is actually not ready to be used yet.

service/src/integrationTest/kotlin/fi/espoo/evaka/vasu/VasuTemplateQueriesTest.kt
```diff
@@ -0,0 +1,71 @@
...
+    @AfterEach
+    fun afterEach() {
+        db.transaction { tx ->
+            tx.resetDatabase()
+        }
+    }
+
+    @Test
+    fun `vasu template can be created and updated`() {
+        assertDoesNotThrow {
```
Joosakur, line 27: I think this function is unnecessary. Test fails by default if it throws.

service/src/main/kotlin/fi/espoo/evaka/vasu/VasuTemplateController.kt
Joosakur: Hmm, well I think in practise the start date is 1.8. and it will be finished before that, so might not be a problem.

service/src/integrationTest/kotlin/fi/espoo/evaka/vasu/VasuTemplateQueriesTest.kt
kimifoud: Right, it's unnecessary here. I added them for explicitness in the unit tests and they were left over from test file copy-paste 😄

Joosakur: also kotlin.test does not have it :D `assertThrows` is there as `assertFails`, IIRC

kimifoud: Refactored.

service/src/main/kotlin/fi/espoo/evaka/vasu/VasuTemplateController.kt
kimifoud: Name and language are now noneditable if the template is in use.

Joosakur (approved):
🏅

PR #1120, opened 2021-07-05
EVAKA-TECH: type safe assertions

Joosakur:
#### Summary

- add kotlin.test dependencies
- replace junit assertions with kotlin.test assertions (makes compiler more intelligent)
- add lint rule to disallow junit assertions


service-lib/build.gradle.kts
```diff
@@ -23,6 +23,8 @@ dependencies {
 
     // Kotlin + core
     implementation(kotlin("stdlib-jdk8"))
+    api(kotlin("test"))
```
Gekkio, line 26: These sound like test-only dependencies, but `api` in Gradle means the deps are a core part of the API of the project and must be included in order to run our code. In practice this means the dependencies and all their transitive dependencies will be bundled in the docker image, and will be in all classpaths (not just tests).

Instead of using `api` here, add them to `testImplementation` in service-lib and to both `testImplementation` and `integrationTestImplementation` in the service projects

Joosakur: done (service-lib did not have integrationTestImplementation available)

Gekkio (approved):

PR #1121, opened 2021-07-06
Fix broken weak citizen user JSON serialization/deserialization + JWT serialization

Gekkio:
#### Summary

- drop support for legacy JWT/JSON formats, which haven't been used since April 2021
- add missing weak citizen serialization/deserialization
- use an enum "tag" instead of a magic string: if a new variant is added, we get a nice compilation error that shows where the new variant needs to be handled


service/src/main/kotlin/fi/espoo/evaka/shared/auth/JwtTokens.kt
```diff
@@ -28,37 +28,23 @@ fun DecodedJWT.toAuthenticatedUser(): AuthenticatedUser? = this.subject?.let { s
...
-        "evaka_type",
-        when (this) {
-            is AuthenticatedUser.SystemInternalUser -> "system"
-            is AuthenticatedUser.MobileDevice -> "mobile"
-            is AuthenticatedUser.Employee -> "employee"
-            is AuthenticatedUser.Citizen -> "citizen"
-            is AuthenticatedUser.WeakCitizen -> "citizen_weak"
+    .withClaim("evaka_type", this.type.toString())
+    .also {
+        if (this is AuthenticatedUser.Employee) {
```
mikkopiu, line 44: Why don't other users have their roles attached?

Gekkio: Because none of the other user types have configurable roles and the "roles" are hard-coded based on the user type. The only reason those special roles even exist is legacy reasons where we needed to have one unified big role array for permission checks.
Once we implement action-based permissions, `END_USER`/`CITIZEN_WEAK`/`MOBILE_DEVICE` roles can be removed completely because they no longer have a purpose.

mikkopiu (approved):

PR #1122, opened 2021-07-06
Clarify and refactor group placement query

Gekkio:
#### Summary

- remove broken check. It never worked, because a daycare placement ID was passed where a daycare ID was expected
- query used in two places with different requirements -> refactor to two simple queries instead


kimifoud (approved):
LGTM 👍

PR #1128, opened 2021-07-07
PoC for running placement optimization in evaka

Joosakur:
#### Summary
This won't be much of use yet but no harm in merging either


apigw/src/shared/config.ts
```diff
@@ -99,6 +99,10 @@ export const evakaServiceUrl = required(
   process.env.EVAKA_SERVICE_URL ??
     ifNodeEnv(['local', 'test'], 'http://localhost:8888')
 )
+export const evakaAIUrl = required(
+  process.env.EVAKA_AI_URL ??
+    ifNodeEnv(['local', 'test'], 'http://localhost:8889')
+)
```
mikkopiu, line 105: I wouldn't make this required as Tampere won't be doing this testing and we might want to disable it at times.

mikkopiu (changes requested):

apigw/src/shared/config.ts
Joosakur: indeed!

Joosakur: like this? https://github.com/espoon-voltti/evaka/pull/1128/commits/743d7fca400161040a7f483bae370f653fc4701f

mikkopiu (approved):

PR #1138, opened 2021-07-09
Refactor react-select usages to Combobox

kimifoud:
#### Summary

Replace some react-selects with Combobox. Multi-select needs its own component.

Combine `VtjPersonSearch` and `PersonSearch` because they shared 90% of the functionality.


frontend/src/employee-frontend/components/common/PersonSearch.tsx
```diff
@@ -104,27 +102,56 @@ function PersonSearch({
...
+    }: PersonDetails): string =>
+      `${formatName(firstName, lastName, i18n)} (${dateOfBirth.format()})${
+        streetAddress ? `\n${streetAddress}` : ''
+      }`,
+    [i18n]
+  )
+
+  const filterItems = useCallback(
+    (_, items: PersonDetails[]) =>
+      searchFrom === 'vtj'
```
mikkopiu, line 126: Mainly thinking out loud: instead of a bunch of conditionals littered in a single component (with the risk of forgetting them in some branch), would it be better to have:

- `AbstractPersonSearch`, which takes in everything that is different between the implementations as parameters (e.g. the search method)
- `DbPersonSearch` (awful name, I know), which has no elements of its own but just passes the correct parameters to `AbstractPersonSearch`
- `VtjPersonSearch`, which does the the same but with VTJ methods/parameters

kimifoud: Thank you for thinking out loud! It turned out much better, imo.

frontend/src/employee-frontend/components/child-information/placements/CreatePlacementModal.tsx
```diff
@@ -32,10 +32,9 @@ export interface Props {
 
 interface Form {
   type: PlacementType
-  unitId: UUID | undefined
   startDate: LocalDate
   endDate: LocalDate
-  ghostUnit: boolean
+  unit: { id: string; name: string; isGhost: boolean } | null
```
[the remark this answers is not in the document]
kimifoud: Reverted to the original `ghostUnit: boolean` naming which seems to be used everywhere.

frontend/src/employee-frontend/components/common/PersonSearch.tsx
mikkopiu: Excellent stuff

mikkopiu (approved):

PR #1139, opened 2021-07-09
EVAKA-4000 handle people without ssn

hruoho:
#### Summary

Each person that will be sent to varda should have either SSN or OID in our DB. If a person doesn't have SSN in DVV they are first manually inserted into Varda and their OID then manually updated to eVaka. Then we can use their OID as an identifier when communicating with Varda.

### DIFF
--- a/service/src/main/kotlin/fi/espoo/evaka/varda/VardaUpdateServiceV2.kt
+++ b/service/src/main/kotlin/fi/espoo/evaka/varda/VardaUpdateServiceV2.kt
@@ -423,7 +423,7 @@ fun getChildVardaGuardians(db: Database.Connection, childId: UUID): List<VardaGu
 }
 
 fun getVardaGuardian(client: VardaClient, ssn: String?, oid: String?): VardaGuardian {
-    val person = client.getPersonFromVardaByOidOrSSN(ssn, oid) ?: error("VardaUpdate: Couldn't fetch guardian from Varda")
+    val person = client.getPersonFromVardaBySsnOrOid(ssn, oid) ?: error("VardaUpdate: Couldn't fetch guardian from Varda")
     return VardaGuardian(
         henkilotunnus = ssn,
         henkilo_oid = person.personOid,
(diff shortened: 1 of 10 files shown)


service/src/main/kotlin/fi/espoo/evaka/varda/VardaUpdateServiceV2.kt
```diff
@@ -370,153 +375,159 @@ fun addServiceNeedDataToVarda(db: Database.Connection, vardaClient: VardaClient,
...
+    )
+}
+
+fun getFeeReceiverGuardians(db: Database.Connection, childId: UUID, headOfFamilyId: UUID): List<VardaGuardianWithId> {
+    val guardians = getChildVardaGuardians(db, childId)
+    if (guardians.isEmpty()) error("VardaUpdate: could not create fee data: child has no guardians")
+
+    // If head of family is not a guardian, fee data is not supposed to be sent to varda (https://wiki.eduuni.fi/display/OPHPALV/Huoltajan+tiedot)
+    if (guardians.none { it.id == headOfFamilyId }) {
+        error("VardaUpdate: could not create fee data: head of family is not a guardian")
```
mikkopiu, line 478: Is this actually an error? I mean I understand the control flow here pretty much currently requires it to be but purely from a business logic POV -- isn't this just data that shouldn't be sent (maybe worth info-level logging, though)? Or am I misunderstanding the linked docs?

Maybe what I'm really asking is (assuming this shouldn't be an actual error): should these cases be filtered out earlier in the flow?

mikkopiu (changes requested):

service/src/main/kotlin/fi/espoo/evaka/varda/VardaUpdateServiceV2.kt
hruoho: That's a good question. I think we would want to get notified on these at least on some level. But on the other hand, apart from perhaps a bunch of cases when these are indications of bad data, logging these as exceptions is probably just noise.

service/src/main/kotlin/fi/espoo/evaka/varda/integration/VardaClient.kt
```diff
@@ -136,17 +136,20 @@ class VardaClient(
...
 
-    fun getPersonFromVardaBySSN(ssn: String): VardaPersonResponse? {
+    fun getPersonFromVardaBySsnOrOid(ssn: String?, oid: String?): VardaPersonResponse? {
+        check(ssn != null || oid != null) {
+            "Both params ssn and oid shouln't be null"
+        }
+
         logger.info { "Fetching person from Varda" }
-        data class PersonSearchRequest(val henkilotunnus: String)
+        data class PersonSearchRequest(val henkilotunnus: String?, val henkilo_oid: String?)
```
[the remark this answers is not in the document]
hruoho: You are right, I think that we should use an anonymous object instead

service/src/main/kotlin/fi/espoo/evaka/varda/VardaUpdateServiceV2.kt
mikkopiu: I'm OK with keeping these as errors for now (assuming there's no other clear solution) and just keeping it in mind, and maybe later removing it once all data is in a more acceptable condition.

service/src/main/kotlin/fi/espoo/evaka/varda/integration/VardaClient.kt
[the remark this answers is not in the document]
hruoho: Hmm decided to lift this data class up after all and use it as function param type

service/src/main/kotlin/fi/espoo/evaka/varda/VardaUpdateServiceV2.kt
hruoho: But considering that the data is handled childwise I think we should do the filtering in a very early step, before the actual processing of service needs.

hruoho: Also if we did that, then this exception would be justified because each processed child should have a guardian as their head of family

mikkopiu: Agreed to do the pre-filtering later and merge this as-is

mikkopiu (approved):

PR #1140, opened 2021-07-09
[EVAKA] Self-host Montserrat and Open Sans

mikkopiu:
#### Summary
- Frontend: replace Google Fonts CDN with self-hosted assets
    - Instead of relying on the Google Fonts CDN, self-host Montserrat and Open Sans (only the versions actually in use)
    - As Google officially only provides `.ttf` variants of its fonts as a download (other formats are provided via CDN), these have been downloaded from the unofficial but excellent https://google-webfonts-helper.herokuapp.com
    - **NOTE:** For now, we decided with UX that we won't allow customization before someone actually requests it
        - No need to require treVaka to copy ~300 font files now
        - It should be pretty straightforward based on my "reverted" commits in this branch to re-create later
    - All of our officially supported browsers support WOFF and/or WOFF2, so only these variants are included (everything that supports TLS 1.2, supports at least WOFF)
        - Previously, Google CDN dynamically decided which variants it provided based on request metadata
    - For Storybook, it didn't make sense to go through the configuration effort, so just use the newer CDN version (CSS2)
    - As maintenance-page is Espoo-specific (and a Parcel app for now), just explicitly load the Espoo version of the customized assets (like with the favicon)
- Also **actually** loads italic variants and correct weights of both Montserrat and Open Sans instead of using browser-rendered versions
    - For Montserrat, only Regular (400) gets an italic variant as no other weights use/should use italics
    - For Open Sans, all weights get italic variants
- Proxy: don't allow Google Fonts CDN in CSP
- Also add some missing license headers

##### Font diff

Previously (ignore icons, free vs. pro):
Screenshot 2021-07-14 at 13-24-46 Hakemus - Testinen Testi - Varhaiskasvatus

New (ignore icons):
Screenshot 2021-07-14 at 13-24-36 Hakemus - Karhula Jari-Petteri Mukkelis-Makkelis Vetelä-Viljami Eelis-Juhani - Varhaiskas

- Tighter italics
- Actually using Montserrat Semi-bold (600) instead of Bold (700) browser-adjusted to 600


proxy/files/etc/nginx/conf.d/nginx.conf.template
```diff
@@ -144,7 +144,7 @@ server {
   add_header X-DNS-Prefetch-Control off;
   add_header Report-To '{"group","csp-endpoint","max_age":31536000,"endpoints":[{"url":"https://$host/api/csp"}]}';
 
-  set $contentSecurityPolicyBase "block-all-mixed-content; upgrade-insecure-requests; default-src 'self'; font-src 'self' data: https://fonts.gstatic.com; style-src 'self' 'unsafe-inline' https://fonts.googleapis.com; img-src 'self' data: https://maps.gstatic.com https://maps.googleapis.com https://a.tile.openstreetmap.org https://b.tile.openstreetmap.org https://c.tile.openstreetmap.org; connect-src 'self' https://o318158.ingest.sentry.io https://api.digitransit.fi; object-src 'none'; report-uri /api/csp; report-to csp-endpoint";
+  set $contentSecurityPolicyBase "block-all-mixed-content; upgrade-insecure-requests; default-src 'self'; font-src 'self' data:; style-src 'self' 'unsafe-inline'; img-src 'self' data: https://maps.gstatic.com https://maps.googleapis.com https://a.tile.openstreetmap.org https://b.tile.openstreetmap.org https://c.tile.openstreetmap.org; connect-src 'self' https://o318158.ingest.sentry.io https://api.digitransit.fi; object-src 'none'; report-uri /api/csp; report-to csp-endpoint";
```
mikkopiu, line 147: Is Google Maps still in use somewhere, or should everything related to it be dropped?

hruoho (approved):

proxy/files/etc/nginx/conf.d/nginx.conf.template
mikkopiu: It is not

PR #1143, opened 2021-07-12
Refactor Select component to use Combobox instead of react-select

kimifoud:
#### Summary

Refactor `Select` component to use `Combobox` instead of `react-select`.


hruoho (approved):
🏅

PR #1150, opened 2021-07-13
[EVAKA] Frontend: make all feature flags and configs customizable

mikkopiu:
#### Summary

**IMPORTANT:** This brings quite a lot of changes into customization types so forks will need to be updated accordingly **but** now there's a lot more to customize and it's all in a single place :tada:

- Frontend: make all feature flags customizable
    - Instead of having separate feature flags "internally" in the core and some in `lib-customizations`, make all feature flags customizable
        - One downside here is that if flags are set in a runtime-dependent way (e.g. environment-specific like in the `espoo` theme), modules and their dependencies cannot be dropped from builds via tree shaking
    - There are basically two types of feature flags currently in eVaka:
        1. Production ready features that some implementations might not want to enable
        2. Features in development (now under: `experimental.*`) that can be enabled in implementations in order to test e.g. in non-production environments
            - These were un-customizable previously and were generally the way things have been tested in Espoo while still merging changes to `master` constantly -- now they're just in a different place **and** can even be tested by other implementations without having to wait for production-readiness
    - This also allows us to drop Espoo-specific assumptions about environments and their relations to URLs/domains/etc in relation to feature flags as `lib-customizations` will only export a single interface of flags and any environment specific logic is left up to the implementations
        - Espoo's environment specific configs are now located in the "espoo" theme and provided as an example for other implementations
    - Also make feature flags read-only with
    - **NOTE**: Citizen and employee both now obey a single `messaging` feature flag
        - As discussed, makes no sense to disable just one side of this
- Frontend: replace isNotProduction with a feature flag
    - Add new flag: `experimental.mobileDailyNotes`
        - Not sure if that's the best name but from the contexts it was used in, that's what it looked like
    - Decided to make the flag experimental as previously the feature was explicitly checked to be disabled in production-like environments (staging & prod)
    - NOTE: As documented, the flag can be short-circuited in some places through "isPilotUnit" (or equivalent) checks
- Frontend: make all configs customizable
    - I.e. Sentry configs
        - As Sentry is entirely optional, match this in config types
    - Create new customization module `employeeMobile` so that all frontend apps have their own configs
    - Separate from feature flags as configs can be app (+implementation) specific and feature flags generally span all apps per implementation
    - Drop the unused history functions (legacy stuff, hasn't been used in a while)
- Remove unused `attachments` feature flag
    - Introduced but never actually used
- Documentation for frontend configuration

##### Update instructions for forks

In your theme (highly recommend looking at the examples in the `espoo` theme):

- Create `employeeMobile.tsx` that exports `EmployeeMobileCustomizations` typed object
    - **NOTE:** As `sentry` is fully optional, your `appConfig` can just be `{}` if you aren't using Sentry
- Add `appConfig` to your `citizen.tsx` and `employee.tsx` modules
- Set and export all the new feature flags:
    - `evakaLogin: boolean`
    - `financeBasicsPage: boolean`
    - `messaging: boolean`
    - `vasu: boolean`
    - `voucher.valueDecisionsPage: boolean`
    - _Optionally_: `experimental.mobileDailyNotes: boolean`
    - `experimental.ai` ei **very** experimental right now and not recommended to be activated (can be skipped as its type is optional)
- If you need per-environment _feature flag configs_:
    1. Create some method/module for finding out your environment in runtime (e.g. `espoo/env.ts`)
    2. Use that method to split your feature flags per environment dynamically (e.g. `espoo/featureFlags.tsx`)
- If you were using the previous `configs/index` or `configs` exports somewhere in your customizations, replace with imports from the new location under customizations:

    ```diff
    - import { featureFlags } from '../config'
    + import { featureFlags } from 'lib-customizations/employee' // or citizen or employeeMobile
    ```


kimifoud (approved):
Nice! 👍 Is `isNotProduction()` now redundant too?

mikkopiu:
> Nice! +1 Is `isNotProduction()` now redundant too?

:D Totally forgot about cleaning that up, got so wrapped in the bigger picture :facepalm: I'll do that now...

mikkopiu:
Created a new flag: `experimental.mobileDailyNotes` (that's what it sounded like based on where `isNotProduction()` was used + commit messages) to replace `isNotProduction()`: enabled by default and disabled in staging+prod -> enabled in non-production environments like before and disabled for **non-pilot-units** in staging+prod (the flag is short circuited by `isPilotUnit()` checks if it is falsy).

hruoho (approved):

PR #1151, opened 2021-07-13
EVAKA-4203 Validate vasu content when saving

kimifoud:
#### Summary

- Prevent updating template content if template is already in use 
- Validate that vasu content structurally matches the template content when saving


hruoho (approved):
Nice 👍

PR #1159, opened 2021-07-19
EVAKA-HOTFIX handle application state changes in placement and decision draft views

jaahon:
#### Summary

Redirect to the applications list on application state conflict. If the placement plan or decision draft views were opened in separate tabs and the application's state has changed meanwhile, the views would either appear to work (placement plans) or just show a crude error message (decision draft).


hruoho (approved):
🙏

PR #1162, opened 2021-07-20
Rewrite messaging tests with Playwright

jaahon:
#### Summary

The new tests verify all the same things even if they are not completely identical.


frontend/src/e2e-playwright/pages/employee/messages/messages-page.ts
```diff
@@ -53,4 +75,52 @@ export default class MessagesPage {
...
+
+  async draftNewMessage(title: string, content: string) {
+    await this.#newMessageButton.click()
+    await waitUntilEqual(() => this.isEditorVisible(), true)
+    await this.#inputTitle.fill(title)
+    await this.#inputContent.fill(content)
+    await this.#receiverSelection.click()
+    await this.page.keyboard.press('Enter')
+    // Draft is not saved immediately so we wait for a bit.
+    await this.page.waitForTimeout(500)
```
mikkopiu, line 87: Is there any indicator on the page that we could use? Or maybe something disappears that existed before pressing Enter; the editor itself? Random sleeps are always a bit iffy.

jaahon: There isn't, but it seems like it wouldn't be a huge task to add some kind of indicator. I'll do that.

mikkopiu (approved):
:+1:

PR #1163, opened 2021-07-21
Do inactive people query in one database query

jaahon:
#### Summary

Querying one at a time took 7.5 hours in production even if the load was minimal.


mikkopiu (approved):
GLHF

PR #1164, opened 2021-07-22
Rewrite finance e2e tests with Playwright

jaahon:
#### Summary

The tests aren't completely identical, I added more assertions and removed some redundant tests.


frontend/src/e2e-playwright/specs/4_finance/fee-decisions.spec.ts
```diff
@@ -0,0 +1,82 @@
...
+      fixtures.daycareFixture.id
+    )
+  ])
+
+  page = await (await newBrowserContext({ acceptDownloads: true })).newPage()
+
+  await insertEmployeeFixture({
+    id: config.financeAdminAad,
+    externalId: `espoo-ad:${config.financeAdminAad}`,
+    email: 'lasse.laskuttaja@espoo.fi',
```
mikkopiu, line 51: Supernit: I guess this doesn't need to a real address -> I'd prefer something definitely non-existent like `billy.biller@billing.test` (`.test` will [never resolve](https://www.rfc-editor.org/rfc/rfc6761.html#section-6.2)), just to avoid the unnecessary risk of extra spam to espoo.fi

frontend/src/e2e-playwright/specs/4_finance/invoices.spec.ts
```diff
@@ -0,0 +1,155 @@
...
+      feeDecision.validDuring.end?.formatIso()
+    )
+  ])
+
+  page = await (await newBrowserContext({ acceptDownloads: true })).newPage()
+
+  await insertEmployeeFixture({
+    id: config.financeAdminAad,
+    externalId: `espoo-ad:${config.financeAdminAad}`,
+    email: 'lasse.laskuttaja@espoo.fi',
```
mikkopiu, line 75: Same nit as above

frontend/src/e2e-playwright/specs/4_finance/invoices.spec.ts
```diff
@@ -0,0 +1,155 @@
...
+    await invoicesPage.navigateBackToInvoices()
+  })
+
+  test('Add a new invoice row, modify its amount and price and persist the changes', async () => {
+    await invoicesPage.createInvoiceDrafts()
+    await invoicesPage.openFirstInvoice()
+    await invoicesPage.assertInvoiceRowCount(1)
+    await invoicesPage.addNewInvoiceRow('12345', 10, 100)
+    await invoicesPage.navigateBackToInvoices()
+    await invoicesPage.assertInvoiceTotal(289 + 10 * 100)
```
mikkopiu, line 113: What's the origin of 289 here? Or rather: if it's something meaningful (like a base free etc), this would be more understandable as a variable / reference to another variable.

frontend/src/e2e-playwright/specs/4_finance/invoices.spec.ts
```diff
@@ -0,0 +1,155 @@
...
+  })
+
+  test('Invoices are toggled and sent', async () => {
+    await invoicesPage.createInvoiceDrafts()
+    await invoicesPage.toggleAllInvoices(true)
+    await invoicesPage.sendInvoices()
+    await invoicesPage.assertInvoiceCount(0)
+    await invoicesPage.showSentInvoices()
+    await invoicesPage.assertInvoiceCount(1)
+  })
```
mikkopiu, line 131: Would it be too much to ask for more than one invoice here to ensure the toggle _all_ actually works?

frontend/src/e2e-playwright/specs/4_finance/value-decisions.spec.ts
```diff
@@ -57,48 +61,67 @@ beforeEach(async () => {
   ])
 
   page = await (await newBrowserContext({ acceptDownloads: true })).newPage()
-  await employeeLogin(page, 'ADMIN')
+
+  await insertEmployeeFixture({
+    id: config.financeAdminAad,
+    externalId: `espoo-ad:${config.financeAdminAad}`,
+    email: 'lasse.laskuttaja@espoo.fi',
```
mikkopiu, line 68: Same as above

mikkopiu (changes requested):
Mainly nits about the actual tests and not so much about the rewrite, so take with a grain of a spice of your choosing.

frontend/src/e2e-playwright/specs/4_finance/fee-decisions.spec.ts
```diff
@@ -0,0 +1,82 @@
...
+      fixtures.daycareFixture.id
+    )
+  ])
+
+  page = await (await newBrowserContext({ acceptDownloads: true })).newPage()
+
+  await insertEmployeeFixture({
+    id: config.financeAdminAad,
+    externalId: `espoo-ad:${config.financeAdminAad}`,
+    email: 'lasse.laskuttaja@test.com',
```
mikkopiu, line 51: Actually, it's `.test` as in `.com` (the spec can be a bit confusing as it's written as `.test.` but that's just the canonical way of writing domains; the last dot is just generally hidden/considered OK to leave out -- `example.com` and `example.com.` are equivalent):
```suggestion
    email: 'lasse.laskuttaja@whatever-can-go-here-and-should-never-resolve.test',
```

mikkopiu (changes requested):
One more adjustment, otherwise LGTM

frontend/src/e2e-playwright/specs/4_finance/fee-decisions.spec.ts
jaahon: Alright, should read more carefully 😄

mikkopiu: Hmmh, looks like these addresses are littered in quite a few places (e.g. dev-data SQL) -- I'm OK with leaving this as tech debt if it looks to be too much effort to update everything.

jaahon: Yeah I won't do completely now, doesn't seem like a top priority right now.

mikkopiu (approved):

PR #1167, opened 2021-07-22
Remove duplicate Chip component

jaahon:
#### Summary

Had to replace the `DataList` component as well since it applied styled to all `<label>` elements.


mikkopiu (approved):
LGTM

PR #1168, opened 2021-07-22
EVAKA inactive people clean up

jaahon:
#### Summary

Add `absence` and `child_attendance` tables to the archive data check to be more explicit since attendance data should be archived along with placement data.
Make all the references to `person` and `child` tables either `ON DELETE CASCADE` if the data doesn't have to be archived or `ON DELETE RESTRICT` if the data should be archived to make automatic data clean up easier.

Try to fix a new e2e test's flakyness.


mikkopiu (approved):
Don't really know of a good way to test/verify this but at least manually skimming through the re-created constraints seemed to match the originals (where appropriate), so I guess :heavy_check_mark:

PR #1183, opened 2021-08-02
Make voucher value age coefficient configurable

jaahon:
#### Summary

The migration sets the value as `1.55` for all existing rows, which is Espoo's coefficient.


Gekkio (approved):
:gem:

PR #1184, opened 2021-08-02
EVAKA-4208 remove roles from inactive employees

jaahon:
#### Summary

Add a scheduled job that removes all roles from employees that haven't logged in last 3 months. For employees that do not have `last_login` data, use 1.6.2021 as a default since `last_login` has been tracked since May.

### DIFF
--- a/service/src/integrationTest/kotlin/fi/espoo/evaka/pis/InactiveEmployeesRoleResetIntegrationTest.kt
+++ b/service/src/integrationTest/kotlin/fi/espoo/evaka/pis/InactiveEmployeesRoleResetIntegrationTest.kt
@@ -1,0 +1,150 @@
+package fi.espoo.evaka.pis
+
+import fi.espoo.evaka.PureJdbiTest
+import fi.espoo.evaka.resetDatabase
+import fi.espoo.evaka.shared.auth.UserRole
+import fi.espoo.evaka.shared.auth.insertDaycareAclRow
+import fi.espoo.evaka.shared.dev.DevCareArea
+import fi.espoo.evaka.shared.dev.DevDaycare
+import fi.espoo.evaka.shared.dev.DevEmployee
+import fi.espoo.evaka.shared.dev.insertTestCareArea
+import fi.espoo.evaka.shared.dev.insertTestDaycare
+import fi.espoo.evaka.shared.dev.insertTestEmployee
+import fi.espoo.evaka.shared.domain.HelsinkiDateTime
+import org.junit.jupiter.api.BeforeEach
+import org.junit.jupiter.api.Test
+import java.time.LocalDate
+import java.time.LocalTime
+import java.time.temporal.ChronoUnit
+import kotlin.test.assertEquals
+import kotlin.test.assertTrue
+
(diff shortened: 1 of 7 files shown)


service/src/integrationTest/kotlin/fi/espoo/evaka/pis/InactiveEmployeesRoleResetIntegrationTest.kt
```diff
@@ -0,0 +1,150 @@
...
+import org.junit.jupiter.api.BeforeEach
+import org.junit.jupiter.api.Test
+import java.time.LocalDateTime
+import java.time.ZonedDateTime
+import java.time.temporal.ChronoUnit
+import kotlin.test.assertEquals
+import kotlin.test.assertTrue
+
+class InactiveEmployeesRoleResetIntegrationTest : PureJdbiTest() {
+    private val FirstOfAugust2021 = ZonedDateTime.of(LocalDateTime.of(2021, 8, 1, 3, 15), helsinkiZone).toInstant()
```
Gekkio, line 23: `private val FirstOfAugust2021` vs `private val firstOfAugust2021`?

service/src/main/kotlin/fi/espoo/evaka/pis/EmployeeQueries.kt
```diff
@@ -298,3 +302,24 @@ fun Database.Read.isPinLocked(employeeId: UUID): Boolean =
...
+    WHERE coalesce(last_login, :defaultLastLogin) < :now::timestamp with time zone - interval '3 months'
+    RETURNING id
+), deleted_daycare_acl AS (
+    DELETE FROM daycare_acl WHERE employee_id IN (SELECT id FROM reset_employee)
+)
+DELETE FROM daycare_group_acl WHERE employee_id IN (SELECT id FROM reset_employee)
+"""
+)
+    .bind("now", now)
+    .bind("defaultLastLogin", ZonedDateTime.of(LocalDateTime.of(2021, 6, 1, 0, 0), helsinkiZone).toInstant())
```
Gekkio, line 324: Would `HelsinkiDateTime.of(LocalDate.of(2021, 6, 1), LocalTime.of(0, 0))` be simpler? No need to manually fiddle with zones or cast to Instant, since it works with JDBI correctly

Gekkio (approved):

service/src/integrationTest/kotlin/fi/espoo/evaka/pis/InactiveEmployeesRoleResetIntegrationTest.kt
jaahon: Thanks, I was so focused on capitalizing the months that I accidentally capitalized everything 😄

service/src/main/kotlin/fi/espoo/evaka/pis/EmployeeQueries.kt
jaahon: Yeah, that's a good idea. I've never used it so it didn't even cross my mind.

PR #1191, opened 2021-08-04
EVAKA-4203: prevent changes to vasu template content once it is in use

Joosakur:
#### Summary
readonly mode in frontend for used vasu templates

### DIFF
--- a/frontend/src/employee-frontend/components/vasu/templates/VasuTemplateEditor.tsx
+++ b/frontend/src/employee-frontend/components/vasu/templates/VasuTemplateEditor.tsx
@@ -62,6 +62,8 @@ export default React.memo(function VasuTemplateEditor() {
   useEffect(() => loadTemplate(id), [id, loadTemplate])
   useWarnOnUnsavedChanges(dirty, i18n.vasuTemplates.unsavedWarning)
 
+  const readonly = !(template.isSuccess && template.value.documentCount === 0)
+
   function moveSection(index: number, dir: 'up' | 'down') {
     setDirty(true)
     setTemplate((res) =>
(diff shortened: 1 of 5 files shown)


frontend/src/employee-frontend/components/vasu/templates/VasuTemplateEditor.tsx
```diff
@@ -423,6 +435,7 @@ export default React.memo(function VasuTemplateEditor() {
                               }
                               text={i18n.vasuTemplates.addNewQuestion}
                               icon={faPlus}
+                              disabled={readonly}
```
jaahon, line 438: Would it make more sense to drop these buttons completely if the template is readonly?

jaahon (commented):
I think backend should check that there aren't any documents created off the template when the edit is saved.

frontend/src/employee-frontend/components/vasu/templates/VasuTemplateEditor.tsx
Joosakur: https://github.com/espoon-voltti/evaka/blob/60f8035d0c6ec607f6465ad807a8e3bdb577a648/service/src/main/kotlin/fi/espoo/evaka/vasu/VasuTemplateController.kt#L149 backend validation actually is there already. You were probably looking at the other PUT endpoint which updates e.g. template date range and allows limited edits even after template is in use.

jaahon: Sorry, my bad.

jaahon (approved):

PR #1195, opened 2021-08-05
EVAKA-TECH: use action based access control - part 2/x

Joosakur:


service/src/main/kotlin/fi/espoo/evaka/shared/security/AccessControl.kt
```diff
@@ -88,11 +100,16 @@ class AccessControl(private val permittedRoleActions: PermittedRoleActions, priv
...
-    fun requirePermissionFor(user: AuthenticatedUser, action: Action.VasuDocument, id: UUID) {
+    fun requirePermissionFor(user: AuthenticatedUser, action: Action.VasuDocument, id: VasuDocumentId) {
         val roles = acl.getRolesForVasuDocument(user, id).roles
         assertPermission(roles, action, permittedRoleActions::vasuDocumentActions)
     }
 
+    fun requirePermissionFor(user: AuthenticatedUser, action: Action.VasuTemplate, @Suppress("UNUSED_PARAMETER") id: VasuTemplateId) {
+        // VasuTemplate actions in Espoo are global so the id parameter is ignored
+        assertPermission(user.roles, action, permittedRoleActions::vasuTemplateActions)
+    }
```
jaahon, line 111: I must admit I'm not very familiar with the principles of this new access control code. If the actions are "global" why not put them under `Action.Global`? Or at least remove the unused parameter from this function.

jaahon (commented):
The backup pickup endpoints aren't backwards compatible. It doesn't seem like a lot of work to leave compatible endpoints that can be cleaned up later when we know no older versions of frontend exist. Should we do that so we don't have to break evaka temporarily?

service/src/main/kotlin/fi/espoo/evaka/shared/security/AccessControl.kt
Joosakur: They are currently global for Espoo but may not be for other cities, so the idea is that they can override this function and control it more tightly as long as the real scope and the target id is available. I guess the unused parameter looks more weird now since this hasn't yet been placed behind an interface.

jaahon: Why are `CREATE_VASU_TEMPLATE` and `READ_VASU_TEMPLATE` global but these are not? 

Until we actually have an interface that this function must adhere to I don't see a reason to add dead code.

I'm just trying to understand this and I'm probably just out of the loop.

Joosakur: Yeah no problem! The difference is that those two (creating template, listing templates) don't have any id related to them, for example in api paths. Thus there's no possible scoped context for them so therefore those actions go under global/root scope. Updating/deleting a specific template however has the id in path and belongs under this specific scope. Similarly listing all applications (search) would belong under `Action.Global`, updating an application belongs under `Action.Application` while listing applications of some child belongs under `Action.Child`.

Regarding dead code, here's what I discussed with gekkio, sorry I did not have it on a public channel:
Screenshot 2021-08-06 at 12 22 02

jaahon (approved):

PR #1209, opened 2021-08-12
EVAKA-4203: misc improvements to Vasu

Joosakur:

### DIFF
--- a/frontend/src/employee-frontend/components/child-information/VasuAndLeops.tsx
+++ b/frontend/src/employee-frontend/components/child-information/VasuAndLeops.tsx
@@ -27,6 +27,7 @@ import { VasuStateChip } from '../common/VasuStateChip'
 import {
   createVasuDocument,
   getVasuDocumentSummaries,
+  updateDocumentState,
   VasuDocumentSummary
 } from '../vasu/api'
 import { getLastPublished } from '../vasu/vasu-events'
@@ -34,6 +35,7 @@ import {
   getVasuTemplateSummaries,
   VasuTemplateSummary
 } from '../vasu/templates/api'
+import { RequireRole } from '../../utils/roles'
 
 const StateCell = styled(Td)`
   display: flex;
@@ -89,7 +91,13 @@ const InitializationContainer = styled.div`
 
 const getValidVasuTemplateSummaries = () => getVasuTemplateSummaries(true)
 
-function VasuInitialization({ childId }: { childId: UUID }) {
+function VasuInitialization({
+  childId,
+  allowCreation
+}: {
+  childId: UUID
+  allowCreation: boolean
+}) {
   const { i18n } = useTranslation()
   const { setErrorMessage } = useContext(UIContext)
   const history = useHistory()
@@ -110,6 +118,7 @@ function VasuInitialization({ childId }: { childId: UUID }) {
             resolveLabel: i18n.common.ok
           })
           setInitializing(false)
+          setTemplates(undefined)
         },
         loading: () => setInitializing(true),
         success: (id) => history.push(`/vasu/${id}`)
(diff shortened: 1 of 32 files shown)


frontend/src/employee-frontend/components/child-information/VasuAndLeops.tsx
```diff
@@ -213,16 +224,34 @@ const VasuAndLeops = React.memo(function VasuAndLeops({
...
+                />
+              )}
             </Td>
           </Tr>
         ))
       )
       .getOrElse(null)
 
+  const allowCreation = vasus.isSuccess
+    ? vasus.value.find((doc) => doc.documentState !== 'CLOSED') === undefined
```
jaahon, line 252: I think `vasus.value.every(doc => doc.documentState === 'CLOSED')` communicates the intent better. 

Also, as more of a fyi, you could always use `vasus.map(...).getOrElse(...)` instead of checking for the state manually.

frontend/src/employee-frontend/components/vasu/templates/VasuTemplatesPage.tsx
```diff
@@ -82,6 +82,7 @@ export default React.memo(function VasuTemplatesPage() {
                         <IconButton
                           icon={faCopy}
                           onClick={() => setTemplateToCopy(template)}
+                          disabled={true} /*needs fixing*/
```
jaahon, line 85: What does this mean?

frontend/src/lib-components/molecules/ExpandingInfo.tsx
```diff
@@ -71,6 +71,8 @@ export default function ExpandingInfo({
   const { colors } = useTheme()
   const [expanded, setExpanded] = useState<boolean>(false)
 
+  if (info === null) return <div>{children}</div>
```
jaahon, line 74: I don't think this kind of optional rendering belongs in this component. Conceptually I don't think it makes any sense so use this component when `info` is `null`. I think conditional rendering should be at the using end.

service/src/main/kotlin/fi/espoo/evaka/vasu/OphQuestion.kt
```diff
@@ -19,32 +19,193 @@ data class OphQuestionOption(
     val name: Map<VasuLanguage, String>
 )
 
-fun getDefaultTemplateContent(lang: VasuLanguage) = VasuContent(
+fun getDefaultTemplateContent(@Suppress("UNUSED_PARAMETER") lang: VasuLanguage) = VasuContent(
```
jaahon, line 22: I'm confused, isn't this annotation completely unnecessary?

Joosakur: ah yes, now it is, will remove

frontend/src/employee-frontend/components/vasu/templates/VasuTemplatesPage.tsx
Joosakur: the template copying doesnt make sense anymore after some changes, going to discuss with Kimi should we remove it or how to fix. I can remove this button for now though instead of disabling

frontend/src/lib-components/molecules/ExpandingInfo.tsx
Joosakur: The conditional rendering of such wrapper component caused tons of annoying boilerplate. I agree this way feels a bit dirty too though.. Should I build some higher order component to do the conditional wrapping?

jaahon: Yeah I would create a new component somewhere around the vasu code that would conditionally wrap other components in the `ExpandingInfo` component.

jaahon (approved):

PR #1210, opened 2021-08-13
Change checkbox label font size from 15 px to 16 px

akheron:
#### Summary

It was changed in the design system already a long time ago.

### DIFF
--- a/frontend/src/lib-components/atoms/form/Checkbox.tsx
+++ b/frontend/src/lib-components/atoms/form/Checkbox.tsx
@@ -18,7 +18,7 @@ const Wrapper = styled.div`
   cursor: pointer;
 
   label {
-    font-size: 15px;
+    font-size: 1rem;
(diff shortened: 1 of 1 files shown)


frontend/src/lib-components/atoms/form/Checkbox.tsx
```diff
@@ -18,7 +18,7 @@ const Wrapper = styled.div`
   cursor: pointer;
 
   label {
-    font-size: 15px;
+    font-size: 16px;
```
jaahon, line 21: I think using `1rem` is a better idea here, because the reason this didn't match the spec is that font size was increased from 15px to 16px at one point.

akheron: Good point! Fixed.

jaahon (approved):

PR #1216, opened 2021-08-16
EVAKA-TECH: functional feature flags

Joosakur:
#### Summary
- rename mobileDailyNotes flag to employeeMobile and promote from experimental
- remove static pilot unit list and move under feature flag functions

### DIFF
--- a/frontend/src/employee-frontend/components/Header.tsx
+++ b/frontend/src/employee-frontend/components/Header.tsx
@@ -79,7 +79,7 @@ const UnreadCount = styled.span`
 const Header = React.memo(function Header({ location }: RouteComponentProps) {
   const { i18n } = useTranslation()
   const { user, loggedIn } = useContext(UserContext)
-  const { hasPilotAccess, accounts } = useContext(MessageContext)
+  const { accounts } = useContext(MessageContext)
   const [popupVisible, setPopupVisible] = useState(false)
 
   const unreadCount = useMemo<number>(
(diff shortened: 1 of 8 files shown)


frontend/src/employee-frontend/components/Header.tsx
```diff
@@ -208,7 +208,7 @@ const Header = React.memo(function Header({ location }: RouteComponentProps) {
               </RequireRole>
 
               {featureFlags.messaging &&
-                (featureFlags.experimental?.mobileDailyNotes ||
+                (featureFlags.experimental?.employeeMobile ||
                   hasPilotAccess) && (
```
jaahon, line 212: You probably forgot the change this conditional, just `hasPilotAccess` should be enough.

Joosakur: Yeah, actually tried a slightly different approach there now.

frontend/src/employee-frontend/components/Header.tsx
```diff
@@ -207,23 +207,21 @@ const Header = React.memo(function Header({ location }: RouteComponentProps) {
...
-                      data-qa="messages-nav"
-                    >
-                      {i18n.header.messages}
-                      {unreadCount > 0 && (
-                        <UnreadCount>{unreadCount}</UnreadCount>
-                      )}
-                    </NavbarLink>
-                  </RequireRole>
-                )}
+              {accounts.isSuccess && accounts.value.length > 0 && (
```
jaahon, line 210: I could be wrong, but doesn't every employee and group have an account?

Joosakur: Yeah need to check with @mapehe, I'm not sure how this should work. The account list excludes group accounts outside pilot units, but if every emploee has a personal account then this won't work.

PR #1217, opened 2021-08-16
Refactor fee decision template address model

patari:
#### Summary

To support multiple address rows in fee decisions.

### DIFF
--- a/service/src/main/kotlin/fi/espoo/evaka/decision/DecisionAddress.kt
+++ b/service/src/main/kotlin/fi/espoo/evaka/decision/DecisionAddress.kt
@@ -4,6 +4,8 @@
 
 package fi.espoo.evaka.decision
 
+import fi.espoo.evaka.invoicing.domain.PersonData
+import fi.espoo.evaka.invoicing.domain.addressUsable
 import fi.espoo.evaka.pis.service.PersonDTO
 import fi.espoo.evaka.shared.message.IMessageProvider
 import fi.espoo.evaka.shared.message.MessageLanguage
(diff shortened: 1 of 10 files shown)


service/src/main/kotlin/fi/espoo/evaka/decision/DecisionAddress.kt
```diff
@@ -46,4 +48,21 @@ data class DecisionSendAddress(
...
+                if (addressUsable(person.streetAddress, person.postalCode, person.postOffice)) {
+                    DecisionSendAddress(
+                        person.streetAddress!!, person.postalCode!!, person.postOffice!!,
+                        person.streetAddress, "${person.postalCode} ${person.postOffice}", ""
+                    )
+                } else {
+                    when (lang) {
+                        "sv" -> messageProvider.getDefaultFeeDecisionAddress(MessageLanguage.SV)
+                        else -> messageProvider.getDefaultFeeDecisionAddress(MessageLanguage.FI)
+                    }
```
jaahon, line 64: I think this default address should be passed to this function or the whole function should be moved someplace else because `DecisionAddress` is not used only with fee decisions and the default address depends on the decision.

jaahon: Just to make it less likely that the wrong default address is used accidentally.

patari: Yeah, something like this https://github.com/espoon-voltti/evaka/pull/1217/commits/cdb895564ea8c47781f77699a6c7e0793eeec91d?

jaahon: Yeah that looks good 👍

jaahon (approved):

PR #1235, opened 2021-08-18
Move from yarn 2 to yarn 1

akheron:
#### Summary

This enables us to use dependabot, which doesn't support yarn 2.

It would be harder to move npm now, because its dependency resolution is stricter and our repo has dependency conflicts that npm cannot resolve.


akheron:
It [seems](https://github.com/dependabot/dependabot-core/issues/1297#issue-478320093) that dependabot actually understands the `yarn.lock` format of yarn 2, and works fine with `nodeLinker: node-modules`, so downgrade to yarn 1 is not needed.

PR #1236, opened 2021-08-19
EVAKA-4240: mobile first calendar

Joosakur:

### DIFF
--- a/frontend/src/citizen-frontend/App.tsx
+++ b/frontend/src/citizen-frontend/App.tsx
@@ -21,6 +21,7 @@ import Header from './header/Header'
 import { Localization } from './localization'
 import MapView from './map/MapView'
 import MessagesPage from './messages/MessagesPage'
+import CalendarPage from './calendar/CalendarPage'
 import { MessageContextProvider } from './messages/state'
 import GlobalErrorDialog from './overlay/Error'
 import GlobalInfoDialog from './overlay/Info'
@@ -72,6 +73,11 @@ export default function App() {
                       exact
                       path="/messages"
                       component={requireAuth(MessagesPage, false)}
+                    />
+                    <Route
+                      exact
+                      path="/calendar"
+                      component={requireAuth(CalendarPage, false)}
                     />
                     <Route path="/" component={RedirectToMap} />
                   </Switch>
--- a/frontend/src/citizen-frontend/calendar/CalendarListView.tsx
+++ b/frontend/src/citizen-frontend/calendar/CalendarListView.tsx
@@ -1,0 +1,113 @@
+// SPDX-FileCopyrightText: 2017-2021 City of Espoo
+//
+// SPDX-License-Identifier: LGPL-2.1-or-later
+
+import React from 'react'
+import {
+  FixedSpaceColumn,
+  FixedSpaceRow
+} from 'lib-components/layout/flex-helpers'
+import LocalDate from 'lib-common/local-date'
+import { useTranslation } from '../localization'
+import styled from 'styled-components'
+import colors from 'lib-customizations/common'
+
+const HistoryOverlay = styled.div`
+  position: absolute;
+  top: 0;
+  left: 0;
+  width: 100%;
+  height: 100%;
+  z-index: 99;
+  opacity: 0.3;
+  background-color: ${colors.blues.lighter};
+`
+
+const DayColumn = styled(FixedSpaceColumn)`
+  width: 48px;
+  color: ${colors.blues.dark};
+  font-weight: 600;
+  font-size: 16px;
+`
+
+const DayDiv = styled(FixedSpaceRow)<{ today: boolean }>`
+  position: relative;
+  padding: 8px 16px;
+  height: 80px;
+  border-bottom: 1px solid ${colors.greyscale.lighter};
+  ${(p) =>
+    p.today
+      ? `
+    border-left: 6px solid ${colors.brandEspoo.espooTurquoise};
+    padding-left: 10px;
+  `
+      : ''}
+`
+
+interface DayProps {
+  date: LocalDate
+}
+
+const DayElem = React.memo(function DayElem({ date }: DayProps) {
+  const i18n = useTranslation()
+
+  return (
+    <DayDiv alignItems="center" today={date.isToday()}>
+      <DayColumn spacing="xxs">
+        <div>
+          {i18n.common.datetime.weekdaysShort[date.getIsoDayOfWeek() - 1]}
+        </div>
+        <div>{date.format('dd.MM.')}</div>
+      </DayColumn>
+      <div>09:00 - 15:30</div>
+      {date.isBefore(LocalDate.today()) && <HistoryOverlay />}
+    </DayDiv>
+  )
+})
+
+const WeekDiv = styled.div`
+  display: flex;
+  justify-content: center;
+  align-items: center;
+  padding: 16px 0 8px;
+  background-color: ${colors.blues.lighter};
+  color: ${colors.blues.dark};
+  font-weight: 600;
+  font-size: 14px;
+  border-bottom: 1px solid ${colors.greyscale.lighter};
+`
+
+interface WeekProps {
+  startDate: LocalDate
+}
+
+const WeekElem = React.memo(function WeekElem({ startDate }: WeekProps) {
+  const dates = [...Array(5).keys()].map((n) => startDate.addDays(n))
+
+  return (
+    <div>
+      <WeekDiv>Vk {startDate.getIsoWeek()}</WeekDiv>
+      <div>
+        {dates.map((d) => (
+          <DayElem date={d} key={d.formatIso()} />
+        ))}
+      </div>
+    </div>
+  )
+})
+
+export default React.memo(function CalendarListView() {
+  const currentWeekStartDate = LocalDate.today().startOfWeek()
+
+  const dates = [...Array(5).keys()].map((n) =>
+    currentWeekStartDate.addWeeks(n)
+  )
+
+  return (
+    <FixedSpaceColumn spacing={'zero'}>
+      {dates.map((d) => (
+        <WeekElem startDate={d} key={d.formatIso()} />
+      ))}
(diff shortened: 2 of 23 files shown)


service/src/main/kotlin/fi/espoo/evaka/reservations/ReservationControllerCitizen.kt
```diff
@@ -0,0 +1,77 @@
...
+@RestController
+class ReservationControllerCitizen {
+    @GetMapping("/citizen/reservations")
+    fun getReservations(
+        db: Database.Connection,
+        user: AuthenticatedUser,
+        @RequestParam @DateTimeFormat(iso = DateTimeFormat.ISO.DATE) from: LocalDate,
+        @RequestParam @DateTimeFormat(iso = DateTimeFormat.ISO.DATE) to: LocalDate,
+    ): List<DailyReservationData> {
+        Audit.AttendanceReservationRead.log(targetId = user.id)
```
jaahon, line 30: Should you somehow specify that it's a guardian reading their own children's reservations? Employees need a pretty much identical audit event.

frontend/src/citizen-frontend/calendar/WeekElem.tsx
```diff
@@ -0,0 +1,39 @@
...
+  weekNumber: number
+  data: DailyReservationData[]
+}
+
+export default React.memo(function WeekElem({ weekNumber, data }: WeekProps) {
+  const i18n = useTranslation()
+  return (
+    <div>
+      <WeekDiv>
+        {i18n.common.datetime} {weekNumber}
```
jaahon, line 18: Shouldn't this be `i18n.common.datetime.weekShort`?

frontend/src/citizen-frontend/calendar/WeekElem.tsx
```diff
@@ -0,0 +1,39 @@
+import React from 'react'
+import styled from 'styled-components'
+import colors from 'lib-customizations/common'
+import { DailyReservationData } from './api'
+import DayElem from './DayElem'
+import { useTranslation } from '../localization'
+
+export interface WeekProps {
+  weekNumber: number
+  data: DailyReservationData[]
```
jaahon, line 10: I would prefer a more descriptive name like `dailyReservations` or something like that.

frontend/src/citizen-frontend/calendar/CalendarPage.tsx
```diff
@@ -0,0 +1,51 @@
...
+        <ContentArea opaque paddingVertical="zero" paddingHorizontal="zero">
+          {data.mapAll({
+            loading() {
+              return <Loader />
+            },
+            failure() {
+              return <div>{i18n.common.errors.genericGetError}</div>
+            },
+            success() {
+              return <CalendarListView data={data.getOrElse([])} />
```
jaahon, line 43: `success` gets the `Result`'s value as an argument.

frontend/src/citizen-frontend/App.tsx
```diff
@@ -76,6 +77,11 @@ export default function App() {
                         component={requireAuth(MessagesPage, false)}
                       />
                     )}
+                    <Route
+                      exact
+                      path="/calendar"
+                      component={requireAuth(CalendarPage, false)}
+                    />
```
jaahon, line 84: Should this route be disabled when the feature is not enabled?

frontend/src/citizen-frontend/calendar/CalendarListView.tsx
```diff
@@ -0,0 +1,38 @@
...
+    } else {
+      return [
+        ..._.dropRight(weekly),
+        {
+          ...last,
+          data: [...last.data, daily]
+        }
+      ]
+    }
+  }, [] as WeekProps[])
```
jaahon, line 29: Doesn't metter here, but you could set `weeklyData`'s type to `WeekProps[]` (like this `const weeklyData: WeekProps[] = ...`) so you wouldn't have to do a cast.

jaahon, line 29: I would extract this whole reduce into a function with a descriptive name like `partitionByWeek` because it's kind of hard to understand what's going on.

frontend/src/citizen-frontend/calendar/DayElem.tsx
```diff
@@ -0,0 +1,84 @@
...
+const DayColumn = styled(FixedSpaceColumn)`
+  width: 48px;
+  color: ${colors.blues.dark};
+  font-weight: 600;
+  font-size: 16px;
+`
+
+const DayDiv = styled(FixedSpaceRow)<{ today: boolean; holiday: boolean }>`
+  position: relative;
+  padding: 8px 16px;
```
jaahon, line 55: Should probably use `defaultMargins` here.

frontend/src/citizen-frontend/calendar/DayElem.tsx
```diff
@@ -0,0 +1,84 @@
...
+      {date.isBefore(LocalDate.today()) && <HistoryOverlay />}
+    </DayDiv>
+  )
+})
+
+const DayColumn = styled(FixedSpaceColumn)`
+  width: 48px;
+  color: ${colors.blues.dark};
+  font-weight: 600;
+  font-size: 16px;
```
jaahon, line 50: Isn't font size effectively `16px` anyway? And if it isn't, use `1rem` instead.

frontend/src/citizen-frontend/calendar/DayElem.tsx
```diff
@@ -0,0 +1,84 @@
...
+          )
+          .join(', ')}
+      </div>
+      {date.isBefore(LocalDate.today()) && <HistoryOverlay />}
+    </DayDiv>
+  )
+})
+
+const DayColumn = styled(FixedSpaceColumn)`
+  width: 48px;
```
jaahon, line 47: Would a margin work better instead of a fixed width?

frontend/src/citizen-frontend/calendar/DayElem.tsx
```diff
@@ -0,0 +1,84 @@
...
+  color: ${colors.blues.dark};
+  font-weight: 600;
+  font-size: 16px;
+`
+
+const DayDiv = styled(FixedSpaceRow)<{ today: boolean; holiday: boolean }>`
+  position: relative;
+  padding: 8px 16px;
+  height: 80px;
```
jaahon, line 56: I feel like fixed height is a bad idea here because it breaks easily and doesn't seem necessary in my opinion.

frontend/src/citizen-frontend/calendar/DayElem.tsx
```diff
@@ -0,0 +1,84 @@
...
+    border-left: 6px solid ${colors.brandEspoo.espooTurquoise};
+    padding-left: 10px;
+  `
+      : ''}
+
+  ${(p) =>
+    p.holiday
+      ? `
+    border-left: 6px solid ${colors.accents.red};
+    padding-left: 10px;
```
jaahon, line 70: Shouldn't just the text be red?

frontend/src/citizen-frontend/calendar/DayElem.tsx
```diff
@@ -0,0 +1,84 @@
...
+const DayDiv = styled(FixedSpaceRow)<{ today: boolean; holiday: boolean }>`
+  position: relative;
+  padding: 8px 16px;
+  height: 80px;
+  border-bottom: 1px solid ${colors.greyscale.lighter};
+  ${(p) =>
+    p.today
+      ? `
+    border-left: 6px solid ${colors.brandEspoo.espooTurquoise};
+    padding-left: 10px;
```
jaahon, line 62: You could use a transparent border with days other than today so you won't have to calculate the new padding.

jaahon (commented):
Sorry, I got a bit carried away with my review 😅 

Most of the comments are just nitpicking, you can just disregard most of them.

frontend/src/citizen-frontend/calendar/WeekElem.tsx
Joosakur: whoops, indeed

frontend/src/citizen-frontend/calendar/CalendarPage.tsx
Joosakur: oh!

Joosakur:
Thanks for detailed review! Will get back to this next week

frontend/src/citizen-frontend/App.tsx
Joosakur: A bit difficult to access user response here reliably, so I'll instead add `if(!user || !user.accessibleFeatures.reservations) return null` to the component itself

frontend/src/citizen-frontend/calendar/DayElem.tsx
Joosakur: I think width is better here to make sure that the next "column" is aligned

jaahon: Yeah, I didn't think about the fact that different dates have different widths because the font isn't monoscape 😄

In that case I would use `rem` units to define the width.

frontend/src/citizen-frontend/calendar/DayElem.tsx
Joosakur: yeah, and in the latest spec actually grey

frontend/src/citizen-frontend/calendar/WeekElem.tsx
```diff
@@ -0,0 +1,42 @@
...
+
+const WeekDiv = styled.div`
+  display: flex;
+  justify-content: center;
+  align-items: center;
+  padding: 16px 0 8px;
+  background-color: ${colors.brandEspoo.espooTurquoiseLight};
+  color: ${colors.blues.dark};
+  font-weight: 600;
+  font-size: 14px;
```
jaahon, line 40: Should use relative units here as well, especially since this UI is used by citizens.

jaahon (approved):

PR #1244, opened 2021-08-20
EVAKA-4240 unit attendance reservations

jaahon:
#### Summary

Minimal table of unit's this week's attendance reservations.

There is some code duplicated from https://github.com/espoon-voltti/evaka/pull/1236 that I'll remove when that PR is merged.


frontend/src/lib-customizations/types.d.ts
```diff
@@ -111,6 +111,7 @@ interface BaseFeatureFlags {
    */
   experimental?: {
     ai?: boolean
+    attendanceReservation?: boolean
```
Joosakur, line 114: I don't think we should use these anymore. Maybe rebase after I merge and then on unit page you can check the backend response in the same way as I did with the "add mobile device" feature.

Joosakur (changes requested):
Couple comments already

frontend/src/lib-customizations/types.d.ts
jaahon: I personally see a difference between feature flags and "pilot units". But is the idea to get rid of frontend feature flags completely?

Joosakur: I'm not sure. If the feature is such that it will be released in phases and/or never to some units, then I think it makes sense to design it already to use the db based unit specific model. In those cases the additional global flag seems just redundant and confusing to me. But if the feature is for example for service workers or otherwise released fully in a big bang once it's ready then the frontend feature flag is the way to go.

frontend/src/employee-frontend/components/unit/unit-reservations/UnitAttendanceReservationsPage.tsx
```diff
@@ -0,0 +1,131 @@
...
+  )
+
+  useEffect(
+    () => loadAttendanceReservations(unitId, dateRange),
+    [unitId, dateRange, loadAttendanceReservations]
+  )
+
+  return (
+    <Container>
+      <Gap size="L" />
```
Joosakur, line 49: I feel this could use a return button (same way as in absence calendar)

Joosakur (approved):

PR #1245, opened 2021-08-23
Add new provider type EXTERNAL_PURCHASED

reynders:
#### Summary
- Add new provider type "EXTERNAL_PURCHASED", which means external entities providing purchased services. For example in varda this means that the external company has already been created in varda so evaka provides only child data to varda (not the company data)


hruoho (commented):
Shouldn't this be added to `VardaUnitProviderType` too? In `service/src/main/kotlin/fi/espoo/evaka/varda/VardaUnits.kt`

reynders:
Continued in EVAKA-4000_external_purchased_provider_type_2

PR #1247, opened 2021-08-23
Update fee/voucher value decision pdf variables

patari:
#### Summary

Currently it is not possible to show validity end in voucher value decision. Also fee decision's weird `showValidTo` implementation doesn't seem to work correctly for Tampere use case.

This PR adds validity end to pdf variables to be able use them in templates.


mikkopiu (approved):
LGTM, pushed to CI

mikkopiu (approved):
:heavy_check_mark: on test fixes, missed those...

PR #1253, opened 2021-08-24
EVAKA-4240 attendance reservations in employee mobile

jaahon:
#### Summary

Show attendance reservations in child list and child page in employee mobile. No feature flags or pilot feature checks are used (should they?).

Also refactor:
  * existing code under attendance package to use `HelsinkiDateTime` instead of `Instant`s and `OffsetDateTime`s and pass the query date as parameter.
  * mobile child page


Joosakur (approved):
preapproved

PR #1254, opened 2021-08-24
Frontend enum code generation 

Gekkio:
#### Summary

- generate TS types for manually chosen backend enums -> no need to maintain typings manually
- add `gradlew codegen` task for regenerating all files
- add `gradlew codegenCheck` task that fails if any file is not up-to-date (= regenerating would cause changes)
- rename absence `CareType` -> `AbsenceCareType` in backend to match the frontend name
- rename `ApplicationGuardianAgreementStatus` -> `OtherGuardianAgreementStatus` in frontend to match the backend name
- replace several manual enum types with generated ones


akheron (approved):

akheron:
🏆

PR #1261, opened 2021-08-25
EVAKA-4000 client should throw

hruoho:
#### Summary

Add new versions of POST methods that throw on failed request and instead of just logging it.

### DIFF
--- a/service/src/main/kotlin/fi/espoo/evaka/varda/VardaUpdateServiceV2.kt
+++ b/service/src/main/kotlin/fi/espoo/evaka/varda/VardaUpdateServiceV2.kt
@@ -403,11 +403,11 @@ fun sendDecisionToVarda(client: VardaClient, vardaChildId: Long?, evakaServiceNe
 }
 
 fun sendPlacementToVarda(client: VardaClient, vardaDecisionId: Long, evakaServiceNeedInfoForVarda: EvakaServiceNeedInfoForVarda): Long {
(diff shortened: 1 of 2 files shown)


service/src/main/kotlin/fi/espoo/evaka/varda/VardaUpdateServiceV2.kt
```diff
@@ -545,7 +548,13 @@ fun sendVoucherDecisionToVarda(
         lahdejarjestelma = client.sourceSystem
     )
 
-    return client.createFeeData(requestPayload)?.id
+    val res = try {
+        client.createFeeDataV2(requestPayload)
+    } catch (e: Exception) {
+        error("VardaUpdate: cannot send voucher data for ${evakaServiceNeedInfoForVarda.id}: varda client threw ${e.localizedMessage}")
```
reynders, line 554: ..for service need ${evakaSer...

service/src/main/kotlin/fi/espoo/evaka/varda/integration/VardaClient.kt
```diff
@@ -279,6 +296,23 @@ class VardaClient(
         }
     }
 
+    fun createDecision2(newDecision: VardaDecision): VardaDecisionResponse {
```
reynders, line 299: fun createDecision2 -> fun createDecisionV2

reynders (approved):

PR #1273, opened 2021-08-26
Search people using POST methods instead GET

ilja-git:
#### Summary

Using POST methods in person searches so we do not get problematic entries into logs. The GET methods are replaced with POST-JSON equivalents.

This will change the internal API.

### DIFF
--- a/service/src/main/kotlin/fi/espoo/evaka/application/ApplicationControllerV2.kt
+++ b/service/src/main/kotlin/fi/espoo/evaka/application/ApplicationControllerV2.kt
@@ -40,7 +40,6 @@ import fi.espoo.evaka.shared.domain.Forbidden
 import fi.espoo.evaka.shared.domain.NotFound
 import fi.espoo.evaka.shared.security.AccessControl
 import fi.espoo.evaka.shared.security.Action
-import org.springframework.format.annotation.DateTimeFormat
 import org.springframework.http.HttpStatus
 import org.springframework.http.ResponseEntity
 import org.springframework.web.bind.annotation.GetMapping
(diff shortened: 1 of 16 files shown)


service/src/main/kotlin/fi/espoo/evaka/application/ApplicationControllerV2.kt
```diff
@@ -561,6 +540,26 @@ data class PaperApplicationCreateRequest(
...
+    val area: String?,
+    val units: String?,
+    val basis: String?,
+    val type: ApplicationTypeToggle,
+    val preschoolType: String?,
+    val status: String?,
+    val dateType: String?,
+    val distinctions: String?,
+    @DateTimeFormat(iso = DateTimeFormat.ISO.DATE) val periodStart: LocalDate?,
+    @DateTimeFormat(iso = DateTimeFormat.ISO.DATE) val periodEnd: LocalDate?,
```
jaahon, line 557: These `@DateTimeFormat` annotations aren't be needed.

service/src/main/kotlin/fi/espoo/evaka/pis/controllers/EmployeeController.kt
```diff
@@ -149,20 +149,26 @@ class EmployeeController {
...
-        @RequestParam(required = false) searchTerm: String?,
+        @RequestBody body: SearchEmployeeRequest
     ): ResponseEntity<Paged<EmployeeWithDaycareRoles>> {
         Audit.EmployeesRead.log()
         user.requireOneOfRoles(UserRole.ADMIN)
+        val maxPageSize = 5000
+        if (body.pageSize != null && body.pageSize > maxPageSize) throw BadRequest("Maximum page size is $maxPageSize")
         return db.read { tx ->
-            getEmployeesPaged(tx, page ?: 1, (pageSize ?: 50).coerceAtMost(100), searchTerm ?: "")
+            getEmployeesPaged(tx, body.page ?: 1, (body.pageSize ?: 50).coerceAtMost(100), body.searchTerm ?: "")
```
jaahon, line 163: Page size is already at most `100` (see `(body.pageSize ?: 50).coerceAtMost(100)`) so the first page size check is kind of conflicting with it.

jaahon (commented):
You forgot to change the `getOrCreatePersonBySsn` endpoint in `PersonController`, which was the original offender 😄

jaahon (approved):

PR #1278, opened 2021-08-28
EVAKA-TECH: type codegen

Joosakur:


service/codegen2/src/main/kotlin/evaka/codegen2/cli/Generate.kt
```diff
@@ -0,0 +1,183 @@
...
+import LocalDate from "../local-date";
+import FiniteDateRange from "../finite-date-range";
+import DateRange from "../date-range";
+import {DailyServiceTimes} from "../api-types/child/common";
+
+"""
+
+val tsMapping = mapOf(
+    "kotlin.String" to "string",
+    "java.util.UUID" to "string",
```
akheron, line 39: We have a `UUID` alias in the frontend, it should probably be used here.

service/codegen2/src/main/kotlin/evaka/codegen2/cli/Generate.kt
```diff
@@ -0,0 +1,183 @@
...
+    "fi.espoo.evaka.shared.Id" to "string",
```
akheron, line 40: And here

service/codegen2/src/main/kotlin/evaka/codegen2/cli/Generate.kt
```diff
@@ -0,0 +1,183 @@
...
+    "java.time.LocalDate" to "string",
+    "java.time.LocalTime" to "string",
+    "fi.espoo.evaka.shared.domain.FiniteDateRange" to "FiniteDateRange",
+    "fi.espoo.evaka.shared.domain.DateRange" to "DateRange",
+    "kotlin.Int" to "number",
+    "kotlin.Long" to "number",
+    "kotlin.Double" to "number",
+    "java.math.BigDecimal" to "number",
+    "kotlin.Boolean" to "boolean",
+    "java.time.LocalDate" to "LocalDate",
```
akheron, line 50: `java.time.LocalDate` is listed twice, this latter instance is correct

service/codegen2/src/main/kotlin/evaka/codegen2/cli/Generate.kt
```diff
@@ -0,0 +1,183 @@
...
+        if (analyzed is AnalyzedClass.DataClass) {
+            analyzed.properties.forEach { prop ->
+                if (!knownClasses.contains(prop.type.qualifiedName) && waiting.none { it == prop.type }) {
+                    waiting.addLast(prop.type)
+                }
+            }
+        }
+    }
+
+    return header + analyzedClasses.sortedBy { it.name }.joinToString("\n\n") { it.toTs() }
```
akheron, line 82: Sorting 👍

service/codegen2/src/main/kotlin/evaka/codegen2/cli/Generate.kt
```diff
@@ -0,0 +1,183 @@
...
+    abstract fun toTs(): String
+
+    class DataClass(
+        name: String,
+        val properties: List<AnalyzedProperty>
+    ) : AnalyzedClass(name) {
+        override fun toTs(): String {
+            return """/**
+* Generated from $name
+*/
```
akheron, line 152: JSDoc 👍

akheron (changes requested):
Is there a reason this cannot be put in the same `codegen` project as the enum codegen?

Also, I can now see why @Gekkio fears that this needs to be maintained quite a bit. But I also don't see a way to avoid that really, unless some actually useful stuff can be digged from the internals of Jackson somehow.

service/.gitignore
```diff
@@ -11,6 +11,8 @@
 /lib/
 /codegen/build/
 /codegen/target/
+/codegen2/build/
+/codegen2/target/
```
akheron, line 15: These are useless now

akheron (approved):

PR #1280, opened 2021-08-30
Set application service need to placement

patari:
#### Summary

https://github.com/espoon-voltti/evaka/pull/871 added support for application service need (i.e. citizen selects service need in the application form). This PR adds support to automatically persist that service need to placement when citizen accepts the decision.

### DIFF
--- a/service/src/main/kotlin/fi/espoo/evaka/placement/PlacementService.kt
+++ b/service/src/main/kotlin/fi/espoo/evaka/placement/PlacementService.kt
@@ -12,10 +12,12 @@ import fi.espoo.evaka.serviceneed.ServiceNeed
 import fi.espoo.evaka.serviceneed.clearServiceNeedsFromPeriod
 import fi.espoo.evaka.serviceneed.getServiceNeedsByChild
 import fi.espoo.evaka.serviceneed.getServiceNeedsByUnit
+import fi.espoo.evaka.serviceneed.insertServiceNeed
 import fi.espoo.evaka.shared.DaycareId
 import fi.espoo.evaka.shared.GroupId
 import fi.espoo.evaka.shared.GroupPlacementId
 import fi.espoo.evaka.shared.PlacementId
+import fi.espoo.evaka.shared.ServiceNeedOptionId
 import fi.espoo.evaka.shared.auth.AclAuthorization
 import fi.espoo.evaka.shared.db.Database
 import fi.espoo.evaka.shared.db.mapPSQLException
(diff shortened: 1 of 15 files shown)


service/src/main/kotlin/fi/espoo/evaka/serviceneed/ServiceNeed.kt
```diff
@@ -47,10 +47,10 @@ data class ServiceNeedOptionSummary(
 )
 
 data class ServiceNeedConfirmation(
-    val employeeId: UUID,
-    val firstName: String,
-    val lastName: String,
-    val at: HelsinkiDateTime
+    val employeeId: UUID?,
```
Joosakur, line 50: I recall seeing an annotation @PropagateNull, which could be used to make `val confirmed: ServiceNeedConfirmation` to be null instead of having these fields be separately nullable.

Joosakur: e.g. here https://github.com/espoon-voltti/evaka/blob/master/service/src/main/kotlin/fi/espoo/evaka/messaging/message/Message.kt#L70

service/src/main/kotlin/fi/espoo/evaka/placement/PlacementService.kt
```diff
@@ -12,31 +12,42 @@ import fi.espoo.evaka.serviceneed.ServiceNeed
...
 import org.jdbi.v3.core.kotlin.mapTo
 import org.jdbi.v3.json.Json
 import java.time.LocalDate
 import java.util.UUID
 
+data class ApplicationServiceNeed(
+    val optionId: ServiceNeedOptionId,
+    val shiftCare: Boolean,
+    val confirmedBy: UUID?,
+    val confirmedAt: HelsinkiDateTime?
```
Joosakur, line 39: are these ever not null? I assume these could be removed and hardcoded to null on line 65

Joosakur (commented):
Good stuff! Just a couple minor comments

service/src/main/kotlin/fi/espoo/evaka/serviceneed/ServiceNeed.kt
patari: How to achieve same with json-query, like https://github.com/espoon-voltti/evaka/blob/master/service/src/main/kotlin/fi/espoo/evaka/placement/PlacementService.kt#L409-L414 ?

service/src/main/kotlin/fi/espoo/evaka/placement/PlacementService.kt
patari: Fixed in https://github.com/espoon-voltti/evaka/pull/1280/commits/3819bf989cf2aa1923c6422de1a0ade7bd26a2cf

service/src/main/kotlin/fi/espoo/evaka/serviceneed/ServiceNeed.kt
patari: Fixed in https://github.com/espoon-voltti/evaka/pull/1280/commits/fc279664a0c806ff732401c8659afc08a284e93f

Joosakur (approved):

huhtari:
Conflicting files: service/src/main/resources/migrations.txt

PR #1282, opened 2021-08-30
EVAKA-HOTFIX: Fix authorization of global scope actions for scoped roles

Joosakur:

### DIFF
--- a/service/src/main/kotlin/fi/espoo/evaka/shared/security/AccessControl.kt
+++ b/service/src/main/kotlin/fi/espoo/evaka/shared/security/AccessControl.kt
@@ -147,7 +147,12 @@ WHERE employee_id = :userId
     }
 
     fun requirePermissionFor(user: AuthenticatedUser, action: Action.Global) {
-        assertGlobalPermission(user, action, permittedRoleActions::globalActions)
+        assertPermission(
+            user = user,
+            getAclRoles = { acl.getRolesForAnyUnit(user).roles },
+            action = action,
+            mapping = permittedRoleActions::globalActions
+        )
     }
 
     fun getPermittedRoles(action: Action.Global): Set<UserRole> {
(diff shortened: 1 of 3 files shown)


service/src/main/kotlin/fi/espoo/evaka/shared/security/AccessControl.kt
```diff
@@ -145,7 +145,12 @@ WHERE employee_id = :userId
     }
 
     fun requirePermissionFor(user: AuthenticatedUser, action: Action.Global) {
-        assertGlobalPermission(user, action, permittedRoleActions::globalActions)
+        assertPermission(
+            user = user,
+            getAclRoles = { acl.getRolesForAnyUnit(user).roles },
+            action = action,
+            mapping = permittedRoleActions::globalActions
+        )
```
akheron, line 153: The downside here is that the SQL query is run even if the action doesn’t depend on any scoped roles.

akheron: Actually it seems they aren’t after all 👍

Joosakur: Are you sure? It's supposed to be a lambda that is only evaluated when necessary

service/src/main/kotlin/fi/espoo/evaka/shared/security/AccessControl.kt
```diff
@@ -331,7 +341,7 @@ WHERE employee_id = :userId
         return result
     }
 
-    private inline fun <reified A, reified I> ActionConfig<A>.hasGlobalPermission(
+    private inline fun <reified A, reified I> ActionConfig<A>.hasPermissionThroughGlobalRole(
```
Joosakur, line 344: these functions are new to me and I don't know their purpose so I don't know if i broke them

Gekkio (approved):

PR #1284, opened 2021-08-30
EVAKA-4275: Fix voucher value decision layout

mapehe:
#### Summary

See [this](https://voltti.atlassian.net/browse/EVAKA-4275) for details.


service/src/main/resources/WEB-INF/templates/fee-decision/voucher-value-decision.html
```diff
@@ -100,7 +103,7 @@ <h1 th:text="#{text.title}"></h1>
         <tbody>
             <tr>
                 <td th:text="#{decision.amountOfVoucherForProvider}"></td>
-                <td class="text-right"><strong th:text="${serviceProviderValue}"></strong></td>
+                <td class="text-right"><p th:text="${serviceProviderValue}"></p></td>
```
jaahon, line 106: `<span>` would be a better element to use here.

PR #1287, opened 2021-08-30
EVAKA-4240 citizen reservation inline editor

jaahon:
#### Summary

Add inline editor to citizen reservation day view where single day reservations can be updated quickly.
Also,
* Show month with first available end date when opening date picker in reservation modal
* Fix a bug in citizen reservation overwrite
* Show same reservations only once

### DIFF
--- a/frontend/src/citizen-frontend/calendar/DayView.tsx
+++ b/frontend/src/citizen-frontend/calendar/DayView.tsx
@@ -2,25 +2,45 @@
 //
 // SPDX-License-Identifier: LGPL-2.1-or-later
 
-import React, { Fragment, useMemo } from 'react'
+import React, { Fragment, useEffect, useMemo, useState } from 'react'
 import styled from 'styled-components'
-import { faChevronLeft, faChevronRight } from 'lib-icons'
+import { faCheck, faChevronLeft, faChevronRight, faPen } from 'lib-icons'
 import { DATE_FORMAT_TIME_ONLY, formatDate } from 'lib-common/date'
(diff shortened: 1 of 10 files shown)


frontend/src/citizen-frontend/calendar/DayView.tsx
```diff
@@ -40,53 +60,256 @@ export default React.memo(function DayView({ date, data, selectDate }: Props) {
...
+                  )}`}</span>
+                )}
+                <Label>Toteuma</Label>
+                <span>–</span>
+              </Grid>
+            </Fragment>
+          ))}
+      {confirmationModal ? (
+        <InfoModal
+          title="Haluatko tallentaa muutokset?"
```
mikkopiu, line 170: Shouldn't these be localized? These will after all be read by screenreaders.

mikkopiu (changes requested):
Missing some localizations and IMO the logic here is complex enough that some E2E tests would be nice to have.

frontend/src/citizen-frontend/calendar/DayView.tsx
jaahon: Yeah, everything should be localized when it's enabled in production. And this is an actual "pop up" modal so it's definitely visible to everyone.

jaahon:
> Missing some localizations and IMO the logic here is complex enough that some E2E tests would be nice to have.

We are planning to write e2e-tests when everything's finished and ready for production.

mikkopiu (approved):
OK with above comment replies

Joosakur (approved):

PR #1296, opened 2021-09-01
New frontend access control - part 3

Gekkio:
#### Summary

- implement top-level action checks on child information page (= load data and show main sections only if permitted to do so)
- upgrade to Kotlin 1.5.30


frontend/src/employee-frontend/components/ChildInformation.tsx
```diff
@@ -67,17 +68,55 @@ const HeadOfFamilyLink = styled(Link)`
...
+function requirePermittedAction(
+  Component: React.FunctionComponent<SectionProps>,
+  ...actions: Action.Child[]
+): React.FunctionComponent<SectionProps> {
+  return function Section({ id, startOpen }: SectionProps) {
+    const { permittedActions } = useContext<ChildState>(ChildContext)
+    if (actions.some((action) => permittedActions.has(action))) {
+      return <Component id={id} startOpen={startOpen} />
+    } else {
+      return null
```
Joosakur, line 85: nitpick: If there's two actions and you have permission for one of them, then this renders the component. After a second thought I suppose that is intentional, and it is then the component's responsibility to check on a more granular level which parts of the data to load/display (as you seem to have done), but maybe the naming could still be clarified to distinct between require permission for all vs some of these actions.

Joosakur (approved):

PR #1303, opened 2021-09-02
E2E tests in containers

ilja-git:
#### Summary

Run E2E inside containers.


.circleci/config.yml
```diff
@@ -376,6 +385,41 @@ commands:
...
+          working_directory: *workspace_compose
+          command: |
+            ecr-login
+            # The image is not expected to chage too often so we use :latest for all cache
+            E2E_IMAGE="307238562370.dkr.ecr.eu-west-1.amazonaws.com/evaka/e2e-<< parameters.test_runner >>:latest"
+            docker pull "$E2E_IMAGE"
+            docker build -t "$E2E_IMAGE" --cache-from "$E2E_IMAGE" -f "./e2e/<< parameters.test_runner >>.Dockerfile" ./e2e/
+            docker push "$E2E_IMAGE"
+      - run:
+          name: Build and push E2E << parameters.test_runner >> image
```
mikkopiu, line 411: Copy-pasted name, should be something like:
```suggestion
          name: Package E2E << parameters.test_runner >> image
```

.circleci/config.yml
```diff
@@ -376,6 +385,41 @@ commands:
...
+          working_directory: *workspace_compose
+          command: |
+            ecr-login
+            # The image is not expected to chage too often so we use :latest for all cache
+            E2E_IMAGE="307238562370.dkr.ecr.eu-west-1.amazonaws.com/evaka/e2e-<< parameters.test_runner >>:latest"
+            docker pull "$E2E_IMAGE"
+            docker build -t "$E2E_IMAGE" --cache-from "$E2E_IMAGE" -f "./e2e/<< parameters.test_runner >>.Dockerfile" ./e2e/
+            docker push "$E2E_IMAGE"
```
mikkopiu, line 409: One downside I can see to this, is that now this could not be run as-is on the theoretical "forks-only workflow without secrets" but not sure that it's a use case worth over-complicating this over (and rather just create some alternative solution).

.circleci/config.yml
```diff
@@ -376,6 +385,41 @@ commands:
...
+      - setup_remote_docker:
+          version: *remote_docker_version
+          docker_layer_caching: false
+      - run:
+          name: Build and push E2E << parameters.test_runner >> image
+          working_directory: *workspace_compose
+          command: |
+            ecr-login
+            # The image is not expected to chage too often so we use :latest for all cache
+            E2E_IMAGE="307238562370.dkr.ecr.eu-west-1.amazonaws.com/evaka/e2e-<< parameters.test_runner >>:latest"
```
mikkopiu, line 406: Suggestion: move the Docker registry part of image names to a pipeline parameter (like `workspace_root`).

.circleci/config.yml
```diff
@@ -400,101 +444,20 @@ commands:
...
-          steps:
-            - run:
-                name: Install Playwright dependencies
-                command: |
-                  sudo apt-get update && sudo apt-get --yes install --no-install-recommends libgbm1
       - restore_frontend_deps
       - restore_frontend_comm_deps
+      - restore_cache:
+          name: Restore Docker image from cache
+          key: << parameters.test_runner >>-image-v2-{{ .Environment.CIRCLE_WORKFLOW_WORKSPACE_ID }}
```
mikkopiu, line 451: Did you test whether this + `docker load` is faster than just pulling the pushed image from ECR? That of course incurs some costs but worth testing at least.

frontend/package.json
```diff
@@ -26,9 +26,9 @@
     "e2e-employee-2": "testcafe 'chrome --start-fullscreen' src/e2e-test/specs/2_employee-2/*.spec.ts",
     "e2e-mobile": "testcafe 'chrome --start-fullscreen' src/e2e-test/specs/6_mobile/*.spec.ts",
     "e2e-messaging": "testcafe 'chrome --start-fullscreen' src/e2e-test/specs/7_messaging/*.spec.ts",
-    "e2e-playwright": "jest --testTimeout 60000 --runInBand src/e2e-playwright/specs",
-    "e2e-ci-testcafe": "BASE_URL=http://localhost:9999 testcafe 'chrome:headless:emulation:width=1920;height=1080' -s takeOnFails=true --video videos --video-options failedOnly=true --stop-on-first-fail",
-    "e2e-ci-playwright": "BASE_URL=http://localhost:9999 jest --testTimeout 60000 --runInBand --bail 1"
+    "e2e-playwright": "LANG=fi-FI LANGUAGE=fi_FI jest --testTimeout 60000 --runInBand src/e2e-playwright/specs",
```
mikkopiu, line 29: Why the `LANG` and `LANGUAGE` envs?

.circleci/config.yml
ilja-git: We can also build and save it on local too. Something like 
- restore from cache
- if cache
-- docker load
-- docker build with cache
- else
-- docker build without cache 
- store to cache

frontend/package.json
ilja-git: Noticed that if user has different LANGUAGE or LANG environment variables set then those are passed to chromium and then tests fail.

mikkopiu: Weird, 'cause I've always used `LANG=en_US.UTF-8` locally and never had those problems. And besides, I think we should actually then fix those tests to either a) not make assumptions about user language and rely on `data-qa` labels, or b) explicitly select a language in eVaka.

Did this break in CI, too, without the variables?

.circleci/config.yml
ilja-git: The pull was slow but so is load. Tried it first and used this afterwards to speed up, but this did not speed up as mutch as I hoped. Did not make real comparision. If we want to use only local image as discussed above then we need to use this.

mikkopiu: :heavy_check_mark: I was hoping pulling would be way faster (considering how abysmally slow CircleCI's cache restores can be). But in that case, I think going full-local would be the best option. Maybe compare pull vs. cache restore speeds for that, too.

frontend/package.json
ilja-git: CI uses LANG: https://github.com/espoon-voltti/evaka/blob/master/.circleci/config.yml#L462

Why I added those here was because I spend some time to find out why tests were failing locally but not in CI. Also noticed that you need overwrite both of those if those are defined. That LANG requirement has not been long.

.circleci/config.yml
ilja-git: Created flag to swich between these.

.circleci/config.yml
ilja-git: Unified the parameters for job and parametrized the registry.

mikkopiu (approved):
LGTM!

PR #1318, opened 2021-09-07
EVAKA-4240 absences in citizen calendar

jaahon:
#### Summary

Let citizens mark absences for their children and show absences in citizen calendar.


service/src/main/resources/db/migration/V138__absence_modified_by_citizen.sql
```diff
@@ -0,0 +1,9 @@
+ALTER TABLE absence ALTER COLUMN modified_by DROP NOT NULL;
+ALTER TABLE absence RENAME COLUMN modified_by TO modified_by_deprecated;
+ALTER TABLE absence ADD COLUMN modified_by_employee_id uuid REFERENCES employee(id);
+ALTER TABLE absence ADD COLUMN modified_by_guardian_id uuid REFERENCES person(id);
+
+UPDATE absence SET modified_by_employee_id = (SELECT id FROM employee WHERE modified_by_deprecated = id::text OR modified_by_deprecated = concat(first_name, ' ', last_name));
```
Joosakur, line 6: subquery could return multiple if multiple employees have same name. unlikely, but could add limit 1 to pick random instead of fail in some environment with bunch of testi testaaja employees

Joosakur (approved):

PR #1337, opened 2021-09-09
EVAKA-4240 citizen reservation e2e tests

jaahon:
#### Summary

Basic e2e tests for citizen reservations.

Also fix a bug where person logging in into evaka for the first time might not get all relevant features.

### DIFF
--- a/frontend/src/e2e-playwright/pages/citizen/citizen-calendar.ts
+++ b/frontend/src/e2e-playwright/pages/citizen/citizen-calendar.ts
@@ -1,0 +1,142 @@
+// SPDX-FileCopyrightText: 2017-2021 City of Espoo
+//
+// SPDX-License-Identifier: LGPL-2.1-or-later
+
+import { waitUntilEqual } from 'e2e-playwright/utils'
+import { Checkbox, Combobox, RawTextInput } from 'e2e-playwright/utils/element'
+import FiniteDateRange from 'lib-common/finite-date-range'
+import LocalDate from 'lib-common/local-date'
+import { Page } from 'playwright'
+
+export default class CitizenCalendarPage {
+  constructor(
+    private readonly page: Page,
+    private readonly type: 'desktop' | 'mobile'
+  ) {}
+
+  #openCalendarActionsModal = this.page.locator(
+    '[data-qa="open-calendar-actions-modal"]'
+  )
+
+  #childCheckbox = (childId: string) =>
+    new Checkbox(this.page, `[data-qa="child-${childId}"]`)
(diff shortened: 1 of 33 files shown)


frontend/src/e2e-playwright/pages/citizen/citizen-calendar.ts
```diff
@@ -0,0 +1,142 @@
...
+    await this.#startDateInput.clear()
+    await this.#startDateInput.type(dateRange.start.format())
+    await this.#endDateInput.type(dateRange.end.format())
+    await this.#repetitionCombobox.fill('Viikoittain')
+    await this.#repetitionCombobox.findItem('Viikoittain').click()
+    await weeklyTimes.reduce(async (promise, { startTime, endTime }, index) => {
+      await promise
+      await this.#weeklyStartTimeInputs[index].type(startTime)
+      await this.#weeklyEndTimeInputs[index].type(endTime)
+    }, Promise.resolve())
```
Joosakur, line 93: What's the purpose of reduce over e.g. forEach or a basic for-loop? The function also doesn't return anything, so on first item it will `await Promise.resolve()` and after that it will `await undefined`, which also confuses me.

frontend/src/lib-customizations/espoo/citizen/assets/i18n/en.tsx
```diff
@@ -148,40 +148,40 @@ const en: Translations = {
     backToSearch: 'Back to search'
   },
   calendar: {
-    title: 'Calendar',
-    holiday: 'Holiday',
-    absent: 'Absent',
+    title: 'Kalenteri',
```
Joosakur, line 151: So what's our plan with i18n during piloting? There surely are some non-Finnish speakers there so I don't think we can release this without at least some level of translations.

Joosakur (approved):

frontend/src/e2e-playwright/pages/citizen/citizen-calendar.ts
jaahon: Async functions always return promises. So that function returns `Promise<void>`. So in effect we wait for the previous start and end times to be written into the inputs before taking the next weekday.

frontend/src/lib-customizations/espoo/citizen/assets/i18n/en.tsx
jaahon: Didn't we agree to just use Finnish for now? I think it's better to use Finnish only instead of incomplete and possibly misleading translations.

PR #1340, opened 2021-09-09
Compile TypeScript outside webpack

akheron:
### Summary

Before this change, `.ts` files were transpiled to JavaScript _by_ webpack/ts-loader without type checking. Type checking was done separately with `webpack-plugin-fork-ts-checker`.

After this change, `.ts` files are compiled with `tsc` to disk, and the resulting `.js` files are bundle with `webpack`.

Pros:
- No more errors like `x.js was not compiled from x.ts`
- All changes are picked up from `lib-common`, `lib-components`, etc. without the need to restart `yarn dev`
- No need to destroy caches by `yarn clean` anymore, `yarn dev` just works
- No need for separate `yarn type-check` as the normal build process now checks the validity of typings

Cons:
- Type errors prevent a build, even with `yarn dev`. All type errors need to be fixed before `.js` files are emitted and subsequently bundled, because it's how `tsc --build` [works](https://www.typescriptlang.org/docs/handbook/project-references.html#caveats).
- Changing asset files (like images) don't trigger a rebuild. You have to change a `.ts` file or restart `yarn dev`. This could also be enhanced, but not sure how to support all platforms.
- When saving a `.ts` file, webpack doesn't start a new build immediately, but only after `tsc` has emitted the `.js` files. So if you're used to refreshing the browser in the same second, you'll still get the old app.

**How it works?**

Building the frontend now works like this:
1. Use `tsc --build` to compile TypeScript files under `src/` and write to `dist/`.
2. Copy asset files as-is from `src/` to `dist/`.
3. Bundle the `.js` files and assets under `dist/` with webpack to `dist/bundle/`.

In development, `tsc` and `dev-server.js` are run in parallel. A complete `tsc` build is waited for first because otherwise there's nothing to bundle!

All test code (unit, testcafe, playwright) is still compiled by the test runner. All jest-based tests could already be run from the files under `dist/`, but it seems to be impossible with TestCafe.

**Misc changes:**

- Remove `ts-loader` and `fork-ts-checker-webpack-plugin` (no longer needed)
- Remove project internal non-`lib-*` imports (these now seem to cause problems with global value identities)
- Use webpack/jest module aliases for `lib-*` imports (because we cannot really link to directories under `dist/` in `package.json` as they don't exist before the build)


akheron:
Closed in favor of #1370

PR #1342, opened 2021-09-09
EVAKA-HOTFIX fee decision send conflict feedback

jaahon:
#### Summary

* Fix regression where finance decision/invoice send errors were ignored.
* Show a better error message when fee decision send fails because of existing decisions waiting for manual sending.


hruoho (approved):
Very nice, thank you! 🙏

PR #1350, opened 2021-09-10
Store citizen ui lang in local storage

jaahon:


Gekkio:
IIRC localStorage functions can throw exceptions in some browsers and some scenarios (Safari incognito?), so error handling is probably needed.

jaahon:
> IIRC localStorage functions can throw exceptions in some browsers and some scenarios (Safari incognito?), so error handling is probably needed.

Right, didn't think of that. Thanks!

reynders (approved):

PR #1351, opened 2021-09-10
Require protocol in unit url

jaahon:
#### Summary

This seems like the easiest and simplest solution to the problem of urls without protocol being interpreted as relative urls.

I'm not sure if we even want to allow plain http :smile:

### DIFF
--- a/frontend/src/lib-customizations/espoo/employee/assets/i18n/fi.ts
+++ b/frontend/src/lib-customizations/espoo/employee/assets/i18n/fi.ts
@@ -2653,6 +2653,7 @@ export const fi = {
       daycareType: 'Varhaiskasvatuksen tyyppi puuttuu',
       capacity: 'Kapasiteetti on virheellinen',
       costCenter: 'Kustannuspaikka puuttuu',
+      url: 'URL-osoitteessa pitää olla https://- tai http://-etuliite',
       visitingAddress: {
         streetAddress: 'Käyntiosoitteen katuosoite puuttuu',
         postalCode: 'Käyntiosoitteen postinumero puuttuu',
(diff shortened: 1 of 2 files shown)


frontend/src/lib-customizations/espoo/employee/assets/i18n/fi.ts
```diff
@@ -2653,6 +2653,7 @@ export const fi = {
       daycareType: 'Varhaiskasvatuksen tyyppi puuttuu',
       capacity: 'Kapasiteetti on virheellinen',
       costCenter: 'Kustannuspaikka puuttuu',
+      url: 'URL-osoitteessa pitää olla http(s)://-etuliite',
```
mikkopiu, line 2656: As people editing these might not be familiar with what this is, should we be even more explicit like:
```suggestion
      url: 'URL-osoitteessa pitää olla https://- tai http://-etuliite',
```

mikkopiu (commented):
Regarding "do we even want to allow plain HTTP": In Espoo's production, there are currently at least a few non-HTTPS links that don't have HTTPS equivalent links available (old sites that have never been updated).

But not only that, there is at least one email address in the link field and I'm not sure what we want to do with those; maybe we need an alternative input for those or just disallow them completely and tell the existing unit to update their info?

frontend/src/lib-customizations/espoo/employee/assets/i18n/fi.ts
jaahon: Yeah, good point.

jaahon:
> But not only that, there is at least one email address in the link field and I'm not sure what we want to do with those; maybe we need an alternative input for those or just disallow them completely and tell the existing unit to update their info?

I had no idea. I'm not sure if email makes sense in the current use case. Sure it might work as a link, but I don't think it makes sense in the current context. I'll ask what to do with those.

image

jaahon:
The email was changed into the unit's home page and it seems like there is no requirement to support email addresses.

mikkopiu (approved):
I think you should inform Tampere about this one, too; don't know what state their unit data is in right now.

PR #1352, opened 2021-09-10
EVAKA-4240: move absences and reservations under calendar tab

Joosakur:


frontend/src/employee-frontend/components/unit/TabCalendar.tsx
```diff
@@ -0,0 +1,100 @@
...
+interface Props {
+  unitId: UUID
+}
+
+export default React.memo(function TabCalendar({ unitId }: Props) {
+  const { i18n } = useTranslation()
+  const { unitInformation } = useContext(UnitContext)
+  const [mode, setMode] = useState<Mode>('month')
+  const defaultGroup = useQuery().get('group')
+  const [groupId, setGroupId] = useState<UUID | 'no-group' | null>(defaultGroup)
```
jaahon, line 45: You might want to check that `defaultGroup` is a valid id. Also, the query parameter is not synced with the UI, which might get a bit confusing.

Joosakur: I guess it should "work" even if it's not valid - the group just won't be found from the results and won't be shown. Or how should it behave? Redirect to no query param?

I'll try to implement the syncing. Do you happen to remember good example code for that?

jaahon: There's a `useSyncQueryParams` hook for that.

jaahon: I think the `group` value should be reset to the default value if it's not a valid selection.

jaahon: By the way, the select component dissappears if you pick the "no selection" option.

Joosakur: had some trouble with infinite loops despite using memo, so added some extra conditions to that hook https://github.com/espoon-voltti/evaka/pull/1352/commits/04d8bb3d8e70b85a25bc1beffed5a29731cc90df

frontend/src/employee-frontend/components/UnitPage.tsx
```diff
@@ -148,6 +154,11 @@ export default React.memo(function UnitPage() {
               />
             )}
           />
+          <RouteWithTitle
+            exact
+            path="/units/:id/calendar"
+            component={() => <TabCalendar unitId={id} />}
```
jaahon, line 160: This arrow function is different every time `UnitPage` is rerendered and thus it's a different component each time. In practice this mounts and unmounts `TabCalendar` as many times as `UnitPage` is rerendered. This means for example that the data used by the component is requested 5 times. To prevent that either use `Route`'s `render` or `children` props or just get the `unitId` prop from path using `useParam`.

frontend/src/employee-frontend/components/unit/TabCalendar.tsx
jaahon: The problems were probably because of the thing I just commented on. The extra conditions don't have an effect if you use `useMemo`.

Still, I got an idea for an improvement to `useSyncQueryParams` so you wouldn't have to use `useMemo` with it:
```
/**
 * Keeps current URL's query params up to date with the queryParams object
 */
export function useSyncQueryParams(queryParams: Record<string, string>) {
  const history = useHistory()
  const query = new URLSearchParams(queryParams).toString()
  useEffect(() => {
    const { pathname } = history.location
    const url = query ? `${pathname}?${query}` : pathname
    history.replace(url)
  }, [history, query])
}
```

frontend/src/employee-frontend/components/unit/TabCalendar.tsx
```diff
@@ -0,0 +1,108 @@
...
+  const { id: unitId } = useParams<{ id: UUID }>()
+  const { unitInformation } = useContext(UnitContext)
+  const [mode, setMode] = useState<Mode>('month')
+  const [selectedDate, setSelectedDate] = useState<LocalDate>(LocalDate.today())
+
+  const groupParam = useQuery().get('group')
+  const defaultGroup =
+    groupParam && (groupParam === 'no-group' || UUID_REGEX.test(groupParam))
+      ? groupParam
+      : null
```
jaahon, line 50: I'm sorry that I forgot to comment on this earlier, but originally I was thinking about checking that the query param matches one of the groups queried from back end, not that it looks like a uuid. Even if the id is a valid uuid the ui shows that the first group is selected but fetching the data failed. But it's not a must have so you can leave it out.

jaahon (approved):

PR #1361, opened 2021-09-13
Validate mandatory varda values for varda unit

reynders:


frontend/src/lib-common/utils/string.ts
```diff
@@ -0,0 +1,7 @@
+// SPDX-FileCopyrightText: 2017-2020 City of Espoo
+//
+// SPDX-License-Identifier: LGPL-2.1-or-later
+
+export function isNullOrEmpty(str?: string | null): boolean {
+  return !str || str.length == 0
```
hruoho, line 6:
```suggestion
  return !str || str.trim().length == 0
```

Gekkio:
We already have validation + trim-before-save mechanism in the editor :thinking:

PR #1364, opened 2021-09-14
Evaka 4000 external purchased provider type 2

reynders:
Add a new unit provider type "external purchased" for daycares which are operated completely outside evaka


frontend/src/citizen-frontend/map/SearchSection.tsx
```diff
@@ -147,20 +147,24 @@ export default React.memo(function SearchSection({
...
-                    }}
-                  />
-                )
-              )}
+              {(
+                [
+                  'MUNICIPAL',
+                  'PURCHASED',
+                  'EXTERNAL_PURCHASED'
+                ] as ProviderTypeOption[]
```
jaahon, line 155: Not really related to this PR, but using `as const` instead of `as ProviderTypeOption[]` is better because with the latter you could pass values that aren't actually a part of `ProviderTypeOption`.

jaahon (commented):
I don't think citizens care about the distinction between "ostopalvelu" and "nettobudjetoituostopalvelu", so I think the map filter should combine the two. There's a layout problem anyway if you keep the two separate. The problem could be fixed by putting all the options in same flex row and using flex wrap.

Anyway, I couldn't think of any pitfalls that adding this new value would have.

reynders:
> I don't think citizens care about the distinction between "ostopalvelu" and "nettobudjetoituostopalvelu", so I think the map filter should combine the two. There's a layout problem anyway if you keep the two separate. The problem could be fixed by putting all the options in same flex row and using flex wrap.
> 
> Anyway, I couldn't think of any pitfalls that adding this new value would have.

It was agreed with the product management that as external purchased units are not open for applications those do not need to be supported as a separate unit type on the map.

PR #1386, opened 2021-09-16
Add pdf fonts customization support

patari:
#### Summary

Currently specific fonts are required to be available in application resources even if not used in templates (see https://github.com/Tampere/trevaka/commit/15e731fec8bcf5107cc339af10be45c80459f545). This adds support to load all available fonts from `resources/ttf` directory.

As a side note this also fixes Montserrat bold for you because old implementation was

```
val montserrat = getResourcePath("ttf/Montserrat-Regular.ttf")
val montserratBold = getResourcePath("ttf/Montserrat-Regular.ttf")
```

but looking at your template files it seems that bold Montserrat is not even used so no harm here (fonts are embedded to PDF only if used)


mikkopiu (approved):
Nice find!

PR #1388, opened 2021-09-16
EVAKA-4288 family initialization improvements

jaahon:
#### Summary

* Add instructions about adding family memebers to applications
* Add dependants living in the same address as the applicant to application's "other children" list by default
* Fix family initialization not adding child's other guardian to the family
* Show family overview without income to unit supervisors

This PR does not fix cases where family initialization fails when the applicants partner is already the head of the family.

### DIFF
--- a/service/src/main/kotlin/fi/espoo/evaka/application/ApplicationStateService.kt
+++ b/service/src/main/kotlin/fi/espoo/evaka/application/ApplicationStateService.kt
@@ -26,6 +26,7 @@ import fi.espoo.evaka.daycare.domain.ProviderType
 import fi.espoo.evaka.daycare.getActivePreschoolTermAt
 import fi.espoo.evaka.daycare.getClubTerms
 import fi.espoo.evaka.daycare.getDaycare
+import fi.espoo.evaka.daycare.getPreschoolTerms
 import fi.espoo.evaka.daycare.getUnitApplyPeriods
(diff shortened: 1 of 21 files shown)


service/src/main/kotlin/fi/espoo/evaka/application/ApplicationStateService.kt
```diff
@@ -96,6 +99,66 @@ class ApplicationStateService(
...
+        type: ApplicationType,
+        form: ApplicationForm
+    ): ApplicationForm {
+        val today = dateNow()
+
+        val startDate = when (type) {
+            ApplicationType.PRESCHOOL -> tx.getPreschoolTerms().firstOrNull { it.extendedTerm.start.isAfter(today) }
+                ?.extendedTerm?.start
+            ApplicationType.CLUB -> tx.getClubTerms().firstOrNull { it.term.start.isAfter(today) }
+                ?.term?.start
```
Joosakur, line 134: Should one still be able to apply to club after term has started? I don't remember

Joosakur (approved):
🎖️

service/src/main/kotlin/fi/espoo/evaka/application/ApplicationStateService.kt
jaahon: Yeah, but this only affects the default date. So if the term has started we don't fill a default start date.

PR #1395, opened 2021-09-17
Add more action based permission checks

akheron:
#### Summary

- Implement action based permission checks for a few more simple cases: Fee thresholds, absences, application notes
- While at it, simplify the application notes controller a bit
- Use `ActionConfig` to check scoped roles for applications
- Refactor attachment permission checks
  - Use action based permissions for both citizen and employee
  - Add actions for each "parent" type: application, income statement
  - Sanitize permissions for attachment deletion


service/src/main/kotlin/fi/espoo/evaka/attachment/AttachmentQueries.kt
```diff
@@ -42,15 +37,15 @@ fun Database.Transaction.insertAttachment(
...
                     it
                         .bind("applicationId", attachTo.applicationId)
                         .bindNullable("incomeStatementId", null as IncomeStatementId?)
-                is AttachToIncomeStatement ->
+                is AttachmentParent.IncomeStatement ->
                     it
                         .bindNullable("applicationId", null as ApplicationId?)
                         .bind("incomeStatementId", attachTo.incomeStatementId)
-                is AttachToNothing ->
+                is AttachmentParent.None ->
```
kimifoud, line 48: Instead of binding nullables and bloating each branch when a new parent type is added, we could use `bindKotlin` and use a data class with null defaults.

```kt

data class AttachmentParent(
    val applicationId: ApplicationId? = null,
    val incomeStatementId: IncomeStatementId? = null,
    val fooId: FooId? = null,
)
```

```kt
        .bindKotlin(when (attachTo) {
            is AttachToApplication ->
                AttachmentParent(applicationId = attachTo.applicationId)
            is AttachToIncomeStatement ->
                AttachmentParent(incomeStatementId = attachTo.incomeStatementId)
            is AttachToNothing ->
                AttachmentParent()
        })
```

kimifoud: Naming needs some work but you get the idea. 😄

akheron: Great idea, fixed!

service/src/main/kotlin/fi/espoo/evaka/attachment/AttachmentQueries.kt
```diff
@@ -65,26 +56,36 @@ fun Database.Transaction.insertAttachment(
...
     )
     .bind("id", id)
-    .mapTo<Attachment>()
+    .map { row ->
+        val applicationId = row.mapColumn<ApplicationId?>("application_id")
+        val incomeStatementId = row.mapColumn<IncomeStatementId?>("income_statement_id")
+        val attachedTo =
+            if (applicationId != null) AttachmentParent.Application(applicationId)
+            else if (incomeStatementId != null) AttachmentParent.IncomeStatement(incomeStatementId)
+            else AttachmentParent.None
```
kimifoud, line 71: Hmmm.. `attachedTo` could be a generated column, see `message_account_type`. Or if the value is not needed in db queries, could it be a calculated value in `Attachment` data class?

kimifoud (approved):

service/src/main/kotlin/fi/espoo/evaka/attachment/AttachmentQueries.kt
akheron: We talked about this through another channel and came to the conclusion that this is the best way to do it 😄

service/src/main/kotlin/fi/espoo/evaka/attachment/AttachmentQueries.kt
```diff
@@ -9,54 +9,45 @@ import fi.espoo.evaka.shared.AttachmentId
...
+import org.jdbi.v3.core.kotlin.bindKotlin
 import org.jdbi.v3.core.kotlin.mapTo
 import java.util.UUID
 
-sealed interface AttachTo
-
-data class AttachToApplication(val applicationId: ApplicationId) : AttachTo
-data class AttachToIncomeStatement(val incomeStatementId: IncomeStatementId) : AttachTo
-object AttachToNothing : AttachTo
+data class AttachmentParentColumn(
```
Gekkio, line 18: Could this be `private`? Is there a need to expose this class? Or you could just move the class inside the the `insertAttachment` function if it's the only place where it's needed. Local classes don't need the `private` modified since they're only visible within the local scope (e.g. a function).

service/src/main/kotlin/fi/espoo/evaka/attachment/AttachmentQueries.kt
```diff
@@ -65,26 +56,36 @@ fun Database.Transaction.insertAttachment(
...
         """
-        SELECT id, name, content_type
+        SELECT id, name, content_type, uploaded_by_employee, uploaded_by_person, application_id, income_statement_id
         FROM attachment
         WHERE id = :id
         """
     )
     .bind("id", id)
-    .mapTo<Attachment>()
+    .map { row ->
```
Gekkio, line 65: This could be simplified a bit using the query-specific row mapper technique I mentioned a couple of weeks ago :nerd_face: 

Since we only want to specialize handling of one field that has a specific type, something like this should work:

```kotlin
.registerColumnMapper(AttachmentParent::class.java) { row ->
  val applicationId = row.mapColumn<ApplicationId?>("application_id")
  val incomeStatementId = row.mapColumn<IncomeStatementId?>("income_statement_id")
  if (applicationId != null) AttachmentParent.Application(applicationId)
  else if (incomeStatementId != null) AttachmentParent.IncomeStatement(incomeStatementId)
  else AttachmentParent.None
}
.mapTo<Attachment>()
```

= "JDBI, if you encounter a need to map an AttachmentParent somewhere, use this piece of code to do it"

service/src/main/kotlin/fi/espoo/evaka/shared/security/AccessControl.kt
```diff
@@ -424,19 +497,37 @@ WHERE employee_id = :userId
         throw Forbidden("Permission denied")
     }
 
+    private inline fun <reified A> hasPermissionThroughRoles(
```
Gekkio, line 500: I'm not 100% sure, but couldn't all these functions be normal non-inline non-reified functions? Usually reified type parameters are used when there's some need to operate on the type directly (e.g. in this case calling `A::class` would require a reified function). But it looks like we're just doing simple checks based on values and functions here

Gekkio (approved):
:+1:

service/src/main/kotlin/fi/espoo/evaka/attachment/AttachmentQueries.kt
Gekkio: Never mind, the API is a bit different than I thought, so it would make things a bit too complicated

Gekkio:
```kotlin
.registerRowMapper(AttachmentParent::class.java) { rs, ctx ->
  val applicationId = ctx.mapColumn<ApplicationId?>(rs, "application_id")
  val incomeStatementId = ctx.mapColumn<IncomeStatementId?>(rs, "income_statement_id")
  if (applicationId != null) AttachmentParent.Application(applicationId)
  else if (incomeStatementId != null) AttachmentParent.IncomeStatement(incomeStatementId)
  else AttachmentParent.None
}
```

Actually, it's probably not that bad :thinking: 

:shrug:

service/src/main/kotlin/fi/espoo/evaka/attachment/AttachmentQueries.kt
akheron: Moved inside the function 👍

service/src/main/kotlin/fi/espoo/evaka/attachment/AttachmentQueries.kt
akheron: Great suggestion, changed!

PR #1398, opened 2021-09-17
EVAKA-4273 finalize preparatory decisions

jaahon:
#### Summary

* Remove (broken) part that details the required attachments for correction requests on preschool and preparatory decisions
* Fix the layout of address info on decision corrections part
* Clarify that the date following child name on decisions is the child's date of birth
* Update preparatory decision texts


mikkopiu (approved):
LGTM though a bit weird to have that SV file with Finnish texts :P

PR #1406, opened 2021-09-20
EVAKA-4214: citizen sees message attachments

Joosakur:

### DIFF
--- a/frontend/src/lib-components/molecules/FileDownloadButton.tsx
+++ b/frontend/src/lib-components/molecules/FileDownloadButton.tsx
@@ -12,7 +12,7 @@ import { UUID } from 'lib-common/types'
 import {IconDefinition} from "@fortawesome/fontawesome-svg-core";
 import {FixedSpaceRow} from "../layout/flex-helpers";
 import {FontAwesomeIcon} from "@fortawesome/react-fontawesome";
-import {faFile, faFileImage, faFilePdf} from "../../lib-icons";
+import {fileIcon} from "./FileUpload";
 
 const DownloadButton = styled.button`
   background: none;
@@ -24,18 +24,6 @@ const DownloadButton = styled.button`
   text-align: start;
   text-decoration: none;
 `
-
-function getIcon(contentType: string): IconDefinition {
-  switch (contentType) {
-    case 'pdf':
-      return faFilePdf
-    case 'jpeg':
-    case 'png':
-      return faFileImage
-    default:
-      return faFile
-  }
-}
 
 interface FileDownloadButtonProps {
   file: Attachment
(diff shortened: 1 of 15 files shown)


frontend/src/lib-components/molecules/FileDownloadButton.tsx
```diff
@@ -21,10 +25,23 @@ const DownloadButton = styled.button`
   text-decoration: none;
 `
 
+function getIcon(contentType: string): IconDefinition {
```
kimifoud, line 28: `FileUpload.tsx` exports `fileIcon` function in `lib-components/molecules`

Joosakur: oh, thanks 👍

Joosakur:
e2e test passes locally but fails in ci. need to investigate..

service/src/main/resources/db/migration/R__message_account_access_view.sql
```diff
@@ -1,22 +1,29 @@
 DROP VIEW IF EXISTS message_account_access_view;
 
-CREATE VIEW message_account_access_view(employee_id, account_id) AS (
-    SELECT employee.id AS employee_id, acc.id AS account_id
+CREATE VIEW message_account_access_view(account_id, employee_id, person_id) AS (
```
kimifoud, line 3: I'd rather leave this view as it is (or maybe rename it to `employee_account_access_view`) and just check for person access directly in the queries. Or maybe create a separate view for citizen account access. There's no use-case where we lookup with both ids at the same time and this implementation has performance implications.

service/src/main/kotlin/fi/espoo/evaka/messaging/message/MessageQueries.kt
```diff
@@ -220,9 +225,17 @@ SELECT
...
     msg.recipient_id,
     msg.recipient_name,
-    msg.recipient_account_type
+    msg.recipient_account_type,
+    (
+        SELECT coalesce(jsonb_agg(json_build_object(
+           'id', att.id,
+           'name', att.name,
+           'contentType', att.content_type
+        )) FILTER ( WHERE att.id IS NOT NULL ), '[]')
```
kimifoud, line 234: What is the `FILTER` clause needed for? 🤔

kimifoud (changes requested):

service/src/main/kotlin/fi/espoo/evaka/messaging/message/MessageQueries.kt
Joosakur: I think it otherwise produced an object with null values

kimifoud: Ach, the coalesce default should probably be `'[]'::jsonb`.

Joosakur: huh, that works

service/src/main/resources/db/migration/R__message_account_access_view.sql
```diff
@@ -20,4 +20,4 @@ CREATE VIEW message_account_access_view(employee_id, account_id) AS (
     FROM message_account acc
         JOIN daycare_group_acl gacl ON gacl.daycare_group_id = acc.daycare_group_id
     WHERE acc.active = TRUE
-);
+);
```
kimifoud, line 23: 🤓 
```suggestion
);

```

kimifoud (approved):

PR #1411, opened 2021-09-20
Citizen income to a dropdown menu

mapehe:
#### Summary

Create citizen user dropdown menu (see the attached picture). Also update styling in the language dropdown.

image

### DIFF
--- a/frontend/src/citizen-frontend/header/DesktopNav.tsx
+++ b/frontend/src/citizen-frontend/header/DesktopNav.tsx
@@ -8,11 +8,11 @@ import useCloseOnOutsideClick from 'lib-components/utils/useCloseOnOutsideClick'
 import { defaultMargins, Gap } from 'lib-components/white-space'
 import colors from 'lib-customizations/common'
 import {
-  faCheck,
   faChevronDown,
   faChevronUp,
   faLockAlt,
-  faSignIn
+  faSignIn,
+  faSignOut
 } from 'lib-icons'
 import React, { useCallback, useState } from 'react'
 import { NavLink } from 'react-router-dom'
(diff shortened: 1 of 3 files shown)


akheron:
Layout is still broken between 1000 and 1200 px 😢 

Näyttökuva 2021-9-23 kello 11 43 17

akheron:
In my opinion (and according to the ui design), the dropdown menu should be aligned to the right edge of the screen. Now it seems to be aligned to the left edge of the user's name.

Näyttökuva 2021-9-23 kello 11 46 39

mapehe:
Thanks for the feedback @akheron ! 💎 

> Layout is still broken between 1000 and 1200 px 😢
> 
> Näyttökuva 2021-9-23 kello 11 43 17

This pull request is an intermediate step and will not suffice to fix this. I'm working on a further improvement, that should completely resolve the issue, but I thought it would make sense to split it to two more manageable PRs.

> In my opinion (and according to the ui design), the dropdown menu should be aligned to the right edge of the screen. Now it seems to be aligned to the left edge of the user's name.

Thanks for the feedback, checked Figma and tbh I'm not really sure how it's supposed to work _exactly_. Tried a couple of things and, in e4b6014, settled for some simple CSS trickery to force the dropdown to be at least as wide as the person's name.

akheron:
It still looks a bit weird when the user's name is really short, because the logout item is wrapped to two lines.

Näyttökuva 2021-9-24 kello 8 30 57

My suggestion to fix all these issues:

Näyttökuva 2021-9-24 kello 8 41 24

- `position: relative` to the `div` that contains both the user name and the dropdown
- `position: absolute; right: 0; min-width: <enough>` to the dropdown

Result:

Näyttökuva 2021-9-24 kello 8 46 22
Näyttökuva 2021-9-24 kello 8 46 00

mapehe:
Thanks for the feedback @akheron.

To me, on the other hand, the case with a really long name doesn't look that great if we just have, say `width: 200px`. How about `width: 100%, min-width: 200px`?

akheron:
Well, I think `min-width: 100%` doesn't look good and makes the dropdown look different for different users, but I'm not a visual designer 🤷 😄

If you want, you could remove the wrapping `<div style={{ position: 'relative'` altogether and move the `position: relative` to the `UserMenu` root div.

mapehe:
Fair enough. Changed to `width: 200px`. Also removed the wrapping div as suggested.

akheron (approved):

PR #1412, opened 2021-09-21
Bump up service dependency versions

reynders:
#### Summary
ktlint = "0.41.0" -> "0.42.1"
flyway = "7.12.1" -> "7.15.0"
kotlin = "1.5.30" -> "1.5.31"
kotlinter = "3.4.5" -> "3.6.0"
springBoot = "2.5.3" -> "2.5.4"


Gekkio (commented):
- this only updates gradle plugins and none of the actual dependencies. Actual dependency versions are defined in evaka-bom
- message-service?
- the indentation changes required by the new ktlint version look wrong...it seems that latest version is very buggy and based on Github discussions, many other people are avoiding it. Should we wait for the next version too?

reynders:
> * this only updates gradle plugins and none of the actual dependencies. Actual dependency versions are defined in evaka-bom

Aa, that is why I could not find them from my service Idea project

> * message-service?

Bumped

> * the indentation changes required by the new ktlint version look wrong...it seems that latest version is very buggy and based on Github discussions, many other people are avoiding it. Should we wait for the next version too?

Reverted

Gekkio (approved):
:+1:

PR #1418, opened 2021-09-22
Load Espoo logos from remote storage, replace with placeholders

mikkopiu:
#### Summary

- Improve REUSE compliance with a few key changes:
    - Update `LICENSE` to reflect our REUSE compliance instead of the default license
    - Remove all the remaining city of Espoo logos from the repository and instead load them from private storage during CI builds
        - Background: with REUSE, all files require some license (also, this prevents us from using reuse-tool lint in CI without big hacks) and city of Espoo logos aren't licensed for anyone else to use
        - Replace logos with generated placeholders (very obvious)
- Move Espoo's unlicenseable customizations to new top-level directory, `espoo-customizations`
    - Create symbolic links back to where the files are actually used so that the files can be replaced in CI transparently
- CI: Download Espoo customizations from S3 with new script (similarly to Font Awesome Pro fonts)
    - This is a step that should be skipped if a forks-only workflow is implemented (and the builds should still work as-is)
- Service: load Espoo logo from template to dedupe usage
- Add missing license headers
- Add short docs for miscellaneous repo directories in top-level README
- Un-exclude Espoo logos in license scripts
    - The versions in the repo should now be placeholders that can be licensed

#### Screenshots

New placeholders:
image
image


Gekkio (approved):

mxmehl:
Hi, REUSE coordinator here! I just found this repo being REUSE compliant, thanks for the great work!

> * Background: with REUSE, all files require some license (also, this prevents us from using reuse-tool lint in CI without big hacks) and city of Espoo logos aren't licensed for anyone else to use

I am sad to hear this causes a lot of extra work. How about adding a custom license that says exactly this? REUSE mandates all files to have a license, not necessarily a Free Software license (which obviously is the preferred choice). Here you find info on how this could be achieved: https://reuse.software/faq/#custom-license

PR #1420, opened 2021-09-22
CI: don't push builds from branches to S3

mikkopiu:
#### Summary
- No reason to push branch builds to all S3 buckets, just incurs unnecessary costs


mikkopiu:
Closing, this would prevent making deployments from any changes which we are moving towards.

PR #1410, opened 2021-09-20
Improve database indexing and index use with date columns

Gekkio:

PR #1424, opened 2021-09-23
[EVAKA-4293] Prevent parallel invoice generation

Gekkio:
#### Summary

Prevent parallel invoice generation, refactor code:

- database lock blocks parallel operations. Explicit locking always adds a risk of deadlocks, but I think this is worth a try
- disable UI button during the operation to avoid multiple clicks
- simplify some JDBI mapping cases
- rename period to range in some places

*Note: depends on https://github.com/espoon-voltti/evaka/pull/1410*, so those commits are included but are not really meant to be a part of this pull request


jaahon (approved):
Everything on top of #1410 looks good 👍

PR #1448, opened 2021-09-29
Message draft receiver bugfix

mapehe:
#### Summary

Even though message draft receivers are saved and recovered correctly, a mistake in a hook causes them to clear on load.

### DIFF
--- a/frontend/src/employee-frontend/components/messages/MessageEditor.tsx
+++ b/frontend/src/employee-frontend/components/messages/MessageEditor.tsx
@@ -190,13 +190,29 @@ export default React.memo(function MessageEditor({
       const nestedAccount = nestedAccounts.find(
         (account) => account.account.id === message.sender.value
       )
-      if (nestedAccount && !isNestedGroupMessageAccount(nestedAccount)) {
-        setReceiverTree(availableReceivers)
-      } else if (nestedAccount && isNestedGroupMessageAccount(nestedAccount)) {
-        const groupId = nestedAccount.daycareGroup.id
-        const selection = getSubTree(availableReceivers, groupId)
-        if (selection) {
-          setReceiverTree(selection)
+      if (!nestedAccount) {
+        throw 'Invalid sender id'
+      } else {
+        if (!isNestedGroupMessageAccount(nestedAccount)) {
+          setReceiverTree((previousReceivers) =>
+            getSelectedBottomElements(previousReceivers).reduce(
+              (acc, id) =>
+                updateSelector(acc, { selectorId: id, selected: true }),
+              availableReceivers
+            )
+          )
+        } else {
+          const groupId = nestedAccount.daycareGroup.id
+          const selection = getSubTree(availableReceivers, groupId)
+          if (selection) {
+            setReceiverTree((previousReceivers) =>
+              getSelectedBottomElements(previousReceivers).reduce(
+                (acc, id) =>
+                  updateSelector(acc, { selectorId: id, selected: true }),
+                selection
+              )
+            )
+          }
         }
       }
     },
--- a/frontend/src/employee-frontend/components/messages/MessageEditor.tsx
+++ b/frontend/src/employee-frontend/components/messages/MessageEditor.tsx
@@ -192,27 +192,26 @@ export default React.memo(function MessageEditor({
       )
       if (!nestedAccount) {
         throw new Error('Selected sender was not found in accounts')
+      }
+      if (!isNestedGroupMessageAccount(nestedAccount)) {
+        setReceiverTree((previousReceivers) =>
+          getSelectedBottomElements(previousReceivers).reduce(
+            (acc, id) =>
+              updateSelector(acc, { selectorId: id, selected: true }),
+            availableReceivers
+          )
+        )
       } else {
-        if (!isNestedGroupMessageAccount(nestedAccount)) {
+        const groupId = nestedAccount.daycareGroup.id
+        const selection = getSubTree(availableReceivers, groupId)
+        if (selection) {
           setReceiverTree((previousReceivers) =>
             getSelectedBottomElements(previousReceivers).reduce(
               (acc, id) =>
                 updateSelector(acc, { selectorId: id, selected: true }),
-              availableReceivers
+              selection
             )
           )
-        } else {
-          const groupId = nestedAccount.daycareGroup.id
-          const selection = getSubTree(availableReceivers, groupId)
-          if (selection) {
-            setReceiverTree((previousReceivers) =>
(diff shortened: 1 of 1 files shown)


frontend/src/employee-frontend/components/messages/MessageEditor.tsx
```diff
@@ -159,14 +159,21 @@ export default React.memo(function MessageEditor({
...
-        ...old,
-        recipientAccountIds,
-        recipientIds: selected.map((s) => s.value),
-        recipientNames: selected.map((s) => s.label)
-      }))
+      if (
+        selected
+          .map((s) => s.value)
+          .sort()
+          .toString() !== message.recipientIds.sort().toString()
```
Gekkio, line 166: `.sort()` mutates the array in place. Are you 100% sure this is ok here? Mutating `selected.map((s) => s.value)` is ok, since we know calling map creates a new array, but it's not obvious to me if it's ok to mutate `message.recipientIds`.

Also, consider using something like lodash `isEqual` to compare arrays instead of comparing their string representations

mapehe:
Thanks for the feedback @Gekkio 💎 Made improvements accordingly in 54fbb7a.

frontend/src/employee-frontend/components/messages/MessageEditor.tsx
```diff
@@ -285,8 +306,7 @@ export default React.memo(function MessageEditor({
     [nestedAccounts, availableReceivers, selectedUnit.value]
   )
 
-  const sendEnabled =
-    !sending && draftState === 'clean' && areRequiredFieldsFilled(message)
```
kimifoud, line 289: Why was the `draftState === 'clean'` predicate removed?

frontend/src/employee-frontend/components/messages/MessageEditor.tsx
```diff
@@ -159,14 +160,21 @@ export default React.memo(function MessageEditor({
...
-        recipientAccountIds,
-        recipientIds: selected.map((s) => s.value),
-        recipientNames: selected.map((s) => s.label)
-      }))
+      if (
+        !isEqual(
+          new Set(message.recipientIds),
+          new Set(selected.map((option) => option.value))
+        )
+      ) {
```
kimifoud, line 168: Is this actually needed to fix the problem? This adds a nasty two way dependency, because you are updating the same state that triggers the effect.

kimifoud: Or is this just prevent an infinite loop because now the other effect relies on recipients too?

frontend/src/employee-frontend/components/messages/MessageEditor.tsx
```diff
@@ -187,20 +195,33 @@ export default React.memo(function MessageEditor({
   )
   useEffect(
     function updateReceiversOnSenderChange() {
+      const previousReceivers = message.recipientIds
```
kimifoud, line 198: IMHO it is wrong to read from `message.recipientIds` here since its value is deduced from the `receiverTree`. This creates a nasty two-way data flow.
I believe you could/should use the previous value provided by `setReceiverTree` function and get rid of the `!isEqual()` check altogether.

kimifoud (changes requested):

frontend/src/employee-frontend/components/messages/MessageEditor.tsx
```diff
@@ -187,20 +195,33 @@ export default React.memo(function MessageEditor({
...
       const nestedAccount = nestedAccounts.find(
         (account) => account.account.id === message.sender.value
       )
       if (nestedAccount && !isNestedGroupMessageAccount(nestedAccount)) {
-        setReceiverTree(availableReceivers)
+        setReceiverTree(
```
kimifoud, line 203: Moreover, this if-else-if logic could be simplified by returning early or throwing `if (!nestedAccount)`.

frontend/src/employee-frontend/components/messages/MessageEditor.tsx
mapehe: I wonder if it makes sense to require the user to wait for clean state before send, since the draft will be deleted anyway.

frontend/src/employee-frontend/components/messages/MessageEditor.tsx
mapehe: Thanks. Refactored as suggested.

frontend/src/employee-frontend/components/messages/MessageEditor.tsx
mapehe: Thanks. Refactored as suggested.

frontend/src/employee-frontend/components/messages/MessageEditor.tsx
kimifoud: > Or is this required because of the attachments?

Basically yes. Otherwise they may be out of sync.

frontend/src/employee-frontend/components/messages/MessageEditor.tsx
```diff
@@ -190,13 +190,29 @@ export default React.memo(function MessageEditor({
...
-      if (nestedAccount && !isNestedGroupMessageAccount(nestedAccount)) {
-        setReceiverTree(availableReceivers)
-      } else if (nestedAccount && isNestedGroupMessageAccount(nestedAccount)) {
-        const groupId = nestedAccount.daycareGroup.id
-        const selection = getSubTree(availableReceivers, groupId)
-        if (selection) {
-          setReceiverTree(selection)
+      if (!nestedAccount) {
+        throw 'Invalid sender id'
+      } else {
```
kimifoud, line 195: Throwing a plain string is not a good practice even though it is allowed by the language. You should throw an error with a good error message instead for debuggability. The `else` is redundant after early return.

frontend/src/employee-frontend/components/messages/MessageEditor.tsx
```diff
@@ -190,13 +190,29 @@ export default React.memo(function MessageEditor({
...
       )
-      if (nestedAccount && !isNestedGroupMessageAccount(nestedAccount)) {
-        setReceiverTree(availableReceivers)
-      } else if (nestedAccount && isNestedGroupMessageAccount(nestedAccount)) {
-        const groupId = nestedAccount.daycareGroup.id
-        const selection = getSubTree(availableReceivers, groupId)
-        if (selection) {
-          setReceiverTree(selection)
+      if (!nestedAccount) {
+        throw 'Invalid sender id'
```
kimifoud, line 194:
```suggestion
        throw new Error('Selected sender was not found in accounts')
```

kimifoud (approved):

PR #1455, opened 2021-09-30
Refactor `renderResult()`

akheron:
#### Summary

- Use translations for the active language in `renderResult()`
- Re-introduce `<UnwrapResult>`. `renderResult()` now uses it to be able to use the translations from the context. Hooks can only be called from components.
- `<UnwrapResult>` can be used directly for the cases where the loading or failure state rendering needs to be customized. `renderResult()` is more readable and should be used for the simple case.

### DIFF
--- a/frontend/src/citizen-frontend/async-rendering.tsx
+++ b/frontend/src/citizen-frontend/async-rendering.tsx
@@ -1,0 +1,22 @@
+import React from 'react'
+import { Result } from 'lib-common/api'
+import {
+  genericUnwrapResult,
+  UnwrapResultProps
+} from 'lib-components/async-rendering'
+import { useTranslation } from './localization'
+
+export function UnwrapResult<T>(props: UnwrapResultProps<T>) {
+  const t = useTranslation()
+  return genericUnwrapResult({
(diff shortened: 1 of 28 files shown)


frontend/src/citizen-frontend/async-rendering.tsx
```diff
@@ -0,0 +1,28 @@
...
+    ...props,
+    failureMessage: t.common.errors.genericGetError
+  })
+}
+
+export interface RenderResultOptions<T> {
+  result: Result<T>
+  loading?: () => React.ReactElement | null
+  failure?: () => React.ReactElement | null
+}
```
kimifoud, line 21: These interfaces are not used anywhere.

kimifoud (approved):
LGTM! 🎖️ Just remove the unused interfaces.

frontend/src/citizen-frontend/async-rendering.tsx
akheron: Ah, true! I first tried to use `renderResult` for the case where loading and failure rendering can be customized, but it was not good.

PR #1457, opened 2021-09-30
Update last_login on employee pin check

mapehe:
#### Summary

Update `last_login` on a successful employee pin check.


Gekkio (approved):
:+1:

PR #1458, opened 2021-09-30
Support for shift care units in employee mobile

jaahon:
#### Summary

Add support for shift care units in employee mobile by exluding shift care units from automatic departure markings and taking multi day attendances into account in child listing.

Also update employee mobile child list to primarily show information about child reservations or service times.


service/src/main/kotlin/fi/espoo/evaka/attendance/ChildAttendanceQueries.kt
```diff
@@ -75,22 +74,23 @@ fun Database.Transaction.insertAbsence(
...
-fun Database.Read.getChildAttendance(childId: UUID, unitId: DaycareId, date: LocalDate): ChildAttendance? {
+fun Database.Read.getChildAttendance(childId: UUID, unitId: DaycareId, now: HelsinkiDateTime): ChildAttendance? {
     // language=sql
     val sql =
         """
         SELECT id, child_id, unit_id, arrived, departed
         FROM child_attendance
-        WHERE child_id = :childId AND unit_id = :unitId AND tstzrange(:todayStart, :todayEnd, '[)') @> arrived
+        WHERE child_id = :childId AND unit_id = :unitId
+        AND (arrived::date = :date OR tstzrange(arrived, departed) @> :departedThreshold OR (arrived::date = departed::date AND arrived::date = :date))
```
Gekkio, line 84: `arrived::date = :date` is listed in the first and third ORed condition, so the third condition seems to be unnecessary

service/src/main/kotlin/fi/espoo/evaka/attendance/ChildAttendanceQueries.kt
```diff
@@ -296,12 +296,14 @@ fun Database.Read.fetchChildrenAttendances(unitId: DaycareId, date: LocalDate):
             ca.arrived,
             ca.departed
         FROM child_attendance ca
-        WHERE (ca.departed IS NULL OR ca.departed::date = :date) AND ca.child_id IN ($placedChildrenSql)
+        WHERE ca.child_id IN ($placedChildrenSql)
+        AND (ca.arrived::date = :date OR tstzrange(ca.arrived, ca.departed) @> :departedThreshold OR (ca.arrived::date = ca.departed::date AND ca.arrived::date = :date))
```
Gekkio, line 301: `ca.arrived::date = :date` is listed in the first and third ORed condition, so the third condition seems to be unnecessary

service/src/main/kotlin/fi/espoo/evaka/attendance/ChildAttendanceQueries.kt
jaahon: Yeah, you are right. I'm not sure what I was thinking anymore, so I can't say if the first condition is incomplete.

jaahon: But either way, `arrived::date = :date OR tstzrange(arrived, departed) @> :departedThreshold` should be fine here.

PR #1462, opened 2021-09-30
Add VTJ WSDL and other missing licenses

mikkopiu:
#### Summary
- Make **ALL** files REUSE compliant by fixing the last known exception: VTJ WSDLs missing licensing info (+ miscellaneous other files missing headers)
- VTJ WSDLs:
    - Some files originate from X-Road, i.e. should be under their MIT license
    - Add best guess for VTJ and Suomi.fi WSDL licenses based on licensing info found in other public code repositories (e.g. from Opetushallitus)
        - We are still awaiting official word on all of these but these are our current best guesses, and can just be fixed if these were incorrect (not like an incorrect license would be valid)
- Add short README about the whole vtjclient directory
    - Document XSD and WSDL file origins in comments
    - Drop the vtj-xsd files that haven't been used anywhere in a while
- Remove exceptions from license scripts
    - And we are finally ready to implement license checks in CI


service/vtjclient/src/main/resources/wsdl/xml.xsd
```diff
@@ -1,5 +1,12 @@
 <?xml version='1.0'?>
 <?xml-stylesheet href="../2008/09/xsd.xsl" type="text/xsl"?>
+
+<!--
+SPDX-FileCopyrightText: 2017-2020 City of Espoo
```
ilja-git, line 5: Should we use 2021 here?

mikkopiu: Yes, fixing

service/vtjclient/README.md
```diff
@@ -0,0 +1,19 @@
+<!--
+SPDX-FileCopyrightText: 2017-2020 City of Espoo
```
ilja-git, line 2: 2021 instead 2020

mikkopiu: That's what I get for copy-pasting...

mikkopiu:
Fixed the dates and added a few more missing license headers

ilja-git (approved):
LGTM

PR #1465, opened 2021-10-01
EVAKA-HOTFIX Fix playwright image build

akheron:
#### Summary

Remove `deb.nodesource.com` from apt sources, because its sertificate is outdated and we don't need it as a part of our own build.


mikkopiu (approved):
Approved as a temporary fix, hopefully can be reverted soon (once base image updates)

PR #1468, opened 2021-10-01
CI: REUSE compliance checks

mikkopiu:
#### Summary
- CI: implement REUSE license compliance checking in CI
    - Use the official reuse-tool image as the executor
        - Using our own wrapper should no longer really be necessary as there aren't any custom exceptions -> no need for custom logic
    - Now that all files have licensing data available, this can finally be added to CI so that license headers will never be forgotten again
    - Downside here is that now CI will nag on every commit until the headers have been added but the best solution honestly will be to just either configure the `add-license-headers.sh` script to pre-commit hooks or in IDEs -- a small pain now but saves a big pain later
- CI: update all images
    - Nothing breaking, just keeping up-to-date
- CI: use smaller executors where possible
    - And be explicit with `resource_class`
    - Also, fix machine executors' `resource_class` configs; they worked due to backwards compatibility but were still in the wrong place
- Project has now been registed with the REUSE API for active compliance checking (and a neat badge for our README!): https://api.reuse.software/info/github.com/espoon-voltti/evaka

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -44,7 +44,7 @@ aliases:
   - &remote_docker_version "20.10.7"
   - &yarn_version '1.22.\*'
   - &shellcheck_image_version v0.7.2
-  - &reuse_image fsfe/reuse:0.13.0
+  - &reuse_image fsfe/reuse:0.13.0 # NOTE: Update bin/add-license-headers.sh to match
   - &nodejs_image << pipeline.parameters.nodejs_image >>
   - &openjdk_image cimg/openjdk:11.0
   - &ubuntu_machine_image ubuntu-2004:202107-02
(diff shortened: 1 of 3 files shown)


akheron (approved):

.circleci/config.yml
```diff
@@ -817,11 +817,11 @@ jobs:
...
       - restore_repo
-      - setup_authenticated_remote_docker
+      - run: reuse lint
       - run:
-          name: Check REUSE licensing status
-          command: |
-            ./bin/add-license-headers.sh --lint-only
+          when: on_fail
+          name: Install dependencies
+          command: apk add curl jq
```
akheron, line 839: Why is stuff installed on failure? 🤔

mikkopiu: The silly Slack orb needs `curl` and `jq`installed and those parts only run on failure -> no need to waste time otherwise. The Slack orb has auto-install for those deps during its run for most other systems (even Homebrew!) but for some reason not for `apk` (although it even has a branch for it).

PR #1469, opened 2021-10-01
Reduce accidental misuse of Database API

Gekkio:
#### Summary

`db: Database` as a parameter is only meant for functions that need to connect to the database multiple times. In all other cases, `db: Database.Connection` is the right thing to use.

### DIFF
--- a/service/src/main/kotlin/fi/espoo/evaka/shared/db/Database.kt
+++ b/service/src/main/kotlin/fi/espoo/evaka/shared/db/Database.kt
@@ -56,20 +56,6 @@ class Database(private val jdbi: Jdbi) {
     }
 
     /**
-     * Opens a database connection in read mode, runs the given function, and closes the connection.
-     *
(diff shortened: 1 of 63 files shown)


akheron (approved):

service/src/main/kotlin/fi/espoo/evaka/shared/db/Database.kt
```diff
@@ -41,45 +41,40 @@ class Database(private val jdbi: Jdbi) {
...
-     */
-    fun <T> read(f: (db: Read) -> T): T = connect { it.read(f) }
-
-    /**
-     * Opens a database connection, runs the given function within a transaction, and closes the connection.
-     *
-     * Throws `IllegalStateException` if a connection is already open
-     */
-    fun <T> transaction(f: (db: Transaction) -> T): T = connect { it.transaction(f) }
-
```
akheron, line 71: 🙌

PR #1477, opened 2021-10-04
Use JSX elements instead of parameterless functions + support for l10n customizations

apeltola:
Modified l10n files so that parameterless functions are replaced with direct JSX elements.
React.isValidElement used for detecting JSX elements in localization files.


frontend/src/lib-customizations/common.tsx
```diff
@@ -50,4 +52,19 @@ export const absenceColours = {
   PRESENCE: colors.greyscale.white
 }
 
+/* eslint-disable @typescript-eslint/no-explicit-any */
+export const translationsMergeCustomizer = (
```
apeltola, line 56: This is the most important change: allows overriding JSX elements in city specific implementations.

akheron: You can avoid the `eslint-disable` like this:

```
 export const translationsMergeCustomizer = (
  origValue: Record<string, unknown>,
  customizedValue: unknown
): unknown => {
```

akheron (changes requested):

akheron:
Please merge on top of the latest master to get the build to pass

frontend/src/lib-customizations/common.tsx
apeltola: Maybe even better to have it like this?

```
export const translationsMergeCustomizer = (
  origValue: Record<string, unknown>,
  customizedValue: Record<string, unknown>
): Record<string, unknown> | undefined => {
```

akheron: Yeah, sure 👍

akheron (approved):

PR #1492, opened 2021-10-08
Replace some nulls with defaults in the raw report

Gekkio:
#### Summary

The semantics were previously accidentally changed, because the old code used manual mapping to Java primitive types (`rs.getBoolean`, `rs.getDouble`), which return `false`/`0.0` if the column is NULL. The new automated mapping did not do this, because the fields were nullable in the target data class.

So, remove nullability to clearly communicate we don't want empty cells in the report, and explicitly coalesce to defaults in the SQL query.


jaahon:
> So something like `coalesce(sno.some_flag, default_sno.some_flag, false)`? I noticed the default service need option was used in some other places in the same query but not here for some reason

Probably, yeah. I guess the only question is should the default be used only when there are no other options or in other cases as well?

PR #1508, opened 2021-10-12
Retry E2E tests

ilja-git:
#### Summary

Retry E2E tests


tuniemela (approved):
Why they fail in the first place? Anyway, if this works to the problem, let's use it :+1:

PR #1531, opened 2021-10-15
Add evaka_user table and change most someaction_by columns to point to it

Gekkio:
#### Summary

* evaka_user rows are normally created during login (citizen/employee/mobile device)
* cache a textual name in the table, so we don't need complicated joins in most use cases
* `fee_decision` / `voucher_value_decision` table `approved_by`, and `service_need` table `confirmed_by` were not changed, because multiple places assume separate `first_name` and `last_name` fields, and changing this is complicated
* add foreign key indexes to tables. Not strictly necessary, but it's a good default, and if we ever delete from `evaka_user`, this avoids full table scans in all related tables
* deduplicated `resetDatabase()` function in Kotlin code. There's no need to have two, and now we can easily add the system user after database resets
* add new `evaka_user` rows for all `employee` / `person` table rows referenced from any column we're changing here
* add `UNKNOWN` rows for legacy `absence` table rows which only have a textual name and no "real" foreign key reference


jaahon (approved):
Merge with caution 😱

PR #1536, opened 2021-10-19
Notify the user when the app has changed

akheron:
#### Summary

You can see the toast if you start the frontend like this:

```
APP_COMMIT=foobar yarn dev
```


frontend/src/lib-components/molecules/modals/ReloadNotification.tsx
```diff
@@ -0,0 +1,66 @@
+// SPDX-FileCopyrightText: 2017-2021 City of Espoo
```
akheron, line 1: This file should be somewhere else than `molecules/modals` 🤔

akheron: Moved to `molecules`

reynders (approved):

PR #1551, opened 2021-10-22
Scroll to message thread bottom

mapehe:
#### Summary

Scroll to the bottom upon opening a message thread so that the newest message is displayed right away.

### DIFF
--- a/frontend/src/citizen-frontend/messages/ThreadView.tsx
+++ b/frontend/src/citizen-frontend/messages/ThreadView.tsx
@@ -11,6 +11,7 @@ import React, {
   useContext,
   useEffect,
   useMemo,
+  useRef,
   useState
 } from 'react'
 import styled from 'styled-components'
@@ -27,7 +28,6 @@ import { MessageContext } from './state'
 import { formatDate } from 'lib-common/date'
 import { faReply } from '@fortawesome/free-solid-svg-icons'
 import InlineButton from 'lib-components/atoms/buttons/InlineButton'
-import { MessageType } from 'lib-common/generated/enums'
 import HorizontalLine from 'lib-components/atoms/HorizontalLine'
 import { FixedSpaceColumn } from 'lib-components/layout/flex-helpers'
 import FileDownloadButton from 'lib-components/molecules/FileDownloadButton'
@@ -43,14 +43,33 @@ const TitleRow = styled.div`
     margin-top: ${defaultMargins.L};
   }
 `
+
+const StickyTitleRow = styled(TitleRow)`
+  position: sticky;
+  top: 0;
+  padding: 15px;
+  background: white;
+  max-height: 100px;
+`
+
+const StickyTitleRowTitle = styled(H2)`
+  top: 0;
+  padding: 15px;
+  background: white;
+  overflow: scroll;
+  max-height: 100px;
+`
+
 const SenderName = styled.div`
   font-weight: ${fontWeights.semibold};
 `
+
 const SentDate = styled.div`
   font-size: 14px;
   font-weight: ${fontWeights.semibold};
   color: ${colors.greyscale.dark};
 `
+
 const MessageContent = styled.div`
   padding-top: ${defaultMargins.s};
   white-space: pre-line;
@@ -60,26 +79,12 @@ const ReplyToThreadButton = styled(InlineButton)`
   padding-left: 28px;
 `
 
-function SingleMessage({
-  title,
-  type,
-  message
-}: {
-  message: Message
-  type?: MessageType
-  title?: string
-}) {
+function SingleMessage({ message }: { message: Message }) {
   const i18n = useTranslation()
   const { setErrorMessage } = useContext(OverlayContext)
 
   return (
     <MessageContainer>
-      {title && type && (
-        <TitleRow>
-          <H2 data-qa="thread-reader-title">{title}</H2>
-          <MessageTypeChip type={type} labels={i18n.messages.types} />
-        </TitleRow>
-      )}
       <TitleRow>
         <SenderName>{message.sender.name}</SenderName>
         <SentDate>{formatDate(message.sentAt)}</SentDate>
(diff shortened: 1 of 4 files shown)


frontend/src/employee-frontend/components/messages/SingleThreadView.tsx
```diff
@@ -192,29 +216,54 @@ export function SingleThreadView({
...
-            title={idx === 0 ? title : undefined}
-            type={idx === 0 ? type : undefined}
-            index={idx}
-            onAttachmentUnavailable={onAttachmentUnavailable}
-          />
+          <>
+            {!replyEditorVisible && idx === messages.length - 1 && (
+              <div style={{ position: 'relative' }}>
+                <span
+                  style={{ position: 'absolute', top: '-100px' }}
```
mapehe, line 229: This `-100px` is a little (dirty?) hack to scroll correctly to the top of the last message, if it's very long.

frontend/src/citizen-frontend/messages/ThreadView.tsx
```diff
@@ -142,6 +136,14 @@ export default React.memo(function ThreadView({
 
   useEffect(() => setReplyEditorVisible(false), [threadId])
 
+  const messagesEndRef = useRef<HTMLSpanElement>(null)
+  const scrollToBottom = () => {
+    messagesEndRef.current?.scrollIntoView({ behavior: 'smooth' })
+  }
+  useEffect(() => {
+    scrollToBottom()
+  })
```
jaahon, line 145: This `useEffect` should have `[]` as its dependency array so that the view isn't scrolled every time it's rerendered.

frontend/src/citizen-frontend/messages/ThreadView.tsx
```diff
@@ -202,8 +215,10 @@ export default React.memo(function ThreadView({
               data-qa="message-reply-editor-btn"
               text={i18n.messages.replyToThread}
             />
+            <Gap size={'m'} />
           </>
         ))}
+      {replyEditorVisible && <span ref={messagesEndRef} />}
```
jaahon, line 221: Remove this.

frontend/src/citizen-frontend/messages/ThreadView.tsx
```diff
@@ -43,6 +43,14 @@ const TitleRow = styled.div`
     margin-top: ${defaultMargins.L};
   }
 `
+const StickyTitleRow = styled(TitleRow)`
+  position: sticky;
+  top: 0;
+  padding: 15px;
+  background: white;
+  height: 100px;
```
jaahon, line 51: Use a max-height instead of a fixed height if you want to make sure that the title row doesn't take up all the space. Also, you need to give them to the `<h2>` element.

frontend/src/citizen-frontend/messages/ThreadView.tsx
```diff
@@ -171,13 +173,24 @@ export default React.memo(function ThreadView({
...
-          key={message.id}
-          message={message}
-          title={idx === 0 ? title : undefined}
-          type={idx === 0 ? type : undefined}
-        />
+        <>
+          {idx === messages.length - 1 && !replyEditorVisible && (
+            <div style={{ position: 'relative' }}>
+              <span
+                style={{ position: 'absolute', top: '-100px' }}
```
jaahon, line 187: Instead of using a fixed amount of `100px` you should calculate the offset from `<StickyTitleRow>`'s box height. Also, consider using styled-components for consistency.

frontend/src/citizen-frontend/messages/ThreadView.tsx
mapehe: I actually think it makes sense to scroll again on re-render, because re-render should only happen when changing threads or when a new message is sent. In both cases it makes sense to scroll again.

frontend/src/citizen-frontend/messages/ThreadView.tsx
mapehe: This was committed intentionally. I think it scrolls nicely to the bottom when the user wants to create a new message or a new message is sent.

frontend/src/citizen-frontend/messages/ThreadView.tsx
mapehe: Thanks, will do

frontend/src/citizen-frontend/messages/ThreadView.tsx
mapehe: Thanks, will do.

frontend/src/citizen-frontend/messages/ThreadView.tsx
jaahon: Ah ok. Sorry, I just misunderstood the intention.

frontend/src/citizen-frontend/messages/ThreadView.tsx
jaahon: Oh ok, I guess it's just preference so it's fine then 😄

Should you mark those conditions explicitly since there's no guarantee that the component doesn't rerender on some other condition in the future?

frontend/src/citizen-frontend/messages/ThreadView.tsx
jaahon: Yeah, maybe. I don't know how likely it is that the title is super long, but very long titles can block the whole view since the header is now sticky.

mapehe: Hmm yeah, how about a scrollable title box with a max-height? Too wild?

jaahon: Yeah, that's what I tried to suggest in my first comment 😄

mapehe: Oops sorry... need to work on my reading comprehension I guess 😅

frontend/src/citizen-frontend/messages/ThreadView.tsx
mapehe: Added `[messages, replyEditorVisible]`, which seems to work pretty well.

jaahon (approved):

PR #1582, opened 2021-10-29
Remove redundant daycare.organization_oid

reynders:
#### Summary
- Replace daycare.oph_organization_oid with env property because it is a per municipal setting
- Replace varda_organizer with env properties because those are per municipal constant information
- Remove daycare dev data and related information because the data is stale and really hard to refactor
- Move oph related config to separate Config


Gekkio (commented):
Overall looks good! :+1: 
There's one thing I'm not so sure about: all the oph env variables are mandatory *and* we provide Espoo by default, even in Trevaka. IMHO it would be better if those variables would be mandatory only if Varda/Koski is used, and there would be no defaults.

Here's what I would do:
- add new `OphEnv` type, similar to the other `*Env` types and move the oph fields there. The fields can be non-null, because the whole `OphEnv` bean will be lazy and is only used if needed
- add `OphEnv` bean to `EnvConfig.kt`
- make Koski/Varda integration read the fields from `OphEnv`
- add Espoo default values by doing either A or B:

A) Create `application-espoo_evaka.yaml`, which should be picked up automatically in Espoo only, and add the oph field defaults there

or

B) Add defaults using `MapPropertySource` bean to `EspooConfig`. Something like:

```
@Bean
fun espooDefaults() = MapPropertySource("espoo-defaults", mapOf(
  "evaka.foo.bar" to "value",
))
```

service/src/main/kotlin/fi/espoo/evaka/EvakaEnv.kt
```diff
@@ -362,6 +362,22 @@ data class ScheduledJobsEnv(val jobs: Map<ScheduledJob, ScheduledJobSettings>) {
     }
 }
 
+data class OphEnv(
+    val ophOrganizerOid: String,
```
Gekkio, line 366: Nit: I think the `oph` prefix in the fields is now unnecessary since the fields are grouped in a type that provides some context.

In other words:

```kotlin
data class Person(
  val firstName: String
  val lastName: String
)
```

vs

```kotlin
data class Person(
  val personFirstName: String
  val personLastName: String
)
```

Gekkio (approved):

PR #1589, opened 2021-11-01
EVAKA time inputs

jaahon:
#### Summary

Replace native browser time inputs with our own `TimeInput`. Also shorten the time validation info texts to not break the layout.


kimifoud (approved):
🏅 RIP `TimeWrapper`

PR #1596, opened 2021-11-02
Fix promise rejection error when fetching auth status in mobile frontend

kimifoud:
#### Summary

- Move auth status from App level to UserContext
- Refactor auth status to use Result in order to handle errors gracefully and display loading status
- Fix FullScreenDimmedSpinner to actually use the whole screen estate
- Simplify RedirectToMainPage

### DIFF
--- a/frontend/src/employee-mobile-frontend/components/ensureAuthenticated.tsx
+++ b/frontend/src/employee-mobile-frontend/components/ensureAuthenticated.tsx
@@ -2,9 +2,9 @@
 //
 // SPDX-License-Identifier: LGPL-2.1-or-later
 
+import { FullScreenDimmedSpinner } from 'lib-components/molecules/FullScreenDimmedSpinner'
 import React, { ComponentType, useContext } from 'react'
 import { Redirect } from 'react-router-dom'
-import { FullScreenDimmedSpinner } from 'lib-components/molecules/FullScreenDimmedSpinner'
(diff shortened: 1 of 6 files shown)


frontend/src/employee-mobile-frontend/components/ensureAuthenticated.tsx
```diff
@@ -4,16 +4,23 @@
...
   return function Authenticated(props: P) {
-    const { loggedIn } = useContext(UserContext)
+    const { loggedIn, user } = useContext(UserContext)
 
     if (!loggedIn) {
       return <Redirect to="" />
     }
 
-    return <Component {...props} />
+    const showSpinner = user.isLoading || (user.isSuccess && user.isReloading)
```
akheron, line 18: Are you sure you want to show the full screen spinner when reloading the user? 🤔

akheron (approved):

frontend/src/employee-mobile-frontend/components/ensureAuthenticated.tsx
kimifoud: Well, tbh we can get rid of it 😄

kimifoud: Come to think of it, the loading state and spinner is already handled in the landing page component. We can just remove the spinner from here altogether.

PR #1599, opened 2021-11-02
Enable Java datadog profiling

ilja-git:
#### Summary

Enable Java Datadog profiling using environment variable `DD_PROFILING_ENABLED`. Requires Datadog agent running on AWS node.

### DIFF
--- a/service/Dockerfile
+++ b/service/Dockerfile
@@ -5,17 +5,22 @@
 FROM evaka-base:latest
 
 USER root
-RUN set -eux \
-    && apt-get update && apt-get -y --no-install-recommends install \
+RUN apt-get update \
+ && apt-get -y --no-install-recommends install \
        curl \
        openjdk-11-jre-headless \
-    && rm -rf /var/lib/apt/lists/*
+ && rm -rf /var/lib/apt/lists/*
+
+ARG DD_JAVA_AGENT_VERSION="0.89.0"
+ARG DD_JAVA_AGENT_SHA256="432c110ddd4b8d769bce394b9df62fe5c627e9d1fb30250fb2af9c034569abb3"
+
+RUN curl -sSfo /opt/dd-java-agent.jar "https://repo1.maven.org/maven2/com/datadoghq/dd-java-agent/${DD_JAVA_AGENT_VERSION}/dd-java-agent-${DD_JAVA_AGENT_VERSION}.jar" \
+ && echo "${DD_JAVA_AGENT_SHA256}  /opt/dd-java-agent.jar" | sha256sum -c -
 
 USER evaka
 WORKDIR /home/evaka
 
 COPY --chown=evaka:evaka "entrypoint.sh" "entrypoint.sh"
-RUN chmod +x "entrypoint.sh"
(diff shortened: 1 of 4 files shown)


service/Dockerfile
```diff
@@ -6,21 +6,22 @@ FROM evaka-base:latest
...
 
 COPY --chown=evaka:evaka "entrypoint.sh" "entrypoint.sh"
-RUN chmod +x "entrypoint.sh"
 COPY --chown=evaka:evaka "target/org/" "org/"
 COPY --chown=evaka:evaka "target/META-INF" "META-INF/"
 COPY --chown=evaka:evaka "target/BOOT-INF/lib" "BOOT-INF/lib/"
 COPY --chown=evaka:evaka "target/BOOT-INF/classes" "BOOT-INF/classes/"
 
+RUN curl -sSfo dd-java-agent.jar "https://repo1.maven.org/maven2/com/datadoghq/dd-java-agent/0.9.0/dd-java-agent-0.9.0.jar"
+
```
kimifoud, line 24: `0.9.0` seems to be a 2018 version while latest is `0.89.0`

service/entrypoint.sh
```diff
@@ -17,5 +17,14 @@ fi
 
 # Run as exec so the application can receive any Unix signals sent to the container, e.g.,
 # Ctrl + C.
-# shellcheck disable=SC2086
-exec java -cp . -server $JAVA_OPTS org.springframework.boot.loader.JarLauncher "$@"
+if [ "${DD_PROFILING_ENABLED:-false}" = "true" ]; then
+  export DD_ENV="${DD_ENV:-invalid}"
```
kimifoud, line 21: Where is `DD_ENV` defined? Should this use `VOLTTI_ENV` or something?

service/Dockerfile
ilja-git: Good point. I'll change that.

service/entrypoint.sh
ilja-git: DD_ENV comes environment variables. 

https://docs.datadoghq.com/tracing/setup_overview/setup/java/?tab=containers

We could use VOLTTI_ENV in future insted DD_ENV.

This is still draft to get build to test the metric collection from nodes https://docs.datadoghq.com/agent/amazon_ecs/?tab=awscli that is needed for posting data to datadog.

kimifoud: Yeah I was just curious and gave some drive-by comments, no worries 🙂 

`DD_ENV` is supposed to be set by the user, right? Doesn't that line try to assign the current `DD_ENV` value to `DD_ENV` itself, thus always ending up as `invalid`..?

ilja-git: Thanks for input. Good points. :slightly_smiling_face: 

`export DD_ENV="${DD_ENV:-invalid}"` is bash trick. So it will set `DD_ENV` to `DD_ENV` when set and if not set then to `invalid`. I changed default value to it to `VOLTTI_ENV` environment variable as you suggested as it might work out of the box without need to define new variable in ecs-definition.

tuniemela (approved):

PR #1601, opened 2021-11-03
Fix flaky test

reynders:

### DIFF
--- a/frontend/src/e2e-playwright/specs/6_mobile/child-attendances.spec.ts
+++ b/frontend/src/e2e-playwright/specs/6_mobile/child-attendances.spec.ts
@@ -49,18 +49,12 @@ beforeEach(async () => {
   listPage = new MobileListPage(page)
   childPage = new MobileChildPage(page)
   childAttendancePage = new ChildAttendancePage(page)
-
-  const mobileSignupUrl = await pairMobileDevice(
-    employee.data.id!, // eslint-disable-line
-    fixtures.daycareFixture.id
-  )
-  await page.goto(mobileSignupUrl)
 })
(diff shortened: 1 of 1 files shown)


frontend/src/e2e-playwright/specs/6_mobile/child-attendances.spec.ts
```diff
@@ -86,6 +86,7 @@ const checkAbsenceTypeSelectionButtonsExistance = async (
   arrivalTime = '08:15',
   departureTime = '16:00'
 ) => {
+  await page.reload()
```
kimifoud, line 89: 🤔

jaahon: Would a `page.goto` to the list page be more clear in intent? @kimifoud

kimifoud: Should the fixtures be created before the actual test and first navigation?

kimifoud (changes requested):
I would still want to get rid of the redundant `page.reload()` call.

Maybe remove and extract the following code in `beforeEach`
```
async function pairDeviceAndLoadPage() {
  const mobileSignupUrl = await pairMobileDevice(
    employee.data.id!, // eslint-disable-line
    fixtures.daycareFixture.id
  )
  await page.goto(mobileSignupUrl)
}
```

and call it in each test after `createPlacement()`. Maybe as the first thing in `checkAbsenceTypeSelectionButtonsExistence`.

kimifoud (approved):

PR #1623, opened 2021-11-05
Evaka 4000 varda error report add created

reynders:
#### Summary
- Show varda error age in varda errors report.
- Age is calculated from either  
-- varda_reset_child.created if reset_timestamp is null (e.g. reset failed)
-- varda_service_need.created otherwise


hruoho (approved):
I don't know what have been discussed about the titles etc. but technically this seems fine 👍

PR #1631, opened 2021-11-09
Use design system margins

kimifoud:
#### Summary

Add missing 40px "XL" margin between 32 and 48. New X5L is now 120px.
All current margins of XL and above are now slightly narrower, but seemed to be good all around.

### DIFF
--- a/frontend/src/citizen-frontend/map/UnitListItem.tsx
+++ b/frontend/src/citizen-frontend/map/UnitListItem.tsx
@@ -55,7 +55,7 @@ const Wrapper = styled.div`
 
   &:after {
     content: '';
-    width: calc(100% - ${dM.XXL});
+    width: calc(100% - ${dM.X3L});
     background: ${colors.greyscale.lighter};
     height: 1px;
     display: block;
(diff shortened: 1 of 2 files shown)


reynders (approved):

jaahon (commented):
Every place that uses margins larger than `L` is going to have different margins now. Is that intended?

Edit: never mind, I should read before I comment 😂

kimifoud:
> Every place that uses margins larger than L is going to have different margins now. Is that intended?
> Edit: never mind, I should read before I comment 😂

The change from `48px` to `40px` looked good to me in all places that I checked. It was mainly used in `Gap`s and `FixedSpaceColumn` spacing.

Same goes for `64px` to `48px` change. The previous 64 pixel `XXL` spacing was only used on a couple of employee pages and the `BaseModal`. The thinner margin looks better imo.

I couldn't find any broken layouts after this change.

jaahon:
> > Every place that uses margins larger than L is going to have different margins now. Is that intended?
> > Edit: never mind, I should read before I comment 😂
> 
> The change from `48px` to `40px` looked good to me in all places that I checked. It was mainly used in `Gap`s and `FixedSpaceColumn` spacing.
> 
> Same goes for `64px` to `48px` change. The previous 64 pixel `XXL` spacing was only used on a couple of employee pages and the `BaseModal`. The thinner margin looks better imo.
> 
> I couldn't find any broken layouts after this change.

Alright :shipit:

PR #1633, opened 2021-11-09
Show partner name in invoicing decision information

reynders:
#### Summary
- If fee decision has partner set, and all the children in the decision are both parents' children, show the partner in details UI and .pdf
- If voucher value decision has partner set, and child in the decision is both parent's child, show the partner in details UI and .pdf
Screenshot 2021-11-12 at 11 56 38


frontend/src/employee-frontend/components/voucher-value-decision/VoucherValueDecisionHeading.tsx
```diff
@@ -100,6 +102,18 @@ export default React.memo(function VoucherValueDecisionHeading({
               </Link>
             )
           },
+          ...(partner
```
jaahon, line 105: This should also have the `partner && isElementaryFamily` check.

jaahon (commented):
I think moving the `isElementaryFamily` check closer to the actual database query would be a good idea. I feel like now it's too easy to forget.

jaahon (approved):

PR #1635, opened 2021-11-10
Whitelist and validate attachment file extensions

kimifoud:
#### Summary

- Model allowed/known contentTypes as an enum class
- File extension must match the content type
- Render contentType/extension related error messages in the file upload component

kuva

### DIFF
--- a/service/src/main/kotlin/fi/espoo/evaka/attachment/AttachmentsController.kt
+++ b/service/src/main/kotlin/fi/espoo/evaka/attachment/AttachmentsController.kt
@@ -9,8 +9,10 @@ import fi.espoo.evaka.BucketEnv
 import fi.espoo.evaka.EvakaEnv
 import fi.espoo.evaka.application.ApplicationStateService
 import fi.espoo.evaka.application.utils.exhaust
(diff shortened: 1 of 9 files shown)


service/src/main/kotlin/fi/espoo/evaka/attachment/AttachmentsController.kt
```diff
@@ -265,19 +277,18 @@ class AttachmentsController(
...
+            ContentType.PDF,
+            ContentType.MSWORD,
+            ContentType.MSWORD_DOCX,
+            ContentType.OPEN_DOCUMENT_TEXT,
+            ContentType.TIKA_MSOFFICE,
+            ContentType.TIKA_OOXML,
+        )
 
-    val pedagogicalDocumentAllowedAttacmentContentTypes = defaultAllowedAttachmentContentTypes + listOf(
-        "video/*",
```
reynders, line 280: The explicit list of allowed video and audio formats is more restricted than this, but I agree it is better to have those listed out and then updated when real need arises

reynders (approved):

PR #1639, opened 2021-11-10
EVAKA-HOTFIX better props typing for Select

jaahon:
#### Summary

Fix `<Select/>` to require passing `getItemValue` prop when the type parameter is not a subtype of the type `string | number`.


kimifoud (approved):
💎

PR #1640, opened 2021-11-10
Vasu translations

zch:
#### Summary
Show all sections of the Vasu document in the language of the document. I.e. even if the UI is in Finnish, we want to show the document sections translated to Swedish if the document language is Swedish.


frontend/src/lib-customizations/espoo/employee/assets/i18n/vasu/fi.ts
```diff
@@ -0,0 +1,47 @@
+// SPDX-FileCopyrightText: 2017-2020 City of Espoo
```
kimifoud, line 1:
```suggestion
// SPDX-FileCopyrightText: 2017-2021 City of Espoo
```

🤓

frontend/src/employee-frontend/components/vasu/templates/VasuTemplateEditor.tsx
```diff
@@ -65,6 +66,16 @@ export default React.memo(function VasuTemplateEditor() {
 
   const readonly = !(template.isSuccess && template.value.documentCount === 0)
 
+  const translations = useMemo(
+    () =>
+      template
+        ? (vasuTranslations[
+            template.getOrElse({ language: 'FI' }).language.toLowerCase()
+          ] as VasuTranslations)
+        : vasuTranslations.fi,
```
kimifoud, line 75: `template` is always defined so the ternary is redundant. How about 

```ts
vasuTranslations[
    template.map((t) => t.language.toLowerCase() as Lang).getOrElse('fi')
]
```
Would be even better if we had a type-safe `lang` to `vasu lang` mapping fn somewhere.

frontend/src/employee-frontend/components/vasu/sections/VasuHeader.tsx
```diff
@@ -36,18 +37,26 @@ const Confidential = styled.div`
...
     },
     documentState,
-    templateName
+    templateName,
+    language
   }
 }: Props) {
   const { i18n } = useTranslation()
+  const translations = vasuTranslations[
+    language.toLowerCase()
```
kimifoud, line 57: Come to think of it, the language mapping function would be very nice because the vasu language might not always be a superset of UI languages, and the type casts are much harder to find when adding new languages.

frontend/src/lib-customizations/employee.tsx
```diff
@@ -31,7 +34,7 @@ export {
   unitProviderTypes
 }
 
-export type Lang = 'fi'
+export type Lang = 'fi' | 'sv'
```
jaahon, line 37: I think it would make more sense to not add `'sv'` as a "UI language" since it doesn't seem to provide any benefit. What do you think?

frontend/src/employee-frontend/components/vasu/sections/VasuHeader.tsx
kimifoud: `function langToVasuLang(language: Lang): VasuLanguage` with explicit mapping and `fi` as fallback?

kimifoud (changes requested):
Looks good. Let's just get rid of the type casting.

frontend/src/employee-frontend/components/vasu/sections/VasuHeader.tsx
zch: Excellent point

frontend/src/lib-customizations/employee.tsx
zch: Combined with Kim's comments I think you might be right that it would be a good idea to not add Swedish here 👍

frontend/src/employee-frontend/components/vasu/sections/VasuHeader.tsx
zch: Reworked it a bit so that instead of having a mapping function, the `language` of the VasuDocument / VasuTemplate is directly the key to `vasuTranslations`

kimifoud (approved):
👍

PR #1641, opened 2021-11-10
Fix messaging bulletin regression

mapehe:


hruoho:
What's the tldr for this? How to verify this?

hruoho (approved):
Seems legit, went through this in a call. Agreed to add a few e2e-tests in another PR.

PR #1649, opened 2021-11-12
Refresh employee's unread message counts periodically

kimifoud:
#### Summary

- Add `useInterval` hook for fetching unread message counts every minute
- Instead of manually tracking the unread count after opening the thread, just mark the thread read and refresh unread counts
- Simplify `unreadCount` logic in `Header`
- Also load message accounts only in the message context, do not expose refresh callback
  - They are always needed in the header
  - There is no use fetching them again after message page is mounted

### DIFF
--- a/frontend/src/lib-common/utils/useAxiosResponseInterceptor.ts
+++ b/frontend/src/lib-common/utils/useAxiosResponseInterceptor.ts
@@ -2,23 +2,23 @@
 //
 // SPDX-License-Identifier: LGPL-2.1-or-later
 
-import axios, { AxiosInstance } from 'axios'
-import { useEffect, useState } from 'react'
+import axios, { AxiosError, AxiosInstance, AxiosResponse } from 'axios'
+import { useEffect } from 'react'
 
-export function useAxiosResponseInterceptor(client: AxiosInstance): {
-  lastResponse: number
-} {
-  const [lastResponse, setLastResponse] = useState(Date.now())
-
+export function useAxiosResponseInterceptor(
+  client: AxiosInstance,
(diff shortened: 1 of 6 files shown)


frontend/src/lib-common/utils/usePeriodicRefresh.ts
```diff
@@ -0,0 +1,56 @@
...
+    (err: unknown) => {
+      if (axios.isAxiosError(err) && err.response) {
+        setLastResponse(Date.now())
+      }
+      return Promise.reject(err)
+    }
+  )
+  return () => {
+    client.interceptors.response.eject(id)
+  }
```
jaahon, line 55: You need to wrap this in a `useEffect` if you want the interceptor eject to happen automatically on unmount.

frontend/src/lib-common/utils/useAxiosResponseInterceptor.ts
```diff
@@ -0,0 +1,32 @@
+// SPDX-FileCopyrightText: 2017-2021 City of Espoo
+//
+// SPDX-License-Identifier: LGPL-2.1-or-later
+
+import axios, { AxiosInstance } from 'axios'
+import { useEffect, useState } from 'react'
+
+export function useAxiosResponseInterceptor(client: AxiosInstance): {
+  lastResponse: number
+} {
```
jaahon, line 10: I don't know if it makes sense to have this as an exported custom hook as it is. I would either move it to `usePeriodicRefresh` and not export it or change it so that it receives a callback as a parameter and returns its result so that it could be reused in cases where you might need a response interceptor to do something.

jaahon (approved):

PR #1654, opened 2021-11-15
Fix child attendances to round the clock backup care

jaahon:
#### Summary

Slightly modified version of https://github.com/espoon-voltti/evaka/pull/1611. The tests are updated slightly and the query is split into three distinct cases:
* children with a placement to the queried unit but no attendance or backup placement
* children with a backup placement to the queried unit but no attendance
* children with an ongoing or recently ended attendance to the queried unit


hruoho (approved):
Seems leaner than the earlier take :+1:

PR #1657, opened 2021-11-16
Replace react-select with Combobox in map view

jaahon:
#### Summary

Functionally they should be identical and visually they are pretty much the same (and closer to other comboboxes).


kimifoud (approved):
👏

PR #1659, opened 2021-11-16
Refactor child context

akheron:
#### Summary

- Move most data out of `ChildContext` to be managed by the components that use the data
- Data that's used commonly by many components were left, but is now fetched by `ChildContextProvider`
- Add `React.memo`, `renderResult`, etc. usages to affected components

### DIFF
--- a/frontend/src/employee-frontend/components/child-information/GuardiansAndParents.tsx
+++ b/frontend/src/employee-frontend/components/child-information/GuardiansAndParents.tsx
@@ -2,47 +2,31 @@
 //
 // SPDX-License-Identifier: LGPL-2.1-or-later
 
-import React, { useState } from 'react'
+import React, { useContext, useState } from 'react'
 import { useTranslation } from '../../state/i18n'
-import { useEffect } from 'react'
-import { Loading } from 'lib-common/api'
-import { useContext } from 'react'
-import Loader from 'lib-components/atoms/Loader'
 import { Table, Tbody, Td, Th, Thead, Tr } from 'lib-components/layout/Table'
 import _ from 'lodash'
 import { Link } from 'react-router-dom'
 import { PersonJSON } from 'lib-common/generated/api-types/pis'
 import { formatName } from '../../utils'
 import { NameTd } from '../PersonProfile'
-import { ChildContext } from '../../state'
-import { getPersonGuardians } from '../../api/person'
 import { getAge } from 'lib-common/utils/local-date'
 import { CollapsibleContentArea } from 'lib-components/layout/Container'
 import { H2, H3 } from 'lib-components/typography'
 import FridgeParents from './FridgeParents'
 import { Gap } from 'lib-components/white-space'
-import { UUID } from 'lib-common/types'
+import { UnwrapResult } from '../async-rendering'
+import { ChildContext } from '../../state'
 
 interface Props {
-  id: UUID
   startOpen: boolean
 }
 
-const GuardiansAndParents = React.memo(function Guardians({
-  id,
-  startOpen
-}: Props) {
+export default React.memo(function Guardians({ startOpen }: Props) {
   const { i18n } = useTranslation()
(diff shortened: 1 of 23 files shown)


frontend/src/employee-frontend/components/child-information/GuardiansAndParents.tsx
```diff
@@ -96,15 +56,42 @@ const GuardiansAndParents = React.memo(function Guardians({
               <Th>{i18n.personProfile.streetAddress}</Th>
             </Tr>
           </Thead>
-          <Tbody>{renderGuardians()}</Tbody>
+          <Tbody>
+            <UnwrapResult result={guardians}>
```
jaahon, line 60: Why not use `renderResult` here? Also, it would a good idea to wrap the whole `<table>` element in it since `<div>`s shouldn't be `<tbody>`'s children.

frontend/src/employee-frontend/components/child-information/MessageBlocklist.tsx
```diff
@@ -73,68 +75,43 @@ const MessageBlocklist = React.memo(function ChildDetails({
...
-                          !checked
-                        ).then((res) => {
-                          if (res.isFailure) {
-                            setErrorMessage({
-                              type: 'error',
-                              title: i18n.common.error.unknown,
-                              text: i18n.common.error.saveFailed,
-                              resolveLabel: i18n.common.ok
-                            })
+            <UnwrapResult result={recipients}>
```
jaahon, line 78: Same thing as in `GuardiansAndParents.tsx`.

frontend/src/employee-frontend/components/child-information/VasuAndLeops.tsx
```diff
@@ -279,13 +250,58 @@ const VasuAndLeops = React.memo(function VasuAndLeops({
       >
         <VasuInitialization childId={childId} allowCreation={allowCreation} />
         <Table>
-          <Tbody>{renderSummaries()}</Tbody>
+          <Tbody>
+            <UnwrapResult result={vasus}>
```
jaahon, line 254: Same thing as in `GuardiansAndParents.tsx`.

frontend/src/employee-frontend/state/child.tsx
```diff
@@ -2,220 +2,141 @@
...
+    [id]
   )
-  const [applications, setApplications] = useState<
-    Result<PersonApplicationSummary[]>
-  >(defaultState.applications)
-  const [vasus, setVasus] = useState<Result<VasuDocumentSummary[]>>(
-    defaultState.vasus
+  const [parentships] = useApiState(
+    async (): Promise<Result<Parentship[]>> =>
+      requireRole(roles, 'SERVICE_WORKER', 'UNIT_SUPERVISOR', 'FINANCE_ADMIN')
```
jaahon, line 109: Change this to a `permittedActions` check now that you touched it 😁

frontend/src/employee-frontend/components/child-information/GuardiansAndParents.tsx
akheron: Yeah, `renderResult` requires wrapping the whole table, that's why I initially used `UnwrapResult`. Now changed all three to use `renderResult`.

frontend/src/employee-frontend/state/child.tsx
akheron: Not sure how to do that, it needs some investigation. The problem is that while it's easy to check whether someone is a `UNIT_SUPERVISOR` for a child, it's not really supported for adults. So if I change the check in the service to be based on actions, unit supervisors stop seeing parentships for adults.

jaahon: Ah, I didn't realize that service was still doing role checks. It could do different checks depending on the parameter but that's getting really far away from the scope of this PR so no need to do it.

jaahon (approved):

PR #1670, opened 2021-11-17
Make sure guardians with email addresses get pedadoc notifications

zch:
#### Summary
The sending of pedagogical documentation notifications did not work if
one of the guardians of the child did not have an email address
configured. This makes sure that even if one guardian is missing an
email address, all the other guardians and heads of the child are sent
the notification.

### DIFF
--- a/service/src/main/kotlin/fi/espoo/evaka/pedagogicaldocument/PedagogicalDocumentNotificationService.kt
+++ b/service/src/main/kotlin/fi/espoo/evaka/pedagogicaldocument/PedagogicalDocumentNotificationService.kt
@@ -17,7 +17,10 @@ import fi.espoo.evaka.shared.db.mapColumn
 import fi.espoo.evaka.shared.db.updateExactlyOne
 import fi.espoo.evaka.shared.domain.HelsinkiDateTime
 import fi.espoo.evaka.shared.utils.dateNow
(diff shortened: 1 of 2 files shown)


service/src/main/kotlin/fi/espoo/evaka/pedagogicaldocument/PedagogicalDocumentNotificationService.kt
```diff
@@ -83,11 +86,14 @@ class PedagogicalDocumentNotificationService(
                 }
                 .list()
         } catch (e: Exception) {
+            logger.error(e) { "Could not get pedagogical documentation notifications" }
```
jaahon, line 89: Rather than log the error should the `try ... catch` be removed altogether? Is it expected that the query fails?

jaahon: And if it's expected sometimes I would handle it explicitly and rethrow every other error.

zch: Failure was expected before this fix 😆 
But good point, removed it 👍

jaahon (approved):

PR #1673, opened 2021-11-18
Varda improvements

reynders:
- Ensure varda child is eventually deleted when there are no service needs to be sent from evaka

### DIFF
--- a/service/src/main/kotlin/fi/espoo/evaka/varda/VardaUpdateService.kt
+++ b/service/src/main/kotlin/fi/espoo/evaka/varda/VardaUpdateService.kt
@@ -120,7 +120,13 @@ fun updateVardaChild(db: Database.Connection, client: VardaClient, serviceNeedDi
 
         if (!db.read { it.hasVardaServiceNeeds(evakaChildId) }) {
             logger.info("VardaUpdate: deleting child for not having service needs ($evakaChildId)")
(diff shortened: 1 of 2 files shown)


service/src/main/kotlin/fi/espoo/evaka/varda/VardaUpdateService.kt
```diff
@@ -120,7 +120,15 @@ fun updateVardaChild(db: Database.Connection, client: VardaClient, serviceNeedDi
...
         if (!db.read { it.hasVardaServiceNeeds(evakaChildId) }) {
             logger.info("VardaUpdate: deleting child for not having service needs ($evakaChildId)")
-            deleteChildDataFromVardaAndDb(db, client, evakaChildId.raw)
+
+            if (!deleteChildDataFromVardaAndDb(db, client, evakaChildId.raw)) {
+                // If child has 0 service needs in varda, and delete fails, we need to retry it via the reset mechanism
+                db.transaction {
+                    it.createUpdate("UPDATE varda_reset_child SET reset_timestamp = null WHERE evaka_child_id = :id")
+                        .bind("id", evakaChildId)
+                        .execute()
```
akheron, line 129: This is the only SQL in this file. Should it be moved to `VardaQueries.kt` or somewhere? 🤔

akheron (approved):

PR #1674, opened 2021-11-18
Fix a flaky E2E test

akheron:
#### Summary

The fix here is to type to the combo box before attempting to select anything from it. Just opening the combo box doesn't work in about 1/20 of the times.

While searching for the fix, refactored the tests to use locators and the unit access control components to use `renderResult`, `useApiState`, factored out inline components, etc.

### DIFF
--- a/frontend/src/employee-frontend/components/unit/tab-unit-information/UnitAccessControl.tsx
+++ b/frontend/src/employee-frontend/components/unit/tab-unit-information/UnitAccessControl.tsx
@@ -2,43 +2,40 @@
 //
 // SPDX-License-Identifier: LGPL-2.1-or-later
 
-import { sortBy } from 'lodash'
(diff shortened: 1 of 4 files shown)


frontend/src/employee-frontend/components/unit/tab-unit-information/UnitAccessControl.tsx
```diff
@@ -381,7 +376,20 @@ const AddAclLabel = styled.p`
   margin-bottom: 0;
 `
 
-function AddAcl({
+interface AclOption {
+  label: string
+  value: string
+}
```
kimifoud, line 382: One day we shall export a common `LabelValue` interface to rule them all 😄

kimifoud (approved):
🆗 👍

PR #1683, opened 2021-11-18
Hide old assistance need information when child goes to preschool

reynders:
#### Summary
- If child is or has been in preschool, hide pre preschool service needs from non admins
- Add E2E tests for child information page assistance needs

### DIFF
--- a/service/src/main/kotlin/fi/espoo/evaka/assistanceneed/AssistanceNeedController.kt
+++ b/service/src/main/kotlin/fi/espoo/evaka/assistanceneed/AssistanceNeedController.kt
@@ -9,9 +9,6 @@ import fi.espoo.evaka.daycare.controllers.utils.created
 import fi.espoo.evaka.daycare.controllers.utils.noContent
 import fi.espoo.evaka.daycare.controllers.utils.ok
 import fi.espoo.evaka.pis.service.PersonService
-import fi.espoo.evaka.placement.Placement
-import fi.espoo.evaka.placement.PlacementType
-import fi.espoo.evaka.placement.getPlacementsForChild
 import fi.espoo.evaka.shared.AssistanceNeedId
 import fi.espoo.evaka.shared.auth.AuthenticatedUser
 import fi.espoo.evaka.shared.db.Database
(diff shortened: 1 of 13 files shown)


service/src/main/kotlin/fi/espoo/evaka/assistanceneed/AssistanceNeedController.kt
```diff
@@ -54,7 +59,23 @@ class AssistanceNeedController(
     ): ResponseEntity<List<AssistanceNeed>> {
         Audit.ChildAssistanceNeedRead.log(targetId = childId)
         accessControl.requirePermissionFor(user, Action.Child.READ_ASSISTANCE_NEED, childId)
-        return assistanceNeedService.getAssistanceNeedsByChildId(db, childId).let(::ok)
+        return assistanceNeedService.getAssistanceNeedsByChildId(db, childId).filter { assistanceNeed ->
+            // If child is or has been in preschool, do not show pre preschool daycare assistance needs for non admin
+            if (!user.isAdmin) {
```
akheron, line 64: Perhaps something like `if (accessControl.hasPermissionFor(user, Action.Child.READ_ASSISTANCE_NEED_FULL, childId)` would be better than a hard-coded check for admin.

service/src/main/kotlin/fi/espoo/evaka/assistanceneed/AssistanceNeedController.kt
```diff
@@ -54,7 +59,23 @@ class AssistanceNeedController(
...
+                val preschoolPlacements = getChildPreschooPlacements(db, childId)
+                if (preschoolPlacements.any { preschoolPlacement ->
+                    assistanceNeed.startDate.isBefore(preschoolPlacement.startDate)
+                }
+                ) false else true
+            } else
+                true
+        }.let(::ok)
```
akheron, line 72: If you remove `ResponseEntity<>` from return type, the `ok` wrapping should not be needed.

service/src/main/kotlin/fi/espoo/evaka/assistanceneed/AssistanceNeedController.kt
reynders: Did a whole revamp by moving this stuff to access control side

PR #1698, opened 2021-11-22
Remove AbsencesContext

akheron:


frontend/src/employee-frontend/components/absences/AbsenceModal.tsx
```diff
@@ -0,0 +1,92 @@
...
+  onSave: () => void
+  saveDisabled: boolean
+  onCancel: () => void
+  selectedAbsenceType: AbsenceType | null
+  setSelectedAbsenceType: (value: AbsenceType | null) => void
+  selectedCareTypeCategories: CareTypeCategory[]
+  updateCareTypes: (value: CareTypeCategory) => void
+}) {
+  const { i18n } = useTranslation()
+  return show ? (
```
kimifoud, line 39: In react-world, I prefer short circuit evaluation `show &&` instead of ternaries.

kimifoud: Tbh I would leave the `show` out of props and just short circuit eval in the parent container `showModal && <AbsenceModal ..`

kimifoud (approved):
LGTM!

frontend/src/employee-frontend/components/absences/AbsenceModal.tsx
akheron: This works in the parent but not here, because a component cannot return `false`.

kimifoud: Right, you'd have to wrap it in a fragment. Anyway, I think it's unidiomatic to pass `show` to the modal.

akheron: You're right, fixed!

PR #1707, opened 2021-11-22
Use <UnwrapResult> in citizen navi

akheron:
#### Summary

This way we don't show the login button (or anything else) while loading the auth status.

This also fixes a flaky e2e test, caused by clicking the language select just when the auth status arrives and the user name is rendered, and thus hitting the user menu instead.

### DIFF
--- a/frontend/src/citizen-frontend/header/DesktopNav.tsx
+++ b/frontend/src/citizen-frontend/header/DesktopNav.tsx
@@ -16,15 +16,16 @@ import {
   faSignOut,
   faUser
 } from 'lib-icons'
-import React, { useCallback, useState } from 'react'
+import React, { useCallback, useContext, useState } from 'react'
 import { NavLink, useHistory } from 'react-router-dom'
 import styled, { useTheme } from 'styled-components'
 import { fontWeights } from 'lib-components/typography'
-import { useUser } from '../auth/state'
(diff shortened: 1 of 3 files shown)


frontend/src/citizen-frontend/header/DesktopNav.tsx
```diff
@@ -35,65 +36,72 @@ export default React.memo(function DesktopNav({
...
-                ) : (
-                  ''
-                )}
-              </StyledNavLink>
-            )}
-            {user.accessibleFeatures.reservations && (
-              <StyledNavLink to="/calendar" data-qa="nav-calendar">
-                {t.header.nav.calendar}
-              </StyledNavLink>
+    <UnwrapResult result={user} loading={() => null}>
```
jaahon, line 42: Since `loading={() => null}` returns `null` when called the `<SpinnerSegment />` is rendered anyway. Could you fix that in `<UnwrapResult>`? 😄

jaahon (approved):

frontend/src/citizen-frontend/header/DesktopNav.tsx
akheron: Oops :D That explains why there were two spinners in mobile widths. I thought it was an exisiting feature.

PR #1714, opened 2021-11-23
Add expand / collapse buttons to new message desktop view

reynders:
#### Summary
Add expand / collapse button to desktop new message view

### DIFF
--- a/frontend/src/lib-components/employee/messages/MessageEditor.tsx
+++ b/frontend/src/lib-components/employee/messages/MessageEditor.tsx
@@ -13,7 +13,12 @@ import MultiSelect from 'lib-components/atoms/form/MultiSelect'
 import Radio from 'lib-components/atoms/form/Radio'
 import { FixedSpaceRow } from 'lib-components/layout/flex-helpers'
 import { defaultMargins, Gap } from 'lib-components/white-space'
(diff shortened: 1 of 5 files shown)


frontend/src/lib-components/employee/messages/MessageEditor.tsx
```diff
@@ -418,15 +425,39 @@ export default React.memo(function MessageEditor({
...
-          icon={faTimes}
-          onClick={onCloseHandler}
-          white
-          data-qa="close-message-editor-btn"
-        />
+        <HeaderButtonContainer>
+          {expandedView ? (
+            <IconButton
+              icon={faDownLeftAndUpRightToCenter}
+              onClick={() => setExpandedView(!expandedView)}
```
akheron, line 439: This could be a good exercise for the "memo everything" stuff we discussed at some point.

Define a new function:
```
const toggleExpandedView = useCallback(
  () => setExpandedView((prev) => !prev),
  []
)
```
and use it:
```
<IconButton ... onClick={toggleExpandedView} ... />
```

This way, `IconButton` doesn't have to re-renderer each time something else changes.

On the other hand, in this component it doesn't matter as much because there's a lot of other things causing re-rendering of other child components.

akheron (approved):

PR #1717, opened 2021-11-24
Use Ruby templates to generate dynamic CI configuration

ilja-git:
#### Summary

To work around problems with CircleCI use dynamic configuration generation with Ruby templates.


.circleci/test.sh
```diff
@@ -0,0 +1,45 @@
...
+ && tar xzf circleci-cli.tar.gz \
+ && sudo mv "circleci-cli_${CIRCLECI_CLI_VERSION}_linux_amd64/circleci" /usr/local/bin \
+ && sudo chown 0:0 /usr/local/bin/circleci \
+ && rm -r circleci-cli.tar.gz "circleci-cli_${CIRCLECI_CLI_VERSION}_linux_amd64/"
+WORKDIR /ci
+EOF
+
+CHECK_COMMAND="erb -T- /ci/template.yml | tee /dev/tty | circleci --skip-update-check config validate -"
+
+if test -z "${1:-}" || [ "${1:-}" = "fail" ]; then
```
akheron, line 26: What's the point of `${x:-}`? It seems to me that it just substitutes an empty string with an empty string 🤔 

If `${1:-}` is actually required, then this would be a simpler way:
```sh
arg=${1:-}
if [ -z "$arg" -o "$arg" = "fail" ]; then
```

.circleci/test.sh
```diff
@@ -0,0 +1,45 @@
...
+    docker run -v "$(pwd):/ci" \
+        -e CIRCLE_BRANCH=master \
+        -it "$TEST_IMAGE" bash -c "$CHECK_COMMAND"
+fi
+
+if test -z "${1:-}" || [ "${1:-}" = "pull" ]; then
+    docker run -v "$(pwd):/ci" \
+        -e CIRCLE_BRANCH=pull/1234 \
+        -it "$TEST_IMAGE" bash -c "$CHECK_COMMAND"
+fi
```
akheron, line 45: If `$1` is empty, all three `docker run` commands are run. Is this expected?

.circleci/test.sh
```diff
@@ -0,0 +1,45 @@
...
+
+if test -z "${1:-}" || [ "${1:-}" = "master" ]; then
+    docker run -v "$(pwd):/ci" \
+        -e CIRCLE_BRANCH=master \
+        -it "$TEST_IMAGE" bash -c "$CHECK_COMMAND"
+fi
+
+if test -z "${1:-}" || [ "${1:-}" = "pull" ]; then
+    docker run -v "$(pwd):/ci" \
+        -e CIRCLE_BRANCH=pull/1234 \
```
akheron, line 43: `pull/1234`? 🤔

.circleci/test.sh
ilja-git: End result is the same, but maybe nicer with variable. I'll change it.

Point is that we want empty string - not undefined, because we use "u" in set -eu that will fail if we have "undefined" variable so "$1" would fail if argument is not provided.

.circleci/test.sh
ilja-git: It will run without variables all test and with variable single test.

.circleci/test.sh
ilja-git: The script is just for sanity testing of the template generator, so it will fake the CI functionality locally. The pull request from forks come in format: pull/<pull-request-number> so CIRCLE_BRANCH is set to this variable.

akheron (approved):

PR #1719, opened 2021-11-24
EVAKA-4347 Remove assistance need descriptions and other basis descriptions

akheron:
#### Summary

Requires a manual migration BEFORE DEPLOY in staging and prod:

```
INSERT INTO assistance_basis_option (value, name_fi, display_order)
  VALUES ('OTHER', 'Muu peruste', 999);

INSERT INTO assistance_basis_option_ref (need_id, option_id)
SELECT id, (SELECT id FROM assistance_basis_option WHERE value = 'OTHER')
FROM assistance_need WHERE other_basis <> '';
```


Gekkio (commented):
Would it be simpler to make other basis just a normal option in the `assistance_basis_option` table? Tampere doesn't use the other basis feature, so you could just do this manually in Espoo environments, and then drop the column in the actual migration.

In other words:
- manual migration in Espoo environments:
  - add new `assistance_basis_option` "OTHER"
  - add references to this new option when `other_basis != ''`
- drop other_basis column and all special handling (e.g. in the report), including the feature flag
- profit

akheron:
Yes, true! Now that the other basis option doesn't have a description, it's just a normal basis.

frontend/src/employee-frontend/components/reports/AssistanceNeedsAndActions.tsx
```diff
@@ -329,11 +308,6 @@ function AssistanceNeedsAndActions() {
                   <Td>
                     {reducePropertySum(filteredRows, (r) => r.otherBasisCount)}
                   </Td>
-                  {featureFlags.assistanceBasisOtherEnabled && (
```
Gekkio, line 332: This removes the wrong column because the feature flag has been originally put in the wrong column :see_no_evil:

akheron: Yeah, and the query was also a bit wrong. The report now works after the latest push.

Gekkio (approved):
> +38
> -195

:heart_eyes:

PR #1727, opened 2021-11-24
Citizen personal details only with strong auth

kimifoud:
#### Summary

Require strong auth for viewing the personal details page and updating the details. Currently the page can be accessed but update request will fail with 403.

Lock icons for _weak citizen_ are coming in #1726


jaahon:
I think we should also hide sensitive data from /auth/status response if the intent is to hide it without strong authentication.

jaahon (approved):

PR #1734, opened 2021-11-25
Use cache headers for child images

kimifoud:
#### Summary

Fetching child images is the most time consuming request in the whole application. Utilizing browser cache headers should significantly reduce the time spent fetching images and thus reduce thread utilization.


akheron:
When an existing child image is replaced by a new image, does the URL of the image change? If it doesn't, the users will never see the updated image.

kimifoud:
> When an existing child image is replaced by a new image, does the URL of the image change? If it doesn't, the users will never see the updated image.

Images are immutable in S3. Each one gets a new id. https://github.com/espoon-voltti/evaka/blob/5c1900eeef24f60c4860488e30df25d8642f99fe/service/src/main/kotlin/fi/espoo/evaka/childimages/ChildImage.kt#L24

akheron (approved):

PR #1740, opened 2021-11-29
Serve child images through nginx

akheron:
#### Summary

- Add an internal nginx endpoint that can proxy any URL
- When a child image request comes, generate a presigned S3 URL
- Use [X-Accel-Redirect](https://www.nginx.com/resources/wiki/start/topics/examples/x-accel/) to the internal endpoint to serve the presigned URL to the client through nginx only

This way, slow clients don't consume a thread from service while downloading a child image. nginx uses async i/o and is able to handle thousands of concurrent requests, whereas service chokes already for a few hundred.


kimifoud (approved):
🚀 🌕

PR #1741, opened 2021-11-29
EVAKA-4283 employee mobile units list for unit supervisors

jaahon:
#### Summary

Add a units list for unit supervisors with multiple units.

Also small refactorings in employee mobile. Use the same `<TopBar />` component in all employee mobile views apart from the new message view (where using the same component would require a bigger refactoring).


frontend/src/employee-mobile-frontend/components/UnitList.tsx
```diff
@@ -0,0 +1,66 @@
...
+          <Gap size="s" />
+        </ContentArea>
+      ))}
+    </>
+  )
+})
+
+const CenteredH1 = styled(H1)`
+  text-align: center;
+`
```
kimifoud, line 60: All headings have an optional `centered` property.

frontend/src/employee-mobile-frontend/components/UnitList.tsx
```diff
@@ -0,0 +1,66 @@
...
+                [0, 0]
+              )
+
+              return (
+                <UnitContainer key={id} to={`/units/${id}`}>
+                  <H2 noMargin>{name}</H2>
+                  <Gap size="xxs" />
+                  <FixedSpaceColumn spacing="xxs">
+                    <span>{`${i18n.units.children} ${present}/${total}`}</span>
+                  </FixedSpaceColumn>
```
kimifoud, line 46: I think this could be simplified by adding `gap: ${defaultMargins.xxs};` to `UnitContainer` and removing the `FixedSpaceColumn` here altogether since it does not seem to add any benefit..?

frontend/src/employee-mobile-frontend/components/attendances/AttendanceList.tsx
```diff
@@ -2,113 +2,146 @@
...
+      totalPresent: groupChildren.filter((ac) => ac.status === 'PRESENT')
+        .length,
+      totalDeparted: groupChildren.filter((ac) => ac.status === 'DEPARTED')
+        .length,
+      totalAbsent: groupChildren.filter((ac) => ac.status === 'ABSENT').length
+    }),
+    [groupChildren]
+  )
+
+  const tabs = useMemo(() => {
```
kimifoud, line 62: 👌

service/src/main/kotlin/fi/espoo/evaka/attendance/MobileUnitController.kt
```diff
@@ -12,33 +12,47 @@ import fi.espoo.evaka.shared.auth.AccessControlList
...
-    @GetMapping("/units/{unitId}")
+    @GetMapping
+    fun getUnitInfos(
+        db: Database.Connection,
+        user: AuthenticatedUser,
+        evakaClock: EvakaClock,
+        @RequestParam unitIds: List<DaycareId>
+    ): List<UnitInfo> {
+        Audit.UnitRead.log(targetId = unitIds)
+        unitIds.forEach { unitId -> acl.getRolesForUnit(user, unitId).requireOneOfRoles(UserRole.MOBILE) }
```
kimifoud, line 40: This has the potential to translate to hundreds of queries each opening and closing their own connection, and should be rethinked.

kimifoud: Moreover, I do not like the idea of adding attendance data to `UnitInfo` because the attendance numbers are not needed in most of the use cases, and the data might become stale quickly which could encourage someone to just refresh the unit data instead of using a more approriate solution.
Previously all unit data was refreshed on all page interactions and was included in `AttendanceResponse`. @Joosakur and I refactored long-lived unit data to a separate endpoint and context so that it can be kept in memory without constant need of refreshing.

Could the unit-level attendances just be fetched in a separate request when needed?

kimifoud (changes requested):
Just a quick nonextensive 👀 review.

kimifoud (approved):
Needs to be rebased but LGTM! 👍

frontend/src/employee-mobile-frontend/api/unit.ts
```diff
@@ -76,3 +77,18 @@ export function getMobileUnitInfo(unitId: string): Promise<Result<UnitInfo>> {
...
+  unitIds: string[]
+): Promise<Result<UnitStats[]>> {
+  return client
+    .get<JsonOf<UnitStats[]>>(
+      `/mobile/units/stats?unitIds=${unitIds.join(
+        ','
+      )}&useRealtimeStaffAttendance=${String(
+        featureFlags.experimental?.realtimeStaffAttendance ?? false
+      )}`
+    )
```
kimifoud, line 91:
```suggestion
    .get<JsonOf<UnitStats[]>>('/mobile/units/stats', {
      params: {
        unitIds: unitIds.join(','),
        useRealtimeStaffAttendance:
          featureFlags.experimental?.realtimeStaffAttendance ?? false
      }
    })
```

PR #1742, opened 2021-11-29
Store open unit groups to local storage for persistence over navigation

reynders:

### DIFF
--- a/frontend/src/employee-frontend/components/UnitPage.tsx
+++ b/frontend/src/employee-frontend/components/UnitPage.tsx
@@ -26,6 +26,7 @@ import { UUID } from 'lib-common/types'
 import TabApplicationProcess from './unit/TabApplicationProcess'
 import styled from 'styled-components'
 import { fontWeights } from 'lib-components/typography'
+import useLocalStorage from 'lib-common/utils/useLocalStorage'
 
 const UnitPage = React.memo(function UnitPage({ id }: { id: UUID }) {
   const { i18n } = useTranslation()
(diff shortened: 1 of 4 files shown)


frontend/src/employee-frontend/components/UnitPage.tsx
```diff
@@ -49,6 +55,51 @@ const UnitPage = React.memo(function UnitPage({ id }: { id: UUID }) {
...
+      .join(',')
+  }
+
+  useEffect(() => {
+    const openList = openGroupsToStringList(openGroups)
+    openList.length > 0
+      ? history.push(`?open_groups=${openList}`)
+      : history.push('?')
+  }, [openGroups, history])
+
```
akheron, line 102: The state is now stored in two places: URL and React component state.

Could it be like this instead (partly pseudocode):
```
const openGroups = useMemo(
  () => getOpenGroupsFromQueryParams(location.search),
  [location.search]
)
const setOpenGroups = useCallback((openGroups: Record<string, boolean>) => 
  history.push(`?open_groups=${openGroupsToStringList(openGroups)}`)
)
```

The functions `getOpenGroupsFromQueryParams` and `openGroupsToStringList` could be moved outside of the component function, too.

jaahon: I give a +1 to that idea. We have a `useQuery` hook that is useful in this case. 

Also, I think `history.replace` would be a better idea than `history.push` so that browser history doesn't get filled up from simply opening and closing group collapsibles.

reynders: Without the component state the open groups are lost when navigating between the unit tabs. I added a test for this case.

akheron (approved):

PR #1743, opened 2021-11-29
Prevent pedagogical document email notifications when there is no content

kimifoud:
#### Summary

- Require either description or uploaded attachment before scheduling the email job
- Track email job status in email_job_created_at column to create the job only once
- Fix create document insert to actually save the description given in request body
- Set description not null in the database while at it
- Replace artisan mapPedagogicalDocument fn with `@Nested`
- Re-use transaction in handleFileUpload for transactional business logic after a successful operation


service/src/main/kotlin/fi/espoo/evaka/pedagogicaldocument/PedagogicalDocumentNotificationService.kt
```diff
@@ -46,43 +46,73 @@ class PedagogicalDocumentNotificationService(
...
+        return this.createQuery(
+            """
+SELECT EXISTS(
+    SELECT 1
+    FROM pedagogical_document doc
+    LEFT JOIN attachment a ON doc.id = a.pedagogical_document_id
+    WHERE
+        doc.id = :id AND
+        doc.email_job_created_at IS NULL AND 
+        (LENGTH(COALESCE(doc.description, '')) > 0 OR a.id IS NOT NULL)
```
akheron, line 95: `COALESCE` is not needed anymore?

akheron (approved):

PR #1745, opened 2021-11-29
Try to fix flakiness in invoices e2e tests

kimifoud:
#### Summary

- Use memoization in InvoiceFilters to prevent re-rendering of static elements
- Instead of reloading the page, switch tabs back and forth to reload newly created fixture data


frontend/src/e2e-playwright/specs/4_finance/invoices.spec.ts
```diff
@@ -168,6 +172,7 @@ describe('Invoices', () => {
     await invoicesPage.freeTextFilter(adultFixtureWihtoutSSN.firstName)
     await invoicesPage.assertInvoiceCount(1)
     await invoicesPage.toggleAllInvoices(true)
+    await invoicesPage.assertInvoiceCount(1)
```
jaahon, line 175: I don't understand the idea behind adding this check. The data isn't realoaded after `await invoicesPage.toggleAllInvoices(true)` so the previous `await invoicesPage.assertInvoiceCount(1)` should be enough.

kimifoud: 🤦 It's so nice to have PR reviews.

frontend/src/employee-frontend/components/invoices/InvoiceFilters.tsx
```diff
@@ -169,4 +174,4 @@ function InvoiceFilters() {
   )
 }
 
-export default InvoiceFilters
+export default React.memo(InvoiceFilters)
```
akheron, line 177: I prefer `export default React.memo(function InvoiceFilters(...) { ... })` but it's a matter of taste I guess 😄

kimifoud:
> > Wait for table data to be loaded before sending invoices
> 
> Which part does the waiting? 🤔

Ok, I'm gonna quit now. Summary was outdated.

akheron (approved):

PR #1747, opened 2021-11-29
Display citizen rejected decision for two weeks

mapehe:

### DIFF
--- a/frontend/src/employee-frontend/components/unit/tab-placement-proposals/PlacementProposalRow.tsx
+++ b/frontend/src/employee-frontend/components/unit/tab-placement-proposals/PlacementProposalRow.tsx
@@ -78,6 +78,7 @@ type Props = {
     otherReason?: string
   ) => undefined | void
   loadUnitData: () => void
+  rejectedByCitizen: boolean
 }
 
 export default React.memo(function PlacementProposalRow({
@@ -85,7 +86,8 @@ export default React.memo(function PlacementProposalRow({
   confirmationState,
   submitting,
   onChange,
-  loadUnitData
+  loadUnitData,
+  rejectedByCitizen
 }: Props) {
   const { i18n } = useTranslation()
   const [modalOpen, setModalOpen] = useState(false)
@@ -145,7 +147,7 @@ export default React.memo(function PlacementProposalRow({
         data-qa={`placement-proposal-row-${placementPlan.applicationId}`}
       >
         <Td data-qa="child-name">
-          {confirmationState !== 'REJECTED_BY_CITIZEN' ? (
+          {!rejectedByCitizen ? (
             <Link to={`/child-information/${placementPlan.child.id}`}>
               {childName}
             </Link>
@@ -157,16 +159,12 @@ export default React.memo(function PlacementProposalRow({
         </Td>
         <Td
           data-qa="child-dob"
-          color={
-            confirmationState === 'REJECTED_BY_CITIZEN'
-              ? colors.greyscale.medium
-              : undefined
-          }
+          color={rejectedByCitizen ? colors.greyscale.medium : undefined}
         >
           {placementPlan.child.dateOfBirth.format()}
         </Td>
         <Td data-qa="placement-duration">
-          {confirmationState !== 'REJECTED_BY_CITIZEN' ? (
+          {!rejectedByCitizen ? (
             `${placementPlan.period.start.format()} - ${placementPlan.period.end.format()}`
           ) : (
             <FlexRow>
@@ -190,7 +188,7 @@ export default React.memo(function PlacementProposalRow({
         </Td>
         <Td data-qa="application-link">
           <CenteredDiv>
-            {confirmationState !== 'REJECTED_BY_CITIZEN' ? (
+            {!rejectedByCitizen ? (
               <a
                 href={`${getEmployeeUrlPrefix()}/employee/applications/${
                   placementPlan.applicationId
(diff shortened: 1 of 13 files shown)


frontend/src/employee-frontend/components/unit/tab-placement-proposals/PlacementProposalRow.tsx
```diff
@@ -117,18 +143,49 @@ export default React.memo(function PlacementProposalRow({
...
-            )}
-          </Link>
+          {confirmationState !== 'REJECTED_BY_CITIZEN' ? (
+            <Link to={`/child-information/${placementPlan.child.id}`}>
+              {formatName(
+                placementPlan.child.firstName,
+                placementPlan.child.lastName,
+                i18n,
+                true
+              )}
```
kimifoud, line 153: Child name could be extracted to a variable to avoid repetition.

frontend/src/employee-frontend/components/unit/tab-placement-proposals/PlacementProposalRow.tsx
```diff
@@ -141,23 +198,32 @@ export default React.memo(function PlacementProposalRow({
...
+              >
+                <IconButton
+                  onClick={() => undefined}
+                  icon={faFileAlt}
+                  altText={i18n.personProfile.application.open}
+                />
+              </a>
+            ) : (
               <IconButton
                 onClick={() => undefined}
```
kimifoud, line 217: I'm no UX expert but it would feel more intuitive to have an empty cell instead of a button that cannot ever be interacted with.

kimifoud (changes requested):
Review the frontend only, since I do not fully understand the consequences of adding a new status to the flow.

frontend/src/employee-frontend/components/unit/tab-placement-proposals/PlacementProposalRow.tsx
```diff
@@ -46,6 +49,29 @@ const CenteredRow = styled(Tr)`
...
+  min-width: 16px;
+  min-height: 16px;
+  font-size: 16px !important;
+  border-radius: 100%;
+  color: ${colors.greyscale.white};
+`
+
+const CircleIconOrange = styled(CircleIcon)`
+  background-color: ${colors.accents.orange};
+`
```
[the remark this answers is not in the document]
mapehe: My understanding is that this is a `PlacementProposalRow` specific thing.

mapehe:
Thanks for the feedback @kimifoud, made changes accordingly 🙏

kimifoud (changes requested):
As discussed, it's preferred to derive the status from decision instead of adding a new state.

frontend/src/employee-frontend/components/unit/tab-placement-proposals/PlacementProposalRow.tsx
```diff
@@ -57,20 +78,29 @@ type Props = {
     otherReason?: string
   ) => undefined | void
   loadUnitData: () => void
+  rejectedByCitizen: boolean
```
kimifoud, line 81: The `placementPlan` is already in the props so having this as a separate field is unnecessary

service/src/main/kotlin/fi/espoo/evaka/placement/PlacementPlanQueries.kt
```diff
@@ -144,11 +145,25 @@ fun Database.Read.getPlacementPlans(unitId: DaycareId, from: LocalDate?, to: Loc
...
+WHERE pp.unit_id = :unitId AND (
+                                (
+                                    a.status = ANY(:statuses::application_status_type[])
+                                    AND a.status != 'REJECTED'::application_status_type
+                                    AND deleted = false
+                                )
+                                OR (
+                                    d.status = 'REJECTED'::decision_status
+                                    AND a.status = ANY(:statuses::application_status_type[])
+                                    AND a.status = 'REJECTED'::application_status_type
```
kimifoud, line 162: Having two `a.status` predicates here is slightly confusing.
`a.status = ANY(:statuses::application_status_type[])` is common for both sides of `OR` so it should be extracted to top level.

```
(a.status != 'REJECTED'::application_status_type AND deleted = false)
OR
(a.status = 'REJECTED'::application_status_type AND d.status = ...)
```

service/src/main/kotlin/fi/espoo/evaka/placement/PlacementPlanQueries.kt
```diff
@@ -135,7 +135,8 @@ fun Database.Read.getPlacementPlans(unitId: DaycareId, from: LocalDate?, to: Loc
         val dateOfBirth: LocalDate,
         val unitConfirmationStatus: PlacementPlanConfirmationStatus,
         val unitRejectReason: PlacementPlanRejectReason?,
-        val unitRejectOtherReason: String?
+        val unitRejectOtherReason: String?,
+        val rejectedByCitizen: Boolean?
```
kimifoud, line 139: Isn't this always present in the result? Should be `Boolean`.

service/src/main/kotlin/fi/espoo/evaka/placement/PlacementPlanQueries.kt
kimifoud: This query likely does full table scans. That might already be the case in production code but adding a timestamp interval filtering on an non-indexed column does not help. At the very least, the performance needs to be monitored after production deployment.

mapehe: Thanks for the feedback @kimifoud. This query is also used in fetching the plans that are being confirmed by the host. The current (confusing) common condition is because `A and (B or C)` is not equal to `(A or B) and (A or C)`.

mapehe:
Thanks for the feedback @kimifoud. Updated accordingly.

kimifoud (approved):

PR #1751, opened 2021-12-01
EVAKA-4256 enable mobile messaging as pilot feature

hruoho:
#### Summary

- Add enum & include it in the unit request
- Hide messaging components if feature not enabled
- Also remove unread count indicator for now (being redesigned)


jaahon (approved):
I think you should also add this feature in `espoo-dev-data.sql` to make local development as painless as possible.

PR #1759, opened 2021-12-01
Add child income mvp

patari:
#### Summary

Tampere is migrating child income data from previous system and needs to display results in the UI to verify them.


service/src/main/kotlin/fi/espoo/evaka/invoicing/controller/IncomeController.kt
```diff
@@ -51,7 +52,11 @@ class IncomeController(
     @GetMapping
     fun getIncome(db: Database.Connection, user: AuthenticatedUser, @RequestParam personId: PersonId): Wrapper<List<Income>> {
         Audit.PersonIncomeRead.log(targetId = personId)
-        accessControl.requirePermissionFor(user, Action.Person.READ_INCOME, personId)
+        if (!accessControl.hasPermissionFor(user, Action.Person.READ_INCOME, personId) &&
+            !accessControl.hasPermissionFor(user, Action.Child.READ_INCOME, personId)
+        ) {
+            throw Forbidden("Permission denied")
+        }
```
jaahon, line 59: I don't think a separate `Child.READ_INCOME` action is necessary so this check could be left as is.

frontend/src/employee-frontend/components/ChildInformation.tsx
```diff
@@ -142,6 +145,9 @@ const layouts: Layouts<typeof components> = {
     },
     { component: 'assistance', open: false },
     { component: 'applications', open: false },
+    ...(featureFlags.childIncomeEnabled
+      ? [{ component: 'income' as keyof typeof components, open: false }]
```
jaahon, line 149: Using `as keyof typeof components` turns off type checking so using `as const` is preferable.

jaahon (commented):
Works nicely 👌

service/src/main/kotlin/fi/espoo/evaka/invoicing/controller/IncomeController.kt
patari: Yeah, I trusted too much to type in `ChildInformation.tsx#requireOneOfPermittedActions`. I guess https://github.com/espoon-voltti/evaka/pull/1759/commits/1996cce57a7c842ec6eb6ffb94a55f90dfde9fbc is the correct fix?

frontend/src/employee-frontend/components/ChildInformation.tsx
patari: Wau, fixed

jaahon (approved):

PR #1764, opened 2021-12-02
Vasu followup edit comment

zch:
#### Summary
Implements editing of vasu followup comments if you are the author or have roles allowing editing other people's comments.


frontend/src/employee-frontend/components/vasu/components/FollowupQuestion.tsx
```diff
@@ -41,7 +41,17 @@ const FollowupEntryElement = React.memo(function FollowupEntryElement({
...
+    () => (editedEntry ? editedEntry : entry),
+    [entry, editedEntry]
+  )
+
+  const editAllowed =
+    user?.id === entry.authorId ||
+    roles.includes('ADMIN') ||
+    roles.includes('UNIT_SUPERVISOR') ||
+    roles.includes('STAFF') ||
+    roles.includes('SPECIAL_EDUCATION_TEACHER')
```
akheron, line 55: New role-based checks in the frontend 😬 This should be changed to use `getPermittedActions` in the service and send the results to frontend somehow.

service/src/main/kotlin/fi/espoo/evaka/vasu/VasuController.kt
```diff
@@ -150,18 +153,23 @@ class VasuController(
...
+        accessControl.requirePermissionFor(user, Action.VasuDocument.UPDATE_OWN_FOLLOWUP_ENTRY, id)
         db.transaction { tx ->
             val vasu = tx.getVasuDocumentMaster(id) ?: throw NotFound("vasu $id not found")
             val entry = findFollowupEntryWithId(vasu, entryId) ?: throw NotFound("Entry with $entryId not found")
 
-            if (entry.authorId?.raw  != user.id) {
+            val updateAnyAllowedBy = Action.VasuDocument.UPDATE_ANY_FOLLOWUP_ENTRY.defaultRoles().toTypedArray()
+            val userCanUpdateAnyEntry = user.hasOneOfRoles(*updateAnyAllowedBy)
+
+            if (entry.authorId != user.id && !userCanUpdateAnyEntry) {
```
akheron, line 164: Okay, you changed this here. It would still be ideal if the actual check was done in the `accessControl` side instead of the controller.

frontend/src/employee-frontend/components/vasu/components/FollowupQuestion.tsx
```diff
@@ -117,6 +122,7 @@ const FollowupEntryEditor = React.memo(function FollowupEntryEditor({
     } else {
       onChange({
         ...entry,
+        id: uuid(),
```
akheron, line 125: Not sure generating an `uuid` in the frontend is a good thing. There must be a reason why there's no other place in the codebase that does this 🤔 

I think at least PostgreSQL has been set up in a way that we use (maybe?) UUID v1 that are based on timestamp and mac address so that they cannot collide even theoretically.

akheron: Is the `id` even used anywhere? It wasn't there before this commit but this commit adds no usages.

zch: The ID is needed to be able to edit the comment after it has been saved. I tried two approaches:
1. Generating the ID here in the frontend when creating an entry
2. Generating the ID in the service, and reloading the document afterwards

With 2. we increase the complexity of the code quite a bit and I got into a bunch of corner cases and timing issues. Would require a bit bigger refactor in order for this case to work correctly... So I went with 1. for now.

PR #1780, opened 2021-12-07
Moves vasu followup comment access control to `AccessControl.kt`

zch:

### DIFF
--- a/service/src/main/kotlin/fi/espoo/evaka/shared/security/AccessControl.kt
+++ b/service/src/main/kotlin/fi/espoo/evaka/shared/security/AccessControl.kt
@@ -49,6 +49,7 @@ import fi.espoo.evaka.shared.PedagogicalDocumentId
 import fi.espoo.evaka.shared.PersonId
 import fi.espoo.evaka.shared.PlacementId
 import fi.espoo.evaka.shared.ServiceNeedId
+import fi.espoo.evaka.shared.VasuDocumentFollowupEntryId
 import fi.espoo.evaka.shared.VasuDocumentId
 import fi.espoo.evaka.shared.VasuTemplateId
 import fi.espoo.evaka.shared.auth.AccessControlList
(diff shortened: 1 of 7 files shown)


service/src/main/kotlin/fi/espoo/evaka/shared/security/AccessControl.kt
```diff
@@ -732,6 +733,24 @@ WHERE employee_id = :userId
...
+            is AuthenticatedUser.Employee -> user.globalRoles
+            else -> user.roles
+        }
+        if (globalRoles.any { it == UserRole.ADMIN })
+            return
+
+        Database(jdbi).connect { db ->
+            db.read { tx ->
+                val entry = tx.getVasuFollowupEntry(id, entryId)
+                if (entry.authorId != user.id && !user.hasOneOfRoles(*Action.VasuDocument.UPDATE_ANY_FOLLOWUP_ENTRY.defaultRoles().toTypedArray())) {
```
zch, line 747: We need to check both the roles and whether the user is the author (`user.id == authorId`) of a comment with id `entryId` in a vasu document with id `id`.

`ActionConfig` doesn't quite cut it when we need to compare this many IDs, and I don't think it's a good idea to make a generic solution for one instance of the problem.

service/src/main/kotlin/fi/espoo/evaka/shared/security/AccessControl.kt
```diff
@@ -732,6 +734,30 @@ WHERE employee_id = :userId
...
 
+    fun requirePermissionFor(user: AuthenticatedUser, action: Action.VasuDocumentFollowup, id: VasuDocumentFollowupEntryId) {
+        if (action != Action.VasuDocumentFollowup.UPDATE_FOLLOWUP_ENTRY) {
+            throw Forbidden()
+        }
+
+        val globalRoles = when (user) {
+            is AuthenticatedUser.Employee -> user.globalRoles
+            else -> user.roles
+        }
```
akheron, line 745: I think that `globalRoles` never includes `UNIT_SUPERVISOR`, `SPECIAL_EDUCATION_TEACHER` or `GROUP_STAFF`. You would have to use `acl.getRolesForChild` to get these scoped roles for a specific child.

akheron: And that can obviously be done only after reading the vasu document from database to check which child "owns" it.

akheron:
Otherwise looks good now! 🏅

akheron (approved):

PR #1788, opened 2021-12-09
EVAKA-4359 Add weekend to daily service times

akheron:
#### Summary

Saturday and sunday are always shown as options, but only around the clock units will use them.


hruoho (approved):
Great 👍

PR #1798, opened 2021-12-10
EVAKA-4204 preschool curriculum templates

jaahon:
#### Summary

* Add a permanent migration for curriculum related tables (and use the more generic "curriculum" name instead of "vasu"). Repeatable migrations delete all data when any changes are done, which isn't ideal since vasu is being actively tested in staging.
* Use generated api types in vasu code.
* Add support for creating leops templates and hide it behind a feature flag for now.


kimifoud (approved):
💯

PR #1815, opened 2021-12-15
Propagate all props directly to DatePickerDeprecated input

Gekkio:
#### Summary

Fixes at least two problems that happened since we didn't propagate
necessary props to our custom input: disabled state didn't work, enter
did not close the popup.

Since we're replacing an `<input>` with a styled `<input>`, we can just
propagate all props as is.


akheron:
You could remove the whole `CustomInput` and pass the input like this:
```
customInput={<StyledInput />}
```
The refs are not used and can be removed.

PR #1832, opened 2021-12-16
EVAKA-4256 Use max width for message container

hruoho:

### DIFF
--- a/frontend/src/employee-mobile-frontend/components/messages/ThreadView.tsx
+++ b/frontend/src/employee-mobile-frontend/components/messages/ThreadView.tsx
@@ -150,6 +150,7 @@ const RoundIconButton = styled.button`
 
 const MessageContainer = styled.div`
   border-radius: ${defaultMargins.s};
+  max-width: 90%;
   ${(p: { ours: boolean }) =>
     p.ours
       ? `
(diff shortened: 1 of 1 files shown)


frontend/src/employee-mobile-frontend/components/messages/ThreadView.tsx
```diff
@@ -150,6 +150,7 @@ const RoundIconButton = styled.button`
 
 const MessageContainer = styled.div`
   border-radius: ${defaultMargins.s};
+  max-width: 90%;
```
kimifoud, line 153: Is this to prevent the `kasdjfoashkdasdhasdhkaskdjaskhjdaskjfsjofashjodasjkasjkhdasfhdj` overflow? In that case, `overflow-wrap: anywhere;` could be considered too.

jaahon (commented):
I think you should use `word-break: break-word` as well

frontend/src/employee-mobile-frontend/components/messages/ThreadView.tsx
```diff
@@ -150,6 +150,9 @@ const RoundIconButton = styled.button`
...
+  word-break: break-word;
```
jaahon, line 154: You should probably use this only for `MessageContent` since otherwise it can for example cut the message times (happened to me locally).

hruoho: Ok thanks!

Also it seems that word-break: break-word; is deprecated: https://developer.mozilla.org/en-US/docs/Web/CSS/word-break

jaahon: Yeah seems like that's the right call then!

PR #1833, opened 2021-12-16
Resolve income statement care area correctly

akheron:
#### Summary

Use fridge partner relationships in addition to being a head of a child to find placements to determine the care area.

### DIFF
--- a/service/src/main/kotlin/fi/espoo/evaka/incomestatement/IncomeStatementController.kt
+++ b/service/src/main/kotlin/fi/espoo/evaka/incomestatement/IncomeStatementController.kt
@@ -11,6 +11,7 @@ import fi.espoo.evaka.shared.Paged
 import fi.espoo.evaka.shared.PersonId
 import fi.espoo.evaka.shared.auth.AuthenticatedUser
 import fi.espoo.evaka.shared.db.Database
+import fi.espoo.evaka.shared.domain.HelsinkiDateTime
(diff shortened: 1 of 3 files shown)


service/src/main/kotlin/fi/espoo/evaka/incomestatement/IncomeStatementController.kt
```diff
@@ -96,6 +97,6 @@ class IncomeStatementController(
         Audit.IncomeStatementsAwaitingHandler.log()
         accessControl.requirePermissionFor(user, Action.Global.FETCH_INCOME_STATEMENTS_AWAITING_HANDLER)
         val areasList = areas.split(",").filter { it.isNotEmpty() }
-        return db.read { it.fetchIncomeStatementsAwaitingHandler(areasList, page, pageSize) }
+        return db.read { it.fetchIncomeStatementsAwaitingHandler(LocalDate.now(), areasList, page, pageSize) }
```
kimifoud, line 100: You should use `HelsinkiDateTime.now().toLocalDate()` so that server timezone does not matter.

kimifoud (approved):
Performance needs to be monitored once this is deployed.

service/src/main/kotlin/fi/espoo/evaka/incomestatement/IncomeStatementController.kt
akheron: Fixed!

PR #1847, opened 2021-12-17
mapToPaged: Run the SQL query once instead of three times

akheron:


jaahon (commented):
This is very surprising 😄. Would changing the lambda's parameter to `List<WithCount<T>>` be a good idea to require that the query as been run before calling this function? Or make this an extension function on `ResultIterable<WithCount<T>>`?

jaahon (approved):

PR #1857, opened 2021-12-20
Show pre preschool assistance need and actions to SEO

reynders:
#### Summary
- In addition to admin, also SEO should see pre preschool assistance needs and actions

### DIFF
--- a/service/src/main/kotlin/fi/espoo/evaka/shared/security/AccessControl.kt
+++ b/service/src/main/kotlin/fi/espoo/evaka/shared/security/AccessControl.kt
@@ -97,6 +97,29 @@ WHERE employee_id = :userId
         "attachment.id",
         permittedRoleActions::attachmentActions,
     )
+
+    private val assistanceAction = ActionConfig(
+        """
+SELECT ac.id, role
+FROM child_acl_view acl
+JOIN assistance_action ac ON acl.child_id = ac.child_id
+WHERE acl.employee_id = :userId
+        """.trimIndent(),
+        "ac.id",
+        permittedRoleActions::assistanceActionActions
+    )
+
+    private val assistanceNeed = ActionConfig(
+        """
+SELECT an.id, role
+FROM child_acl_view acl
+JOIN assistance_need an ON acl.child_id = an.child_id
+WHERE acl.employee_id = :userId
+        """.trimIndent(),
+        "an.id",
+        permittedRoleActions::assistanceNeedActions
+    )
+
     private val backupCare = ActionConfig(
         """
 SELECT bc.id, role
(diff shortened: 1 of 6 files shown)


service/src/main/kotlin/fi/espoo/evaka/shared/security/AccessControl.kt
```diff
@@ -389,7 +412,7 @@ WHERE employee_id = :userId
             is AuthenticatedUser.Employee -> when (action) {
                 Action.AssistanceAction.READ_PRE_PRESCHOOL_ASSISTANCE_ACTION ->
                     // If child is or has been in preschool, do not show pre preschool daycare assistance actions for non admin
-                    if (user.isAdmin) true
+                    if (user.isAdmin || user.roles.contains(UserRole.SPECIAL_EDUCATION_TEACHER)) this.assistanceAction.hasPermission(user, action, id)
```
akheron, line 415: I think this should be:
```
if (user.isAdmin || (user.roles.contains(...) && this.assistanceAction.hasPermission(user, action, id)) return true
```
Otherwise admin doesn't get the permission.

service/src/main/kotlin/fi/espoo/evaka/shared/security/AccessControl.kt
```diff
@@ -419,7 +442,7 @@ WHERE employee_id = :userId
             is AuthenticatedUser.Employee -> when (action) {
                 Action.AssistanceNeed.READ_PRE_PRESCHOOL_ASSISTANCE_NEED ->
                     // If child is or has been in preschool, do not show pre preschool daycare assistance needs for non admin
-                    if (user.isAdmin) true
+                    if (user.isAdmin || user.roles.contains(UserRole.SPECIAL_EDUCATION_TEACHER)) this.assistanceNeed.hasPermission(user, action, id)
```
akheron, line 445: Same here

service/src/main/kotlin/fi/espoo/evaka/shared/security/Action.kt
```diff
@@ -153,7 +153,7 @@ sealed interface Action {
     enum class AssistanceAction(private val roles: EnumSet<UserRole>) : ScopedAction<AssistanceActionId> {
         UPDATE(SERVICE_WORKER, UNIT_SUPERVISOR, SPECIAL_EDUCATION_TEACHER),
         DELETE(SERVICE_WORKER, UNIT_SUPERVISOR, SPECIAL_EDUCATION_TEACHER),
-        READ_PRE_PRESCHOOL_ASSISTANCE_ACTION;
+        READ_PRE_PRESCHOOL_ASSISTANCE_ACTION(SPECIAL_EDUCATION_TEACHER);
```
akheron, line 156: This change is not really needed because the defaults are not used

service/src/main/kotlin/fi/espoo/evaka/shared/security/AccessControl.kt
reynders: Well that hasPermission function does allow everything for admin (there is even a test for that), but I agree it is better to be explicit than hide it underneath

akheron: Ok, okay. Didn't see that.

PR #1862, opened 2021-12-21
EVAKA-FIX citizen message editor height in percentage

hruoho:
#### Summary

Replace vh with percentage to avoid height problems with ios safari (experimental)


jaahon (approved):
Works on my iphone 👍

PR #1912, opened 2021-12-30
EVAKA-4174 show child capacity in group view

hruoho:
#### Summary

Screenshot 2021-12-30 at 13 36 51


service/src/main/kotlin/fi/espoo/evaka/placement/PlacementService.kt
```diff
@@ -273,6 +274,40 @@ private fun checkAclAuth(aclAuth: AclAuthorization, placement: Placement) {
...
+    return this.createQuery(
+        """
+        SELECT
+            ch.id child_id,
+            MAX(COALESCE(an.capacity_factor, 1)) assistance_need_factor,
+            MAX(CASE
+                WHEN u.type && array['FAMILY', 'GROUP_FAMILY']::care_types[] THEN 1.75
+                WHEN extract(YEARS FROM age(ch.date_of_birth)) < 3 THEN 1.75
+                ELSE 1.0
+            END) AS age_factor
```
jaahon, line 296: This other factor is affected not only by age and unit but also by service need. So the `ELSE` branch should use `service_need_option`'s `occupancy_coefficient`.

service/src/main/kotlin/fi/espoo/evaka/units/UnitsView.kt
```diff
@@ -112,7 +116,8 @@ class UnitsView(private val accessControl: AccessControl) {
                         placements.flatMap { placement ->
                             placement.groupPlacements.mapNotNull { groupPlacement -> groupPlacement.id }
                         }
-                    )
+                    ),
+                    unitChildrenCapacityFactors = if (accessControl.hasPermissionFor(user, Action.Unit.READ_CHILD_CAPACITY_FACTORS, unitId)) capacities else listOf()
```
jaahon, line 120: It would make more sense to do the access control check before querying the data to minimize unnecessary work.

frontend/src/employee-frontend/components/unit/tab-groups/groups/Group.tsx
```diff
@@ -216,6 +224,16 @@ export default React.memo(function Group({
...
 
+  const showChildOccupancyFactor =
+    !unit.type.includes('CLUB') &&
+    requireRole(
+      roles,
+      'ADMIN',
+      'UNIT_SUPERVISOR',
+      'SPECIAL_EDUCATION_TEACHER',
+      'SERVICE_WORKER'
+    )
```
jaahon, line 235: You shouldn't use these role checks in front end. I think the simplest check would be to check that the data exists.

frontend/src/e2e-playwright/specs/5_employee/unit-groups.spec.ts
```diff
@@ -184,7 +204,25 @@ describe('Unit groups - staff', () => {
...
+        endDate: placementEndDate.format('yyyy-MM-dd')
+      })
+      .save()
+
+    const groupsPage = await loadUnitGroupsPage()
+    await groupsPage.childOccupancyFactorColumnHeading.waitUntilHidden()
+    await waitUntilEqual(
+      () => groupsPage.childOccupancyFactorColumnData.count(),
+      0
+    )
```
jaahon, line 226: I think it would be a good idea to abstract this behind a single `groupsPage` method.

service/src/main/kotlin/fi/espoo/evaka/placement/PlacementService.kt
hruoho: Changed this to `coalesce(sno.occupancy_coefficient, default_sno.occupancy_coefficient, 1)`

jaahon (approved):
Good job! I think you could rename `ageFactor` to something else now, but otherwise everything looks good to me.

PR #1927, opened 2022-01-04
E2E test for mobile departure

reynders:
#### Summary
- E2E test for marking present child departed

### DIFF
--- a/frontend/src/e2e-playwright/specs/6_mobile/child-attendances.spec.ts
+++ b/frontend/src/e2e-playwright/specs/6_mobile/child-attendances.spec.ts
@@ -193,7 +193,50 @@ describe('Child mobile attendances', () => {
   })
 })
 
+const assertAttendanceCounts = async (
+  coming: number,
+  present: number,
+  departed: number,
+  absent: number,
+  total: number
+) =>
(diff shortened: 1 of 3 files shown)


frontend/src/e2e-playwright/specs/6_mobile/child-attendances.spec.ts
```diff
@@ -193,7 +193,50 @@ describe('Child mobile attendances', () => {
...
+  departed: number,
+  absent: number,
+  total: number
+) =>
+  await waitUntilEqual(() => listPage.getAttendanceCounts(), {
+    coming,
+    present,
+    departed,
+    absent,
+    total
```
kimifoud, line 208: Come to think of it, total could be omitted from params and just summed here `total: coming + present + departed + absent`

kimifoud (commented):
The original task was to create an e2e test for realtime _staff_ attendance which is lacking altogether. 😬

reynders:
> The original task was to create an e2e test for realtime _staff_ attendance which is lacking altogether. 😬

And I thought it was attendances made _by_ the staff :D. Well, will add those

PR #1935, opened 2022-01-04
Leops template changes and improvements

reynders:
#### Summary
- fix texts
- add and use new multi select text selection
- e2e test for creating a leops template and filling it

### DIFF
--- a/frontend/src/e2e-playwright/specs/5_employee/leops.spec.ts
+++ b/frontend/src/e2e-playwright/specs/5_employee/leops.spec.ts
@@ -75,6 +75,7 @@ describe('Child Information - Leops documents section', () => {
     await vasuTemplatesList.selectType.selectOption('PRESCHOOL')
     await vasuTemplatesList.okButton.click()
     await new VasuTemplateEditPage(page).saveButton.click()
+    await vasuTemplatesList.templateTable.waitUntilVisible()
   }
 
   test('Can create a leops template', async () => {
--- a/frontend/src/e2e-playwright/specs/5_employee/leops.spec.ts
+++ b/frontend/src/e2e-playwright/specs/5_employee/leops.spec.ts
@@ -8,16 +8,12 @@ import ChildInformationPage, {
 import { employeeLogin } from 'e2e-playwright/utils/user'
 import config from 'e2e-test-common/config'
 import {
-  insertDaycarePlacementFixtures,
+  deleteVasuTemplates,
   insertDefaultServiceNeedOptions,
   resetDatabase
 } from 'e2e-test-common/dev-api'
 import { initializeAreaAndPersonData } from 'e2e-test-common/dev-api/data-init'
-import {
-  createDaycarePlacementFixture,
-  Fixture,
-  uuidv4
-} from 'e2e-test-common/dev-api/fixtures'
+import { Fixture, uuidv4 } from 'e2e-test-common/dev-api/fixtures'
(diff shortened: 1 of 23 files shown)


frontend/src/e2e-playwright/specs/5_employee/leops.spec.ts
```diff
@@ -0,0 +1,106 @@
...
+
+  const createLeopsTemplate = async (templateName: string) => {
+    await new EmployeeNav(page).openAndClickDropdownMenuItem('vasu-templates')
+    const vasuTemplatesList = new VasuTemplatesListPage(page)
+    await vasuTemplatesList.addTemplateButton.click()
+    await vasuTemplatesList.nameInput.fill(templateName)
+    await vasuTemplatesList.selectType.selectOption('PRESCHOOL')
+    await vasuTemplatesList.okButton.click()
+    await new VasuTemplateEditPage(page).saveButton.click()
+  }
```
jaahon, line 79: It's a good thing that curriculum template creation is tested as well, but I think it should be tested in a separate test.

frontend/src/e2e-playwright/specs/5_employee/leops.spec.ts
```diff
@@ -0,0 +1,106 @@
...
+
+  test('Can add a new vasu document', async () => {
```
jaahon, line 81: This name could be better.

frontend/src/employee-frontend/components/vasu/components/MultiSelectQuestion.tsx
```diff
@@ -54,9 +65,14 @@ export function MultiSelectQuestion({
         <ValueOrNoRecord
           text={options
             .filter((option) => selectedValues.includes(option.key))
-            .map((o) => o.name)
+            .map((o) =>
+              textValue && textValue[o.key]
+                ? `${o.name} (${textValue[o.key]})`
```
jaahon, line 70: At least in the current use case, the emphasis is a bit backwards here in my opinion. The "text value" is more important than the option's "name" (eg. "muualle, minne?"). I know it's problematic to omit the "name" completely because this implementation is so generic that it might be important later. I don't know what to replace it with either, maybe `${o.name}: ${textValue[o.key]}` or something.

I'm not necessarily suggesting that you should change the model, but did you think about the option of modeling the solutions as an "other" option with a text field that could be optionally included at the end of a `MultiSelectQuestion`?

service/src/main/kotlin/fi/espoo/evaka/vasu/Vasu.kt
```diff
@@ -280,10 +280,11 @@ sealed class VasuQuestion(
         val options: List<QuestionOption>,
         val minSelections: Int,
         val maxSelections: Int?,
-        val value: List<String>
+        val value: List<String>,
+        val textValue: Map<String, String>? = emptyMap()
```
jaahon, line 284: I think `textValue` field shouldn't be nullable. It just makes using it harder and json deserialization will use the default value if it's missing from the json.

jaahon (approved):
Mostly everything looks good. I left a couple of comments, but they don't necessarily have to be addressed.

frontend/src/e2e-playwright/specs/5_employee/leops.spec.ts
reynders: I added a separate test for the leops template creation. I use this also for other tests because having a template fixture might cause having to update it whenever there is an unrelated changed in the framework.

PR #1938, opened 2022-01-04
Minor service code cleanup

Gekkio:


service/src/main/kotlin/fi/espoo/evaka/Audit.kt
```diff
@@ -157,7 +157,6 @@ enum class Audit(
     MessagingMessageReceiversRead("messaging.message.receivers.read"),
     MessagingReceivedMessagesRead("messaging.messages.received-messages.read"),
     MessagingSentMessagesRead("messaging.messages.sent-messages.read"),
-    MessagingGetReceivers("messaging.messages.receivers.read"),
```
akheron, line 160: Should we keep old audit events for future reference? 🤔

akheron (approved):

PR #1948, opened 2022-01-05
Remove Testcafe residues

akheron:
#### Summary

While at it, add `e2e-test-common` to the list of `lib-*` packages that require absolute importing.


jaahon (commented):
Why not just move stuff under `e2e-test-common` to `e2e-playwright`?

akheron:
I might do that as a separate thing. We could also discuss it first.

PR #1965, opened 2022-01-10
Evaka 4384 mobile bottom navi employee

zch:
#### Summary
- Solid icons for the selected item and light icons for options in the bottom navbar.
- Refactors weird layout in the employee mobile staff page that caused scrolling issues. Now you should no longer be required to scroll if the screen is big enough to fit everything.


frontend/src/employee-mobile-frontend/components/staff/StaffPage.tsx
```diff
@@ -78,15 +80,15 @@ export default React.memo(function StaffPage() {
   const today = useMemo(() => LocalDate.today(), [])
 
   return (
-    <>
+    <FixedSpaceColumn spacing="zero" style={{ height: '100vh' }}>
```
kimifoud, line 83: These layout changes make this tab behave inconsistently with the other tabs. The top bar seems to be squeezed a little bit.

image

Either revert the layout change or make all tabs behave similarly. This could also be done in a separate PR.

kimifoud (changes requested):
👍 for the icon changes.

frontend/src/employee-mobile-frontend/components/staff/StaffPage.tsx
zch: Fixed this by adding a `min-height` rule to the top bar

kimifoud: Well, that's just one instance affected by the squeezing problem 😅 

image
image

I think all pages should be wrapped in the same container components, or we'll be always battling these small differences.

zch: Splitting this off to a separate PR

kimifoud (approved):

PR #1968, opened 2022-01-10
Convert a lot of UUIDs to type-safe Ids

Gekkio:


akheron:
Rebasing onto master should fix the flaky test

PR #1972, opened 2022-01-10
EVAKA-4385 update input info visuals

jaahon:
#### Summary

* Use `1rem` font size in input placeholder and info texts and a darker orange with warning text.
* Add proper padding to input's right side when the input contains a button, icon or a symbol.
* Let info text continue to multiple rows, but do not let it stretch vertically.
* Fix input positioning when focus borders are added.

### DIFF
--- a/frontend/src/employee-mobile-frontend/components/attendances/components.tsx
+++ b/frontend/src/employee-mobile-frontend/components/attendances/components.tsx
@@ -116,7 +116,7 @@ export const TimeWrapper = styled.div`
   display: flex;
   flex-direction: column;
   align-items: center;
-  font-size: 20px;
+  font-size: 60px;
   color: ${colors.main.dark};
   font-weight: ${fontWeights.semibold};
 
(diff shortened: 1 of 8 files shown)


frontend/src/employee-mobile-frontend/components/attendances/components.tsx
```diff
@@ -116,7 +116,7 @@ export const TimeWrapper = styled.div`
   display: flex;
   flex-direction: column;
   align-items: center;
-  font-size: 20px;
+  font-size: 60px;
```
kimifoud, line 119: Some problems with `TimeWrapper`. These might have been there before these changes
image

kimifoud (commented):
Do we want to limit the info message width to input width? Maybe?

Old:
image

New:
image

kimifoud:
image
The info texts need a little air.

kimifoud:
Old
image

New
image

jaahon:
> Do we want to limit the info message width to input width? Maybe?

Yeah, I remember there being discussion about that, although I can't find the discussion anywhere. Previously the info texts broke layout when there was anything to the right of the input.

> The info texts need a little air.

I'm not sure where to add that "air". I think it would be nice if the info text didn't break spacing between components, but then it would always change the layout.

jaahon:
I removed the negative margin from input info completely so now spacing between elements should always be the same. This causes unpleasant shifts in layout, but they were already happening because previously the negative margin wasn't large enough.

I tried to animate it, but gave up because it wasn't that straightforward with the dynamic text content and it didn't look very good anyway.

kimifoud (approved):
LGTM! 👍

PR #1973, opened 2022-01-10
Use shorter `AsyncButton` animations in e2e tests

akheron:
#### Summary

This makes e2e faster, because animations took seconds and now they take milliseconds.

Some tests needed fixing that relied on seeing one of the transient `AsyncButton` states.

### DIFF
--- a/frontend/src/e2e-playwright/pages/employee/finance-basics.ts
+++ b/frontend/src/e2e-playwright/pages/employee/finance-basics.ts
@@ -2,9 +2,9 @@
 //
 // SPDX-License-Identifier: LGPL-2.1-or-later
 
-import { waitUntilEqual } from 'e2e-playwright/utils'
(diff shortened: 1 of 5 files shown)


frontend/src/e2e-playwright/specs/4_finance/fee-decisions.spec.ts
```diff
@@ -87,6 +87,7 @@ describe('Fee decisions', () => {
       enduserGuardianFixture,
       enduserChildFixtureKaarina
     )
+    await page.pause()
```
kimifoud, line 90: 🤔

akheron: Oops

kimifoud (approved):
🥇

PR #1976, opened 2022-01-11
Varda: remove unnecessarily strict application condition

jaahon:
#### Summary

Not all applications and placement proposals go through unit supervisors so this condition for finding a placement's application is unnecessarily strict. Just noticed this by accident.


hruoho:
Hmm good catch! I still wonder though should we have _some_ condition there to prevent placements that will be rejected 🤔

jaahon:
> Hmm good catch! I still wonder though should we have _some_ condition there to prevent placements that will be rejected 🤔

But doesn't `a.status IN ('ACTIVE')` already guarantee that? If a placement plan is rejected, it's replaced by a new one.

hruoho:
> > Hmm good catch! I still wonder though should we have _some_ condition there to prevent placements that will be rejected 🤔
> 
> But doesn't `a.status IN ('ACTIVE')` already guarantee that? If a placement plan is rejected, it's replaced by a new one.

Ah yes... I believe it does 👍

hruoho (approved):
Let's ship it in and see what happens :)

PR #1977, opened 2022-01-11
Fix scrolling issues on employee mobile staff page

zch:
#### Summary
The staff page contained unnecessary scrolling when viewed on a screen
big enough to fit everything. This change refactors the layouts of all
the pages by moving the top bar and the bottom navigation bar to a
separate component, which wraps the content of each separate page.

The vertical scrolling issues were fixed by using flexbox to expand the
staff page content to full height. The previous solution set the height
of the content area to equal the full height of the screen, which caused
unnecessary vertical scrolling.

The horizontal scrolling (on some browsers) was caused by negative
margins on the cancel / confirm buttons. These were removed and the
paddings inside of the buttons were removed to allow for relatively good
results even on screens only 300px wide.


akheron (approved):
Very nice 💎

kimifoud (commented):
Nice! I think `StaffAttendancesPage` should be using this too.

PR #1987, opened 2022-01-11
Reset child from varda error report

reynders:


hruoho (approved):
Apart from the unresolved conflict, this looks good to me 👍

PR #1995, opened 2022-01-13
Replace invoice agreement_type with area_id

Gekkio:
#### Summary

* `invoice` table: `agreement_type` replaced by `area_id`
* `Invoice` type: `agreementType: Int` replaced by `areaId: AreaId`
* `InvoiceDetailed` type: `areaId: AreaId` added so it has both. Agreement type is joined from care_area in queries


service/src/main/kotlin/fi/espoo/evaka/invoicing/service/InvoiceService.kt
```diff
@@ -28,11 +29,10 @@ import java.time.Instant
 import java.time.LocalDate
 import java.util.UUID
 
-data class DaycareCodes(val areaCode: Int?, val costCenter: String?, val subCostCenter: String?)
+data class DaycareCodes(val areaId: AreaId, val areaCode: Int?, val costCenter: String?, val subCostCenter: String?)
```
jaahon, line 32: `val areaCode: Int?` is now unused.

jaahon (approved):

PR #1999, opened 2022-01-14
Replace invoice row cost_center with unit_id

Gekkio:

### DIFF
--- a/service/src/main/kotlin/fi/espoo/evaka/invoicing/service/InvoiceService.kt
+++ b/service/src/main/kotlin/fi/espoo/evaka/invoicing/service/InvoiceService.kt
@@ -17,6 +17,7 @@ import fi.espoo.evaka.invoicing.domain.InvoiceStatus
 import fi.espoo.evaka.invoicing.domain.Product
 import fi.espoo.evaka.invoicing.integration.InvoiceIntegrationClient
 import fi.espoo.evaka.shared.AreaId
+import fi.espoo.evaka.shared.DaycareId
 import fi.espoo.evaka.shared.InvoiceId
 import fi.espoo.evaka.shared.InvoiceRowId
 import fi.espoo.evaka.shared.auth.AuthenticatedUser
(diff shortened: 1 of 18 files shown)


service/src/main/kotlin/fi/espoo/evaka/invoicing/service/InvoiceService.kt
```diff
@@ -68,6 +71,7 @@ class InvoiceService(
             tx.updateInvoiceDates(succeeded.map { it.id }, invoiceDate, dueDate)
         }
         tx.setDraftsSent(succeeded.map { it.id to it.number!! }, user.evakaUserId)
+        tx.saveCostCenterFields(succeeded.map { it.id })
         tx.updateToWaitingForSending(withoutSSNs.map { it.id })
```
jaahon, line 75: Cost center information should be saved also for invoices that are manually sent (`withoutSSNs` here).

frontend/src/employee-frontend/components/invoice/InvoiceRowsSectionRow.tsx
```diff
@@ -273,4 +253,89 @@ const UnitPriceInput = React.memo(function UnitPriceInput({
...
+    [unitDetails]
+  )
+  const onChange = useCallback(
+    (value: UUID | null) => update({ unitId: value }),
+    [update]
+  )
+  const getItemLabel = useCallback(
+    (unitId: UUID) => unitDetails[unitId]?.name ?? unitId,
+    [unitDetails]
+  )
```
jaahon, line 290: I think showing the cost center in the item label could be valuable or was it left out to save space?

frontend/src/employee-frontend/components/invoice/InvoiceRowsSectionRow.tsx
```diff
@@ -109,34 +106,19 @@ function InvoiceRowSectionRow({
...
+            unitDetails={unitDetails}
+            update={update}
+          ></UnitCombobox>
         ) : (
-          <div>{subCostCenter}</div>
+          <div>
+            <span>{unit?.name}</span>
+            {unit?.costCenter && (
+              <UnitCostCenter>{unit.costCenter}</UnitCostCenter>
+            )}
```
jaahon, line 120: I think the `savedCostCenter` should be shown here instead.

frontend/src/employee-frontend/components/invoice/InvoiceRowsSectionRow.tsx
Gekkio: There's not much space, and based on a quick UX talk it's probably better leave all the space for the name. If we get feedback that the cost center is important enough to be shown there, we can add it easily later.

PR #2016, opened 2022-01-18
Fix postal info layout

reynders:
#### Summary
Screenshot 2022-01-18 at 15 50 22


frontend/src/employee-frontend/components/unit/unit-details/UnitEditor.tsx
```diff
@@ -164,7 +164,7 @@ const AddressSecondRowContainer = styled.div`
   justify-content: space-between;
 
   & > :first-child {
-    margin-right: 20px;
+    margin-right: 120px;
```
jaahon, line 167: I think the correct solution would be to remove `min-width: 0` from `InputField.tsx -> Wrapper`. I broke it.

PR #2018, opened 2022-01-19
EVAKA-4399 Configurable placement termination unit feature

kimifoud:
#### Summary

Add placement termination to configurable unit features.
Some units (i.e. service voucher units) do not allow placement termination via eVaka.


akheron:
We discussed in yesterday's daily that this should not be a configurable flag. Instead just disable it for voucher units. I guess no-one remembered to update the ticket 😬

akheron (approved):
Maybe at some point the feature should be renamed to `UnitFeatures` or something, because now at the latest all the features have nothing to do with pilot units anymore.

kimifoud:
> Maybe at some point the feature should be renamed to `UnitFeatures` or something, because now at the latest all the features have nothing to do with pilot units anymore.

True that. Reservations could also be a feature that will never be relevant for some (probably outsourced) units.

PR #2029, opened 2022-01-21
Elementary codegen support for maps

jaahon:
#### Summary

No tests yet. Map keys are types as `string` unless the key is an enum and there's extra value in using a union of string literals as the type. Numeric types could also use `number` but that's not done yet.

Converted array types from `T[]` to `Array<T>` so we don't have to bother with parentheses with nested arrays or nullable type parameters.


kimifoud (approved):
🎖️ Nice work!

service/codegen/src/main/kotlin/evaka/codegen/apitypes/Generate.kt
```diff
@@ -188,16 +169,70 @@ ${values.joinToString("\n") { "  | '$it'" }}"""
...
+
+data class TsPlain(val type: KType) : AnalyzedType {
+    override val declarableTypes = listOf(type.jvmErasure)
+    override fun toTs() = toNullableTs(type.isMarkedNullable, tsName(type.jvmErasure))
+}
+
+data class TsArray(val type: KType) : AnalyzedType {
+    private val typeParameter = unwrapCollection(type)
+    override val declarableTypes = listOf(typeParameter)
+    override fun toTs() = toNullableTs(type.isMarkedNullable, "Array<${tsName(typeParameter)}>")
```
akheron, line 198: Why did change `X[]` to `Array<X>`? The diff is unnecessarily large and we use `X[]` in hand-crafted code, too.

jaahon: It might not be a very relevant concern, but collections of nullable types like `List<T?>` would need extra parentheses `(T | null)[]` but `Array<T | null>` works automatically.

jaahon: Arrays are now written as `X[]`. Nullable types weren't handled correctly anyway so I had to fix it separately.

PR #2032, opened 2022-01-21
Highlight attendance time differences in unit calendar week view

kimifoud:
#### Summary

Highlight differences in expected and actual attendance times in week view.

The calculation is "best effort" based on two `HH:mm` local time strings without date part. This is only used for highlighting in the calendar so edge cases are expected and might not matter much. The functionality can be improved once the reservation data model supports reservations that span multiple days.

### DIFF
--- a/frontend/src/employee-frontend/components/unit/unit-reservations/ChildDay.tsx
+++ b/frontend/src/employee-frontend/components/unit/unit-reservations/ChildDay.tsx
@@ -8,17 +8,58 @@ import {
   getTimesOnWeekday,
   isIrregular,
   isRegular,
-  isVariableTime
+  isVariableTime,
+  TimeRange
 } from 'lib-common/api-types/child/common'
 import {
   ChildReservations,
   OperationalDay
 } from 'lib-common/api-types/reservations'
+import { Reservation } from 'lib-common/generated/api-types/reservations'
 import { fontWeights } from 'lib-components/typography'
 import { defaultMargins } from 'lib-components/white-space'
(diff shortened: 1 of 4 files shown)


jaahon (commented):
There is no distinction between a reservation and "service time" anymore. I think you should add that back.

kimifoud:
> There is no distinction between a reservation and "service time" anymore. I think you should add that back. Also, there's no icon shown with the warning.
🤦

frontend/src/employee-frontend/components/unit/unit-reservations/ChildDay.tsx
```diff
@@ -51,34 +94,36 @@ export default React.memo(function ChildDay({ day, childReservations }: Props) {
       ? getTimesOnWeekday(serviceTimes, day.date.getIsoDayOfWeek())
       : null
 
+  const expectedTimeForThisDay = getReservationOrServiceTimeOfDay(
+    dailyData.reservations,
+    serviceTimeOfDay
+  )
+  const maybeAsterisk = expectedTimeForThisDay?.type === 'service-time' && '*'
```
jaahon, line 101: I think the current spec uses ` (s)` instead of an asterisk.

kimifoud: Using ` (s)` introduced some problems with our fixed with container. Circumvented the problem by shrinking the rightmost action column and preventing text wrapping.

jaahon (approved):

PR #2036, opened 2022-01-24
EVAKA-4142 absence service refactor

jaahon:
#### Summary

- Replace vars with vals in AbsenceService
- Remove unused AbsenceType values
- Simplify absences api types and use generated code when possible


zch (approved):
LGTM

PR #2039, opened 2022-01-24
Change title of attachment button to encourage using PDFs

zch:


hruoho (approved):
Let's see if this really makes any difference 😁

PR #2042, opened 2022-01-24
Error: Element inside a table must be wrapped in tr and td

reynders:
#### Summary

There was an error in console that the modal is inside a table but not inside tr and td, so wrap it inside them


kimifoud (commented):
Imho a better solution would be extracting the modal state and its logic from the row component to parent container.

kimifoud (approved):

PR #2048, opened 2022-01-25
Make sure session maxAge is always up to date

Gekkio:
#### Summary

If this is not done, req.session will contain stale expiry timestamp from redis data.


kimifoud (approved):
👌

PR #2058, opened 2022-01-27
Add types to date formatting functions

kimifoud:
#### Summary

- Strictly define allowed date, time and datetime formats
- ~Separate ISO date formatting from human readable formatting~ Replace ISO formatting with `LocalDate`
- Separate time formatting from date formatting

### DIFF
--- a/frontend/src/employee-frontend/components/reports/PresenceReport.tsx
+++ b/frontend/src/employee-frontend/components/reports/PresenceReport.tsx
@@ -1,4 +1,4 @@
-// SPDX-FileCopyrightText: 2017-2020 City of Espoo
+// SPDX-FileCopyrightText: 2017-2022 City of Espoo
 //
 // SPDX-License-Identifier: LGPL-2.1-or-later
 
@@ -63,7 +63,7 @@ function Presences() {
             <ReportDownload
               data={rows.value.map((row) => ({
                 ...row,
-                date: row.date.format('dd/MM/yyyy'),
+                date: row.date.formatIso(),
                 present:
                   row.present === true
                     ? 'kyllä'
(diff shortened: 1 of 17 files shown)


frontend/src/lib-common/date.ts
```diff
@@ -1,25 +1,45 @@
...
+type IsoDate = 'yyyy-MM-dd'
+type IsoDateWithoutDay = 'yyyy-MM'
+type IsoFormat = IsoDate | IsoDateWithoutDay
+
+export function formatIsoDate(
+  date: Date | null | undefined,
+  isoFormat: IsoFormat = 'yyyy-MM-dd'
+): string {
+  return date ? format(date, isoFormat) : ''
+}
```
jaahon, line 39: I don't think we should even have this function since using `Date` instead of `LocalDate` shouldn't be encouraged when only the date matters.

frontend/src/employee-frontend/components/reports/PresenceReport.tsx
```diff
@@ -63,7 +63,7 @@ function Presences() {
             <ReportDownload
               data={rows.value.map((row) => ({
                 ...row,
-                date: row.date.format('dd/MM/yyyy'),
+                date: row.date.formatIso(),
```
jaahon, line 66: This is a breaking change 😅 but I think `dd.MM.yyyy` would be more appropriate here.

frontend/src/lib-common/date.ts
kimifoud: Let's see what I can do 👍

frontend/src/employee-frontend/components/reports/PresenceReport.tsx
kimifoud: Hah, nice catch. It was intentional and I used ISO because it was widely used in the reports and should be more compliant in CSV exports. What do you think?

jaahon: Consistency is always good 👍

jaahon (approved):

PR #2062, opened 2022-01-27
A tool to generate diffs between the current and new invoice generator

reynders:
#### Summary
- A tool to generate a diff of what the current and new invoice generators produce
- Necessary changes in current invoice generator to make diffing possible
- Controller to get the diff
- Basic tests


service/src/main/kotlin/fi/espoo/evaka/invoicing/service/InvoiceGenerator.kt
```diff
@@ -63,23 +80,47 @@ class InvoiceGenerator(private val draftInvoiceGenerator: DraftInvoiceGenerator)
...
+            invoiceData.absences,
+            invoiceData.freeChildren,
+            invoiceData.codebtors
+        )
     }
 
+    data class InvoiceCalculationData(
+        val decisions: Map<PersonId, List<FeeDecision>>,
+        val placements: Map<PersonId, List<Placements>>,
+        val period: DateRange,
```
jaahon, line 115: Nitpicking, but `period` could be removed. It's passed straight into `InvoiceCalculationData` from the function parameter anyway.

jaahon (approved):

PR #2063, opened 2022-01-27
EVAKA-4378 better ui support for days with two reservations

hruoho:
#### Summary

Add better support for overnight reservations.

Screenshot 2022-01-27 at 16 03 32

### DIFF
--- a/frontend/src/employee-frontend/components/unit/unit-reservations/ReservationsTable.tsx
+++ b/frontend/src/employee-frontend/components/unit/unit-reservations/ReservationsTable.tsx
@@ -2,15 +2,16 @@
 //
 // SPDX-License-Identifier: LGPL-2.1-or-later
 
-import classNames from 'classnames'
 import React from 'react'
 import { Link } from 'react-router-dom'
 import styled, { css } from 'styled-components'
 import {
   Child,
-  ChildReservations,
+  ChildRecordOfDay,
+  ChildDailyRecords,
   OperationalDay
 } from 'lib-common/api-types/reservations'
+import { JsonOf } from 'lib-common/json'
 import LocalDate from 'lib-common/local-date'
 import IconButton from 'lib-components/atoms/buttons/IconButton'
 import { Table, Tbody, Td, Th, Thead, Tr } from 'lib-components/layout/Table'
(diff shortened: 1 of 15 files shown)


frontend/src/employee-frontend/components/unit/unit-reservations/ReservationsTable.tsx
```diff
@@ -113,9 +137,8 @@ const CustomTh = styled(Th)<{ shrink?: boolean }>`
     `}
 `
 
-const DateTh = styled(CustomTh)<{ faded: boolean; highlight: boolean }>`
+const DateTh = styled(CustomTh)<{ faded: boolean }>`
   ${(p) => p.faded && `color: ${colors.grayscale.g35};`}
-  ${(p) => p.highlight && `border-bottom: 2px solid ${colors.main.m3};`}
```
jaahon, line 118: Is there a reason you took this highlight out? It's there in UI specs.

service/src/main/kotlin/fi/espoo/evaka/reservations/AttendanceReservationController.kt
```diff
@@ -262,21 +262,33 @@ WHERE child_id = ANY(:childIds)
...
+                                absence = it.absence
+                            )
+                        }
+                    ),
+                    if (dailyData.any { it.reservations.size > 1 }) dailyData.associateBy(
+                        keySelector = { it.date },
+                        valueTransform = {
+                            UnitAttendanceReservations.ChildRecordOfDay(
+                                reservation = it.reservations.getOrNull(1),
+                                attendance = it.attendance,
```
jaahon, line 287: You'll have to similarly like with reservations use only the second attendance of the day, otherwise the same attendance is shown on both rows.

frontend/src/employee-frontend/components/unit/unit-reservations/ReservationsTable.tsx
hruoho: Ah it seems to be included in one figma view and but not in the other :) I was looking at the other...

jaahon: Then I don't know if it should be taken out or not 😄

service/src/main/kotlin/fi/espoo/evaka/reservations/AttendanceReservationController.kt
hruoho: Yes of course, thanks for catching that!

hruoho: Ok now attendances are handled in the similar way than reservations

jaahon (approved):
🎖️

PR #2067, opened 2022-01-30
Custom KeyCloak event logging handler

ilja-git:
#### Summary

Provides custom event logging handlers that can be configured to omit or hash fields.


akheron (approved):
LGTM, although I'm not an expert on this stuff 😄

PR #2074, opened 2022-01-31
Update `passport` and `passport-saml` packages to latest versions

zch:


Gekkio (commented):
Looks fine to me!
Two comments:

1. I think we should get rid of the custom passport-saml types (`shared/passport-saml.d.ts`). Those types were added when passport-saml was JS + manual types, and the library's own types were sometimes wrong. Now that passport-saml is TS and seems to export everything we need, the library types should always be correct and should be used directly. If this causes type mismatches / compilation errors, they are probably real bugs in our code.

2. The xmldom override in package.json and the old xmldom files in the repo should be removed since they are no longer needed. They were only used to force a newer version to be used to avoid a security problem. Passport-saml now uses `@xmldom/xmldom`, which is considered a different library than `xmldom`, so none of the old xmldom stuff applies.

Gekkio (approved):

PR #2084, opened 2022-02-02
Tidy up FullApplicationTest

Gekkio:
#### Summary

This is the base class for *all* full application tests, so it should be minimal and only contain stuff needed for the majority of tests.


kimifoud (approved):
👍

PR #2087, opened 2022-02-02
Update group table service needs

patari:
#### Summary

Show only current service need in group tables.

Note: This doesn't affect Espoo because `groupsTableServiceNeedsEnabled` feature flag is false.


jaahon (approved):
Is it ok that an empty tooltip is show when the child has no service need set?

patari:
> Is it ok that an empty tooltip is show when the child has no service need set?

At least it has been ok with the previous implementation.

PR #2103, opened 2022-02-09
EVAKA-4152 FREE_ABSENCE AbsenceType

kimifoud:
#### Summary

Add `FREE_ABSENCE` absence type to be used with free summer holiday periods. Currently it works just like `FORCE_MAJEURE`.


akheron:
The free absences should probably be just ignored in the invoice generator for now, i.e. treat them like `OTHER_ABSENCE`. I'm pretty sure they'll eventually need separate handling, an own product code etc.

kimifoud:
Will add the enum value in #2122

PR #2109, opened 2022-02-10
EVAKA-4377 reservations inline editing

jaahon:
#### Summary

Add option to edit reservations inline in employee week view. Also fix day highlight change that was lost in some rebase.

Does not include editing attendances since it's a lot more complicated because of multi-day attendances.


reynders (approved):
Agreed with team it is ok to merge this to master

kimifoud (commented):
Navigating between cells triggers save even though nothing has changed. Should that be fixed?

jaahon:
> Navigating between cells triggers save even though nothing has changed. Should that be fixed?

Yeah, I'll do that 👍

kimifoud (approved):

PR #2130, opened 2022-02-16
EVAKA-4162 Highlight holiday periods in citizen's reservation calendar

kimifoud:


frontend/src/citizen-frontend/calendar/CalendarGridView.tsx
```diff
@@ -35,6 +37,7 @@ export default React.memo(function CalendarGridView({
   onCreateReservationClicked,
   onCreateAbsencesClicked,
   onReportHolidaysClicked,
+  holidayPeriods,
```
zch, line 40: Instead of passing this here, you could use the `useHolidayPeriods` hook for a bit less boilerplate.

kimifoud: True, but then I would have a `Result<HolidayPeriod[]>` within the component tree where everything else is static data. This could (in theory) result in multiple renders, but I guess it's worth a try.

zch (approved):

PR #2131, opened 2022-02-16
Dynamic E2E customization overrides

Gekkio:

### DIFF
--- a/frontend/src/lib-customizations/citizen.tsx
+++ b/frontend/src/lib-customizations/citizen.tsx
@@ -4,16 +4,35 @@
 
 // eslint-disable-next-line @typescript-eslint/ban-ts-comment
 // @ts-ignore
-import customizations from '@evaka/customizations/citizen'
+import defaultsUntyped from '@evaka/customizations/citizen'
 import { mergeWith } from 'lodash'
 
-import { translationsMergeCustomizer } from './common'
+import { JsonOf } from 'lib-common/json'
+
+import { mergeCustomizer } from './common'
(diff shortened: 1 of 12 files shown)


frontend/src/lib-customizations/citizen.tsx
```diff
@@ -2,17 +2,36 @@
 //
 // SPDX-License-Identifier: LGPL-2.1-or-later
 
+/// types: node
 // eslint-disable-next-line @typescript-eslint/ban-ts-comment
 // @ts-ignore
-import customizations from '@evaka/customizations/citizen'
+import defaultsUntyped from '@evaka/customizations/citizen'
```
akheron, line 8: Why have these been renamed? The `untyped` part puzzles me 🤔

akheron (approved):
Otherwise looks good 🏅

frontend/src/lib-customizations/citizen.tsx
Gekkio: All customizations follow this basic structure:

```typescript
import defaultsUntyped from '@evaka/customizations/citizen'

const defaults: CitizenCustomizations = defaultsUntyped

const customizations: CitizenCustomizations = overrides
  ? mergeWith({}, defaults, overrides, mergeCustomizer)
  : defaults
```

`defaultsUntyped` has type `any`, because we're importing from an alias that TS compiler can't understand, so we need to then add type for it (and silence eslint warnings from any -> CitizenCustomizations assignment)

akheron (approved):

PR #2136, opened 2022-02-18
Fix a NullPointerException, disallow double clicks

akheron:
#### Summary

Try to fix the root issue by avoiding double clicks (using `AsyncButton`) in placement update save button.

Fix the NPE to another exception by using the correct mapping functions.


kimifoud (approved):
👌

PR #2141, opened 2022-02-21
EVAKA-4162 Hide CTA banner on non-relevant pages

kimifoud:
#### Summary

Banner would be rendered on top of the sub navigation on applications page. Just hide the banner on non-relevant pages instead of trying to adapt the absolute-positioned map view to changing available viewport area.

### DIFF
--- a/frontend/src/citizen-frontend/calendar/HolidayPeriodBanner.tsx
+++ b/frontend/src/citizen-frontend/calendar/HolidayPeriodBanner.tsx
@@ -9,6 +9,7 @@ import FiniteDateRange from 'lib-common/finite-date-range'
 import LocalDate from 'lib-common/local-date'
 import { useDataStatus } from 'lib-common/utils/result-to-data-status'
 import RoundIcon from 'lib-components/atoms/RoundIcon'
+import { desktopMin } from 'lib-components/breakpoints'
 import Container from 'lib-components/layout/Container'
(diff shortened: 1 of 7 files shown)


frontend/src/citizen-frontend/calendar/HolidayPeriodBanner.tsx
```diff
@@ -47,22 +46,24 @@ const HolidayPeriodBanner = React.memo(function HolidayPeriodBanner({
   return (
     <Banner data-qa="holiday-period-banner">
       <RoundIcon content={faTreePalm} size="L" color={colors.status.warning} />
-      <span>
+      <a>
```
zch, line 49: `<a>` without a `href` feels strange. What's the reasoning for using `<a>` instead of `<span>` here?

zch (commented):
Right now the banner is immediately scrolled out of view on the calendar page, which means that it will likely be missed. Maybe we need to make it sticky at this point?

frontend/src/citizen-frontend/calendar/HolidayPeriodBanner.tsx
kimifoud: FFFFuuuuu... Good catch! That was a WIP mistake 😄

zch (approved):

PR #2146, opened 2022-02-23
EVAKA-4411 report access control changes

jaahon:
#### Summary

- Give service workers access to ended placements report
- Give unit supervisors access to occupancy report
- Give unit supervisors access to applications report


frontend/src/employee-frontend/components/Reports.tsx
```diff
@@ -194,7 +194,13 @@ function Reports() {
...
           </RequireRole>
           <RequireRole
-            oneOf={['ADMIN', 'SERVICE_WORKER', 'DIRECTOR', 'REPORT_VIEWER']}
+            oneOf={[
+              'ADMIN',
+              'SERVICE_WORKER',
+              'DIRECTOR',
+              'REPORT_VIEWER',
+              'UNIT_SUPERVISOR'
+            ]}
```
akheron, line 203: Would it be reasonable to add `getPermittedActions` for the reports? 🤔

jaahon: It would probably have to be rewritten completely when access control dsl is finished. So I don't think it's worth it, but I'm not completely against it.

jaahon: Yeah, I'll do that.

akheron (approved):

PR #2151, opened 2022-02-24
EVAKA-4162 Add server-side validation for adding free absences

zch:

### DIFF
--- a/service/src/main/kotlin/fi/espoo/evaka/holidayperiod/HolidayPeriodControllerCitizen.kt
+++ b/service/src/main/kotlin/fi/espoo/evaka/holidayperiod/HolidayPeriodControllerCitizen.kt
@@ -13,6 +13,8 @@ import fi.espoo.evaka.shared.ChildId
 import fi.espoo.evaka.shared.PersonId
 import fi.espoo.evaka.shared.auth.AuthenticatedUser
 import fi.espoo.evaka.shared.db.Database
+import fi.espoo.evaka.shared.domain.BadRequest
+import fi.espoo.evaka.shared.domain.EvakaClock
 import fi.espoo.evaka.shared.domain.FiniteDateRange
 import fi.espoo.evaka.shared.security.AccessControl
 import fi.espoo.evaka.shared.security.Action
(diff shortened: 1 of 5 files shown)


service/src/main/kotlin/fi/espoo/evaka/holidayperiod/HolidayPeriodQueries.kt
```diff
@@ -40,6 +41,19 @@ fun Database.Read.getHolidayPeriod(id: HolidayPeriodId): HolidayPeriod? =
         .mapTo<HolidayPeriod>()
         .firstOrNull()
 
+fun Database.Read.getActiveFreeHolidayPeriod(date: LocalDate): HolidayPeriod? =
+    this.createQuery(
+        """
+$holidayPeriodSelect
+WHERE show_reservation_banner_from < :date
```
kimifoud, line 48:
```suggestion
WHERE show_reservation_banner_from <= :date
```

service/src/main/kotlin/fi/espoo/evaka/holidayperiod/HolidayPeriodControllerCitizen.kt
```diff
@@ -48,7 +51,18 @@ class HolidayPeriodControllerCitizen(private val accessControl: AccessControl) {
...
+                if (activeHolidayPeriod?.freePeriod == null) {
+                    throw NotFound("No active free holiday period found")
+                }
+
                 val childToDateRange = body.childHolidays.entries.filter { it.value != null }.map { it.key to it.value!! }
+                childToDateRange.forEach { (_, dateRange) ->
+                    if (!activeHolidayPeriod.freePeriod.periodOptions.contains(dateRange)) {
+                        throw NotFound("Free holiday period not found")
+                    }
+                }
```
kimifoud, line 64: Using `also` would be slightly more idiomatic Kotlin. Moreover, how about `BadRequest` in both of the `throw` cases instead of `NotFound`?

kimifoud (approved):

PR #2161, opened 2022-02-25
EVAKA-4400 Permission changes

akheron:


service/src/main/kotlin/fi/espoo/evaka/shared/security/AccessControl.kt
```diff
@@ -363,6 +362,7 @@ WHERE employee_id = :userId
             createUnits = hasPermissionFor(user, Action.Global.CREATE_UNIT),
             vasuTemplates = user.hasOneOfRoles(UserRole.ADMIN),
             personalMobileDevice = user.hasOneOfRoles(UserRole.UNIT_SUPERVISOR),
+            pinCode = !user.hasOneOfRoles(UserRole.FINANCE_ADMIN),
```
Gekkio, line 365: A negative check seems a bit strange to me :thinking: 
Why do we explicitly take out FINANCE_ADMIN, but everybody else (including users with no roles at all) have permission?

service/src/main/kotlin/fi/espoo/evaka/attendance/MobileUnitController.kt
```diff
@@ -127,7 +127,7 @@ fun Database.Read.fetchUnitInfo(unitId: DaycareId, date: LocalDate): UnitInfo {
             WHERE dg.daycare_id = :id
             GROUP BY employee_id
         ) group_acl ON acl.employee_id = group_acl.employee_id
-        WHERE acl.daycare_id = :id AND acl.role = ANY('{STAFF, UNIT_SUPERVISOR}')
+        WHERE acl.daycare_id = :id AND acl.role = ANY('{STAFF, UNIT_SUPERVISOR, SPECIAL_EDUCATION_TEACHER}')
```
Gekkio, line 130: Suggestion: remove the role check completely, since all possible role column values are now in the list anyway

service/src/main/kotlin/fi/espoo/evaka/shared/security/AccessControl.kt
akheron: The idea was to communicate intent: Everyone else except FINANCE_ADMIN. Fixed by listing all roles and adding a comment instead.

service/src/main/kotlin/fi/espoo/evaka/attendance/MobileUnitController.kt
akheron: Fixed 👍

service/src/main/kotlin/fi/espoo/evaka/shared/security/AccessControl.kt
Gekkio: Yeah I understood the intent (except for completely role-less users), but it seems like a strange requirement. For example, why does DIRECTOR need access to pin code functionality?

PR #2181, opened 2022-03-04
Fix voucher value report for annulled decisions

jaahon:
#### Summary

Fix voucher value report so that it correctly refunds annulled decisions or decisions that have had their end date updated because a child has left daycare.

Couldn't come up with any new test cases, but I'm still unsure whether these changes take every single case into account.

### DIFF
--- a/service/src/main/kotlin/fi/espoo/evaka/invoicing/service/VoucherValueDecisionService.kt
+++ b/service/src/main/kotlin/fi/espoo/evaka/invoicing/service/VoucherValueDecisionService.kt
@@ -6,6 +6,7 @@ package fi.espoo.evaka.invoicing.service
 
 import fi.espoo.evaka.BucketEnv
 import fi.espoo.evaka.invoicing.client.S3DocumentClient
+import fi.espoo.evaka.invoicing.data.annulVoucherValueDecisions
 import fi.espoo.evaka.invoicing.data.getValueDecisionsByIds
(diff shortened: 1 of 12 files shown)


service/src/main/kotlin/fi/espoo/evaka/invoicing/service/VoucherValueDecisionService.kt
```diff
@@ -164,11 +164,11 @@ ORDER BY start_date ASC
                 when {
                     mergedPlacementPeriods.isEmpty() -> {
                         val annulled = decision.annul()
-                        tx.updateVoucherValueDecisionStatusAndDates(listOf(annulled))
```
kimifoud, line 167: The row above seems to be redundant now

service/src/main/resources/db/migration/V212__voucher_value_decision_annulment_and_validity_timestamps.sql
```diff
@@ -0,0 +1,5 @@
+ALTER TABLE voucher_value_decision
+    ADD COLUMN annulled_at timestamp with time zone,
+    ADD COLUMN validity_updated_at timestamp with time zone;
+
+UPDATE voucher_value_decision SET annulled_at = updated WHERE status = 'ANNULLED'::voucher_value_decision_status;
```
kimifoud, line 5: I reckon nothing can be done to the decision once it is annulled..?

kimifoud (approved):

PR #2194, opened 2022-03-10
EVAKA-4162 Make holiday banner sticky by modifying page layout

zch:
The layout of the page was modified to always keep the header on top without using `position: sticky`. This is done by making the `main` element of the page scrollable, instead of scrolling the whole page including the header.

This MIGHT have an impact on the different pages in the citizen frontend. Everything I tested works, but I appreciate if you take it for a spin!


kimifoud (approved):
Tested pretty much all pages and couldn't break it 🤷 Nice!

PR #2207, opened 2022-03-14
Add more e2e tests for citizen holiday modal

akheron:

### DIFF
--- a/frontend/src/e2e-test/specs/0_citizen/citizen-holiday-reservations.spec.ts
+++ b/frontend/src/e2e-test/specs/0_citizen/citizen-holiday-reservations.spec.ts
@@ -63,6 +63,25 @@ const holidayQuestionnaireFixture = () =>
     ]
   })
 
+async function assertReservationState(
+  calendar: CitizenCalendarPage,
+  startDate: LocalDate,
+  endDate: LocalDate,
+  hasFreeAbsence: boolean
+) {
+  await calendar.waitUntilLoaded()
+
+  let today = startDate
+  while (today.isEqualOrBefore(endDate)) {
+    if (hasFreeAbsence) {
+      await calendar.assertReservations(today, [], false, true)
+    } else {
+      await calendar.assertNoReservationsOrAbsences(today)
+    }
+    today = today.addBusinessDays(1)
+  }
+}
+
 beforeEach(async () => {
   await resetDatabase()
   page = await Page.open({ mockedTime: mockedDate.toSystemTzDate() })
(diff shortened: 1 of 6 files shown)


frontend/src/e2e-test/specs/0_citizen/citizen-holiday-reservations.spec.ts
```diff
@@ -112,22 +150,65 @@ describe('Holiday periods', () => {
...
+
       await enduserLogin(page)
       await new CitizenHeader(page).selectTab('calendar')
       const calendar = new CitizenCalendarPage(page, 'desktop')
 
-      await calendar.assertNoReservationsOrAbsences(LocalDate.of(2035, 12, 26))
+      await assertFreeAbsences(false)
 
-      const holidayModal = await calendar.openHolidayModal()
+      let holidayModal = await calendar.openHolidayModal()
```
kimifoud, line 168: I don't quite get this pattern of returning page objects, because the page object do not hold any internal state.

> _Locator is resolved to the element immediately before performing an action, so a series of actions on the same locator can in fact be performed on different DOM elements. That would happen if the DOM structure between those actions has changed._

Using `let` and returning page objects leads to a false sense of immutability (or something) 😄

frontend/src/e2e-test/specs/0_citizen/citizen-holiday-reservations.spec.ts
```diff
@@ -197,12 +272,30 @@ describe('Holiday periods', () => {
         })
         .save()
 
+      const child2 = await setupAnotherChild(
+        // Not eligible because placement starts after 2022-01-01
+        LocalDate.of(2022, 3, 1),
```
kimifoud, line 277: `2022-01-01` could be extracted to a well-named constant and: `foo.addMonths(2)`

kimifoud (approved):
Nice work! 👍  I'll try to remember these tests the next time I need to setup some test boilerplate.

zch (approved):

frontend/src/e2e-test/specs/0_citizen/citizen-holiday-reservations.spec.ts
akheron: I see it as a way to "mentally track" which page/modal is visible on the "screen". If an action opens a modal or moves to another page, then it returns the page model for that.

frontend/src/e2e-test/specs/0_citizen/citizen-holiday-reservations.spec.ts
akheron: Fixed and add a few more minor clarifications

PR #2232, opened 2022-03-17
Add guardian blocklist mechanism

Gekkio:
#### Summary

If a row is in `guardian_blocklist` table, the corresponding guardian<->child relation is never created when processing VTJ data

### DIFF
--- a/service/src/integrationTest/kotlin/fi/espoo/evaka/pis/controller/PersonControllerIntegrationTest.kt
+++ b/service/src/integrationTest/kotlin/fi/espoo/evaka/pis/controller/PersonControllerIntegrationTest.kt
@@ -16,7 +16,9 @@ import fi.espoo.evaka.shared.auth.AuthenticatedUser
 import fi.espoo.evaka.shared.auth.CitizenAuthLevel
 import fi.espoo.evaka.shared.auth.UserRole
 import fi.espoo.evaka.shared.db.Database
+import fi.espoo.evaka.shared.dev.DevGuardianBlocklistEntry
 import fi.espoo.evaka.shared.dev.DevPerson
+import fi.espoo.evaka.shared.dev.insertTestGuardianBlocklistEntry
 import fi.espoo.evaka.shared.dev.insertTestPerson
 import fi.espoo.evaka.shared.domain.Forbidden
 import org.junit.jupiter.api.Test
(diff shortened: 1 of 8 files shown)


service/src/integrationTest/kotlin/fi/espoo/evaka/pis/controller/PersonControllerIntegrationTest.kt
```diff
@@ -137,6 +141,54 @@ class PersonControllerIntegrationTest : AbstractIntegrationTest() {
...
+        val dependants = controller.getPersonDependants(Database(jdbi), admin, guardianId)
+        assertEquals(3, dependants.size)
+
+        val blockedDependant = dependants.find { it.socialSecurityNumber == "070714A9126" }!!
+        db.transaction { tx ->
+            tx.createUpdate("DELETE FROM guardian WHERE child_id = :id").bind("id", blockedDependant.id).execute()
+            tx.insertTestGuardianBlocklistEntry(DevGuardianBlocklistEntry(guardianId, blockedDependant.id))
+        }
+
+        assertEquals(2, controller.getPersonDependants(Database(jdbi), admin, guardianId).size)
```
jaahon, line 165: I don't think dependants is queried again from VTJ here since `vtjGuardiansQueried` should not be null anymore. You should reset it or change the name of the test if it's intentional.

service/src/main/resources/db/migration/V216__guardian_blocklist.sql
```diff
@@ -0,0 +1,7 @@
+CREATE TABLE guardian_blocklist (
+    guardian_id uuid NOT NULL CONSTRAINT fk$guardian REFERENCES person (id) ON DELETE CASCADE,
+    child_id uuid NOT NULL CONSTRAINT fk$child REFERENCES person (id) ON DELETE CASCADE
+);
+
```
kimifoud, line 5: Is this feature intended to be used only via direct database access? Anyway, consider future proofing by adding the default `created` and `updated` columns and the trigger (maybe even evaka_user).

service/src/main/resources/db/migration/V216__guardian_blocklist.sql
```diff
@@ -0,0 +1,7 @@
...
+CREATE UNIQUE INDEX uniq$guardian_blocklist_guardian_child ON guardian_blocklist (guardian_id, child_id);
+CREATE INDEX idx$guardian_blocklist_child ON guardian_blocklist (child_id);
```
kimifoud, line 7:
```suggestion
CREATE INDEX idx$guardian_blocklist_child ON guardian_blocklist (child_id);

```

jaahon (approved):

PR #2236, opened 2022-03-17
Allow deleting a person while keeping their message account

akheron:
#### Summary

This fixes a rare case where a person has lost guardianship and would be deleted by `InactivePeopleCleanup`, but received messages prevented the deletion.

The message account can be kept because we add a reference to `evaka_user` table. This also makes it possible to remove the `active` column; If the message account doesn't "belong to anyone", then it's inactive.


service/src/main/kotlin/fi/espoo/evaka/pis/service/MergeService.kt
```diff
@@ -115,11 +115,7 @@ class MergeService(private val asyncJobRunner: AsyncJobRunner<AsyncJob>) {
                 """
                     (SELECT count(*) FROM $table WHERE $column = :id)
                 """.trimIndent()
-            }} +
-                (SELECT count(*) FROM message WHERE sender_id = (SELECT id FROM message_account WHERE person_id = :id)) +
-                (SELECT count(*) FROM message_content WHERE author_id = (SELECT id FROM message_account WHERE person_id = :id)) +
-                (SELECT count(*) FROM message_recipients WHERE recipient_id = (SELECT id FROM message_account WHERE person_id = :id)) +
-                (SELECT count(*) FROM message_draft WHERE account_id = (SELECT id FROM message_account WHERE person_id = :id)) AS count;
+            }}
```
akheron, line 118: These are OK. If they exist, they're just left behind when the user is deleted, as `message_account.person_id` is set to NULL when a person is deleted.

service/src/main/resources/db/migration/V216__message_account_evaka_user_id.sql
```diff
@@ -0,0 +1,15 @@
+ALTER TABLE message_account ADD COLUMN evaka_user_id uuid UNIQUE REFERENCES evaka_user (id);
+UPDATE message_account SET evaka_user_id = coalesce(person_id, employee_id);
+UPDATE message_account SET employee_id = NULL WHERE employee_id IS NOT NULL AND NOT active;
+ALTER TABLE message_account DROP COLUMN active;
+
+ALTER TABLE message_account
+    DROP CONSTRAINT message_account_person_id_fkey,
+    ADD CONSTRAINT message_account_person_id_fkey FOREIGN KEY (person_id) REFERENCES person (id) ON DELETE SET NULL,
```
akheron, line 8: This was `ON DELETE CASCADE`, and that was the culprit of the `InactivePeopleCleaup` failure.

service/src/main/resources/db/migration/V218__message_account_evaka_user_id.sql
```diff
@@ -0,0 +1,15 @@
...
+
+ALTER TABLE message_account
+    DROP CONSTRAINT message_account_person_id_fkey,
+    ADD CONSTRAINT message_account_person_id_fkey FOREIGN KEY (person_id) REFERENCES person (id) ON DELETE SET NULL,
+    DROP CONSTRAINT message_account_created_for_fk,
+    ADD CONSTRAINT message_account_owner_fk CHECK (
+        num_nonnulls(daycare_group_id, employee_id, person_id) IN (0, 1) AND
+        (daycare_group_id IS NULL OR evaka_user_id IS NULL) AND
+        (employee_id IS NULL OR evaka_user_id IS NOT NULL) AND
+        (person_id IS NULL OR evaka_user_id IS NOT NULL)
```
kimifoud, line 14: This could be simplified as discussed. Otherwise LGTM!

kimifoud (changes requested):

kimifoud (approved):

service/src/main/resources/db/migration/V219__message_account_evaka_user_id.sql
```diff
@@ -0,0 +1,16 @@
+ALTER TABLE message_account ADD COLUMN evaka_user_id uuid UNIQUE REFERENCES evaka_user (id);
+UPDATE message_account SET evaka_user_id = coalesce(person_id, employee_id);
+UPDATE message_account SET employee_id = NULL WHERE employee_id IS NOT NULL AND NOT active;
+ALTER TABLE message_account DROP COLUMN active;
+
+ALTER TABLE message_account
+    DROP CONSTRAINT message_account_person_id_fkey,
+    ADD CONSTRAINT message_account_person_id_fkey FOREIGN KEY (person_id) REFERENCES person (id) ON DELETE SET NULL,
+    DROP CONSTRAINT message_account_created_for_fk,
+    ADD CONSTRAINT message_account_owner_fk CHECK (
```
kimifoud, line 10: It just crossed my mind that foreign key naming convention is `fk$person`. I don't think we have a convention for check constraints (just one occurrence of `chk$valid_role`).

kimifoud (approved):

PR #2245, opened 2022-03-23
EVAKA-4391 invoice corrections editor

jaahon:
#### Summary

New invoice corrections section in person page that is hidden behind a feature flag for now. The corrections will be used in invoices later.


service/src/main/resources/db/migration/V218__invoice_correction.sql
```diff
@@ -0,0 +1,16 @@
+CREATE TABLE invoice_correction (
+    id uuid PRIMARY KEY DEFAULT ext.uuid_generate_v1mc(),
+    created timestamp with time zone DEFAULT now() NOT NULL,
+    updated timestamp with time zone DEFAULT now() NOT NULL,
+    head_of_family_id uuid NOT NULL REFERENCES person(id),
```
reynders, line 5: Should there be an index on head_of_family?

reynders (approved):

service/src/main/resources/db/migration/V218__invoice_correction.sql
jaahon: Yeah, definitely 👍

PR #2250, opened 2022-03-23
EVAKA-4455 New realtime attendance graph

kimifoud:


zch:
This is a good enough start, but I feel the UX needs to be improved:
- The tooltip contents is not as clear as it could be. We should make it a "custom tooltip" and render the contents ourselves as in the designs.
- The tooltip is way too jumpy
- The circles at the data points make the chart look busy

Based on the two latter points I think we should build the data sets such that there's a data point for each minute of the day and set the data point radius to 0. This would make the chart easier to read and the tooltip wouldn't jump around.

PR #2252, opened 2022-03-25
EVAKA-4455 Implement realtime staff attendance editing

zch:
Can only edit existing attendances, cannot create new ones. Adding of new ones will be implemented when adding the employee table to the group views.

PR #2255, opened 2022-03-25
EVAKA-4455 Fix rendering of employee attendance rows

zch:
A bug caused rendering as many empty attendance records as there were
records for the entire week in the data set for one employee. This makes
sure to never render empty elements for attendance records that are for
different days than the current cell.

### DIFF
--- a/frontend/src/employee-frontend/components/unit/unit-reservations/StaffAttendanceTable.tsx
+++ b/frontend/src/employee-frontend/components/unit/unit-reservations/StaffAttendanceTable.tsx
@@ -5,7 +5,7 @@
 import classNames from 'classnames'
 import { isSameDay } from 'date-fns'
 import { groupBy, sortBy } from 'lodash'
-import React, { useMemo } from 'react'
+import React, { useMemo, useCallback } from 'react'
 import styled from 'styled-components'
 
 import EllipsisMenu from 'employee-frontend/components/common/EllipsisMenu'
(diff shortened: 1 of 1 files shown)


frontend/src/employee-frontend/components/unit/unit-reservations/StaffAttendanceTable.tsx
```diff
@@ -127,19 +135,14 @@ const AttendanceRow = React.memo(function AttendanceRow({
           partialRow={false}
           rowIndex={rowIndex}
         >
-          {attendances?.length > 0 ? (
-            attendances.map((attendance) => (
+          {attendancesForDay(day.date.toSystemTzDate())?.length > 0 ? (
+            attendancesForDay(day.date.toSystemTzDate()).map((attendance) => (
```
akheron, line 139: Is the `.length > 0` check needed? Now this filters and sorts the attendances twice just to see how many there are, and then again to map over them.

It might be even better to precompute the `(operationalDay, attendances)` data once and then iterate over it in the `.map()` call.

zch: Due to the `useCallback` it should do the actual filtering only once and use the memoized result on the second invocation. When merging this into #2252 it will likely be exactly what you suggest, i.e. precomputing and then iterating over the result.

zch: Maybe I'll just do this fix in #2252 and close this?

zch:
including a slightly cleaner fix in #2252

PR #2258, opened 2022-03-28
Add a feature flag for showing PLANNED_ABSENCE for citizens with children in shift care

akheron:


frontend/src/citizen-frontend/calendar/AbsenceModal.tsx
```diff
@@ -81,7 +82,10 @@ export default React.memo(function AbsenceModal({
   const [request, errors] = validateForm(form)
 
   const childrenInShiftCare = useMemo(
```
jaahon, line 84: I think renaming this variable as well would be a good idea.

PR #2270, opened 2022-03-31
Optimize the query for getting threads/messages by account

akheron:
#### Summary

Try to clean up and speed up the 3rd most time consuming SQL query in eVaka. The query itself doesn't take _too long_ to perform (avg ~330 ms in Espoo production), but it's used a lot.

After the changes in this PR, the query is now simpler and more readable (at least to me), but it seems to be even slower than it was.

Let's have a mob review/discussion about the messaging data model at some point.


akheron:
[EXPLAIN ANALYZE from production](https://github.com/espoon-voltti/evaka/files/8512017/messages.txt) of the query after this PR

akheron:
[EXPLAIN ANALYZE from production](https://github.com/espoon-voltti/evaka/files/8562666/old.txt) of the query before this PR

PR #2274, opened 2022-04-01
Try to fix message textarea min height bug on android

reynders:
#### Summary
On Samsum A32 the mobile message editor text area does not respect min-height. Try to fix that according to the internet. Works still on iphone after the change:
IMG_7764


jaahon (commented):
It's the citizen message editor that has the bug, you've now modified employee message editor. I tested it and adding `min-height: 100px;` to the `<textarea>` there should be enough.

PR #2278, opened 2022-04-01
EVAKA-4391 apply correction on invoices

jaahon:
#### Summary

Apply corrections filled in person page on invoices. Currently the application logic is this:
* If the whole correction is an increase apply it completely always
* If the correction is a refund that would make the invoice total negative, either apply a portion of the `amount` if the `unitPrice` is less than the invoice total or apply a portion of the `unitPrice` that would keep the invoice total at least 0.

Also, drop invoice row child date of birth column (wasn't actually used for anything).


service/src/integrationTest/kotlin/fi/espoo/evaka/invoicing/service/InvoiceGeneratorIntegrationTest.kt
```diff
@@ -3545,6 +3548,73 @@ class InvoiceGeneratorIntegrationTest : PureJdbiTest() {
...
+                    unitId = testDaycare.id,
+                    product = productProvider.mapToProduct(PlacementType.DAYCARE),
+                    description = "",
+                    note = ""
+                )
+            )
+        }
+
+        db.transaction { generator.createAndStoreAllDraftInvoices(it, period) }
+        db.transaction { generator.createAndStoreAllDraftInvoices(it, period) }
```
reynders, line 3592: Why is this doubled?

reynders (approved):

service/src/integrationTest/kotlin/fi/espoo/evaka/invoicing/service/InvoiceGeneratorIntegrationTest.kt
jaahon: The test name was supposed to explain it, but I guess asserting twice would be a good idea 😄

PR #2292, opened 2022-04-05
EVAKA-4455 Add staff attendance table to group views

zch:
image

### DIFF
--- a/frontend/src/employee-frontend/components/unit/unit-reservations/UnitAttendanceReservationsView.tsx
+++ b/frontend/src/employee-frontend/components/unit/unit-reservations/UnitAttendanceReservationsView.tsx
@@ -2,12 +2,13 @@
 //
 // SPDX-License-Identifier: LGPL-2.1-or-later
 
-import React, { useMemo, useState, useCallback } from 'react'
+import React, { useMemo, useState, useCallback, useContext } from 'react'
 import styled from 'styled-components'
 
 import { renderResult } from 'employee-frontend/components/async-rendering'
 import LabelValueList from 'employee-frontend/components/common/LabelValueList'
-import { combine, Success } from 'lib-common/api'
+import { UnitContext } from 'employee-frontend/state/unit'
+import { combine } from 'lib-common/api'
 import { Child } from 'lib-common/api-types/reservations'
 import FiniteDateRange from 'lib-common/finite-date-range'
 import {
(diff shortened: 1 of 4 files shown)


frontend/src/employee-frontend/components/unit/unit-reservations/UnitAttendanceReservationsView.tsx
```diff
@@ -161,17 +161,32 @@ export default React.memo(function UnitAttendanceReservationsView({
...
-              reloadReservations={reloadChildReservations}
-            />
+            <>
+              <StaffAttendanceTable
+                operationalDays={childData.operationalDays}
+                staffAttendances={staffData.staff.filter((s) =>
+                  daycareAclRows
+                    .find((r) => r.employee.id === s.employeeId)
+                    ?.groupIds.includes(groupId)
+                )}
```
akheron, line 171: Perhaps not a thing to fix in this PR, but these pretty complex `filter`/`find`/`includes` combos could perhaps be moved to the component top level by employing `combine` and `Result.map`. That way they would also only run once when the `Result` changes instead of on every render.

frontend/src/employee-frontend/components/unit/unit-reservations/attendance-elements.tsx
```diff
@@ -46,7 +46,10 @@ const CustomTh = styled(Th)<{ shrink?: boolean }>`
     `}
 `
 
+// Use min- and max-width to ensure that columns in two tables in the same layout are aligned
 const DateTh = styled(CustomTh)<{ faded: boolean }>`
+  min-width: 150px;
+  max-width: 150px;
```
akheron, line 52: Is there a difference in `min-width` + `max-width` as opposed to just `width`? 🤔

akheron (approved):

akheron (changes requested):
Umm actually, should this be behind a feature flag? Or is it already? 😄

frontend/src/employee-frontend/components/unit/unit-reservations/attendance-elements.tsx
zch: `width` alone will not force the width in this case (I think it's due to how `table`s do their layout, but I might be slightly wrong).

zch:
> Umm actually, should this be behind a feature flag? Or is it already? 😄

Excellent catch 😅

akheron (approved):

PR #2294, opened 2022-04-06
EVAKA-4391 invoice correction notes and deletes

jaahon:
#### Summary

- Make uninvoiced corrections deletable
- Add support for adding notes to invoice corrections
- Hide invoice corrections that have been applied completely


service/src/main/kotlin/fi/espoo/evaka/invoicing/controller/InvoiceCorrectionsController.kt
```diff
@@ -39,8 +46,20 @@ class InvoiceCorrectionsController(private val accessControl: AccessControl) {
             dbc.read { tx ->
                 tx.createQuery(
                     """
-SELECT id, head_of_family_id, child_id, unit_id, product, period, amount, unit_price, description, note
-FROM invoice_correction WHERE head_of_family_id = :personId
+SELECT c.id, c.head_of_family_id, c.child_id, c.unit_id, c.product, c.period, c.amount, c.unit_price, c.description, c.note,
+    i.id AS invoice_id,
+    i.status AS invoice_status
+FROM invoice_correction c
+LEFT JOIN invoice_row r ON c.id = r.correction_id
```
Gekkio, line 53: This query has `invoice_row` with name `r` twice (one normal join, another inside lateral), but one of them doesn't seem to be used? :thinking:

jaahon: Yeah I forgot to remove this part, the query was originally a bit different 😅

Gekkio (approved):

PR #2296, opened 2022-04-06
Fix various placement plan occupancy calculation bugs

Gekkio:

### DIFF
--- a/service/src/integrationTest/kotlin/fi/espoo/evaka/occupancy/OccupancyControllerIntegrationTest.kt
+++ b/service/src/integrationTest/kotlin/fi/espoo/evaka/occupancy/OccupancyControllerIntegrationTest.kt
@@ -231,14 +231,16 @@ class OccupancyControllerIntegrationTest : FullApplicationTest() {
             testDaycare.id,
             applicationId,
             period = FiniteDateRange(LocalDate.of(2021, 3, 1), LocalDate.of(2021, 6, 4)),
-            preschoolDaycarePeriod = FiniteDateRange(LocalDate.of(2021, 3, 1), LocalDate.of(2021, 7, 31))
+            preschoolDaycarePeriod = FiniteDateRange(LocalDate.of(2021, 6, 1), LocalDate.of(2021, 7, 31))
         )
 
         assertEquals(
             OccupancyResponseSpeculated(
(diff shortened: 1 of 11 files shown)


service/src/integrationTest/kotlin/fi/espoo/evaka/occupancy/OccupancyControllerIntegrationTest.kt
```diff
@@ -231,14 +231,16 @@ class OccupancyControllerIntegrationTest : FullApplicationTest() {
...
+            preschoolDaycarePeriod = FiniteDateRange(LocalDate.of(2021, 6, 1), LocalDate.of(2021, 7, 31))
         )
 
         assertEquals(
             OccupancyResponseSpeculated(
                 max3Months = OccupancyValues(sum = 1.75, headcount = 1, caretakers = 1.0, percentage = 25.0),
                 max6Months = OccupancyValues(sum = 2.75, headcount = 2, caretakers = 1.0, percentage = 39.3),
-                max3MonthsSpeculated = OccupancyValues(sum = 2.75, headcount = 2, caretakers = 1.0, percentage = 39.3),
+                // preschool only
+                max3MonthsSpeculated = OccupancyValues(sum = 2.25, headcount = 2, caretakers = 1.0, percentage = 32.1),
```
jaahon, line 242: Why has the percentage changed? 🤔

jaahon (commented):
Apart from that test thing I have no comments. Looks good!

service/src/integrationTest/kotlin/fi/espoo/evaka/occupancy/OccupancyControllerIntegrationTest.kt
Gekkio: The 3-month period now includes only pure preschool (vs preschool+daycare previously), but the amount of caretakers is the same.

-> lower sum (this child goes 1.0 -> 0.5), lower percentage

jaahon: I stared at this for minutes and still didn't notice that `2.75` was changed to `2.75`. Nevermind 😂

jaahon (approved):

PR #2297, opened 2022-04-06
EVAKA-4487 extend group staff acl

jaahon:
#### Summary

- Add group acl drop downs for unit supervisors and special education teachers
- Include all group acl employees in realtime staff attendances


akheron:
Should the UI part be hidden if realtime staff attendances is not in use?

I also wonder how this is supposed to work if/when an employee can in the future have multiple `daycare_acl` roles in a single unit.

jaahon:
> Should the UI part be hidden if realtime staff attendances is not in use?

Group acl affects other things as well, so it should be enabled for everyone.

akheron (approved):

PR #2300, opened 2022-04-06
Debounce application filters

reynders:
#### Summary
Debounce application search filters:
- Area


jaahon (commented):
There are a couple of filters that are still not in the same debounce: `preschoolType`, `allStatuses`, `distinctions`. Just letting you know in case it wasn't intentional.

jaahon (approved):

PR #2301, opened 2022-04-06
EVAKA-HOTFIX use relative age in age calculation

jaahon:
#### Summary

- Always use relative age when determining child age
- Use constant variable for family unit placement coefficient value
- Use service need occupancy coefficients with realtime occupancy


hruoho (approved):
Looks good to me 👍

PR #2304, opened 2022-04-07
EVAKA-4455 Allow normal staff to see the staff attendance in groups

zch:
The staff attendance table includes the attendance times of all employees assigned to a group.

### DIFF
--- a/service/src/main/kotlin/fi/espoo/evaka/attendance/StaffAttendanceQueries.kt
+++ b/service/src/main/kotlin/fi/espoo/evaka/attendance/StaffAttendanceQueries.kt
@@ -241,3 +241,14 @@ fun Database.Read.getExternalStaffAttendancesByDateRange(unitId: DaycareId, rang
     .bind("end", range.end)
     .mapTo<ExternalAttendance>()
     .list()
+
+fun Database.Read.getGroupsForEmployee(employeeId: EmployeeId): List<GroupId> = createQuery(
+    """
+    SELECT daycare_group_id
+    FROM daycare_group_acl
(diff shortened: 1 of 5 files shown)


jaahon (approved):

service/src/main/kotlin/fi/espoo/evaka/attendance/StaffAttendanceQueries.kt
```diff
@@ -241,3 +241,14 @@ fun Database.Read.getExternalStaffAttendancesByDateRange(unitId: DaycareId, rang
...
     .mapTo<ExternalAttendance>()
     .list()
+
+fun Database.Read.getGroupsForEmployee(employeeId: EmployeeId): List<GroupId> = createQuery(
+    """
+    SELECT daycare_group_id
+    FROM daycare_group_acl
+    WHERE employee_id = :employeeId
+    """.trimIndent()
+)
```
akheron, line 251: Would be better to read group memberships for all employees in a single call, something like this:

```
private data class Foo(val employeeId: EmployeeId, val groupIds: List<GroupId>)

fun Database.Read.gepGroupsForEmployees(employeeIds: Set<EmployeeId>): Map<EmployeeId, List<GroupId>> = createQuery("""
    SELECT employee_id, json_array_agg(daycare_group_id) AS group_ids
    FROM daycare_group_acl
    WHERE employee_id = ANY(:employeeIds)
""").bind("employeeIds", employeeIds.toTypedArray()).mapTo<Foo>().associateJotainJotain()
```

PR #2311, opened 2022-04-08
EVAKA-4455 Allow marking new staff attendances in the staff attendance table on group views

zch:
Also includes a commit for optimizing fetching of groups for employees.

Might be easier to review by commit than all at once.

### DIFF
--- a/frontend/src/employee-frontend/components/unit/unit-reservations/StaffAttendanceTable.tsx
+++ b/frontend/src/employee-frontend/components/unit/unit-reservations/StaffAttendanceTable.tsx
@@ -16,8 +16,8 @@ import {
   Attendance,
   EmployeeAttendance,
   ExternalAttendance,
-  UpdateStaffAttendanceRequest,
-  UpdateExternalAttendanceRequest
+  UpsertStaffAttendanceRequest,
+  UpsertExternalAttendanceRequest
 } from 'lib-common/generated/api-types/attendance'
 import { TimeRange } from 'lib-common/generated/api-types/reservations'
 import LocalDate from 'lib-common/local-date'
@@ -51,9 +51,9 @@ interface Props {
   operationalDays: OperationalDay[]
   staffAttendances: EmployeeAttendance[]
   extraAttendances: ExternalAttendance[]
-  saveAttendance: (body: UpdateStaffAttendanceRequest) => Promise<Result<void>>
+  saveAttendance: (body: UpsertStaffAttendanceRequest) => Promise<Result<void>>
   saveExternalAttendance: (
-    body: UpdateExternalAttendanceRequest
+    body: UpsertExternalAttendanceRequest
   ) => Promise<Result<void>>
 }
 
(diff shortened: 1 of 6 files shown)


frontend/src/employee-frontend/components/unit/unit-reservations/StaffAttendanceTable.tsx
```diff
@@ -306,8 +315,41 @@ const AttendanceRow = React.memo(function AttendanceRow({
       <StyledTd partialRow={false} rowIndex={rowIndex} rowSpan={1}>
         {editing ? (
           <EditStateIndicator
-            status={saveRequestStatus}
-            stopEditing={() => setEditing(false)}
+            status={Success.of()}
+            stopEditing={() => {
+              setEditing(false)
+              const savePromises = values.flatMap(({ date, timeRanges }) =>
```
zch, line 321: It might make sense to refactor the endpoint to accept a list of attendances to upsert, but I'd rather do that in a separate PR I think.

service/src/main/kotlin/fi/espoo/evaka/attendance/StaffAttendanceQueries.kt
```diff
@@ -277,13 +278,19 @@ fun Database.Read.getExternalStaffAttendancesByDateRange(unitId: DaycareId, rang
...
-fun Database.Read.getGroupsForEmployee(employeeId: EmployeeId): List<GroupId> = createQuery(
+private data class EmployeeGroups(
+    val employeeId: EmployeeId,
+    @Json
+    val groupIds: List<GroupId>
+)
+fun Database.Read.getGroupsForEmployees(employeeIds: Set<EmployeeId>): Map<EmployeeId, List<GroupId>> = createQuery(
     """
-    SELECT daycare_group_id
+    SELECT employee_id, json_agg(daycare_group_id) AS group_ids
```
akheron, line 288: Contrary to what I previously said `array_agg(...)` would probably be more readable here. Then you can also remove the `@Json` annotation above.

frontend/src/employee-frontend/components/unit/unit-reservations/StaffAttendanceTable.tsx
```diff
@@ -328,8 +315,41 @@ const AttendanceRow = React.memo(function AttendanceRow({
...
```
akheron, line 321: This could be just `values.map(...)` because you're returning a promise from the function in every case. Or you could replace `return Promise.resolve()` in the end with `return []` to take full advantage of the flatmap.

akheron (approved):

frontend/src/employee-frontend/components/unit/unit-reservations/StaffAttendanceTable.tsx
zch: Actually the map on the following row returns an array of promises, which most often contains only one promise, but may contain several in case the employee has several attendance markings for the same date.

PR #2315, opened 2022-04-08
Add type column to the application table

Gekkio:


kimifoud (approved):
Looks good to me! Document type should probably be cleaned up at some point.

PR #2316, opened 2022-04-08
Allow vtj refresh to service workers

reynders:


frontend/src/employee-frontend/components/person-shared/PersonDetails.tsx
```diff
@@ -201,12 +206,16 @@ export default React.memo(function PersonDetails({
...
+              {vtjUpdateOngoing ? (
+                <SpinnerSegment size="xxs" />
+              ) : (
+                <InlineButton
+                  icon={faSync}
+                  onClick={onVtjUpdate}
+                  data-qa="update-from-vtj-button"
+                  text={i18n.personProfile.updateFromVtj}
+                />
+              )}
```
akheron, line 218: `AsyncButton` could be a better option here

PR #2328, opened 2022-04-11
Holiday questionnaire data model changes

akheron:
#### Summary

There are 3 use cases for holiday periods and questionnaires:

1. Add background color to the citizen calendar -> Create a holiday period WITHOUT a deadline.

2. Ask the guardians to mark absences to the calendar -> Create a holiday period WITH a deadline. Also disallows changing reservations/absences after the deadline.

3. Choose an unpaid vacation period for children -> Create a holiday questionnaire.

This commit allows 2 and 3 to coexists more freely be removing the linking of questionnaires to holiday periods. That is, questionnaires can now exists without a holiday period, which removes the need to add multiple overlapping holiday periods.


kimifoud (approved):
🎉

PR #2337, opened 2022-04-12
Always redirect to login page on session expiry

Gekkio:
#### Summary

Previously this was only done if the current page required strong authentication. If it required weak authentication, the browser was redirected directly to the weak authentication login, even if you previously had a strongly authenticated session.


kimifoud (approved):
Just a thought, we could improve to flow later by adding redirect params to `/login` endpoint too.

PR #2340, opened 2022-04-13
EVAKA citizen messages ux improvements

jaahon:
#### Summary

- Make message thread list items buttons for better accessibility
- Save citizen selected message thread in url for better mobile navigation experience
- Add direct link to message thread in notification email

### DIFF
--- a/service/src/main/kotlin/fi/espoo/evaka/messaging/MessageNotificationEmailService.kt
+++ b/service/src/main/kotlin/fi/espoo/evaka/messaging/MessageNotificationEmailService.kt
@@ -9,6 +9,7 @@ import fi.espoo.evaka.EvakaEnv
 import fi.espoo.evaka.daycare.domain.Language
 import fi.espoo.evaka.emailclient.IEmailClient
 import fi.espoo.evaka.shared.MessageId
+import fi.espoo.evaka.shared.MessageThreadId
 import fi.espoo.evaka.shared.async.AsyncJob
 import fi.espoo.evaka.shared.async.AsyncJobRunner
 import fi.espoo.evaka.shared.db.Database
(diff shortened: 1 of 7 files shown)


service/src/main/kotlin/fi/espoo/evaka/messaging/MessageNotificationEmailService.kt
```diff
@@ -42,11 +43,13 @@ class MessageNotificationEmailService(
...
+                m.thread_id,
                 mr.id as message_recipient_id,
                 p.id as person_id,
                 p.email as person_email,
                 coalesce(lower(p.language), 'fi') as language
             FROM message_recipients mr
+            JOIN message m ON m.id = mr.message_id
             JOIN message_account ma ON ma.id = mr.recipient_id 
             JOIN person p ON p.id = ma.person_id
             WHERE mr.message_id = :messageId
```
kimifoud, line 55: Now that the query needs stuff from `message` it would (maybe?) make more sense to `SELECT .. FROM MESSAGE .. WHERE m.id = :messageId`

service/src/main/kotlin/fi/espoo/evaka/messaging/MessageNotificationEmailService.kt
```diff
@@ -105,16 +109,16 @@ class MessageNotificationEmailService(
         return "Uusi viesti eVakassa / Nytt meddelande i eVaka / New message in eVaka$postfix"
     }
 
-    private fun getCitizenMessagesUrl(lang: Language): String {
+    private fun getCitizenMessageUrl(lang: Language, threadId: MessageThreadId?): String {
```
kimifoud, line 112: `threadId` should never be null

jaahon: What do you mean? We have jobs without a saved `threadId`.

kimifoud: Ah, of course 🤦 Maybe we should have versioned job definitions.

kimifoud (approved):

PR #2342, opened 2022-04-13
Change footer font size and add footer to all citizen pages

zch:


Gekkio (commented):
On the messages page, this change causes a scrollbar to always appear even if there's no need for it (Firefox 99). I think it scrolls just 1-2 pixels

Screenshot from 2022-04-20 11-26-53

zch:
> On the messages page, this change causes a scrollbar to always appear even if there's no need for it (Firefox 99). I think it scrolls just 1-2 pixels

Fixed by explicitly setting the height of the footer on desktop. Firefox calculated the height to be 0.5px too high 😅

Gekkio (approved):

PR #2354, opened 2022-04-14
Add separate voucher value for under three year olds

patari:
#### Summary

Our base value is 826 and age coefficient 1.51. 826 * 1.51 = 1247.26 but it should be 1247.


service/src/main/resources/db/migration/V229__voucher_value_base_value_age_under_three.sql
```diff
@@ -0,0 +1,7 @@
+ALTER TABLE voucher_value ADD base_value_age_under_three INTEGER;
+UPDATE voucher_value SET base_value_age_under_three = base_value * age_under_three_coefficient;
+ALTER TABLE voucher_value ALTER COLUMN base_value_age_under_three SET NOT NULL;
+ALTER TABLE voucher_value DROP COLUMN age_under_three_coefficient;
+
+UPDATE voucher_value_decision SET base_value = base_value * age_coefficient;
```
Gekkio, line 6: This probably needs `floor(...)`, because PostgreSQL rounds instead of truncating when converting from numeric to integer:

```
evaka_local=# SELECT (1.4::numeric(4, 2))::integer, (1.5::numeric(4, 2))::integer;
 int4 | int4 
------+------
    1 |    2
```

This works differently than the Kotlin code with `BigDecimal.toInt()`, which truncates the fractional part.

akheron: In the espoo case this is ok, because 87000 * 1.55 = 134850 and no rounding/truncating happens.

PR #2373, opened 2022-04-21
Save and use generated invoice row order

Gekkio:
#### Summary

- save row indexes and use them to sort invoice rows in queries
- row index is mostly a transparent low-level detail, and is not exposed in the `InvoiceRow` / `InvoiceRowDetailed` types
- gaps are allowed and are created if an invoice correction is deleted. This is not a problem, because the sort order is retained, and we never insert rows without first deleting all old ones
- currently the order comes in practice from the invoice_id index in `invoice_row`. In the migration, updating the entire `invoice_row` table in a single SQL statement didn't always result in the correct order, so loop each invoice individually and disable seq scan to try to use this index. This seemed to give correct results, at least based on tests in staging.
- add random shuffling to result lists in various tests to expose invalid assumptions about order
- refactor invoice correction tests to explicitly use months instead of dateranges
- fix broken invoice correction test, which was generating multiple invoices for the same month

### DIFF
--- a/service/src/integrationTest/kotlin/fi/espoo/evaka/invoicing/VoucherValueDecisionIntegrationTest.kt
+++ b/service/src/integrationTest/kotlin/fi/espoo/evaka/invoicing/VoucherValueDecisionIntegrationTest.kt
@@ -546,7 +546,7 @@ class VoucherValueDecisionIntegrationTest : FullApplicationTest(resetDbBeforeEac
             it.createQuery("SELECT * FROM voucher_value_decision")
                 .mapTo<VoucherValueDecision>()
                 .toList()
-        }
+        }.shuffled() // randomize order to expose assumptions
     }
 
     private fun endDecisions(now: LocalDate) {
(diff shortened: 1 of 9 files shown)


service/src/main/kotlin/fi/espoo/evaka/invoicing/domain/FeeDecisions.kt
```diff
@@ -73,6 +73,8 @@ data class FeeDecision(
     override fun isEmpty(): Boolean = this.children.isEmpty()
     override fun annul() = this.copy(status = FeeDecisionStatus.ANNULLED)
     override fun withChildren(children: List<FeeDecisionChild>) = this.copy(children = children)
+
+    fun containsChild(id: ChildId): Boolean = this.children.any { it.child.id == id }
```
akheron, line 77: Is this used anywhere?

Gekkio: Not anymore :sweat_smile: It was part of some old refactoring that I decided to revert

akheron (approved):

PR #2379, opened 2022-04-22
Make the message editing window use the available space

zch:
The message window "avoided" the massive footer, which caused it to be quite small normally. This fix makes it use the available space much more efficiently.


akheron (approved):
Didn't test, but the changes seem ok :)

PR #2386, opened 2022-04-26
Add error messages when JS is disabled or app doesn't load

Gekkio:


frontend/src/body.html
```diff
@@ -0,0 +1,52 @@
...
+    function hasLoaded() {
+      if (app) {
+        var children = app.childNodes
+        for (var i = 0; i < children.length; i++) {
+          var child = children[i]
+          if (child instanceof HTMLElement && child.nodeName !== 'NOSCRIPT') {
+            return true
+          }
+        }
+      }
```
akheron, line 33: Would this become simpler if the `<noscript>` part was outside the app div? I think it could be outside, because JavaScript is not run anyways when its content is shown.

Gekkio: This would conflict with `height: 100%` style defined for the app div, so we'd need some CSS tricks to compensate

frontend/src/citizen-frontend/index.tsx
```diff
@@ -28,4 +28,13 @@ Sentry.init({
...
+if (!window.evaka) {
+  window.evaka = {}
+}
+
 ReactDOM.render(<App />, document.getElementById('app'))
+
+// Let the HTML template inline script know we have loaded successfully
+if (!window.evaka) {
+  window.evaka = {}
+}
```
akheron, line 40: `window.evaka` is now added twice 🤔

akheron (approved):

PR #2396, opened 2022-04-27
Do not fail whole fee decision batch if there are manual deps

reynders:
- When creating fee decisions from drafts, do not fail the whole batch if some have related fee decisions waiting for sending. Instead skip those.
- From the UI point of view this means that the skipped drafts will remain in the search results after the create fee decisions operation is done

### DIFF
--- a/service/src/main/kotlin/fi/espoo/evaka/invoicing/service/FeeDecisionService.kt
+++ b/service/src/main/kotlin/fi/espoo/evaka/invoicing/service/FeeDecisionService.kt
@@ -85,7 +85,7 @@ class FeeDecisionService(
             )
         }
 
-        val (conflicts, waitingForManualSending) = decisions
+        val (conflicts, waitingForSending) = decisions
             .flatMap {
                 tx.lockFeeDecisionsForHeadOfFamily(it.headOfFamilyId)
                 tx.findFeeDecisionsForHeadOfFamily(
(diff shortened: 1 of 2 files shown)


service/src/main/kotlin/fi/espoo/evaka/invoicing/service/FeeDecisionService.kt
```diff
@@ -96,18 +96,24 @@ class FeeDecisionService(
...
 
-        if (conflicts.any { it.status == WAITING_FOR_MANUAL_SENDING }) throw Conflict(
-            "Some heads of family have overlapping fee decisions waiting for manual sending",
-            "WAITING_FOR_MANUAL_SENDING"
-        )
+        if (waitingForManualSending.isNotEmpty()) {
+            logger.info("Warning: when creating fee decisions, skipped ${waitingForManualSending.size} fee decisions because head of family had overlapping fee decisions waiting for manual sending")
+        }
 
         if (conflicts.any { it.status == WAITING_FOR_SENDING }) error("Some heads of family have overlapping fee decisions still waiting for sending")
```
jaahon, line 105: I think it would be intuitive to handle these similarly to the `WAITING_FOR_MANUAL_SENDING` decisions and just filter these decisions out.

service/src/main/kotlin/fi/espoo/evaka/invoicing/service/FeeDecisionService.kt
```diff
@@ -96,18 +96,24 @@ class FeeDecisionService(
...
 
-        val updatedConflicts = updateEndDatesOrAnnulConflictingDecisions(decisions, conflicts)
+        val remainingDecisions = decisions.filter { fd ->
+            waitingForManualSending.none { wfd ->
+                wfd.headOfFamilyId == fd.headOfFamilyId
+            }
+        }
+
+        val updatedConflicts = updateEndDatesOrAnnulConflictingDecisions(remainingDecisions, conflicts)
```
jaahon, line 113: I'm not 100% sure, but I think the `conflicts` should be filtered similarly. So just skip heads of families that have decisions waiting for sending.

jaahon (approved):

PR #2405, opened 2022-04-28
EVAKA-4406 Add utilization to info in units list

zch:
- Refactor the unit list view in the employee mobile frontend to look like the designs
- Add the current unit utilization to the info shown

### DIFF
--- a/service/src/main/kotlin/fi/espoo/evaka/attendance/MobileUnitController.kt
+++ b/service/src/main/kotlin/fi/espoo/evaka/attendance/MobileUnitController.kt
@@ -5,6 +5,7 @@
 package fi.espoo.evaka.attendance
 
 import fi.espoo.evaka.Audit
+import fi.espoo.evaka.occupancy.familyUnitPlacementCoefficient
 import fi.espoo.evaka.shared.DaycareId
 import fi.espoo.evaka.shared.EmployeeId
 import fi.espoo.evaka.shared.GroupId
@@ -149,7 +150,8 @@ data class UnitStats(
     val presentChildren: Int,
     val totalChildren: Int,
     val presentStaff: Int,
-    val totalStaff: Int
+    val totalStaff: Int,
+    val utilization: Double
 )
 
 fun Database.Read.fetchUnitStats(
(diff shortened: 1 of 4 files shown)


service/src/main/kotlin/fi/espoo/evaka/attendance/MobileUnitController.kt
```diff
@@ -180,20 +196,20 @@ WITH present_children AS (
     ) p
     GROUP BY p.unit_id
 ), present_staff AS (
-    SELECT g.daycare_id AS unit_id, sum(sa.count) AS count
+    SELECT g.daycare_id AS unit_id, sum(sa.capacity) AS capacity, sum(sa.count) AS count
     FROM daycare_group g
     JOIN (
-        SELECT sa.group_id, sa.count + sa.count_other AS count FROM staff_attendance sa
+        SELECT sa.group_id, 7 as capacity, sa.count + sa.count_other AS count FROM staff_attendance sa
```
jaahon, line 202: Shouldn't `capacity` be something like `7 * (sa.count + sa.count_other)`?

jaahon (commented):
Apart from the staff capacity thing looks good to me!

service/src/main/kotlin/fi/espoo/evaka/attendance/MobileUnitController.kt
zch: Great catch, thanks!

While testing I decided to add handling to return Infinity utilization if there are kids but no staff (should not happen, but might if someone forgets to add the staff).

service/src/main/kotlin/fi/espoo/evaka/attendance/MobileUnitController.kt
```diff
@@ -210,7 +226,11 @@ SELECT
     coalesce(pc.count, 0) AS present_children,
     coalesce(tc.count, 0) AS total_children,
     coalesce(ps.count, 0) AS present_staff,
-    coalesce(ts.count, 0) AS total_staff
+    coalesce(ts.count, 0) AS total_staff,
+    CASE
+        WHEN coalesce(ps.capacity, 0) = 0 THEN 'Infinity'::REAL
```
jaahon, line 231: Infinity is now shown even if there are no children present. Is that intentional?

jaahon (approved):

PR #2410, opened 2022-04-29
Make `AsyncButton` better and use it more

akheron:
#### Summary

Fix design problems in the `AsyncButton` API:
- Allow making a synchronous action (or nothing at all) in the `onClick` handler. In this case, the `onClick` handler should return `undefined`.
- It the `onClick` callback returns `undefined`, the button never starts spinning. It just acts like a normal button.
- Remove the old cancel mechanism in favor of this simpler way of controlling whether to start an async call or not.
- `onSuccess` and `onFailure` will only be called if an async action was started.
- If the async action returns a `Success`, `onSuccess` gets the payload as a parameter
- If the async action returns a `Failure`, `onFailure` will get that as a parameter
- The async action (promise) must not reject, as is customary with our API functions.

This required many small changes in the usage sites, and a few bigger refactorings too.

Smaller enhancements to `AsyncButton` in this PR:
- Fix a state update after unmount bug
- Call `e.preventDefault()` if `type` is `"submit"`. This way `AsyncButton` is much more useful as the submit button inside a `<form>` element.

Use  `AsyncButton` in many places where a `Button` was used to start an async action

Fix many usage sites of `AsyncButton` to use the enhanced `onSuccess` and `onFailure` callbacks instead of promise chaining.

### DIFF
--- a/frontend/src/lib-components/atoms/buttons/AsyncButton.tsx
+++ b/frontend/src/lib-components/atoms/buttons/AsyncButton.tsx
@@ -20,17 +20,28 @@ import { faCheck, faTimes } from 'lib-icons'
 
 import { StyledButton } from './Button'
 
-export type AsyncClickCallback = (
-  cancel: () => Promise<void>
-) => Promise<void | Result<unknown>>
+const onSuccessTimeout = isAutomatedTest ? 10 : 500
+const clearStateTimeout = isAutomatedTest ? 25 : 2000
 
-type Props = {
+type ButtonState<T> =
+  | { state: 'idle' | 'in-progress' | 'failure' }
+  | { state: 'success'; value: T }
+
+/* eslint-disable @typescript-eslint/no-explicit-any */
+
+const idle: ButtonState<any> = { state: 'idle' }
+const inProgress: ButtonState<any> = { state: 'in-progress' }
+const failure: ButtonState<any> = { state: 'failure' }
+
+/* eslint-enable @typescript-eslint/no-explicit-any */
+
+export interface Props<T> {
   text: string
   textInProgress?: string
   textDone?: string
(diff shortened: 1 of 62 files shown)


frontend/src/lib-components/atoms/buttons/AsyncButton.tsx
```diff
@@ -14,109 +20,157 @@ import { faCheck, faTimes } from 'lib-icons'
...
+            } else if (result.isFailure) {
+              handleFailure(result)
+            } else {
+              handleSuccess(result.value)
+            }
+          })
+          .catch(() => {
+            throw new Error(
+              "BUG: AsyncButton's promise rejected instead of resolving to a Failure"
+            )
```
akheron, line 127: Show the red cross here. Re-throw the original error or send it to sentry.

frontend/src/lib-components/atoms/buttons/AsyncButton.tsx
```diff
@@ -14,109 +20,157 @@ import { faCheck, faTimes } from 'lib-icons'
...
+
+export interface Props<T> {
   text: string
   textInProgress?: string
   textDone?: string
-  onClick: AsyncClickCallback
-  onSuccess: () => void
-  onFailure?: (result?: Failure<unknown> | undefined) => void
+  onClick: () => Promise<Result<T>> | void
+  onSuccess: (value: T) => void
```
akheron, line 43: Add `onSuccessImmediate` that runs right away instead after `onSuccessTimeout`

akheron: This would have benefited only 1 or 2 of the existing AsyncButton usages, so I decided to not add it now. Let's add it later if it's needed.

PR #2413, opened 2022-05-02
Fix div by zero

reynders:


service/src/main/kotlin/fi/espoo/evaka/attendance/MobileUnitController.kt
```diff
@@ -229,7 +229,7 @@ SELECT
     coalesce(ts.count, 0) AS total_staff,
     CASE
         WHEN coalesce(pc.capacity, 0) > 0 AND coalesce(ps.capacity, 0) = 0 THEN 'Infinity'::REAL
```
jaahon, line 231: This `AND` should be `OR`.

jaahon: Nevermind, this doesn't make any sense 😂

jaahon (approved):

zch (approved):

PR #2417, opened 2022-05-03
Feature to use daycare finance handler always

reynders:

### DIFF
--- a/service/src/integrationTest/kotlin/fi/espoo/evaka/invoicing/VoucherValueDecisionIntegrationTest.kt
+++ b/service/src/integrationTest/kotlin/fi/espoo/evaka/invoicing/VoucherValueDecisionIntegrationTest.kt
@@ -9,6 +9,7 @@ import com.github.kittinunf.fuel.jackson.objectBody
 import com.github.kittinunf.fuel.jackson.responseObject
 import fi.espoo.evaka.FullApplicationTest
 import fi.espoo.evaka.insertGeneralTestFixtures
+import fi.espoo.evaka.invoicing.data.approveValueDecisionDraftsForSending
 import fi.espoo.evaka.invoicing.domain.VoucherValueDecision
 import fi.espoo.evaka.invoicing.domain.VoucherValueDecisionStatus
 import fi.espoo.evaka.invoicing.domain.VoucherValueDecisionSummary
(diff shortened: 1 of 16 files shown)


service/src/integrationTest/kotlin/fi/espoo/evaka/invoicing/VoucherValueDecisionIntegrationTest.kt
```diff
@@ -448,6 +449,40 @@ class VoucherValueDecisionIntegrationTest : FullApplicationTest(resetDbBeforeEac
...
     }
 
+    @Test
+    fun `VoucherValueDecision handler is set to the user when decision is not normal`() {
+        val approvedDecision = createReliefDecision(false)
+        assertEquals(testDecisionMaker_1.id.raw, approvedDecision.decisionHandler)
+    }
+
+    @Test
+    fun `VoucherValueDecision handler is set to the daycare handler when forced when decision is not normal`() {
```
hruoho, line 459: Is this named as you intended?

service/src/main/kotlin/fi/espoo/evaka/shared/FeatureConfig.kt
```diff
@@ -41,4 +41,12 @@ data class FeatureConfig(
      * and all the operating days are marked either sick leaves or planned absences.
      */
     val freeSickLeaveOnContractDays: Boolean,
+
+    /** Whether to always use the marked daycare finance decision handler
+     *
+     *  If true, regardless of situation use the handler marked in the daycare settings
+     *  as the handler of a decision.
+     *  If false, use the current user as the handler as speficied in the code
```
hruoho, line 49:
```suggestion
     *  If true, use the handler marked in the daycare settings, whenever available,
     *  as the handler of a decision.
     *  If false, use either the current user or the handler marked 
     *  in the daycare settings, depending on the case.
```

hruoho (approved):
Looks good to me!

service/src/integrationTest/kotlin/fi/espoo/evaka/invoicing/VoucherValueDecisionIntegrationTest.kt
reynders: Unfortunately yes :D.

PR #2418, opened 2022-05-03
Fix a bug in voucher value report

akheron:
#### Summary

The report didn't work correctly when the `valid_to` date of a decision was moved back to an earlier month, if that month had already been freezed.

Fixing this involved changing a bit how the report generating SQL query works. It now explicitly checks that it's refunding/correcting those months that have changed, and explicitly adds newly added decisions as corrections.

This also has the effect that some unneeded refunds/corrections are not made at all (see the test `when one of part month decisions is annulled it is refunded correctly`).


jaahon (approved):
🥇

PR #2424, opened 2022-05-04
Lastname first

reynders:
#### Summary
- Last name first on voucher value report unit UI and report


frontend/src/employee-frontend/components/reports/VoucherServiceProviderUnit.tsx
```diff
@@ -428,8 +428,8 @@ export default React.memo(function VoucherServiceProviderUnit() {
                         <FixedSpaceColumn spacing="xs">
                           <Link to={`/child-information/${row.childId}`}>
                             {formatName(
-                              row.childFirstName,
                               row.childLastName,
+                              row.childFirstName,
                               i18n
                             )}
```
jaahon, line 434: This does the trick, but `formatName` also has an optional 4th parameter `lastNameFirst` 😄

jaahon (approved):

PR #2430, opened 2022-05-05
Allow importing customizations from e2e tests

akheron:
#### Summary

This is required to for example set a single feature flag and use the espoo defaults for others.


akheron:
https://github.com/espoon-voltti/evaka/pull/2431 is the correct fix

PR #2434, opened 2022-05-09
Allow sending messages to own group's backup care children

reynders:


service/src/main/kotlin/fi/espoo/evaka/messaging/MessageQueries.kt
```diff
@@ -584,6 +584,7 @@ data class MessageReceiversResult(
     val receiverLastName: String
 )
 
+// TODO
```
hruoho, line 587: Still necessary? :D

PR #2435, opened 2022-05-10
Fix voucher value decision ending

akheron:
#### Summary

Fix a bug that ends voucher value decisions prematurely if there's no service need in effect. Also rename the function that ends decisions, because an ended placement is not the only reason.

I tried to bundle finding the end date to the same SQL query too, but that seemed too hard, so this PR keeps the old logic for that part.


akheron:
This is not the correct change. The whole voucher value decision model needs to be reworked a bit instead.

PR #2441, opened 2022-05-11
EVAKA-4406 Bugfix: overnight staff attendances

zch:
Attendances like this were impossible to fix using the 'staff' view in the attendances tab
image

- Fixes the rendering to render arrows on days where there were no entries, and end times where there were actual end times.
- Added support for removing the whole entry from a day by leaving the arrival and departure time fields empty

As of this PR a situation like the above can be fixed like this:
Korjaus


akheron (approved):
🏅

PR #2450, opened 2022-05-13
Add inline "load-failed" script to CSP config

Gekkio:
#### Summary

Our CSP config blocks inline scripts, so we need to explicitly whitelist this one


akheron:
Should there be a comment telling where that hash comes from? 🤔

akheron (approved):

PR #2453, opened 2022-05-13
Boilerplate cleanup 05/2022

Gekkio:
#### Summary

- replace hand-crafted types with codegen types
- remove unnecessary JDBI manual mapping code
- remove dead code
- use proper types instead of strings

Bugs found and fixed:

- Decisions report "provider type" column was always empty (would have been a compilation error, but custom types were wrong)


frontend/src/employee-frontend/components/reports/Occupancies.tsx
```diff
@@ -122,9 +122,10 @@ function getDisplayCells(
 ): string[][] {
   return reportRows.map((row) => {
     if (usedValues === 'raw') {
-      const cells = [row.unitName, row.groupName].filter(
-        (cell) => cell !== undefined
-      ) as string[]
+      const cells = [
+        row.unitName,
+        'groupName' in row ? row.groupName : undefined
+      ].filter((cell): cell is string => cell !== undefined)
```
akheron, line 128: Here's a trick to avoid the type annotation and make this a bit shorter:

```
const cells = [
  row.unitName,
  'groupName' in row ? row.groupName : undefined
].flatMap((cell) => cell !== undefined ? cell : [])
```

The applicability of the trick depends on whether the future readers understand that `flatMap` here is actually used for filtering.

On the teoretical side of things, it's actually more correct to use `flatMap`, because most type systems don't allow changing the collection type with `filter`. (In this case the type is changed from `(string | undefined)[]` to `string[]`.)

frontend/src/employee-frontend/components/reports/Occupancies.tsx
```diff
@@ -175,7 +176,7 @@ function getDisplayCells(
             return occupancy?.[usedValues] ?? ''
           }
         })
-      ].filter((cell) => cell !== undefined) as string[]
+      ].filter((cell): cell is string => cell !== undefined)
```
akheron, line 179: Same here

akheron (approved):

frontend/src/employee-frontend/components/reports/Occupancies.tsx
Gekkio: I redid this part of the code and handle this one optional cell in a different way. It also fixes a potential bug where null in the data would lead to a missing cell (= broken table layout)

PR #2454, opened 2022-05-16
Add daycare type to InvoiceRowDetailed objects

vesahonkasaari:
Add the new field to InvoiceRowDetailed and fetch the value from the DB

Oulu and Turku need this information for their invoicing

### DIFF
--- a/service/src/main/kotlin/fi/espoo/evaka/invoicing/domain/Invoices.kt
+++ b/service/src/main/kotlin/fi/espoo/evaka/invoicing/domain/Invoices.kt
@@ -5,6 +5,7 @@
 package fi.espoo.evaka.invoicing.domain
 
 import com.fasterxml.jackson.annotation.JsonIgnoreProperties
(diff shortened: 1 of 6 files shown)


service/src/main/kotlin/fi/espoo/evaka/invoicing/domain/Invoices.kt
```diff
@@ -101,6 +101,7 @@ data class InvoiceRowDetailed(
     val periodEnd: LocalDate,
     val product: ProductKey,
     val unitId: DaycareId,
+    val daycareType: String,
```
akheron, line 104: In the database this is an array of `care_types`, and I think it should be reflected in the Kotlin side, too. That is, the type here would be `List<CareType>`.

akheron (changes requested):
The build also fails because of a compilation error.

akheron (approved):

PR #2459, opened 2022-05-17
Fix typing errors and bugs related to object indexing

akheron:


service/src/main/kotlin/fi/espoo/evaka/application/ApplicationQueries.kt
```diff
@@ -540,25 +536,6 @@ fun Database.Read.getCitizenChildren(citizenId: PersonId): List<CitizenChildren>
         .list()
 }
 
-private val toPersonApplicationSummary: (ResultSet, StatementContext) -> PersonApplicationSummary = { rs, _ ->
```
Gekkio, line 543: Several column names are different, so I don't think you can remove this manual mapping without modifying the queries a bit

akheron: Good catch! Luckily this was also caught by tests. Fixed.

frontend/src/citizen-frontend/localization/state.tsx
```diff
@@ -17,9 +17,7 @@ const getDefaultLanguage: () => Lang = () => {
   if (lang && langs.includes(lang as Lang)) {
     return lang as Lang
   } else {
-    const language = (
-      (window.navigator['userLanguage'] as string) || window.navigator.language
-    ).split('-')[0]
```
akheron, line 22: `navigator.userLanguage` is old-IE specific and can be removed safely

PR #2470, opened 2022-05-19
EVAKA-4406 Show utilization in group selector for realtime attendances

zch:
Implements the new design for the group selector in the employee mobile.

If the realtime staff attendances feature flag is on, the current
utilization will be shown next to the name of each group in the group
selector. In other cases a `present children / total children` count
will be shown.

### DIFF
--- a/service/src/main/kotlin/fi/espoo/evaka/attendance/MobileUnitController.kt
+++ b/service/src/main/kotlin/fi/espoo/evaka/attendance/MobileUnitController.kt
@@ -23,6 +23,7 @@ import org.springframework.web.bind.annotation.RequestMapping
 import org.springframework.web.bind.annotation.RequestParam
 import org.springframework.web.bind.annotation.RestController
 import java.time.LocalDate
+import kotlin.math.roundToInt
 
 @RestController
 @RequestMapping("/mobile/units")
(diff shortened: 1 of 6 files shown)


service/src/main/kotlin/fi/espoo/evaka/attendance/MobileUnitController.kt
```diff
@@ -97,18 +118,84 @@ fun Database.Read.fetchUnitInfo(unitId: DaycareId, date: LocalDate): UnitInfo {
...
         FROM daycare u
-        JOIN daycare_group g on u.id = g.daycare_id AND daterange(g.start_date, g.end_date, '[]') @> :date
+            JOIN daycare_group g ON u.id = g.daycare_id AND daterange(g.start_date, g.end_date, '[]') @> :date
+            LEFT JOIN child c ON c.group_id = g.id
+            LEFT JOIN staff s ON s.group_id = g.id
         WHERE u.id = :unitId
         """.trimIndent()
 
-    val groups = createQuery(groupsSql)
+    @ExcludeCodeGen
```
jaahon, line 173: This annotation doesn't do anything since the type is not used in any api.

service/src/main/kotlin/fi/espoo/evaka/attendance/MobileUnitController.kt
```diff
@@ -97,18 +118,84 @@ fun Database.Read.fetchUnitInfo(unitId: DaycareId, date: LocalDate): UnitInfo {
...
         """
-        SELECT g.id, g.name
+        WITH child AS (
+            SELECT
+                dgp.daycare_group_id as group_id,
+                SUM(COALESCE(an.capacity_factor, 1) * CASE
+                    WHEN dc.type && array['FAMILY', 'GROUP_FAMILY']::care_types[] THEN 1.75
+                    WHEN extract(YEARS FROM age(ca.date, ch.date_of_birth)) < 3 THEN coalesce(sno.occupancy_coefficient_under_3y, default_sno.occupancy_coefficient_under_3y)
+                    ELSE coalesce(sno.occupancy_coefficient, default_sno.occupancy_coefficient)
+                END) AS capacity
```
jaahon, line 128: It's pretty unfortunate that this formula is now duplicated multiple times. Could it be extracted into a function?

zch: Discussed on slack and came to the conclusion not to change this as the duplication is only in this file.

jaahon (approved):

PR #2476, opened 2022-05-19
Customizable invoice generation logic

akheron:


service/src/main/kotlin/fi/espoo/evaka/invoicing/service/DraftInvoiceGenerator.kt
```diff
@@ -161,6 +180,14 @@ class DraftInvoiceGenerator(
             .groupBy { (_, stub) -> stub.child }
             .flatMap { (child, childStubs) ->
                 val separatePeriods = mergePeriods(childStubs)
+
+                val logic = invoiceGenerationLogicChooser.logicForMonth(
+                    separatePeriods.first().first.start.year,
+                    separatePeriods.first().first.start.month,
```
jaahon, line 186: Why not just use `invoicePeriod` instead of `separatePeriods.first().first`?

service/src/main/kotlin/fi/espoo/evaka/invoicing/service/DraftInvoiceGenerator.kt
```diff
@@ -30,13 +30,32 @@ import java.math.BigDecimal
...
+interface InvoiceGenerationLogicChooser {
+    fun logicForMonth(year: Int, month: Month, childId: ChildId): InvoiceGenerationLogic
+}
+
+class DefaultInvoiceGenerationLogic : InvoiceGenerationLogicChooser {
+    override fun logicForMonth(year: Int, month: Month, childId: ChildId): InvoiceGenerationLogic =
+        InvoiceGenerationLogic.Default
+}
+
+val defaultInvoiceGenerationLogic = DefaultInvoiceGenerationLogic()
```
jaahon, line 52: Why do you have a separate `defaultInvoiceGenerationLogic` variable? If the idea is to not create new objects, then you could use kotlin's singletons instead:

```
object DefaultInvoiceGenerationLogic : InvoiceGenerationLogicChooser { ... }
```

service/src/main/kotlin/fi/espoo/evaka/invoicing/service/DraftInvoiceGenerator.kt
akheron: Fixed!

service/src/main/kotlin/fi/espoo/evaka/invoicing/service/DraftInvoiceGenerator.kt
akheron: That was the idea but I didn't know `object` can be used in this way 🙈. Fixed.

jaahon (approved):

PR #2486, opened 2022-05-23
EVAKA-FIX: handle both placement proposal rejections

hruoho:
#### Summary

Handle both placement proposal rejection statuses

### DIFF
--- a/frontend/src/e2e-test/pages/admin/application-workbench-page.ts
+++ b/frontend/src/e2e-test/pages/admin/application-workbench-page.ts
@@ -14,21 +14,21 @@ import ApplicationDetailsPage from './application-details-page'
 export class SearchFilter {
   constructor(private page: Page) {}
 
-  #statusRadioAll = this.page.find('[data-qa="application-status-filter-ALL"]')
+  #statusRadioAll = this.page.findByDataQa('application-status-filter-ALL')
 
   async filterByTransferOnly() {
-    await this.page.find('[data-qa="filter-transfer-only"]').click()
+    await this.page.findByDataQa('filter-transfer-only').click()
   }
 
   async filterByTransferExclude() {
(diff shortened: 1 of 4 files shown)


frontend/src/e2e-test/pages/admin/application-workbench-page.ts
```diff
@@ -54,6 +54,13 @@ export class ApplicationWorkbenchPage {
   #applicationNoteContent = this.page.find(
     '[data-qa="application-note-content"]'
   )
+
+  #applicationPlacementProposalStatusIndicator = this.page.find(
+    '[data-qa="placement-proposal-status"]'
+  )
+  #applicationPlacementProposalStatusTooltip = this.page.find(
+    '[data-qa="placement-proposal-status-tooltip"]'
+  )
```
akheron, line 63: Perhaps worth another PR, but nowadays we have `page.findByDataQa()` helper that removes the need for the `[data-qa="` boilerplate in selectors.

frontend/src/e2e-test/pages/admin/application-workbench-page.ts
```diff
@@ -178,6 +185,19 @@ export class ApplicationWorkbenchPage {
...
+  ) {
+    await waitUntilTrue(async () => {
+      await this.#applicationPlacementProposalStatusIndicator.hover()
+      const tooltip = await this.#applicationPlacementProposalStatusTooltip
+        .innerText
+      const match = tooltip.match(matchingText)
+      return match ? match.length > 0 : false
+    })
+  }
+
```
akheron, line 200: I think using `waitUntilTrue` is not very useful here. Playwright already waits that the status indicator is in the DOM for `.hover()` to succeed. If the tooltip text is incorrect and `waitUntilTrue` retries, hovering that element again doesn't change anything as the tooltip is already visible.

akheron (approved):

PR #2488, opened 2022-05-24
EVAKA-FIX: Allow extended care unit supervisor to read application attachment

hruoho:

### DIFF
--- a/frontend/src/e2e-test/pages/employee/applications/application-read-view.ts
+++ b/frontend/src/e2e-test/pages/employee/applications/application-read-view.ts
@@ -143,6 +143,12 @@ export default class ApplicationReadView {
       .waitUntilVisible()
   }
 
+  async assertExtendedCareAttachmentDoesNotExist(fileName: string) {
+    await this.page
+      .findByDataQa(`extended-care-attachment-${fileName}`)
+      .waitUntilHidden()
+  }
+
   async assertApplicantIsDead() {
     await this.page.find('[data-qa="applicant-dead"]').waitUntilVisible()
   }
--- a/frontend/src/e2e-test/specs/5_employee/application-attachments.spec.ts
+++ b/frontend/src/e2e-test/specs/5_employee/application-attachments.spec.ts
@@ -3,15 +3,25 @@
 // SPDX-License-Identifier: LGPL-2.1-or-later
 
 import config from '../../config'
-import { insertApplications, resetDatabase } from '../../dev-api'
+import {
+  execSimpleApplicationActions,
+  insertApplications,
+  resetDatabase
+} from '../../dev-api'
 import { initializeAreaAndPersonData } from '../../dev-api/data-init'
 import {
   applicationFixture,
   applicationFixtureId,
-  Fixture
+  daycareFixture,
+  enduserChildFixtureKaarina,
+  enduserGuardianFixture,
+  Fixture,
+  uuidv4
 } from '../../dev-api/fixtures'
 import ApplicationsPage from '../../pages/employee/applications'
+import ApplicationReadView from '../../pages/employee/applications/application-read-view'
 import EmployeeNav from '../../pages/employee/employee-nav'
+import { UnitPage } from '../../pages/employee/units/unit'
 import { Page } from '../../utils/page'
 import { employeeLogin } from '../../utils/user'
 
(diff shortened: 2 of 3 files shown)


service/src/main/kotlin/fi/espoo/evaka/shared/security/actionrule/HasUnitRole.kt
```diff
@@ -97,6 +97,27 @@ AND av.id = ANY(:ids)
...
     )
 
+    fun inPlacementPlanUnitWithExtendedCare() = DatabaseActionRule(
+        this,
+        Query<ApplicationId> { tx, user, ids ->
+            tx.createQuery(
+                """
+SELECT av.id, role
+FROM application_view av
+LEFT JOIN placement_plan pp ON pp.application_id = av.id
```
akheron, line 107: I think this should be an inner join (~`LEFT JOIN`~ -> `JOIN`)

akheron (changes requested):

service/src/main/kotlin/fi/espoo/evaka/shared/security/Action.kt
```diff
@@ -261,7 +261,7 @@ sealed interface Action {
         CREATE_NOTE(HasGlobalRole(SERVICE_WORKER), HasUnitRole(SPECIAL_EDUCATION_TEACHER).inPreferredUnitOfApplication(), HasUnitRole(SPECIAL_EDUCATION_TEACHER).inPlacementPlanUnitOfApplication()),
         READ_SPECIAL_EDUCATION_TEACHER_NOTES(HasUnitRole(SPECIAL_EDUCATION_TEACHER).inPreferredUnitOfApplication(), HasUnitRole(SPECIAL_EDUCATION_TEACHER).inPlacementPlanUnitOfApplication()),
 
-        READ_ATTACHMENTS(HasGlobalRole(SERVICE_WORKER)),
+        READ_ATTACHMENTS(HasGlobalRole(SERVICE_WORKER), HasUnitRole(UNIT_SUPERVISOR).inPlacementPlanUnitWithExtendedCare()),
```
Gekkio, line 264: This rule gives access to *all application attachments* for supervisors in extended care units. Isn't the requirement to limit access to only attachments with type `EXTENDED_CARE`?

If so, I think a different approach is needed. My suggestion: remove the READ_ATTACHMENTS Action completely, and instead filter attachments using one `checkPermissionFor` batch operation in the controller.

So, instead of

```kotlin
val attachments =
    if (accessControl.hasPermissionFor(user, Action.Application.READ_ATTACHMENTS, applicationId)) {
        tx.getApplicationAttachments(applicationId)
    } else {
        listOf()
    }
```

do something like this (code is untested):

```kotlin
val attachments = tx.getApplicationAttachments(applicationId).let { allAttachments ->
  val decisions = accessControl.checkPermissionFor(user, Action.Attachment.READ_APPLICATION_ATTACHMENT, allAttachments.map { it.id })
  allAttachments.filter { decisions[it.id].isPermitted() }
}
```

READ_APPLICATION_ATTACHMENT already includes a rule that handles both the `round_the_clock` flag in the daycare *and* the `EXTENDED_CARE` type of the attachment. The only downside is that we fetch the attachments even if the user can't have access to any of them, but this seems like an acceptable trade-off to me.

hruoho: Ah right! That seems to be the proper action to look 👍  Maybe we could just filter `allAttachments` by checking that user has permission for `READ_APPLICATION_ATTACHMENT` for that attachment 🤔

akheron:
Otherwise looks good to me 👍

akheron (approved):

service/src/main/kotlin/fi/espoo/evaka/application/ApplicationControllerV2.kt
```diff
@@ -294,12 +294,11 @@ class ApplicationControllerV2(
...
                         .map { personDTO -> PersonJSON.from(personDTO) }
 
-                val attachments =
-                    if (accessControl.hasPermissionFor(user, Action.Application.READ_ATTACHMENTS, applicationId)) {
-                        tx.getApplicationAttachments(applicationId)
-                    } else {
-                        listOf()
+                val attachments = tx.getApplicationAttachments(applicationId).let { allAttachments ->
+                    allAttachments.filter { attachment ->
+                        accessControl.hasPermissionFor(user, Action.Attachment.READ_APPLICATION_ATTACHMENT, attachment.id)
```
akheron, line 299: You should actually do this like @Gekkio suggested in the previous comment. Calling `accessControl.hasPermissionFor()` in a loop causes as many database queries as there are attachments. You can use `accessControl.checkPermissionFor` by giving it the ids of all attachments, and then act on the the `Map` it returns.

akheron (changes requested):

akheron (approved):

PR #2499, opened 2022-05-30
Use direct file downloads instead of `Blob` trickery

akheron:
#### Summary

Show a custom error message if a file is not "available" in S3. This happens is the file has been deleted, virus scanning is in progress, or the permissions don't otherwise allow access.

Remove `Blob` based file download logic. Instead, open a direct URL of the file in a new window with `window.open` directly in the click handler. This has multiple advantages:


reynders (approved):
Approved after a mob review and code read review

PR #2500, opened 2022-05-30
Only cancel sent transfer applications when terminating

reynders:
- Only cancel sent transfer applications when terminating
- Only clean up sent transfer applications during nightly cleanup. And because sent application cannot have placements nor decisions, no need to try to clean up those


jaahon (commented):
The scheduled job that runs `cancelOutdatedTransferApplications` should be updated as well.

jaahon (approved):

PR #2501, opened 2022-05-30
Relief decision is not reflected on voucher value report co payment

reynders:
#### Summary
- Add E2E test for unit voucher value report

### DIFF
--- a/frontend/src/e2e-test/utils/page.ts
+++ b/frontend/src/e2e-test/utils/page.ts
@@ -186,6 +186,18 @@ export class Element {
     return this.locator.textContent()
   }
 
+  get textContentAsFloat(): Promise<number | null> {
+    return this.locator
+      .textContent()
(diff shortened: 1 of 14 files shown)


frontend/src/e2e-test/utils/page.ts
```diff
@@ -186,6 +186,18 @@ export class Element {
...
+    return this.locator
+      .textContent()
+      .then((str) => (str ? parseFloat(str) : null))
+  }
+
+  get textContentAsInt(): Promise<number | null> {
+    return this.locator
+      .textContent()
+      .then((str) => (str ? parseInt(str) : null))
+  }
```
jaahon, line 199: Is this ever really needed since javascript only has floats? In general, I personally prefer comparing strings since that's what the content is anyway.

jaahon (approved):
Good job!

PR #2509, opened 2022-06-01
EVAKA-4406 Fix differing utilization on unit list view vs group select

zch:
Doubles were truncated to integers, which caused the utilization percentage to differ on the unit list view vs the group selector.

### DIFF
--- a/service/src/integrationTest/kotlin/fi/espoo/evaka/attendance/MobileUnitControllerIntegrationTest.kt
+++ b/service/src/integrationTest/kotlin/fi/espoo/evaka/attendance/MobileUnitControllerIntegrationTest.kt
@@ -1,0 +1,131 @@
+// SPDX-FileCopyrightText: 2017-2022 City of Espoo
+//
+// SPDX-License-Identifier: LGPL-2.1-or-later
+
+package fi.espoo.evaka.attendance
+
+import com.github.kittinunf.fuel.jackson.responseObject
+import fi.espoo.evaka.FixtureBuilder
+import fi.espoo.evaka.FullApplicationTest
+import fi.espoo.evaka.insertGeneralTestFixtures
+import fi.espoo.evaka.placement.PlacementType
+import fi.espoo.evaka.shared.DaycareId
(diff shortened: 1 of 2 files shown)


service/src/integrationTest/kotlin/fi/espoo/evaka/attendance/MobileUnitControllerIntegrationTest.kt
```diff
@@ -0,0 +1,136 @@
...
+    }
+
+    private fun fetchUnitInfo(): UnitInfo {
+        val (_, res, result) = http.get(
+            "/mobile/units/${testDaycare.id}",
+            listOf(Pair("useRealtimeStaffAttendance", true))
+        )
+            .asUser(mobileUser)
+            .withMockedTime(now)
+            .responseObject<UnitInfo>(jsonMapper)
```
jaahon, line 118: I think it's inconsistent that this function doesn't get unit id as parameter, but `fetchUnitStats` does.

jaahon (approved):

service/src/integrationTest/kotlin/fi/espoo/evaka/attendance/MobileUnitControllerIntegrationTest.kt
```diff
@@ -0,0 +1,136 @@
...
+    private val mobileUser = AuthenticatedUser.MobileDevice(MobileDeviceId(UUID.randomUUID()))
+    private val groupId = GroupId(UUID.randomUUID())
+    private val groupName = "Testaajat"
+    private val today = LocalDate.of(2022, 2, 3)
+    private val now = HelsinkiDateTime.of(today, LocalTime.of(12, 5, 1))
+    private val placementStart = now.toLocalDate().minusDays(30)
+    private val placementEnd = now.toLocalDate().plusDays(30)
+    private lateinit var testEmployeeId_1: EmployeeId
+    private lateinit var testEmployeeId_2: EmployeeId
+
```
jaahon, line 51: None of these variables are used outside the data initialisation, so I don't think they need to be declared up here.

PR #2512, opened 2022-06-01
Evaka 4406 more utilization fixes

zch:
- Fix bug where utilization was calculated incorrectly if a child was placed in more than one group during their daycare unit placement
- Ensure that the utilization % always is formatted correctly on the supervisor's unit list view.


service/src/integrationTest/kotlin/fi/espoo/evaka/attendance/MobileUnitControllerIntegrationTest.kt
```diff
@@ -69,13 +73,38 @@ class MobileUnitControllerIntegrationTest : FullApplicationTest(resetDbBeforeEac
                     endDate = placementEnd
                 )
             }
+
+            // Add child with mulitple group placements
```
patari, line 77: typo

patari (approved):

PR #2521, opened 2022-06-02
EVAKA-4406 Fix bug where employee was shown as present in all units

zch:
Fixes an issue where an employee who is present in one unit is shown as present in all units where they have access. This bug was present only in the employee mobile


jaahon (commented):
This fixes the bug, but when you open the staff members page on mobile, it just says that the person arrived at XX:XX. I think that could be improved by showing "is present in another unit" or something.

zch:
> This fixes the bug, but when you open the staff members page on mobile, it just says that the person arrived at XX:XX. I think that could be improved by showing "is present in another unit" or something.

Excellent point. I'll have a look at that as well.

jaahon (approved):

PR #2524, opened 2022-06-03
Add employee number

patari:
#### Summary

Tampere is going to integrate working time monitor system into eVaka and identifiers for employees are either SSN or some employee number. We chose to use employee number which we can get from our SSO with claim name `http://schemas.xmlsoap.org/ws/2005/05/identity/claims/employeenumber`. All of our users have already logged in so we need to update employee number on next login.

This also adds employee name and email update on login.


service/src/main/kotlin/fi/espoo/evaka/pis/Employee.kt
```diff
@@ -14,6 +14,7 @@ data class Employee(
     val lastName: String,
     val email: String?,
     val externalId: ExternalId?,
+    val employeeNumber: String?,
```
Gekkio, line 17: Do we need the field here? As far as I understand, the employee number is not actually used anywhere except login and later in integration-related queries.

service/src/main/kotlin/fi/espoo/evaka/pis/EmployeeQueries.kt
```diff
@@ -24,13 +24,20 @@ data class NewEmployee(
...
     val externalId: ExternalId?,
+    val employeeNumber: String?,
     val roles: Set<UserRole> = setOf()
 )
 
 data class EmployeeUser(
     val id: EmployeeId,
     val firstName: String,
     val lastName: String,
+    val employeeNumber: String?,
```
Gekkio, line 35: Same here: this type is used for stuff that needs to go to the user session, so the employee number is probably not needed. Note how even the external id is missing, because we don't need that in the session

service/src/main/kotlin/fi/espoo/evaka/pis/Employee.kt
patari: Fixed

service/src/main/kotlin/fi/espoo/evaka/pis/EmployeeQueries.kt
patari: Fixed

Gekkio (approved):

Gekkio:
:+1: Needs rebasing

PR #2530, opened 2022-06-08
Oulu-412 Hide some service needs from the citizen

vesahonkasaari:
Add the show_for_citizen field to service_need_option table and use it to
filter the service need options visible in the citizen UI

The show_for_citizen field defaults to true, so the functionality will remain the same for anyone who does not actively choose to set some rows to false to hide them from the citizen

### DIFF
--- a/service/src/integrationTest/kotlin/fi/espoo/evaka/serviceneed/ServiceNeedQueriesTest.kt
+++ b/service/src/integrationTest/kotlin/fi/espoo/evaka/serviceneed/ServiceNeedQueriesTest.kt
@@ -5,7 +5,11 @@
 package fi.espoo.evaka.serviceneed
 
 import fi.espoo.evaka.PureJdbiTest
+import fi.espoo.evaka.insertServiceNeedOptions
 import fi.espoo.evaka.placement.PlacementType
+import fi.espoo.evaka.shared.db.Database
+import fi.espoo.evaka.snPreparatoryDaycare50
+import fi.espoo.evaka.snPreschoolDaycare45
 import org.assertj.core.api.Assertions.assertThat
 import org.junit.jupiter.api.Test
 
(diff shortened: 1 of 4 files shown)


service/src/integrationTest/kotlin/fi/espoo/evaka/serviceneed/ServiceNeedQueriesTest.kt
```diff
@@ -14,4 +16,15 @@ class ServiceNeedQueriesTest : PureJdbiTest(resetDbBeforeEach = true) {
     fun getServiceNeedOptionPublicInfos() {
         assertThat(db.read { tx -> tx.getServiceNeedOptionPublicInfos(PlacementType.values().toList()) }).isEmpty()
     }
+
+    @Test
+    fun getOnlyShownServiceNeedOptionPublicInfos() {
+        db.transaction { tx -> tx.insertServiceNeedOptions() }
+        db.transaction { tx -> tx.updateShowForCitizen() }
+        assertThat(db.read { tx -> tx.getServiceNeedOptionPublicInfos(PlacementType.values().toList()) }).hasSize(10)
```
jaahon, line 24: This test is a bit fragile since nothing guarantees that the amount of service need options stays the same. I would just check that the updated service need options is not returned by the function. You can also just use the service need option's id in the update query and the check.

vesahonkasaari: Agreed, even though in principle nothing completely protects the test from the test data being changed under it. Couldn't see how to use the option's ID since it's randomized in the test data, so I used their names instead.

vesahonkasaari: Oh, right, Monday morning... Done.

service/src/integrationTest/kotlin/fi/espoo/evaka/serviceneed/ServiceNeedQueriesTest.kt
```diff
@@ -14,4 +18,20 @@ class ServiceNeedQueriesTest : PureJdbiTest(resetDbBeforeEach = true) {
...
+        val queriedOptions = db.read { tx -> tx.getServiceNeedOptionPublicInfos(PlacementType.values().toList()) }
+        // I couln't get any of the fancy asssertj list assertions to work in Kotlin, so test the stupid way
+        queriedOptions.forEach {
+            assertThat(it.id).isNotEqualTo(snPreschoolDaycare45.id)
+            assertThat(it.id).isNotEqualTo(snPreparatoryDaycare50.id)
+        }
+    }
+
+    fun Database.Transaction.updateShowForCitizen() {
+        createUpdate("UPDATE service_need_option SET show_for_citizen=FALSE WHERE id='" + snPreschoolDaycare45.id + "' OR id='" + snPreparatoryDaycare50.id + "'").execute()
```
jaahon, line 35: I know there is no risk of sql injection here, but still using a parameterized query is a good idea for consistency. You can use eg. 
```
execute(UPDATE service_need_option SET show_for_citizen=FALSE WHERE id=? OR id=?", snPreschoolDaycare45.id, snPreparatoryDaycare50.id)
```

vesahonkasaari: OK

jaahon (approved):

jaahon:
This branch needs a rebase because there are new migrations in master.

vesahonkasaari:
I rebased and squashed this, better merge it now before the migration scripts get head ahead of this again...

PR #2531, opened 2022-06-09
Make voucher values more customizable

akheron:


service/src/main/kotlin/fi/espoo/evaka/invoicing/service/VoucherValueDecisionGenerator.kt
```diff
@@ -179,13 +177,18 @@ private fun generateNewValueDecisions(
...
-                    if (placement == null || serviceVoucherUnits.none { unit -> unit == placement.unitId }) 0
-                    else calculateVoucherValue(baseValue, capacityFactor, placement.serviceNeed.voucherValueCoefficient)
+                val baseCoPayment =
+                    calculateBaseFee(price, family.getSize(), familyIncomes + listOfNotNull(childIncome))
+                val voucherValue =
+                    if (placement == null || serviceVoucherUnits.none { unit -> unit == placement.unitId }) null
+                    else {
+                        val voucherValue =
+                            voucherValues[placement.serviceNeed.optionId]?.find { it.validity.contains(period) }
+                                ?: error("Missing voucher value for period ${period.start} - ${period.end}, cannot generate voucher value decision")
```
akheron, line 187: Service need option id should be logged here. Otherwise it's hard to know which voucher value is missing.

service/src/main/kotlin/fi/espoo/evaka/invoicing/service/VoucherValueDecisionGenerator.kt
```diff
@@ -200,7 +203,7 @@ private fun generateNewValueDecisions(
                     child = voucherChild,
                     baseCoPayment = baseCoPayment,
                     siblingDiscount = price.siblingDiscountPercent(1),
-                    baseValue = baseValue,
+                    baseValue = 0,
```
akheron, line 206: Can `baseValue` be zero for empty decisions? We cannot have any other number really, because there's no placement and thus no service need.

service/src/main/kotlin/fi/espoo/evaka/invoicing/service/VoucherValueDecisionGenerator.kt
```diff
@@ -109,7 +110,7 @@ private fun generateNewValueDecisions(
     families: List<FridgeFamily>,
     allPlacements: Map<ChildWithDateOfBirth, List<Pair<DateRange, PlacementWithServiceNeed>>>,
     prices: List<FeeThresholds>,
-    voucherValues: List<VoucherValue>,
+    voucherValues: Map<ServiceNeedOptionId, List<ServiceNeedOptionVoucherValue>>,
```
jaahon, line 113: Just noticed that the date ranges of `voucherValues` is missing from `val periods`, can you add them now? 😅

service/src/main/kotlin/fi/espoo/evaka/invoicing/service/VoucherValueDecisionGenerator.kt
jaahon: Yes, I agree.

service/src/main/kotlin/fi/espoo/evaka/invoicing/service/VoucherValueDecisionGenerator.kt
jaahon: Yeah, I think that's fine. It doesn't make sense either way really 😄

service/src/main/kotlin/fi/espoo/evaka/invoicing/service/VoucherValueDecisionGenerator.kt
akheron: Added 👍

service/src/main/kotlin/fi/espoo/evaka/invoicing/service/VoucherValueDecisionGenerator.kt
akheron: Fixed

jaahon (approved):

PR #2537, opened 2022-06-10
Mobile pairing improvements

jolle:
#### Summary
- Removed lowercase characters from the available character pool for generating pairing codes
  - Increased the code length to 12 (from 10) in order to keep the amount of possible combinations ≥ as before
- Device names in the unit view are sorted alphabetically

### DIFF
--- a/frontend/src/employee-frontend/components/unit/tab-unit-information/UnitAccessControl.tsx
+++ b/frontend/src/employee-frontend/components/unit/tab-unit-information/UnitAccessControl.tsx
@@ -323,7 +323,7 @@ const DevicesTable = React.memo(function DevicesTable({
         </Tr>
       </Thead>
       <Tbody>
-        {sortBy(rows, ['name']).map((row) => (
+        {sortBy(rows, (row) => row.name).map((row) => (
           <DeviceRow
             key={row.id}
             row={row}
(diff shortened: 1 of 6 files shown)


frontend/src/employee-frontend/components/unit/tab-unit-information/UnitAccessControl.tsx
```diff
@@ -323,7 +323,7 @@ const DevicesTable = React.memo(function DevicesTable({
         </Tr>
       </Thead>
       <Tbody>
-        {rows.map((row) => (
+        {sortBy(rows, ['name']).map((row) => (
```
Gekkio, line 326: Consider using the function-based variant of `sortBy`, because lodash typings lose typing information in the "field name as string"-variants. For example:

This compiles but doesn't work (AFAIK there's no runtime error, but it just doesn't sort correctly):

```typescript
sortBy(rows, ['lolasdf'])
```

These both give a proper compilation error:

```typescript
sortBy(rows, ({ lolasdf }) => lolasdf)
sortBy(rows, (row) => row.lolasdf)
```

It's a bit longer but better for code maintainability. We still use the "bad variant" in many places, but in new code the function-based variant is better.

reynders (approved):

PR #2539, opened 2022-06-10
Add unit fields: business id, iban, provider id

akheron:


Gekkio (commented):
Should the fields be added to `DevDaycare` too? I'm not sure if we intend to create E2E tests that need them, but usually the DevXX types are roughly 1:1 mappings to the corresponding tables.

akheron:
> Should the fields be added to `DevDaycare` too? I'm not sure if we intend to create E2E tests that need them, but usually the DevXX types are roughly 1:1 mappings to the corresponding tables.

Good catch, fixed!

Gekkio (approved):

PR #2541, opened 2022-06-10
Vasu document read -support

reynders:
#### Summary
- Show total number of different unread documents on citizen front page
- Show number of unread vasu documents on citizen vasu listing heading
- It was agreed that the vasu unread state is the same as the given-permission-to-share -state, so that is used

### DIFF
--- a/frontend/src/citizen-frontend/child-documents/CitizenVasuAndLeops.tsx
+++ b/frontend/src/citizen-frontend/child-documents/CitizenVasuAndLeops.tsx
@@ -39,15 +39,11 @@ const VasuTableContainer = styled.table`
 const VasuTr = styled.tr`
   & td {
     vertical-align: top;
+    padding-bottom: 16px;
+    padding-right: 16px;
   }
 `
 
-const VasuTd = styled.td`
-  padding-bottom: 16px;
-  padding-right: 16px;
-`
-
-const StateTd = styled(VasuTd)``
 const PermissionToShareText = styled.span`
   padding-left: 6px;
 `
@@ -71,6 +67,22 @@ const PermissionToShare = React.memo(function PermissionToShare() {
   )
 })
 
+const LinkTd = styled.td`
+  width: 40%;
+`
+
+const StateTd = styled.td`
+  width: 10%;
+`
+
+const DateTd = styled.td`
+  width: 10%;
+`
+
+const PermissionTd = styled.td`
+  width: 45%;
+`
+
 const VasuTable = React.memo(function VasuTable({
   summary
 }: {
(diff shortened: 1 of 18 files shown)


frontend/src/citizen-frontend/child-documents/vasu/VasuPage.tsx
```diff
@@ -126,8 +131,16 @@ export default React.memo(function VasuPage() {
...
-                onClick={() => givePermissionToShareVasu(id)}
-                onSuccess={() => setGuardianHasGivenPermissionToShare(true)}
+                onClick={() => {
+                  return givePermissionToShareVasu(id).then((res) => {
+                    refreshUnreadVasuDocumentsCount()
+                    return res
+                  })
+                }}
+                onSuccess={() => {
+                  console.log('DDEBUG')
```
jaahon, line 141: Remove this.

frontend/src/citizen-frontend/header/Header.tsx
```diff
@@ -35,30 +35,41 @@ export default React.memo(function Header(props: { ariaHidden: boolean }) {
...
+  }, [refreshUnreadVasuDocumentsCount, refreshUnreadPedagogicalDocumentsCount, user])
 
   const location = useLocation()
   const hideLoginButton = location.pathname === '/login'
 
+  const unreadDocumentCount = useCallback(() => {
+    return (
+      (unreadPedagogicalDocumentsCount || 0) + (unreadVasuDocumentsCount || 0)
+    )
+  }, [unreadPedagogicalDocumentsCount, unreadVasuDocumentsCount])
```
jaahon, line 57: Wrapping this in a `useCallback` hook and then just evaluating it immediately is just a waste of resources and it hurts readability. Just do 
```
const unreadDocumentCount = (unreadPedagogicalDocumentsCount ?? 0) + (unreadVasuDocumentsCount ?? 0)
```

frontend/src/lib-components/layout/Container.tsx
```diff
@@ -109,10 +116,20 @@ export const CollapsibleContentArea = React.memo(
...
         >
           {title}
-          <TitleIcon
-            icon={open ? faChevronUp : faChevronDown}
-            data-qa="collapsible-trigger"
-          />
+          <IconContainer>
+            {countIndicator > 0 && (
+              <>
+                <CircledChar color="white" data-qa="count-indicator">
```
jaahon, line 122: We already have a `RoundIcon` component that could be used here. 

Also, citizen application sections have error indicators that basically this same thing. You could use this same mechanism there if you want to do the refactoring.

frontend/src/citizen-frontend/child-documents/CitizenVasuAndLeops.tsx
```diff
@@ -71,6 +69,22 @@ const PermissionToShare = React.memo(function PermissionToShare() {
...
+  width: 10%;
+`
+
+const DateTd = styled.td`
+  width: 10%;
+`
+
+const PermissionTd = styled.td`
+  width: 45%;
+`
```
jaahon, line 86: 40 + 10 + 10 + 45 = 105 > 100.

frontend/src/lib-components/layout/Container.tsx
reynders: Excellent idea. To be compatible with the underlying roundicon implementation had to use readableColor in icon color and thus it is now black instead of the UX specified white but according to discussion this is as intended

frontend/src/lib-components/atoms/RoundIcon.tsx
```diff
@@ -138,12 +139,19 @@ const RoundIcon = React.memo(function RoundIcon({
...
+  iconColor
 }: RoundIconProps) {
+  const { colors } = useTheme()
+
   function onKeyDown(e: React.KeyboardEvent) {
     if (onClick && (e.key === ' ' || e.key === 'Enter')) onClick(e)
   }
 
+  const readableIconColor = iconColor
+    ? readableColor(iconColor, colors.grayscale.g0, colors.grayscale.g100)
```
jaahon, line 152: I'm pretty sure you're supposed to use `color` here, so it should be `const readableIconColor = readableColor(color, colors.grayscale.g0, colors.grayscale.g100)`. So `iconColor` prop can just be removed.

PR #2546, opened 2022-06-14
Fix realtime staff attendance queries not including Sundays

jaahon:


jaahon:
Oh yeah, the timezone thing 🙈

PR #2553, opened 2022-06-15
EVAKA-FIX serve attachments inline

hruoho:
#### Summary

Serve all uploaded attachments inline. Must test this in non-local environment for nginx not being configured in our compose stack.

### DIFF
--- a/apigw/src/enduser/routes.ts
+++ b/apigw/src/enduser/routes.ts
@@ -12,8 +12,8 @@ router.all('/citizen/*', createProxy())
 
 router.get('/decisions2/:decisionId/download', proxy)
 
-router.get('/attachments/:applicationId/download', proxy)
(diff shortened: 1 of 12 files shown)


jaahon (commented):
Links to attachments in employee-frontend don't work now since the url has changed. By the way why is the filename added to the path? 

It would also be a good idea to deploy this to test env before merging and verify that everything works there.

hruoho:
> Links to attachments in employee-frontend don't work now since the url has changed. By the way why is the filename added to the path?
> 
> It would also be a good idea to deploy this to test env before merging and verify that everything works there.

Name was added as a means to provide a filename for the user when they decide to download the inline attachment on their machine (instead of the file name being plain UUID without extension). Not clear if this even works.

PR #2554, opened 2022-06-15
EVAKA-4469 Add special support section to vasu

jolle:
#### Summary

- adds a new "label" field for checkbox questions to allow a separate question line and checkbox label
- adds question identifiers and dependencies to show certain questions only when other questions are filled in/selected
- adds an optionally enabled field next to radio buttons for specifying a date range
- adds option to continue numbering from previous question in a follow-up question
- new section in the default vasu for special support

_Pending Swedish translations._


jolle:
Swedish translations will be added later, likely in August.

zch (changes requested):
Code looks good, but the margin at the bottom of sections 5 and 6 is way too small, which makes it look wonky.

image

Note that the focus outlines visually flow outside of the sections. Also visible with the radio button after checking the checkbox in section 6:

image

zch (approved):

PR #2562, opened 2022-06-17
Add strict jsx logical expressions eslint rule

jolle:
#### Summary

Adds the `jsx-expressions/strict-logical-expressions` eslint rule to catch potential incorrect behaviour with logical expressions in JSX like `{0 && …}` and `{"" && …}`. Also fixes existing offenders of this rule in the code.

### DIFF
--- a/frontend/src/citizen-frontend/map/UnitDetailsPanel.tsx
+++ b/frontend/src/citizen-frontend/map/UnitDetailsPanel.tsx
@@ -166,7 +166,7 @@ export default React.memo(function UnitDetailsPanel({
             <div>
               {unit.roundTheClock ? t.map.shiftCareYes : t.map.shiftCareNo}
             </div>
-            {unit.url && (
+            {unit.url !== null && (
               <>
                 <Gap size="s" />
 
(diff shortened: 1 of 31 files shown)


frontend/src/citizen-frontend/map/UnitDetailsPanel.tsx
```diff
@@ -166,7 +166,7 @@ export default React.memo(function UnitDetailsPanel({
             <div>
               {unit.roundTheClock ? t.map.shiftCareYes : t.map.shiftCareNo}
             </div>
-            {unit.url && (
+            {unit.url !== null && (
```
jaahon, line 169: I think the correct fix in a lot of cases (such as here) is to convert these into `!!unit.url` to keep functionality the same. Like here, I don't think it makes sense to have a homepage link that points to nowhere.

frontend/src/employee-frontend/components/common/Filters.tsx
```diff
@@ -112,7 +112,7 @@ export function Filters({
   const { i18n } = useTranslation()
   return (
     <FiltersContainer>
-      {setFreeText && searchPlaceholder && (
+      {setFreeText && !!searchPlaceholder && (
```
jaahon, line 115: Not related to your changes, but this condition just seems wrong. Hiding the whole input if a placeholder is missing or empty 😄 

Could you fix this as well?

jolle: Sure! 😄

jaahon (approved):
Good job!

PR #2564, opened 2022-06-17
Scroll to reservation modal inputs manually after focus on devices with a soft keyboard

jolle:
#### Summary

With some mobile devices, the soft keyboard that opens up when selecting an input in the reservation modal may sometimes overlap with the field. To work around this, whenever a field is focused into, the presence of a soft keyboard is detected with a viewport resize event and after which the input field is scrolled into middle of the view. For the calendar input, the scroll is top-aligned so that it's more likely the day picker will be visible.

### DIFF
--- a/frontend/src/citizen-frontend/calendar/ReservationModal.tsx
+++ b/frontend/src/citizen-frontend/calendar/ReservationModal.tsx
@@ -16,6 +16,7 @@ import {
   TimeRanges,
   validateForm
 } from 'lib-common/reservations'
+import { scrollIntoViewSoftKeyboard } from 'lib-common/utils/scrolling'
 import { SelectionChip } from 'lib-components/atoms/Chip'
 import IconButton from 'lib-components/atoms/buttons/IconButton'
 import Select from 'lib-components/atoms/dropdowns/Select'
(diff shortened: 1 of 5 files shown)


frontend/src/citizen-frontend/calendar/ReservationModal.tsx
```diff
@@ -216,6 +217,9 @@ export default React.memo(function ReservationModal({
             )}
             hideErrorsBeforeTouched={!showAllErrors}
             data-qa="start-date"
+            onFocus={(ev) => {
+              scrollIntoViewSoftKeyboard(ev.target, 'start')
```
jaahon, line 221: Did you test these inputs with `'center'` on android emulator? Did the input still get stuck behind the keyboard? On my iphone, using `'start'` scrolls the view so far up that the background shows if using the simplest daily repetition. And `'center'` scrolls the input far enough above the keyboard on my android.

jolle: `'center'` indeed is enough to get the input field to be visible, but with `'start'` almost all of the actual date picker is visible too (which would be a nice). I can't actually see any difference with `'start'` and `'center'` on my iPhone, but if `'center'` works okay on your device (instead of `'start'`), I'll remove this special case even though the date picker will not be completely visible anymore.

jaahon: I actually didn't think about the date picker itself. `'start'` is probably better then 👍

jaahon (approved):

PR #2574, opened 2022-06-20
Add native date picker

jolle:
#### Summary

For Android, use the native date picker. Also refactors the date picker component to use `LocalDate` instead of raw strings, so that min/max date constraints can be consolidated into this one component to prevent repetition for the same thing. Moreover, the native date picker outputs a browser `Date` which is why it is useful to use `LocalDate` in the component state.

### DIFF
--- a/frontend/src/lib-common/utils/usePendingUserInput.ts
+++ b/frontend/src/lib-common/utils/usePendingUserInput.ts
@@ -7,7 +7,7 @@ import { useEffect, useState } from 'react'
 export function usePendingUserInput<T>(
   transform: (input: string) => T | null,
   initialValue?: string
-) {
+): [string, React.Dispatch<React.SetStateAction<string>>, T | undefined] {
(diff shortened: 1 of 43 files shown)


frontend/src/lib-common/utils/usePendingUserInput.ts
```diff
@@ -0,0 +1,29 @@
...
+    const transformed = transform(input)
+
+    if (transformed) {
+      setValidated(transformed)
+    } else {
+      setValidated(undefined)
+    }
+  }, [input, transform])
+
+  return [input, setInput, validated] as [
```
Gekkio, line 24: Can this type assertion be removed if you add just add a return type to the function?

jolle: Yes – changed. 😄

Gekkio (approved):

PR #2580, opened 2022-06-22
EVAKA-4469 Add status header to the top-right of assistance need decision editor

zch:
image

### DIFF
--- a/frontend/src/employee-frontend/components/child-information/assistance-need/decision/common.tsx
+++ b/frontend/src/employee-frontend/components/child-information/assistance-need/decision/common.tsx
@@ -2,9 +2,16 @@
 //
 // SPDX-License-Identifier: LGPL-2.1-or-later
 
+import React from 'react'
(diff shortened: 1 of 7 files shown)


frontend/src/employee-frontend/components/child-information/assistance-need/decision/common.tsx
```diff
@@ -2,13 +2,65 @@
...
+import { theme } from 'lib-customizations/common'
 
 export const FooterContainer = styled.div`
   display: flex;
   justify-content: flex-end;
   align-items: center;
   padding: ${defaultMargins.s};
 `
+
+export const AssistanceNeedDecisionStatusChip: React.FC<{
```
jolle, line 23: Looking at the codebase of the frontend, `React.FC` is not really used; instead, `const X = React.memo(function X(…) { … })` is. Maybe this component and the one below could use the memo approach as well?

jolle (approved):

PR #2583, opened 2022-06-22
Do not break if there are no reservable days

reynders:


frontend/src/lib-customizations/espoo/citizen/assets/i18n/sv.tsx
```diff
@@ -263,6 +263,7 @@ const sv: Translations = {
       dateRangeInfo: (date: LocalDate) =>
         `Du kan göra bokningar fram till ${date.format()}.`,
       missingDateRange: 'Välj barn att reservera',
+      noReservableDays: 'Inga bokningsbara dagar',
```
jaahon, line 266: The rest of the translations use the word 'reservation' instead of 'bokning' so maybe consider using another translation?

PR #2586, opened 2022-06-23
EVAKA-4469 Autosave assistance need decision forms

zch:
* Extracts the autosave functionality from Vasu and uses this in both when editing Vasus and assistance need decisions.

* When creating a new assistance need decision, the backend will always add the child's guardians if the information is not provided by the caller.


service/src/main/kotlin/fi/espoo/evaka/assistanceneed/decision/AssistanceNeedDecisionController.kt
```diff
@@ -29,13 +32,29 @@ class AssistanceNeedDecisionController(
...
-                tx.insertAssistanceNeedDecision(childId, body.decision)
+                var decision: AssistanceNeedDecision = body.decision
+                if (decision.guardianInfo.isEmpty()) {
+                    val guardianIds = tx.getChildGuardians(childId)
+                    decision = body.decision.copy(
+                        guardianInfo = guardianIds.map { gid ->
+                            val guardian = tx.getPersonById(gid)
+                            AssistanceNeedDecisionGuardian(
+                                personId = gid,
+                                name = "${guardian?.lastName} ${guardian?.firstName}",
```
jolle, line 49: Is it necessary to get the guardian info as it is only used for the name here, and this decision is only used for creating a new one? The name specified here won't be saved to the database anyways, so maybe it is unnecessary to do `tx.getPersonById` overall?

PR #2587, opened 2022-06-23
EVAKA-4469 special support decision preview

jolle:
#### Summary

_Should not be merged before #2586_

- added assistance need decision preview page
- added phone number field for preparers
- changed separate assistance service option date range database columns to one daterange-typed column due to issues with null contents
- modified sticky footer to be more accurate to the specified design: footer is from window's side-to-side, not the parent container
  - I'll make a separate PR with some UI fixes that address this issue properly in other places of use; added bare minimum style fixes to keep same/similar UI in existing components
- made assistance need decision GET route return decision-maker, preparer, and unit information


frontend/src/employee-frontend/components/child-information/assistance-need/decision/AssistanceNeededDecisionForm.tsx
```diff
@@ -68,21 +69,41 @@ const formatName = (someone: {
...
+    EmployeeSelectProps,
+    'phoneLabel' | 'selectedPhone' | 'onChangePhone'
+  > {
+  phoneLabel: string
+  selectedPhone: string
+  onChangePhone: (phone: string) => void
+}
+
+const EmployeeSelectWithTitle: React.FC<
+  EmployeeSelectProps | EmployeeSelectWithPhoneProps
```
zch, line 96: As the `EmployeeSelectWithTitle` operates on a data structure that always looks the same (has a `phoneNumber` field), I'd probably have just extended the props to take this into account and added a `hidePhoneNumberField` (or similar) prop.

In this case the phone number could also be added to the `onChange` function, which could simplify the usage points as well.

I think something like that would be a bit simpler. What do you think?

frontend/src/employee-frontend/components/child-information/assistance-need/decision/AssistanceNeededDecisionForm.tsx
```diff
@@ -480,7 +513,26 @@ export default React.memo(function AssistanceNeedDecisionForm({
             selectedTitle={formState.preparedBy1?.title ?? ''}
             onChange={(employee, title) =>
               setFieldVal({
-                preparedBy1: { employeeId: employee?.id ?? null, title }
+                preparedBy1: {
+                  ...(formState.preparedBy1 ?? {
+                    phoneNumber: null
+                  }),
```
zch, line 519: the employee fields in `formState` should always be defined and have `phoneNumber: null` if they are fetched from the backend due to how the data structure is defined (with `@Nested`). Would `...formState.preparedBy1,` be enough?

frontend/src/employee-frontend/components/child-information/assistance-need/decision/assistance-need-decision-form.ts
```diff
@@ -8,15 +8,37 @@ import {
...
+  'selectedUnit' | 'preparedBy1' | 'preparedBy2' | 'decisionMaker'
+> & {
+  selectedUnit: { id: UUID | null } | null
+  preparedBy1: AssistanceNeedDecisionEmployeeForForm | null
+  preparedBy2: AssistanceNeedDecisionEmployeeForForm | null
+  decisionMaker: Omit<
+    AssistanceNeedDecisionEmployeeForForm,
+    'phoneNumber'
+  > | null
+}
```
zch, line 36: I think the `AssistanceNeedDecisionRequest` type should be updated in the controller and the generated result used in the frontend. Right now there's a bit of type confusion on the server as what the `GET` call returns is not what the `POST` or `PUT` calls use.

frontend/src/employee-frontend/components/child-information/assistance-need/decision/api.ts
```diff
@@ -11,6 +11,8 @@ import { UUID } from 'lib-common/types'
 
 import { client } from '../../../../api/client'
 
+import { AssistanceNeedDecisionForm } from './assistance-need-decision-form'
```
zch, line 14: I think this should be generated from the server-side types. See my later comment 👇

service/src/main/kotlin/fi/espoo/evaka/assistanceneed/decision/AssistanceNeedDecision.kt
```diff
@@ -64,8 +66,12 @@ enum class AssistanceNeedDecisionLanguage {
 }
 
 data class AssistanceNeedDecisionEmployee(
+    @PropagateNull
     val employeeId: EmployeeId?,
-    val title: String?
+    val title: String?,
+    val name: String? = null,
+    val email: String? = null,
+    val phoneNumber: String?
```
zch, line 74: Defaulting to `null` would make sense for the `decisionMaker`, or maybe making a separate type for the decision maker as this person actually does not have a phone number.

zch (changes requested):

zch (approved):

PR #2597, opened 2022-06-27
EVAKA-4469 child assistance need decision list

jolle:
#### Summary

Adds a list of assistance need decisions to the child page. From the list decisions can be opened (previewed) and edited/deleted if the decision is a draft or returned for edits.


frontend/src/e2e-test/pages/employee/child-information.ts
```diff
@@ -569,6 +569,36 @@ export class AssistanceNeedSection extends Section {
...
     await waitUntilEqual(() => this.#assistanceNeedRow.count(), expectedCount)
   }
+
+  async waitUntilAssistanceNeedDecisionsLoaded() {
+    await this.page
+      .findByDataQa('table-of-assistance-need-decisions')
+      .waitUntilVisible()
+  }
+
+  assistanceNeedDecisions(nth: number) {
```
Gekkio, line 579: This pattern of returning an object containing promises is a bit problematic, because the calling side has no real control of *when* the information is fetched, and the promises in the returned object must not be used more than once (unless you really want the same cached value multiple times).

I can think of two better choices:

1. await each of the fields in this function, so the returned object is a simple data object
2. return a wrapper class instead, and make getter functions for each of the fields so that each call to a getter returns a new promise

Gekkio (approved):

frontend/src/employee-frontend/components/child-information/Assistance.tsx
```diff
@@ -41,6 +43,12 @@ export default React.memo(function Assistance({ id, startOpen }: Props) {
         {permittedActions.has('READ_ASSISTANCE_ACTION') && (
           <AssistanceAction id={id} />
         )}
+        {permittedActions.has('READ_ASSISTANCE_NEED_DECISIONS') && (
```
zch, line 46: Should also be hidden behind the `specialNeedsDecisions` feature flag (which should probably be renamed to `assistanceNeedDecisions`)

zch (changes requested):

zch (approved):

PR #2609, opened 2022-06-30
EVAKA-4469 assistance need decision decision-maker report

jolle:
#### Summary

- Adds a report for sent assistance need decisions
- Adds unopened decision count indicator to the header and report list view, if there are any
- Adds separate read-only decision viewing page for decision-makers (accessible from the report)
- Adds ability to decide decision as "needs work" for decision-makers


frontend/src/employee-frontend/components/reports/AssistanceNeedDecisionsReport.tsx
```diff
@@ -0,0 +1,216 @@
...
+  const { i18n } = useTranslation()
+  const [report, setReport] = useState<
+    Result<AssistanceNeedDecisionsReportRow[]>
+  >(Loading.of())
+  const [careAreaFilter, setCareAreaFilter] = useState<string>()
+
+  useEffect(() => {
+    setReport(Loading.of())
+    void getAssistanceNeedDecisionsReport().then(setReport)
+  }, [])
```
zch, line 66: Is there a reason for not using `useApiState` here?

frontend/src/employee-frontend/components/reports/AssistanceNeedDecisionsReport.tsx
```diff
@@ -0,0 +1,216 @@
...
+      <ContentArea opaque>
+        <Title size={1}>{i18n.reports.assistanceNeedDecisions.title}</Title>
+        <FilterRow>
+          <FilterLabel>{i18n.reports.common.careAreaName}</FilterLabel>
+          <Wrapper>
+            <Combobox
+              items={[
+                { value: '', label: i18n.common.all },
+                ...report
+                  .map((rs) =>
```
zch, line 114: This seems like a minor thing, but I think it's important to avoid very short variable/parameter names. The reason is that the reader of the code is forced to stop, go back a bit, inspect the type of `report` and `rs` in this case and then map that knowledge onto what the result of this code will be.

As code is written once and read like a million times, it makes sense to optimize for readability.

frontend/src/employee-frontend/components/reports/AssistanceNeedDecisionsReportDecision.tsx
```diff
@@ -0,0 +1,200 @@
...
+                {t.leavePage}
+              </Button>
+            </FixedSpaceRow>
+            {renderResult(
+              assistanceNeedDecision,
+              ({ permittedActions }) =>
+                !permittedActions.includes('DECIDE') ? null : (
+                  <FixedSpaceRow spacing="m">
+                    <DangerAsyncButton
+                      text="Hylkää päätös"
```
zch, line 142: Please extract the string

frontend/src/employee-frontend/components/reports/AssistanceNeedDecisionsReportDecision.tsx
```diff
@@ -0,0 +1,200 @@
...
+                      onClick={() =>
+                        Promise.resolve(Failure.of({ message: '' }))
+                      }
+                      onSuccess={() => reloadDecision()}
+                    />
+                    <AsyncButton
+                      text="Palauta korjattavaksi"
```
zch, line 149: Please extract the string

frontend/src/employee-frontend/components/reports/AssistanceNeedDecisionsReportDecision.tsx
```diff
@@ -0,0 +1,200 @@
...
+                      onClick={() => setReturnForEdit(true)}
+                      onSuccess={() => reloadDecision()}
+                      data-qa="return-for-edit"
+                    />
+                    <AsyncButton
+                      primary
+                      text="Hyväksy päätös"
```
zch, line 156: Please extract the string

zch (changes requested):

frontend/src/employee-frontend/components/child-information/assistance-need/decision/AssistanceNeedDecisionPage.tsx
```diff
@@ -132,19 +59,31 @@ export default React.memo(function AssistanceNeedDecisionPage() {
...
+          ))}
+        </ContentArea>
+        <Gap size="m" />
+        <CollapsibleContentArea
+          title={<H2 noMargin>{t.appealInstructionsTitle}</H2>}
+          open={appealInstructionsOpen}
+          toggleOpen={() => setAppealInstructionsOpen(!appealInstructionsOpen)}
+          opaque
+        >
+          {t.appealInstructions}
```
zch, line 84: This should also be within the `<I18nContext.Provider>` in order to get the Swedish version once available.

frontend/src/employee-frontend/components/reports/AssistanceNeedDecisionsReportDecision.tsx
```diff
@@ -0,0 +1,205 @@
...
+          decisionId={id}
+        />
+      )}
+
+      <Content>
+        <ReturnButton label={i18n.common.goBack} />
+
+        <ContentArea opaque>
+          {renderResult(assistanceNeedDecision, ({ decision }) => (
+            <AssistanceNeedDecisionReadOnly decision={decision} />
```
zch, line 112: This, along with the appeal instructions below, should also be wrapped in an `I18nContext.Provider` using the language of the decision.

zch (changes requested):

zch (approved):

PR #2611, opened 2022-06-30
PoC of set-returning functions for handling placements and backup care

Gekkio:


hruoho (approved):
This is such a delight, thank you 🙏

PR #2614, opened 2022-06-30
Show backupcare groups as citizen message receivers

reynders:

### DIFF
--- a/service/src/main/kotlin/fi/espoo/evaka/messaging/MessageQueries.kt
+++ b/service/src/main/kotlin/fi/espoo/evaka/messaging/MessageQueries.kt
@@ -462,7 +462,7 @@ personal_accounts AS (
 ),
 group_accounts AS (
     SELECT acc.id, g.name, 'GROUP' AS type
-    FROM relevant_placements p
+    FROM placements p
     JOIN daycare_group_placement dgp ON dgp.daycare_placement_id = p.id AND :today BETWEEN dgp.start_date AND dgp.end_date
     JOIN daycare_group g ON g.id = dgp.daycare_group_id
     JOIN message_account acc on g.id = acc.daycare_group_id


service/src/main/kotlin/fi/espoo/evaka/messaging/MessageQueries.kt
```diff
@@ -370,8 +370,40 @@ fun Database.Read.getMessage(id: MessageId): Message {
...
+        AND b.blocked_recipient = g.guardian_id
+    )
+    AND EXISTS (
+        SELECT 1 FROM daycare u
+        WHERE p.unit_id = u.id AND 'MESSAGING' = ANY(u.enabled_pilot_features)
+    )
+
+    UNION
+
+    SELECT p.id, p.unit_id, p.child_id, null
```
Gekkio, line 390: Why is group_id null here but not null in the other backup care case above?

service/src/main/kotlin/fi/espoo/evaka/messaging/MessageQueries.kt
```diff
@@ -417,6 +466,13 @@ group_accounts AS (
     JOIN daycare_group_placement dgp ON dgp.daycare_placement_id = p.id AND :today BETWEEN dgp.start_date AND dgp.end_date
```
Gekkio, line 466: I think this should use `FROM placements p` instead of `relevant_placements`, because we're joining to daycare_group_placement and backup_care rows should not be used

frontend/src/e2e-test/specs/7_messaging/messaging.spec.ts
```diff
@@ -167,6 +191,36 @@ describe('Sending and receiving messages', () => {
...
 
+      test('Citizen sends a message to backup care child', async () => {
+        await insertBackupCareFixtures([
+          {
+            id: uuidv4(),
+            childId,
+            unitId: backupDaycareId,
+            groupId: backupGroupFixtureId,
+            period: {
+              start: LocalDate.todayInSystemTz().formatIso(),
```
Gekkio, line 202: No mocked time = :cry:

hruoho: The problem lies deeper since `LocalDate.todayInSystemTz()` is used in so many places in our fixtures & tests 😢

PR #2615, opened 2022-06-30
EVAKA-4469 Implement changing language of assistance need decisions

zch:


frontend/src/lib-customizations/espoo/employee/assets/i18n/sv.ts
```diff
@@ -2,4 +2,97 @@
 //
 // SPDX-License-Identifier: LGPL-2.1-or-later
 
-export const sv = {}
+import { fi } from './fi'
+
+export const sv = {
```
jolle, line 7: Could this have the type annotation `Translations` to ensure all translations exist here that also exist in the fi-texts?

zch: We could, but the type is already checked and enforced by [frontend/src/lib-customizations/employee.tsx:66-69](https://github.com/espoon-voltti/evaka/pull/2615/files#diff-72cfcd33c8e650c94272539cbdef6dfb8c4851946b17f93c11eb513fedc2653eR66-R69)

jolle (approved):

PR #2623, opened 2022-07-04
EVAKA-4469 Vasu updates and changes

jolle:
#### Summary

- Updated the vasu default template to match the new guidelines
- Made the first section a dynamic section so that it can be customized in the template editor
  - Backwards compatible: added a flag for newly created templates to indicate a dynamic first section so that flagless (i.e., old templates) still have the static basic info section
- Added intervention radio group options, which are not real options but text guides in between options
- Overhauled follow-up questions: the editor is an infinitely growing list of entries, which contains a configurable entry date and text
  - Removed special handling of follow-up entries and now they're sent and modified using the regular vasu modification handler with some special handling
  - Employee names for follow-up entries (created/edited) are not saved in the document JSON, instead only the ID is and the names are filled in when the document is retrieved
  - Employee IDs (and more importantly, names) are always redacted in citizen's version of the vasu
- Changed the color of the draft vasu status chip

### DIFF
--- a/frontend/src/employee-frontend/components/vasu/components/FollowupQuestion.tsx
+++ b/frontend/src/employee-frontend/components/vasu/components/FollowupQuestion.tsx
@@ -2,183 +2,139 @@
 //
 // SPDX-License-Identifier: LGPL-2.1-or-later
 
-import React, { useState, useContext, useCallback } from 'react'
+import React, { useContext, useEffect, useMemo, useState } from 'react'
 import styled, { css } from 'styled-components'
-import { v4 as uuid } from 'uuid'
 
 import { useTranslation } from 'employee-frontend/state/i18n'
 import { Followup, FollowupEntry } from 'lib-common/api-types/vasu'
-import { PermittedFollowupActions } from 'lib-common/api-types/vasu'
+import HelsinkiDateTime from 'lib-common/helsinki-date-time'
 import LocalDate from 'lib-common/local-date'
-import Button from 'lib-components/atoms/buttons/Button'
-import IconButton from 'lib-components/atoms/buttons/IconButton'
 import TextArea from 'lib-components/atoms/form/TextArea'
 import { FixedSpaceRow } from 'lib-components/layout/flex-helpers'
-import { Dimmed, H2, Label } from 'lib-components/typography'
-import { defaultMargins } from 'lib-components/white-space'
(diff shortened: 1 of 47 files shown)


hruoho (commented):
Reviewed this together on the level of UI changes and all seemed good 👍

service/src/main/kotlin/fi/espoo/evaka/vasu/VasuController.kt
```diff
@@ -88,32 +87,59 @@ class VasuController(
...
+                                                    }
+                                                )
+                                            }
+                                        )
+                                    } else {
+                                        question
+                                    }
+                                }
+                            )
+                        }
```
hruoho, line 141: I think it wouldn't hurt to use named functions or variables here to add more semantics for better readability

service/src/main/kotlin/fi/espoo/evaka/vasu/VasuController.kt
```diff
@@ -135,7 +161,51 @@ class VasuController(
...
+                                                            } == true
+                                                            ) || (entry.text == storedEntry.text && entry.date.isEqual(storedEntry.date))
+                                                    )
+                                                        storedEntry?.edited
+                                                    else
+                                                        FollowupEntryEditDetails(
+                                                            editedAt = LocalDate.now(),
+                                                            editorId = EmployeeId(user.rawId()),
+                                                            editorName = null
+                                                        ),
```
hruoho, line 193: Same here: I think it wouldn't hurt to use named functions or variables here to add more semantics for better readability

frontend/src/employee-frontend/components/vasu/components/FollowupQuestion.tsx
```diff
@@ -193,24 +149,60 @@ export default React.memo(function FollowupQuestion({
...
+                              authorId: user?.id,
+                              authorName: user?.name,
+                              createdDate: HelsinkiDateTime.now()
+                            }
+                          : undefined
+                        : entry.entry
+                    )
+                    .filter((entry): entry is FollowupEntry => !!entry)
+                )
+              }
```
hruoho, line 195: I think it wouldn't hurt to use named functions or variables here to add more semantics for better readability

jaahon (approved):

PR #2630, opened 2022-07-06
Fix autosave force-saving for cancellations

jolle:
#### Summary

The autosave force-saving feature did not properly account for cancelled requests. Cancellations happened often in the e2e tests (but not always) in the assistance need decision edit page because the debounce happened to be almost the same time as it takes to click the buttons, thus cancelling the button-clicking request and preventing navigation. This made two assistance need decision e2e tests flaky.


hruoho (approved):
Went this through with the author, all seems good!

PR #2631, opened 2022-07-06
EVAKA-4469 Require unit field for assistance need decisions

jolle:
#### Summary

The assistance need decisions decision-maker report does not display rows that do not have a unit associated with the decision due to a JOIN in the query; as such, the unit is a required field. There are other required fields (from a non-technical aspect) as well, but those will be decided on later. Moreover, support for the validation field info was only added to the unit dropdown and some other easy cases; when other required fields are decided on the rest of the field infos can be implemented in the more complex cases.

### DIFF
--- a/frontend/src/employee-frontend/components/child-information/assistance-need/decision/AssistanceNeedDecisionEditPage.tsx
+++ b/frontend/src/employee-frontend/components/child-information/assistance-need/decision/AssistanceNeedDecisionEditPage.tsx
@@ -2,13 +2,14 @@
 //
 // SPDX-License-Identifier: LGPL-2.1-or-later
 
-import React, { useCallback, useEffect, useMemo } from 'react'
+import React, { useCallback, useEffect, useMemo, useState } from 'react'
 import { useNavigate } from 'react-router-dom'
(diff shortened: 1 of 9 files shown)


frontend/src/employee-frontend/components/child-information/assistance-need/decision/AssistanceNeedDecisionEditPage.tsx
```diff
@@ -26,17 +27,26 @@ import {
...
 import { getPerson } from '../../../../api/person'
 
-import AssistanceNeededDecisionForm from './AssistanceNeededDecisionForm'
+import AssistanceNeededDecisionForm, {
+  FieldInfos
+} from './AssistanceNeededDecisionForm'
 import { useAssistanceNeedDecision } from './assistance-need-decision-form'
 import { FooterContainer, DecisionInfoHeader } from './common'
 
+const requiredFormFields = ['selectedUnit' as const]
```
jaahon, line 44: You should mark the whole array with `as const` so that every additional element doesn't have to be marked with `as const` as well.

jaahon (approved):

PR #2636, opened 2022-07-12
EVAKA-4555 updated reservation modal

jolle:
#### Summary

**#2634 should be merged first.**

- Updated style for the mobile reservation modal (sticky footer, background color)
- Added automatic filling of reservation time inputs if existing reservations overlap among the existing children during the specified date range
- Added ability to mark absences for days in the reservation modal
- Changed Finnish wording for "reservations"
- Removed unnecessary checkboxes in the week input


frontend/src/citizen-frontend/calendar/reservation-modal/utils.ts
```diff
@@ -0,0 +1,76 @@
...
+  TimeRange
+} from 'lib-common/generated/api-types/reservations'
+import { TimeRanges } from 'lib-common/reservations'
+
+export const emptyTimeRange = [
+  {
+    startTime: '',
+    endTime: ''
+  }
+] as [TimeRange]
```
jaahon, line 18: You shouldn't use `as [TimeRange]` because it basically turns off type checking. If you want `emptyTimeRange` to be some other type other than what was inferred, you can use a type annotation: `export const emptyTimeRange: [TimeRange] = ...`.

frontend/src/citizen-frontend/calendar/reservation-modal/RepetitionTimeInputGrid.tsx
```diff
@@ -0,0 +1,69 @@
...
+      )
+    case 'IRREGULAR':
+      return (
+        <TimeInputGrid>
+          <IrregularRepetitionTimeInputGrid {...props} />
+        </TimeInputGrid>
+      )
+    default:
+      assertNever(repetition)
+      return null
```
jaahon, line 58: I don't see the value of this default block. Typescript can already check without it that all cases are handled.

jaahon (approved):

PR #2644, opened 2022-07-18
Suppress CVE-2022-31569

jolle:
#### Summary

Does not affect evaka; misidentification due to a too-eager CPE. See: https://github.com/jeremylong/DependencyCheck/issues/4671, https://github.com/advisories/GHSA-g7h7-rg53-h2g9, https://github.com/github/securitylab/issues/669#issuecomment-1186015526


nagavem:
Itss merged but it still fails to work for my projects
not sure why, would appreciate any assistance

PR #2646, opened 2022-07-18
EVAKA-4287 Add daily service times notification for citizens

jolle:
#### Summary

- Refactored calendar notification system to support multiple/different kinds of notifications
- Added notifications for guardians whenever daily service times are changed or added
  - A modal is shown if previous reservations were removed due to changes to the daily service times instead of a toast


service/src/main/resources/db/migration/V250__daily_service_times_notification.sql
```diff
@@ -0,0 +1,9 @@
+CREATE TABLE daily_service_time_notification (
+  id uuid PRIMARY KEY NOT NULL DEFAULT ext.uuid_generate_v1mc(),
+  guardian_id uuid NOT NULL REFERENCES person (id) ON DELETE CASCADE,
+  daily_service_time_id uuid NOT NULL REFERENCES daily_service_time (id) ON DELETE CASCADE,
```
Gekkio, line 4: Consider adding an index for this column. Without an index, deleting a row from `daily_service_time` triggers a full table scan of all `daily_service_time_notification` rows in order to cascade the delete.

Gekkio (approved):

PR #2649, opened 2022-07-20
EVAKA-4407 employee staff attendance desktop editor

jolle:
#### Summary

- Add browsing of days in the staff attendance details modal
- Add summary of the current day in the staff attendance details modal
- Show a warning if there is a gap between attendances  in the staff attendance details modal
- Fix details modal button to only be visible on days with other attendances
- Add a sum row of staff to the table (TBD: only excluding training and other work (current implementation) or only including present)

### DIFF
--- a/frontend/src/employee-frontend/components/unit/unit-reservations/StaffAttendanceTable.tsx
+++ b/frontend/src/employee-frontend/components/unit/unit-reservations/StaffAttendanceTable.tsx
@@ -5,6 +5,8 @@
 import classNames from 'classnames'
 import groupBy from 'lodash/groupBy'
 import isEqual from 'lodash/isEqual'
+import maxBy from 'lodash/maxBy'
+import minBy from 'lodash/minBy'
 import partition from 'lodash/partition'
 import sortBy from 'lodash/sortBy'
 import React, {
@@ -16,22 +18,29 @@ import React, {
 } from 'react'
 import styled from 'styled-components'
 
-import { Result } from 'lib-common/api'
-import { OperationalDay } from 'lib-common/api-types/reservations'
+import { getStaffAttendances } from 'employee-frontend/api/staff-attendance'
+import { getUnitAttendanceReservations } from 'employee-frontend/api/unit'
+import { Loading, Result } from 'lib-common/api'
+import {
+  OperationalDay,
+  UnitAttendanceReservations
+} from 'lib-common/api-types/reservations'
 import DateRange from 'lib-common/date-range'
 import FiniteDateRange from 'lib-common/finite-date-range'
 import {
   Attendance,
   EmployeeAttendance,
   ExternalAttendance,
-  UpsertStaffAndExternalAttendanceRequest,
-  UpsertStaffAttendance
+  UpsertStaffAttendance,
+  StaffAttendanceResponse,
+  UpsertStaffAndExternalAttendanceRequest
 } from 'lib-common/generated/api-types/attendance'
 import { DaycareGroup } from 'lib-common/generated/api-types/daycare'
 import HelsinkiDateTime from 'lib-common/helsinki-date-time'
 import LocalDate from 'lib-common/local-date'
 import LocalTime from 'lib-common/local-time'
 import { UUID } from 'lib-common/types'
+import { useRestApi } from 'lib-common/utils/useRestApi'
 import RoundIcon from 'lib-components/atoms/RoundIcon'
 import Tooltip from 'lib-components/atoms/Tooltip'
 import IconButton from 'lib-components/atoms/buttons/IconButton'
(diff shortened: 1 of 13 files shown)


service/src/main/kotlin/fi/espoo/evaka/attendance/StaffAttendanceQueries.kt
```diff
@@ -406,6 +407,21 @@ WHERE employee_id = :employeeId AND tstzrange(start_time, end_time) && tstzrange
         .mapTo<PlannedStaffAttendance>()
         .toList()
 
+fun Database.Read.getPlannedStaffAttendanceForDays(
```
Gekkio, line 410: It seems like this function should actually return `Map<EmployeeId, List<PlannedStaffAttendance>>`, since we're using the `PlannedStaffAttendanceWithEmployeeId` type just to filter multiple rows by employeeId.

Something like this might work, and you can then remove `PlannedStaffAttendanceWithEmployeeId` and just use map accesses instead of manual filtering:

```kotlin
.map { row -> Pair(row.mapColumn<EmployeeId>("employee_id"), row.mapRow<PlannedStaffAttendance>()) }
.groupBy({ it.first }, { it.second })
```

JDBI also has reduce functions for this kind of stuff, but IMHO they are a bit clunky and this is more readable

service/src/main/kotlin/fi/espoo/evaka/attendance/StaffAttendanceQueries.kt
```diff
@@ -406,6 +407,21 @@ WHERE employee_id = :employeeId AND tstzrange(start_time, end_time) && tstzrange
...
+    employeeIds: List<EmployeeId>,
+    days: List<LocalDate>
+): List<PlannedStaffAttendanceWithEmployeeId> =
+    createQuery(
+        """
+SELECT start_time AS start, end_time AS end, type, employee_id FROM staff_attendance_plan
+WHERE employee_id = ANY(:employeeIds) AND (date(start_time) = ANY(:days) OR date(end_time) = ANY(:days))
```
Gekkio, line 417: This query can't use an index for the time part (= database fetches all plans for employees regardless of time, and filters one-by-one), but there's a fix:

Take a FiniteDateRange instead of a list of arbitrary dates, convert it to two HelsinkiDateTimes, and construct a tstzrange:

```kotlin
.bind("startTime", HelsinkiDateTime.of(range.start, LocalTime.MIDNIGHT))
.bind("endTime", HelsinkiDateTime.of(range.end.plusDays(1), LocalTime.MIDNIGHT))
```

and

```sql
WHERE employee_id = ANY(:employeeIds) AND (tstzrange(start_time, end_time) && tstzrange(:start_time, :end_time))
```

PS. You can also accept `Collection<EmployeeId>` so you can pass `attendancesByEmployee.keys` directly in the call site without the extra `.toList()` call

frontend/src/employee-frontend/components/unit/unit-reservations/StaffAttendanceTable.tsx
```diff
@@ -425,8 +633,8 @@ const AttendanceRow = React.memo(function AttendanceRow({
   const updateValue = useCallback(
     (date: LocalDate, rangeIx: number, updatedRange: TimeRange) => {
       setValues(
-        values.map((val) => {
-          return val.date === date
+        values.map((val) =>
+          val.date === date
```
Gekkio, line 637: Is this object instance equal check intentional?

frontend/src/employee-frontend/components/unit/unit-reservations/StaffAttendanceTable.tsx
```diff
@@ -119,6 +132,130 @@ export default React.memo(function StaffAttendanceTable({
...
+    ]
+  )
+
+  const personCountSums = useMemo(
+    () =>
+      mapValues(
+        groupBy(
+          staffRows
+            .flatMap(({ attendances, employeeId }) =>
+              uniq(
```
Gekkio, line 215: I'm not sure `uniq` works here as intended...every element coming from the flatMap below is going to be a different LocalDate instance, so uniq will never deduplicate anything.

frontend/src/employee-frontend/components/unit/unit-reservations/StaffAttendanceTable.tsx
```diff
@@ -119,6 +132,130 @@ export default React.memo(function StaffAttendanceTable({
...
+            .concat(
+              extraAttendances
+                .filter(
+                  ({ type, groupId }) =>
+                    type !== 'OTHER_WORK' &&
+                    type !== 'TRAINING' &&
+                    groupFilter(groupId)
+                )
+                .flatMap(({ departed, arrived, name }) =>
+                  uniq(
```
Gekkio, line 242: Same `uniq` problem here

frontend/src/employee-frontend/components/unit/unit-reservations/StaffAttendanceTable.tsx
jolle: I didn't make this check (just removed the redundant braces), but I guess it might be? It's called from the editor component's update callback, which has the exact reference to the date that needs to be updated, so it'll find the exact entry that was requested to be updated.

service/src/main/kotlin/fi/espoo/evaka/attendance/StaffAttendanceQueries.kt
jolle: This makes more sense than my original query too as start_time and end_time forms a range, so if the plan was over 2 days long the query would fail to find the plans in the middle, whereas by using tstzrange the overlap is checked properly. Thanks!

Gekkio (approved):

PR #2657, opened 2022-07-21
EVAKA-4544 Improve unit features page

jolle:
#### Summary

- Add translations for the features
- Show provider type below unit name
- Add filter for specific provider types
- Allow selecting/unselecting all features (of filtered units)
- Add undo with 1-deep history for accidental uses of "select/unselect all"
- Make table scrollable, and the heading sticky

### DIFF
--- a/service/src/main/kotlin/fi/espoo/evaka/daycare/controllers/DaycareController.kt
+++ b/service/src/main/kotlin/fi/espoo/evaka/daycare/controllers/DaycareController.kt
@@ -9,6 +9,7 @@ import fi.espoo.evaka.daycare.CaretakerAmount
 import fi.espoo.evaka.daycare.Daycare
 import fi.espoo.evaka.daycare.DaycareFields
 import fi.espoo.evaka.daycare.UnitFeatures
+import fi.espoo.evaka.daycare.addUnitFeatures
 import fi.espoo.evaka.daycare.createDaycare
 import fi.espoo.evaka.daycare.deleteCaretakers
 import fi.espoo.evaka.daycare.getCaretakers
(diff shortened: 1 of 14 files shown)


service/src/main/kotlin/fi/espoo/evaka/daycare/controllers/DaycareController.kt
```diff
@@ -72,16 +73,24 @@ class DaycareController(
...
         user: AuthenticatedUser,
-        @PathVariable daycareId: DaycareId,
-        @RequestBody features: Set<PilotFeature>
+        @RequestBody request: UpdateFeaturesRequest
     ) {
-        Audit.UnitFeaturesUpdate.log(targetId = daycareId)
-        accessControl.requirePermissionFor(user, Action.Unit.UPDATE_FEATURES, daycareId)
-        db.connect { dbc -> dbc.transaction { it.setUnitFeatures(daycareId, features) } }
+        Audit.UnitFeaturesEnable.log(targetId = request.unitIds)
+        accessControl.requirePermissionFor(user, Action.Unit.ENABLE_FEATURES, request.unitIds)
```
Gekkio, line 83: This should probably use `Action.Unit.DISABLE_FEATURES` in some cases

jolle: I've reverted the naming changes to logging and to the action. The renaming was originally done for separating the enabling and disabling to separate routes, but I didn't end up using that solution; as such, the renaming is no longer relevant.

Gekkio (approved):

PR #2668, opened 2022-07-27
EVAKA-4469 citizen unread assistance need decisions

jolle:
#### Summary

- Adds an unread assistance need decision count indicator to the header/navigation, the list of children page, the child page (collapsible's header), and in the table itself
- Marks decisions read when they are opened

### DIFF
--- a/frontend/src/lib-customizations/espoo/citizen/assets/i18n/sv.tsx
+++ b/frontend/src/lib-customizations/espoo/citizen/assets/i18n/sv.tsx
@@ -1933,6 +1933,7 @@ const sv: Translations = {
     },
     assistanceNeed: {
       title: 'Behov av stöd',
+      unreadCount: 'olästa',
       decisions: {
         title: 'Beslut om behov av stöd',
         form: 'Form',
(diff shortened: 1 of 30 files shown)


service/src/main/resources/db/migration/V251__assistance_need_decision_read.sql
```diff
@@ -0,0 +1,2 @@
+ALTER TABLE assistance_need_decision
+  ADD COLUMN read_guardian_ids uuid[] DEFAULT array[]::uuid[];
```
Gekkio, line 2: I think this data model might have some issues when guardian information changes...if somebody becomes a guardian for a child, won't they get "unread decision" notifications of all the child's decisions since the beginning of time, since the new guardian is not in this list in old decisions?

My quick 1 minute hunch is that this would work better using the opposite semantics: have a list of *unread* guardian ids, and write that during some decision status change event (decision created, or when status changes to accepted/rejected?). Reading the notification should then remove the guardian id from the list, and new guardians won't suddenly get notifications for old stuff because the list is essentially a "snapshot" of guardians at one point of time.

jolle: I had the opposite thinking that when guardians are changed, the new guardians would see the old decisions as unread, but I do agree that in practice it's probably not useful to see very old decisions as new notifications. I'll switch the logic around as you mentioned 👍

frontend/src/lib-customizations/espoo/citizen/assets/i18n/sv.tsx
```diff
@@ -1933,6 +1933,7 @@ const sv: Translations = {
     },
     assistanceNeed: {
       title: 'Behov av stöd',
+      unreadCount: 'oläst',
```
zch, line 1936:
```suggestion
      unreadCount: 'olästa',
```

Gekkio (approved):

PR #2685, opened 2022-08-01
Remove SSNs and children for weak auth citizens

jolle:
#### Summary
The auth status endpoint always returned social security numbers and children even when weakly authenticated.

### DIFF
--- a/apigw/src/enduser/routes/auth-status.ts
+++ b/apigw/src/enduser/routes/auth-status.ts
@@ -9,7 +9,6 @@ import { appCommit } from '../../shared/config'
 export default toRequestHandler(async (req, res) => {
   if (req.user && req.user.id) {
     const data = await getUserDetails(req, req.user.id)
-    data.userType = req.user.userType
     res.status(200).send({ loggedIn: true, user: data, apiVersion: appCommit })
   } else {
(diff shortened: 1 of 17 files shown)


zch (commented):
Looks good to me, but I think @Gekkio should have a look as well to be sure.

Gekkio (commented):
Two minor comments:

We already have `CitizenAuthLevel` for the strong vs weak distinction, so I recommend using this existing terminology (`type` -> `authLevel`) and the enum type.

I'm not sure how useful `Action.Citizen.Person.READ_STRONG_VTJ_DETAILS` is (vs just checking the auth level in the controller), because the name is not a good indicator of what it really does. If that action is permitted for weak auth level in customizations, it doesn't just send all VTJ details to the frontend as the name would indicate, but *also makes the frontend think the user has strong auth level*. This means all the lock icons disappear, but things don't work because the session still has weak auth level even though `isStrong()` in the frontend returns true. This happens because we're mixing "do we have all VTJ data" and "what is the auth level" into one concept: the frontend `User` type hard-codes this rule so that a strong user has all fields and a weak one doesn't. Either we should separate these things, *or* make sure the model is consistent by hard-coding this rule in the backend too (= check auth level instead of using the permitted actions mechanism).

Gekkio (approved):

PR #2690, opened 2022-08-02
Add unit occupancy chart tooltip rounding

jolle:
#### Summary

- Fix an exception being thrown if `staffData` is a zero-length array
- Round occupancy chart tooltip numbers to max 1 decimal

### DIFF
--- a/frontend/src/employee-frontend/components/unit/tab-unit-information/occupancy/OccupancyDayGraph.tsx
+++ b/frontend/src/employee-frontend/components/unit/tab-unit-information/occupancy/OccupancyDayGraph.tsx
@@ -116,19 +116,19 @@ const Graph = React.memo(function Graph({ occupancy, shiftCareUnit }: Props) {
               </tr>
               <tr>
                 <td>{i18n.unit.occupancy.realtime.children}</td>
-                <td>{child}</td>
+                <td>{parseFloat(child.toFixed(1))}</td>
               </tr>
               <tr>
                 <td>{i18n.unit.occupancy.realtime.childrenPresent}</td>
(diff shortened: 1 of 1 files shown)


frontend/src/employee-frontend/components/unit/tab-unit-information/occupancy/OccupancyDayGraph.tsx
```diff
@@ -116,19 +116,19 @@ const Graph = React.memo(function Graph({ occupancy, shiftCareUnit }: Props) {
               </tr>
               <tr>
                 <td>{i18n.unit.occupancy.realtime.children}</td>
-                <td>{child}</td>
+                <td>{parseFloat(child.toFixed(1))}</td>
```
zch, line 119: is the `parseFloat` really needed here? If I understand this correctly it would format the number to a string representing the number with 1 decimal point, then parse it back into a float, which is then converted back to a string where we in some cases _might_ hit a case where the float is formatted using excessive decimals again.

zch (changes requested):

frontend/src/employee-frontend/components/unit/tab-unit-information/occupancy/OccupancyDayGraph.tsx
jolle: `Number.toFixed(…)` always adds decimals (`(2).toFixed(1) === '2.0'`), but we don't want the unnecessary `.0`; `parseFloat((2).toFixed(1)) == 2`  so the rendered number doesn't have the unnecessary decimal.

zch (approved):

PR #2693, opened 2022-08-03
Wrap info button with inline text

jolle:
#### Summary

Especially apparent on mobile devices, the info button was always positioned to the right of wrapped text instead of directly next to the last character of inline text.


jolle:
Consolidated onto #2697

PR #2694, opened 2022-08-03
Improved JDBI reflection implementation

Gekkio:
#### Summary

Instead of passing a Java `Class<T>` to JDBI in various functions, use Kotlin `typeOf<T>()` to extract accurate type information, and pass to JDBI a Java `Type`/`QualifiedType<T>` which contains as accurate information as possible about the type. This greatly reduces the need for different types of functions which were only added to handle different edge cases.

- want to bind something? `.bind("name", value)`. No need to convert lists/sets/etc to array, or use a different function if the value is nullable
- want to map a column? `.mapColumn<Type>("name")`. No need to map to array first and convert manually if you want a list
- want to map result rows? `.mapTo<Type>()`. Supports lists/sets/etc too

JDBI supports all this, but we just never gave it accurate information before


zch (approved):
Looks great!

PR #2698, opened 2022-08-04
Update LEOPS and vasu

jolle:
#### Summary
- Make disclaimer about document sharing to other parties more prominent by adding it to the page of list of documents, and by creating a separate paragraph for the info box in the document itself
- Fix sticky footer to be the full width of the page
- Make child support level to be a multi select with max 2 selections instead of a radio (for LEOPS)


zch (approved):
Technically looks good and seems to work fine

PR #2705, opened 2022-08-04
Use more HelsinkiDateTime in frontend

Gekkio:

### DIFF
--- a/frontend/src/employee-frontend/components/application-page/ApplicationReadView.tsx
+++ b/frontend/src/employee-frontend/components/application-page/ApplicationReadView.tsx
@@ -11,6 +11,7 @@ import {
   ApplicationAttachment,
   ApplicationPersonBasics
 } from 'lib-common/api-types/application/ApplicationDetails'
+import LocalDate from 'lib-common/local-date'
 import InlineButton from 'lib-components/atoms/buttons/InlineButton'
 import ListGrid from 'lib-components/layout/ListGrid'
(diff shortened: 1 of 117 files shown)


frontend/src/employee-frontend/components/application-page/ApplicationReadView.tsx
```diff
@@ -143,12 +144,14 @@ export default React.memo(function ApplicationReadView({
...
+  const attachmentReceivedAt = (
+    attachment: ApplicationAttachment
+  ): LocalDate => {
     const { sentDate } = application.application
-    if (!sentDate) return attachment.receivedAt
-    else if (sentDate.toSystemTzDate() > attachment.receivedAt)
-      return sentDate.toSystemTzDate()
-    else return attachment.receivedAt
+    if (!sentDate) return attachment.receivedAt.toLocalDate()
+    else if (sentDate.isEqualOrAfter(attachment.receivedAt.toLocalDate()))
```
reynders, line 152: The original version reads as sentDate.isAfter(receivedAt), not isEqualOrAfter

reynders (approved):

frontend/src/employee-frontend/components/application-page/ApplicationReadView.tsx
Gekkio: True, the change is unnecessary and confusing...and it doesn't actually matter since we just want max of the two values. I've replaced this if-else with a maxOf function that is IMHO clearer

PR #2706, opened 2022-08-04
Accessibility and UI updates

jolle:
#### Summary

- Changed z-indexes to put holiday CTA banner under the mobile nav
- Updated font weight of H3s
- Modified AsyncButton to be more accessible
- Added aria-expanded to info buttons and expanding containers
- Modified selectable chips to be more accessible
- Modified language menu to hide abbreviations for screen readers, and make screen readers read the language in the actual language
- Added label for the close button in info boxes
- Fixed header double-reading links
- Added alt texts to income statements table
- Joined child name text nodes so that screen readers do not read them separately
- Improved accessibility of the children page by making the list of links a menu
- Fixed checkbox and radio button labels
- Made navigation items more accessible with more accurate labels
- Made modal headers focus first (does not work on iOS VoiceOver)

### DIFF
--- a/frontend/src/citizen-frontend/LoginPage.tsx
+++ b/frontend/src/citizen-frontend/LoginPage.tsx
@@ -104,6 +104,7 @@ export default React.memo(function LoginPage() {
               <ExpandingInfoBox
                 info={i18n.loginPage.applying.infoBoxText}
                 close={() => setShowInfoBoxText(false)}
+                closeLabel={i18n.common.close}
               />
             )}
             <ul>
--- a/frontend/src/citizen-frontend/LoginPage.tsx
+++ b/frontend/src/citizen-frontend/LoginPage.tsx
@@ -97,6 +97,7 @@ export default React.memo(function LoginPage() {
               <ParagraphInfoButton
                 aria-label={i18n.common.openExpandingInfo}
                 onClick={() => setShowInfoBoxText(!showInfoBoxText)}
+                open={showInfoBoxText}
               />
             </P>
             {showInfoBoxText && (
--- a/frontend/src/lib-components/atoms/SrOnly.tsx
+++ b/frontend/src/lib-components/atoms/SrOnly.tsx
@@ -1,0 +1,14 @@
+// SPDX-FileCopyrightText: 2017-2022 City of Espoo
+//
+// SPDX-License-Identifier: LGPL-2.1-or-later
+
+import styled from 'styled-components'
+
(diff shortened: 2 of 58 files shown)


frontend/src/citizen-frontend/LoginPage.tsx
```diff
@@ -73,12 +73,14 @@ export default React.memo(function LoginPage() {
...
                 aria-label={i18n.common.openExpandingInfo}
                 onClick={() => setShowInfoBoxText(!showInfoBoxText)}
+                open={showInfoBoxText}
               />
             </P>
             {showInfoBoxText && (
               <ExpandingInfoBox
                 info={i18n.loginPage.applying.infoBoxText}
                 close={() => setShowInfoBoxText(false)}
+                closeLabel={i18n.common.close}
```
zch, line 83: How would you feel changing `ExpandingInfoBox` to default the close label to this text?

This change is everywhere, and I can't figure out a reason for not making it the default. As it is now, whenever someone adds an `ExpandingInfoBox` to a layout, they need to remember to add the default close label or we have worse accessibility.

frontend/src/lib-components/atoms/SrOnly.tsx
```diff
@@ -0,0 +1,14 @@
+// SPDX-FileCopyrightText: 2017-2022 City of Espoo
+//
+// SPDX-License-Identifier: LGPL-2.1-or-later
+
+import styled from 'styled-components'
+
+export const SrOnly = styled.div`
```
zch, line 7: I'd love for this to be named `ScreenReaderOnly` instead. As it is now, you need to know what Sr means and I was only able to figure this out by the context of this PR.

zch (changes requested):

frontend/src/citizen-frontend/LoginPage.tsx
jolle: `ExpandingInfoBox` is in `lib-components`, and as such doesn't have access to translations.

zch: well that sucks 😆

jolle: Yeah, the same applies for `ReturnButton.label`, `ExpandingInfo.ariaLabel`, etc.; a shared translations file (and i18n system) with some common translation keys between all the frontends would fix this, but I think it's out of scope for this PR.

zch (approved):

PR #2710, opened 2022-08-05
Store mutp token only if all modifications are successful

reynders:


jaahon (approved):
I think a test checking that the token does not get updated on failed person updates would be valuable.

Gekkio (approved):

PR #2725, opened 2022-08-09
Don't show children without active placements in calendar view

zch:


frontend/src/lib-customizations/espoo/citizen/assets/i18n/en.tsx
```diff
@@ -260,6 +260,8 @@ const en: Translations = {
     reservation: 'Registered attendance',
     realized: 'Realized',
     reservationsAndRealized: 'Period of attendance',
+    noActivePlacements:
+      'Your child does not have an active placement in the early childhood education or pre-school on this day.',
```
jaahon, line 264: I'm not sure about this copy. The concept of "active placements" doesn't seem like something that would be obvious to our users. Would something like "Your children are not in early childhood education or pre-school on this day" be better? What do you think?

zch: I like your suggestion, although "Your children are not in early childhood education or pre-school on this day" is also correct for sick days... But it still might be better. I'll run it by Anna 👍

zch: Simplified the copy in all languages 👍

jaahon (approved):

frontend/src/lib-customizations/espoo/citizen/assets/i18n/en.tsx
jaahon: It's very good now!

PR #2736, opened 2022-08-11
EVAKA-4574 Require backup care to be during a placement

jolle:
#### Summary

- Added database trigger to enforce that backup cares must be during a placement
  - Triggers both ways: when placements are modified/deleted, or when backup cares are modified/added
- Added automatic adjusting of backup care validity periods when placements are modified/deleted
- Added UI elements to warn about automatic adjusting of backup care periods whenever relevant

### DIFF
--- a/service/src/main/kotlin/fi/espoo/evaka/backupcare/BackupCareQueries.kt
+++ b/service/src/main/kotlin/fi/espoo/evaka/backupcare/BackupCareQueries.kt
@@ -149,3 +149,105 @@ WHERE id = :id
 """
 ).bind("id", id)
     .execute()
+
+fun Database.Read.getBackupCareChildId(id: BackupCareId): ChildId = createQuery(
+    // language=SQL
+    """
+    SELECT child_id FROM backup_care WHERE id = :id
+    """.trimIndent()
+)
+    .bind("id", id)
+    .mapTo<ChildId>()
+    .one()
+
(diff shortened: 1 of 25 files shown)


service/src/main/kotlin/fi/espoo/evaka/backupcare/BackupCareQueries.kt
```diff
@@ -113,3 +113,105 @@ WHERE id = :id
...
+        )
+            .bind("childId", childId)
+            .bind("range", oldRange)
+            .execute()
+
+        createUpdate(
+            // language=SQL
+            """
+            UPDATE backup_care
+            SET start_date = upper(:range) + INTERVAL '1 day'
```
Gekkio, line 156: ... and the gotcha is that for dateranges, `upper` always returns the *exclusive* endpoint, even if the range was created as inclusive `daterange(..., ..., '[]')`. This happens because inclusivity/exclusivity info is not stored in the range itself, so it's only used during creation and all date ranges are `'[)` internally.

In other words, there seems to be an off-by-one bug here.

jolle: Thanks, I realised this issue in the else-branch below (when the placement is just moved, not deleted), but didn't update these queries too.

PR #2737, opened 2022-08-11
Mark days off in the unit calendar in case of irregular service times

zch:
If a child has irregular daily service times where one or more days are
completely off, these days are now marked as days off in the monthly
unit calendar

image


jaahon (commented):
The child is still considered to be attending when it comes to invoices. I have a feeling that's going to cause a lot of confusion. I think that should be taken into account somehow.

PR #2744, opened 2022-08-12
Fix unit page resetting URL state

jolle:
#### Summary

The main unit page reset the URL state, causing the calendar state to be reset on page load.


jaahon (commented):
I didn't debug this, but if you go to the groups tab, open a couple of group lists, and go back to the attendances tab, the page crashes because of repeating state updates.

jaahon (approved):

PR #2749, opened 2022-08-15
Use update instead of upsert for message drafts

Gekkio:
#### Summary

This is a workaround for some E2E test flakiness caused by draft updates happening *after* a draft has been deleted.


reynders (approved):

service/src/main/kotlin/fi/espoo/evaka/messaging/DraftQueries.kt
```diff
@@ -44,26 +44,24 @@ fun Database.Transaction.initDraft(accountId: MessageAccountId): MessageDraftId
         .one()
 }
 
-fun Database.Transaction.upsertDraft(accountId: MessageAccountId, id: MessageDraftId, draft: UpsertableDraftContent) {
-    this.createUpdate(
+fun Database.Transaction.upsertDraft(accountId: MessageAccountId, id: MessageDraftId, draft: UpsertableDraftContent) =
```
jaahon, line 47: Might want to change the name as well.

PR #2755, opened 2022-08-16
EVAKA-4581 consolidate child pages

jolle:
#### Summary

- Removed child documents page and moved the content from there to under each individual child alongside assistance need decisions, consents, and placement terminations
- Added a collapsible container type for the children page that fills the entire page on mobile, mimicking the appearance of a new page
- Added pagination to the pedagogical documents section
- Added fixes and changes to the spacing/order of elements in the pedagogical documents and vasu/leops sections


zch (approved):
The code looks good, but needs a rebase.

Slight UI nit: not enough margins
image

Also the focus trapped sections on mobile cause weird scrolling (scroll bar moving up/down, but content staying mostly put). Although I don't think that's a problem (at least for now). The scrolling works pretty well if the focus trapped section is big enough to need scrolling.

jolle:
> Slight UI nit: not enough margins
Fixed, this was only happening to the vasu/leops section as the dimmed placeholder text needed to have some spacing.

> Also the focus trapped sections on mobile cause weird scrolling (scroll bar moving up/down, but content staying mostly put). Although I don't think that's a problem (at least for now). The scrolling works pretty well if the focus trapped section is big enough to need scrolling.

The focus trapping only affects tab-based navigation (on mobile it's mostly VoiceOver and other accessibility tools). When there is not enough content to have a scrolling container and the user tries to scroll, it scrolls the underlying page. The mobile-only whole page view is actually akin to a modal; all (non-scrollable) modals have this underlying background scrolling effect and it can't really be avoided.

PR #2765, opened 2022-08-18
Only parse request body when it's actually needed

akheron:
#### Summary

This should solve unexpected occurences of `PayloadTooLargeError`, as well as enhance performance of proxied requests a bit.


Gekkio (approved):
Seems legit

PR #2768, opened 2022-08-18
Interpret days off marked in daily service time as absences

zch:
Adds 3 SQL functions that return the absences of a child / children on a date / in a period. The functions combine data from the `absence` table with data from the `daily_service_times` table to form a comprehensive set of absences for the children in question.


frontend/src/e2e-test/pages/employee/units/unit-attendances-page.ts
```diff
@@ -340,6 +340,13 @@ export class UnitAttendancesPage {
   personCountSum(nth: number) {
     return this.page.findAllByDataQa('person-count-sum').nth(nth).innerText
   }
+
+  async assertAbsence(childId: UUID, date: LocalDate, type: string) {
+    await this.page
+      .findByDataQa(`absence-cell-${childId}-${date.formatIso()}`)
+      .find(`.absence-cell-right-${type}`)
+      .waitUntilVisible()
+  }
```
jaahon, line 349: `unit-diary-page.ts` already has an almost identical function. Could that code be reused to repurposed for this case?

service/src/main/resources/db/migration/V262__child_absences_function.sql
```diff
@@ -0,0 +1,136 @@
...
+            dst.id as id,
+            dst.child_id,
+            d::date as date,
+            'OTHER_ABSENCE'::absence_type as absence_type,
+            dst.updated as modified_at,
+            null::uuid as modified_by,
+            'BILLABLE'::absence_category as category,
+            null::uuid as questionnaire_id
+        FROM daily_service_time as dst,
+             generate_series(lower($2), upper($2), '24 hours') as d
```
jaahon, line 29: You should use `upper($2) - 1` instead of `upper($2)`, because `upper` returns the exclusive end date of the date range (eg. `upper(daterange(current_date, current_date, '[]'))` is `current_date + 1`).

service/src/main/resources/db/migration/V262__child_absences_function.sql
```diff
@@ -0,0 +1,136 @@
...
+        WHERE a.child_id = ANY($1) AND between_start_and_end($2, a.date)
+        UNION
+        SELECT
+            dst.id as id,
+            dst.child_id,
+            d::date as date,
+            'OTHER_ABSENCE'::absence_type as absence_type,
+            dst.updated as modified_at,
+            null::uuid as modified_by,
+            'BILLABLE'::absence_category as category,
```
jaahon, line 26: What about the `NON_BILLABLE` category? We should probably generate absences for those categories depending on the placement type as well.

service/src/main/resources/db/migration/V262__child_absences_function.sql
```diff
@@ -0,0 +1,136 @@
...
+            null::uuid as questionnaire_id
+        FROM daily_service_time as dst,
+             generate_series(lower($2), upper($2), '24 hours') as d
+        WHERE dst.child_id = ANY($1)
+          AND dst.validity_period && $2
+          AND d >= lower(dst.validity_period)
```
jaahon, line 32: Wouldn't `dst.validity_period @> d` be enough here?

jaahon (commented):
I would like to see an integration test for invoices as well. Basically `invoice generation with unknown absences covering period` but with some of the absences being from daily service times.

I also think it would be a good idea to have only one function that contains all the logic so that the functions do not accidentally behave differently in the future.

service/src/main/resources/db/migration/V262__child_absences_function.sql
```diff
@@ -0,0 +1,136 @@
...
+            null::uuid as modified_by,
+            'BILLABLE'::absence_category as category,
+            null::uuid as questionnaire_id
+        FROM daily_service_time as dst,
+             generate_series(lower($1), upper($1), '24 hours') as d
+        WHERE dst.validity_period && $1
+          AND d >= lower(dst.validity_period)
+          AND dst.type = 'IRREGULAR'
+          AND (
+                (extract(dow FROM d) = 0 AND dst.sunday_times IS NULL)
```
jolle, line 126: I think using `CASE … WHEN …` could look cleaner here (and also above), similar to https://github.com/espoon-voltti/evaka/blob/master/service/src/main/kotlin/fi/espoo/evaka/reservations/ReservationQueries.kt#L27-L35.

service/src/main/resources/db/migration/V262__child_absences_function.sql
```diff
@@ -0,0 +1,136 @@
+CREATE FUNCTION absences_in_period(childIds uuid[], period daterange)
```
jolle, line 1: Nit: the indentation in this file is a bit inconsistent, with quite wide indentations in some parts (like in the TABLE parenthesis near the start).

zch: This is what the autoformatting in IDEA does currently 🤷
The formatter could be adjusted, but I'll leave that as a separate thing.

service/src/integrationTest/kotlin/fi/espoo/evaka/invoicing/service/InvoiceGeneratorIntegrationTest.kt
```diff
@@ -1497,6 +1536,47 @@ class InvoiceGeneratorIntegrationTest : PureJdbiTest(resetDbBeforeEach = true) {
...
+            DailyServiceTimes.IrregularTimes(
+                monday = null,
+                tuesday = null,
+                wednesday = null,
+                thursday = null,
+                friday = null,
+                saturday = null,
+                sunday = null,
+                validityPeriod = period
+            )
```
jaahon, line 1555: I think a test where there is a mix of absences generated by daily service times and manually entered absences would be valuable.

Also, I think this test is just wrong. I don't think these daily service times are really supposed to be interpreted in a way that the child is absent every single weekday.

service/src/main/resources/db/migration/V262__child_absences_function.sql
```diff
@@ -0,0 +1,48 @@
...
+               'OTHER_ABSENCE'::absence_type      as absence_type,
+               dst.updated                        as modified_at,
+               null::uuid                         as modified_by,
+               unnest(absence_categories(p.type)) as category,
+               null::uuid                         as questionnaire_id
+        FROM daily_service_time as dst,
+             generate_series(lower($2), upper($2) - 1, '1 day') as d
+                 JOIN placement as p ON d BETWEEN p.start_date AND p.end_date
+        WHERE ($1 IS NULL OR dst.child_id = ANY ($1))
+          AND p.child_id = dst.child_id
```
jaahon, line 34: I think the join on placement table is a more appropriate place for this condition.

service/src/integrationTest/kotlin/fi/espoo/evaka/invoicing/service/InvoiceGeneratorIntegrationTest.kt
zch: This is not a real world case, no. I added this test as a counter point to the other one, so that we can be sure that something actually happens. 

I'll convert this to mixed absences 👍

service/src/main/resources/db/migration/V262__child_absences_function.sql
zch: I had that at first, but it complains about it not being possible to reference `dst` in the join's ON clause.

service/src/integrationTest/kotlin/fi/espoo/evaka/invoicing/service/InvoiceGeneratorIntegrationTest.kt
jaahon: This test just got me thinking that maybe this exact behaviour is not what we want. The constraint to have at least one of the days filled out was recently taken out so this _could_ happen.

service/src/main/resources/db/migration/V262__child_absences_function.sql
jaahon: Oh yeah, it's because you're not joining the generated series on `daily_service_time`. You might as well do that since you have this `AND dst.validity_period @> d::date` condition in the where clause as well.

service/src/integrationTest/kotlin/fi/espoo/evaka/invoicing/service/InvoiceGeneratorIntegrationTest.kt
jolle: It is a real-world scenario where a child has irregular daily service times without any specified times. I am not sure how this should be interpreted, but my best guess is that by default the child is absent on every day, unless a reservation is manually added.

service/src/main/resources/db/migration/V262__child_absences_function.sql
zch: `ERROR: invalid reference to FROM-clause entry for table "dst"
  Hint: There is an entry for table "dst", but it cannot be referenced from this part of the query.`

service/src/integrationTest/kotlin/fi/espoo/evaka/invoicing/service/InvoiceGeneratorIntegrationTest.kt
zch: jolle's suggestion sounds good to me. Just need to check that a reservation will override the absences derived from daily service times.

zch: Tweaked the function to not return absences for days with reservations and added a test for that.

jaahon: I'm sorry about yet another comment, but should we use attendances instead of reservations in this case?

jaahon (approved):

PR #2771, opened 2022-08-18
Handle undefined data gracefully

Gekkio:
#### Summary

There are many scenarios where we have an error and a response, but the response data is undefined. For some reason this happens frequently after the recent axios upgrade

### DIFF
--- a/frontend/src/lib-common/api.ts
+++ b/frontend/src/lib-common/api.ts
@@ -2,7 +2,7 @@
 //
 // SPDX-License-Identifier: LGPL-2.1-or-later
 
-import axios, { AxiosResponse } from 'axios'
+import axios from 'axios'
 
 export interface Response<T> {
   data: T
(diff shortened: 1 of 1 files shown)


frontend/src/lib-common/api.ts
```diff
@@ -84,9 +84,14 @@ export class Failure<T> {
 
   static fromError<T>(e: unknown): Failure<T> {
     if (axios.isAxiosError(e)) {
-      const response: AxiosResponse<{ errorCode?: string }> | undefined =
-        e.response as AxiosResponse<{ errorCode?: string }>
-      return new Failure(e.message, response?.status, response?.data.errorCode)
+      // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
```
jolle, line 87: This rule disabling seems unnecessary, works without it too.

jolle (approved):

PR #2776, opened 2022-08-18
Add child id and unit id to PlacementCancel audit message

akheron:

### DIFF
--- a/service/src/main/kotlin/fi/espoo/evaka/placement/PlacementController.kt
+++ b/service/src/main/kotlin/fi/espoo/evaka/placement/PlacementController.kt
@@ -178,17 +178,24 @@ class PlacementController(
         user: AuthenticatedUser,
         @PathVariable("placementId") placementId: PlacementId
     ) {
-        Audit.PlacementCancel.log(targetId = placementId)
         accessControl.requirePermissionFor(user, Action.Placement.DELETE, placementId)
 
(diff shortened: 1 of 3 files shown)


service/src/main/kotlin/fi/espoo/evaka/placement/PlacementController.kt
```diff
@@ -178,17 +178,25 @@ class PlacementController(
...
         accessControl.requirePermissionFor(user, Action.Placement.DELETE, placementId)
 
         db.connect { dbc ->
             dbc.transaction { tx ->
-                val (childId, startDate, endDate) = tx.cancelPlacement(placementId)
-                asyncJobRunner.plan(
-                    tx,
-                    listOf(AsyncJob.GenerateFinanceDecisions.forChild(childId, DateRange(startDate, endDate)))
-                )
+                tx.cancelPlacement(placementId).let {
```
jaahon, line 185: You could use `also` instead of `let` and remove `it` from the end of the block 🤓

jaahon (approved):

service/src/main/kotlin/fi/espoo/evaka/placement/PlacementController.kt
akheron: Fixed 🤓 🤓 🤓

PR #2778, opened 2022-08-19
Citizen message UI updates

jolle:
#### Summary
- Make the new message button float on mobile (like the button in calendar)
- Fix spacing for desktop and mobile
- Wrap time-related information in time elements

### DIFF
--- a/frontend/src/lib-components/molecules/ThreadListItem.tsx
+++ b/frontend/src/lib-components/molecules/ThreadListItem.tsx
@@ -39,7 +39,8 @@ export const Container = styled.div<{ isRead: boolean; active: boolean }>`
   color: ${(p) => p.theme.colors.grayscale.g100};
   padding: ${defaultMargins.s} ${defaultMargins.m};
   cursor: pointer;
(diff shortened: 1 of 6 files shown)


frontend/src/lib-components/molecules/ThreadListItem.tsx
```diff
@@ -39,7 +39,8 @@ export const Container = styled.div<{ isRead: boolean; active: boolean }>`
   color: ${(p) => p.theme.colors.grayscale.g100};
   padding: ${defaultMargins.s} ${defaultMargins.m};
   cursor: pointer;
-  border: 1px solid ${(p) => p.theme.colors.grayscale.g15};
+  border-top: 1px solid ${(p) => p.theme.colors.grayscale.g15};
```
jaahon, line 42: The last message in the list now doesn't have a bottom border. Is that intentional?

jolle: Yes. I think it looks better without a bottom border for the last item; the Figma design has so many messages the last message is not (fully) visible.

jaahon (approved):

PR #2780, opened 2022-08-19
Fix flaky application decision e2e test

jaahon:
#### Summary

The test `Decision is not sent automatically to the other guardian if the first guardian has restricted details enabled` used to be flaky because evaka-service would be running a family initialization async job at the same time with a guardian VTJ-query causing the one of the processes to fail because of a deadlock. I added a `runAllAsyncJobs` call to the end of the application actions endpoint in dev api to make sure that all the async jobs have been run before doing the next step in e2e test data initialization. But it turned out that `runAllAsyncJobs` was not reliable and it would sometimes return even though there were some async jobs still waiting to be run. I've now fixed it, but I'm not sure if my fix is bulletproof. There haven't been any issues with `runAllAsyncJobs` in ~100 tests that I ran. Previously ~1/20 of the tests would fail.

### DIFF
--- a/service/src/main/kotlin/fi/espoo/evaka/shared/async/AsyncJobRunner.kt
+++ b/service/src/main/kotlin/fi/espoo/evaka/shared/async/AsyncJobRunner.kt
@@ -56,7 +56,7 @@ class AsyncJobRunner<T : AsyncJobPayload>(val payloadType: KClass<T>, private va
         }
     )
     private val periodicRunner: AtomicReference<ScheduledFuture<*>> = AtomicReference()
-    private val runningCount: AtomicInteger = AtomicInteger(0)
+    private val activeWorkerCount: AtomicInteger = AtomicInteger(0)
 
     private val handlersLock: Lock = ReentrantLock()
     private val handlers: ConcurrentHashMap<AsyncJobType<out T>, Registration<*>> = ConcurrentHashMap()
@@ -66,7 +66,7 @@ class AsyncJobRunner<T : AsyncJobPayload>(val payloadType: KClass<T>, private va
         get() = periodicRunner.get() != null
 
     val isBusy: Boolean
-        get() = runningCount.get() > 0
+        get() = activeWorkerCount.get() > 0
 
     inline fun <reified P : T> registerHandler(noinline handler: (db: Database.Connection, clock: EvakaClock, msg: P) -> Unit) =
         registerHandler(AsyncJobType(P::class)) { db, clock, msg -> db.connect { handler(it, clock, msg) } }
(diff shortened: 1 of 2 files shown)


service/src/main/kotlin/fi/espoo/evaka/shared/async/AsyncJobRunner.kt
```diff
@@ -133,6 +133,7 @@ class AsyncJobRunner<T : AsyncJobPayload>(val payloadType: KClass<T>, private va
     private fun runPendingJobs(clock: EvakaClock, maxCount: Int = 1_000) {
         Database(jdbi).connect { db ->
             var remaining = maxCount
+            runningCount.incrementAndGet()
```
Gekkio, line 136: The original meaning of `runningCount` was "number of currently running jobs" and now it's "number of currently active workers". I think this is fine, since we actually don't care about the former. :+1:

service/src/main/kotlin/fi/espoo/evaka/shared/async/AsyncJobRunner.kt
```diff
@@ -141,6 +142,7 @@ class AsyncJobRunner<T : AsyncJobPayload>(val payloadType: KClass<T>, private va
                 remaining -= 1
                 config.throttleInterval?.toMillis()?.run { Thread.sleep(this) }
             } while (job != null && remaining > 0)
+            runningCount.decrementAndGet()
```
Gekkio, line 145: This should be in a finally block, because if any exception is thrown (e.g. `it.claimJob` fails), the count gets messed up.

Gekkio (approved):

service/src/main/kotlin/fi/espoo/evaka/shared/async/AsyncJobRunner.kt
jaahon: Yeah, I guess it's misleading now. Would it make more sense to turn `isBusy` into `AtomicBoolean` and leave `runningCount` as is?

Gekkio: `AtomicBoolean` wouldn't work, because we can have multiple workers in parallel so it wouldn't be clear which one is the last one that is allowed to set it to false. I think the `runningCount` change is ok, but maybe we could rename it to something like `activeWorkerCount` to be absolutely clear what is means. In any case it's not super important, because `runningCount` is only used in the `isBusy` getter

jaahon: Oh yeah, of course 🙈

PR #2781, opened 2022-08-19
Disable consents for children without a placement

jolle:


jaahon (approved):

jaahon (commented):
Adding a test wouldn't be a bad idea.

PR #2797, opened 2022-08-23
Add assistance decision report support for unit supervisors

patari:
#### Summary

In Tampere unit supervisors are assistance decision makers (for municipal units). With this change we can get it working with following customization on our end:

```
Action.Global.READ_ASSISTANCE_NEED_DECISIONS_REPORT -> {
    action.defaultRules.asSequence() + sequenceOf(HasUnitRole(UserRole.UNIT_SUPERVISOR).inAnyUnit())
}
...
Action.AssistanceNeedDecision.DECIDE,
Action.AssistanceNeedDecision.MARK_AS_OPENED -> {
    @Suppress("UNCHECKED_CAST")
    action.defaultRules.asSequence() + sequenceOf(
        HasUnitRole(UserRole.UNIT_SUPERVISOR).andIsDecisionMakerForAssistanceNeedDecision() as ScopedActionRule<in T>
    )
}
```


service/src/main/kotlin/fi/espoo/evaka/reports/AssistanceNeedDecisionsReport.kt
```diff
@@ -26,11 +28,12 @@ class AssistanceNeedDecisionsReport(private val accessControl: AccessControl, pr
     ): List<AssistanceNeedDecisionsReportRow> {
         Audit.AssistanceNeedDecisionsReportRead.log()
         accessControl.requirePermissionFor(user, Action.Global.READ_ASSISTANCE_NEED_DECISIONS_REPORT)
+        val authorizedUnits = acl.getAuthorizedUnits(user, setOf(UserRole.UNIT_SUPERVISOR))
```
patari, line 31: `setOf(UserRole.UNIT_SUPERVISOR)` probably should be customizable but where? Is it possible to read it from `accessControl`?

Gekkio: Rebase, and use the authorization filter technique as here: https://github.com/espoon-voltti/evaka/blob/ddfef2fd7fa5376fb77e71c0f9116bfa1932e4b6/service/src/main/kotlin/fi/espoo/evaka/daycare/controllers/DaycareController.kt#L64

You'll need to add and use a unit-level action (`Action.Unit.READ_ASSISTANCE_NEED_DECISIONS_REPORT`). Unfortunately we can't get rid of the global action yet, because the reports page uses it for visibility. There's no need to do actual access control with the global action anymore, but it should have "compatible" rules so that the report appears correctly on the page.

service/src/main/kotlin/fi/espoo/evaka/shared/security/actionrule/HasUnitRole.kt
```diff
@@ -187,6 +187,28 @@ AND ad.id = ANY(:ids)
         }
     )
 
+    // For Tampere
+    fun andIsDecisionMakerForAssistanceNeedDecision() = DatabaseActionRule(
```
Gekkio, line 191: This needs to be updated after rebase:

- use QueryFragment and the rule function (look at the other rules for examples)
- remove :ids parameter and check from the query

patari: Done!

service/src/main/kotlin/fi/espoo/evaka/reports/AssistanceNeedDecisionsReport.kt
patari: Done!

Gekkio (approved):
:gem:

PR #2798, opened 2022-08-23
Make header sticky on the citizen side

zch:


zch:
> As these changes are applied on desktop as well, the desktop calendar's sticky subheader (with the create absence and attendance buttons) is being hidden under the main header after this change. Either the main header needs to be unstickied on desktop or the `top` CSS property of the subheader needs to be calculated with the main header's height.

As the same problem was visible on desktop, I tweaked the positioning of the calendar header 👍

jolle (approved):
Seems to work well, although I can't be certain some pages may be broken in some cases by this change as the z-index layers are not documented anywhere, and it's possible some elements will go under or over the header now (when not wanted). I didn't find anything broken, so LGTM! 😄

PR #2801, opened 2022-08-24
Fix employee login when external id changes

patari:
#### Summary

When employee gets a new contract or comes back to work
(or whatever) it gets a new external id in SSO for some reason but
employee number (should) stays the same.

Bad thing about this is (unline external id which seems uuid and
probably generated by system) employee number is set by human, so
there is possibility for human error to set same or wrong employee
number to employee.


Gekkio (approved):
Seems ok, although I think this adds a bit of technical debt: the idea is that the external id is the main unique identifier of an employee. Now that employee number is in practice the main identifier and the external id just "tags along", it would be better to return the employee number in the external id field from the apigw (external id is just a string with a prefix, so no need for UUIDs). But this would require a large migration at this point so it's probably not worth doing.

PR #2807, opened 2022-08-25
Revert "Interpret days off marked in daily service time as absences"

zch:
This reverts commit 9a2ea7f23e811e477c9aeb54a36422e35d921125.


service/src/main/resources/migrations.txt
```diff
@@ -260,4 +260,3 @@ V259__early_childhood_education_secretary_acl.sql
 V260__drop_irregular_daily_service_times_constraint.sql
 V261__invoice_correction_person_reference_on_delete_cascade.sql
 V262__child_consent_on_delete_cascade.sql
-V263__child_absences_function.sql
```
akheron, line 263: You must keep the migration file and this line, otherwise future migrations will fail hard.

akheron (changes requested):

akheron (approved):

PR #2812, opened 2022-08-29
Vasu, leops, and assistance need decision text changes

akheron:

### DIFF
--- a/service/src/main/kotlin/fi/espoo/evaka/vasu/OphQuestion.kt
+++ b/service/src/main/kotlin/fi/espoo/evaka/vasu/OphQuestion.kt
@@ -101,7 +101,10 @@ fun getDefaultVasuContent(lang: VasuLanguage) = VasuContent(
                         VasuLanguage.SV -> ""
                     },
                     info = when (lang) {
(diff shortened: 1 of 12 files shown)


service/src/main/resources/WEB-INF/templates/assistance-need/decision_sv.properties
```diff
@@ -53,6 +53,8 @@ statuses.ACCEPTED=godkänt
 statuses.REJECTED=avvisat
 decisionStatus=Beslutet är
 decisionMadeOn=Beslut fattat den
+legalInstructions=Tillämpade bestämmelser
+legalInstructionsText=Lag om småbarnspedagogik, 3 a kap
 jurisdiction=Befogenhet
 jurisdictionText=Delegationsbeslut om beslutanderätten för tjänsteinnehavarna av personalen vid den finskspråkiga småbarnspedagogiken och tillväxt- och lärandegrenen A del 3 § 3 pkt
```
jolle, line 59: `jurisdictionText`, `startDate`, `unitMayChange`, `structuralMotivationOptions.additionalStaff` need to be updated as well

service/src/main/kotlin/fi/espoo/evaka/vasu/OphQuestion.kt
```diff
@@ -849,17 +852,6 @@ fun getDefaultLeopsContent(lang: VasuLanguage) = VasuContent(
                 )
             )
         ),
-        VasuSection(
```
jolle, line 852: As the vasu/LEOPS e2e tests are unfortunately dependent on the rendered question numbers, the removal of this section should be reflected in the LEOPS e2e test as well (namely decreasing the numbers by 1).

jolle (changes requested):

service/src/main/resources/WEB-INF/templates/assistance-need/decision_sv.properties
akheron: Good catch!

jolle (approved):

PR #2816, opened 2022-08-29
Prevent leftover fridge_partner rows using a self-reference constraint

Gekkio:
#### Summary

The two related rows are tied together with a foreign key, which couples them together:

- both rows must always be added in the same transaction
- if one row is deleted, the delete cascades to the other one

**The migration fails if there are any leftover rows in the database, so these must be cleaned up manually**

```
SELECT partnership_id
FROM fridge_partner
GROUP BY partnership_id
HAVING count(indx) != 2
```


akheron (approved):
🏅

PR #2817, opened 2022-08-30
EVAKA-4596 Remove gaps in occupancy graph for overnight attendances

zch:
#### Summary

Overnight attendances are marked as ending on 23:59 on day 1 and
starting 00:00 on day 2, which caused the occupancy graph to drop down
to zero in the minute between 23:59 and 00:00.

This change removes the zero data point between 23:59 and 00:00, and
thus it is no longer rendered in the graph.

### DIFF
--- a/service/src/integrationTest/kotlin/fi/espoo/evaka/occupancy/RealtimeOccupancyTest.kt
+++ b/service/src/integrationTest/kotlin/fi/espoo/evaka/occupancy/RealtimeOccupancyTest.kt
@@ -28,6 +28,7 @@ import java.math.BigDecimal
 import java.time.LocalDate
 import java.time.LocalTime
 import java.util.UUID
+import kotlin.test.assertContentEquals
 import kotlin.test.assertEquals
 import kotlin.test.assertNull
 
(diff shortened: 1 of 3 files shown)


jaahon (commented):
This doesn't really work as expected if there are no datapoints before or after the dropped datapoint. I think there should be a check for that.

service/src/integrationTest/kotlin/fi/espoo/evaka/occupancy/RealtimeOccupancyTest.kt
```diff
@@ -193,6 +193,45 @@ class RealtimeOccupancyTest : FullApplicationTest(resetDbBeforeEach = true) {
...
+
+        val result = getRealtimeOccupancy()
+        val occupancies = result.occupancySeries
+
+        // 23:59 - 00:00, overnight attendances go to zero in the db during this minute
+        val minuteBeforeMidnight = HelsinkiDateTime.Companion.of(date, LocalTime.of(23, 59))
+        val midnight = HelsinkiDateTime.Companion.of(tomorrow, LocalTime.of(0, 0))
+        val beforeMinuteBeforeMidnight = occupancies.findLast { it.time < minuteBeforeMidnight }
+        occupancies.find { it.time == minuteBeforeMidnight }
+            ?.also { assertEquals(beforeMinuteBeforeMidnight?.childCapacity, it.childCapacity) }
```
jaahon, line 229: Using `?.also` in tests is not a good idea because it can make tests pass even when they shouldn't. This test actually fails if you assert that `occupancies.find { it.time == midnight }` is not null.

jaahon (approved):
Good stuff! Sorry about being a pain in the ass 😄

PR #2819, opened 2022-08-30
EVAKA-4596 Render overnight child attendances using arrows

zch:
#### Summary
Render '00:00' and '23:59' using arrows in order to make the table
easier to read. This is similar to how staff attendances are rendered.
image

### DIFF
--- a/frontend/src/employee-frontend/components/unit/unit-reservations/ChildDay.tsx
+++ b/frontend/src/employee-frontend/components/unit/unit-reservations/ChildDay.tsx
@@ -2,7 +2,7 @@
 //
 // SPDX-License-Identifier: LGPL-2.1-or-later
 
-import React, { useMemo } from 'react'
+import React, { useMemo, useCallback } from 'react'
 import styled, { css } from 'styled-components'
 
 import {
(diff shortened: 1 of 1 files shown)


jolle (changes requested):
Although this superficially renders arrows for overnight attendances, in reality if a child has an attendance to 23:59 (not overnight, just happens to be 23:59 – e.g. `17:00–23:59`), this change would show arrows erroneously. To fix this, more substantial changes need to be made, namely that the dates need to be inspected in addition to the time (the dates need to differ).

zch:
> Although this superficially renders arrows for overnight attendances, in reality if a child has an attendance to 23:59 (not overnight, just happens to be 23:59 – e.g. `17:00–23:59`), this change would show arrows erroneously. To fix this, more substantial changes need to be made, namely that the dates need to be inspected in addition to the time (the dates need to differ).

This is true, but in practice such a situation will be extremely unlikely. 

Because of the unlikeliness of a child leaving at exactly 23:59, or arriving at exactly 00:00 and as it would require substantial changes, all the way from the data model up, I don't think it makes sense to spend more effort than this on it. This way we get 99% of the benefit for 1% of the effort.

jolle (approved):

PR #2838, opened 2022-09-02
Send null paattymis_pvm to varda

reynders:
#### Summary
When unit was accidentally marked a closing date which was sent to varda, then the date was removed, it was not sent to varda because null value fields were dropped from the request payload. Allow sending null paattymis_pvm.


jaahon (approved):
I have a feeling we should just remove the `@JsonInclude(JsonInclude.Include.NON_NULL)` annotation completely, but that's left for later.

PR #2845, opened 2022-09-05
Don't show children in between placements as reservable in the citizen calendar

zch:
Children who had a placement, which ended and are waiting for the next placement to start were wrongly shown in the calendar. This change allows gaps in placements to be rendered correctly.

Also added an error message if the citizen tries to store a reservation that overlaps with non reservable days.

### DIFF
--- a/service/src/main/kotlin/fi/espoo/evaka/reservations/ReservationControllerCitizen.kt
+++ b/service/src/main/kotlin/fi/espoo/evaka/reservations/ReservationControllerCitizen.kt
@@ -60,13 +60,18 @@ class ReservationControllerCitizen(
                 }
                 val reservations = tx.getReservationsCitizen(user.id, range, includeWeekends)
                 val deadlines = tx.getHolidayPeriodDeadlines()
-                val placementRange = FiniteDateRange(
-                    children.minOfOrNull { it.placementMinStart } ?: LocalDate.MIN,
-                    children.maxOfOrNull { it.placementMaxEnd } ?: LocalDate.MAX
-                )
-                val reservableDays =
+                val reservableDayRanges =
                     getReservableDays(evakaClock.now(), featureConfig.citizenReservationThresholdHours, deadlines)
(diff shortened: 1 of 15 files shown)


service/src/main/kotlin/fi/espoo/evaka/reservations/ReservationControllerCitizen.kt
```diff
@@ -272,34 +277,40 @@ SELECT
...
     ch.preferred_name,
     ci.id AS image_id,
-    p.placement_min_start,
-    p.placement_max_end,
+    jsonb_agg(
+            jsonb_build_object(
+                    'start', p.placement_start,
+                    'end', p.placement_end
+                ) ORDER BY p.placement_start ASC
+        ) AS placements,
```
akheron, line 285: Could the `jsonb_agg` be moved inside the subquery? That would remove the need of another `GROUP BY` in the outmost query.

service/src/main/kotlin/fi/espoo/evaka/reservations/ReservationControllerCitizen.kt
```diff
@@ -272,34 +277,40 @@ SELECT
...
-        bool_or(p.round_the_clock) AS in_shift_care_unit
-    FROM (
-        SELECT pl.start_date, pl.end_date, unnest(u.operation_days) AS operation_days, u.round_the_clock
+   SELECT
+       p.start_date AS placement_start,
+       p.end_date AS placement_end,
+       array_agg(DISTINCT p.operation_days) AS max_operational_days,
+       bool_or(p.round_the_clock) AS in_shift_care_unit
+   FROM (
+        SELECT pl.start_date, pl.end_date, unnest(u.operation_days) AS operation_days, u.round_the_clock, pl.child_id
```
akheron, line 298: It seems to me that the added `child_id` columns is unused.

akheron (approved):
💎

PR #2847, opened 2022-09-06
Employee preferred first name support

reynders:


frontend/src/employee-frontend/components/employee/EmployeeNicknamePage.tsx
```diff
@@ -0,0 +1,83 @@
...
+  const { i18n } = useTranslation()
+  const [possibleNicknames, setPossibleNicknames] = useState<
+    Result<EmployeeNicknames>
+  >(Loading.of())
+  const [selectedNickname, setSelectedNickname] = useState<string | null>(null)
+
+  useEffect(() => {
+    setPossibleNicknames(Loading.of())
+    void getPossibleNicknames().then(setPossibleNicknames)
+  }, [])
```
jaahon, line 29: All of this could be replaced with `useApiState`.


```
const [possibleNicknames, setPossibleNicknames] = useState<
  Result<EmployeeNicknames>
>(Loading.of())

useEffect(() => {
  setPossibleNicknames(Loading.of())
  void getPossibleNicknames().then(setPossibleNicknames)
}, [])
  ```

File: frontend/src/employee-frontend/components/employee/EmployeeNicknamePage.tsx [hunk 2]
```diff
@@ -0,0 +1,83 @@
...
+            }
+            onChange={(value) => setSelectedNickname(value)}
+            data-qa="select-nickname"
+          />
+          <AsyncButton
+            primary
+            disabled={disableConfirm()}
+            text={i18n.nickname.confirm}
+            onClick={onSave}
+            onSuccess={() => location.reload()}
```
jaahon, line 76: It's a better idea to just refresh the data instead of doing a full page reload.

jaahon (commented):
I know it's a pain in the ass, but I think nickname is not the best name for this data. Nicknames are casual. I think preferred name would be more accurate.

frontend/src/employee-frontend/components/employee/EmployeeNicknamePage.tsx
reynders: The reason is to the the top right corner user name to updated, as required

jaahon: Sure, but you could just refresh that data only.

jaahon (approved):

PR #2849, opened 2022-09-06
Some citizen navigation fixes

jaahon:
#### Summary

- Fix citizen header city logo width (used to work differently on firefox and chrome)
- Make sure citizen mobile menu hides elements under it
- Close citizen mobile menus on any navigation
- Use a blue city logo in citizen header (the new logo has been uploaded to s3 already)
- Show children menu to users without a strong login


akheron:
I find it confusing that the notification bubble is orange, but the exclamation mark in the menu is not.

Screenshot 2022-09-07 at 9 16 04

jaahon:
> I find it confusing that the notification bubble is orange, but the exclamation mark in the menu is not.

I understand, but that's the spec.

akheron:
All pages should scroll to top on mobile upon navigation. When in the calendar page, the browser is always scrolled downwards at least some amount. Then, when navigating to almost any other page, the new page is scrolled to bottom because the browser tries to keep the scroll position.

PR #2851, opened 2022-09-06
Refactor permitted features to actions

patari:
#### Summary

To support customization.

This allows Tampere to give more permitted features to director role.


jaahon (approved):
Master has a suppression for the CVE

patari:
> Latest master has a suppression for the CVE

Rebased.

Do we even need `accessibleFeatures`? Global actions are already returned to UI which could be used instead. Only problems is that in its current state messages-page shouldn't be visible for admins (and actions are always permitted for admins by default)..

jaahon:
> > Latest master has a suppression for the CVE
> 
> Rebased.
> 
> Do we even need `accessibleFeatures`? Global actions are already returned to UI which could be used instead. Only problems is that in its current state messages-page shouldn't be visible for admins (and actions are always permitted for admins by default)..

Yeah probably not. I'll add it to tech debt backlog. Better leave it as it is for now.

PR #2857, opened 2022-09-07
Use EvakaClock as much as possible in the backend

Gekkio:
#### Summary

- pass EvakaClock to (almost) every function that depends on time. This adds boilerplate, but makes things more explicit and *simpler* because dependency on time is not hidden. It's much easier to reduce boilerplate than try to reliably test a large call stack of functions where some functions may call `HelsinkiDateTime.now()` or some similar thing, and some functions use EvakaClock. The solution is to use EvakaClock as much as possible and find other ways to fight boilerplate than magically conjuring current time (= an uncontrolled side effect) in random functions
- replace SQL `current_date`/`clock_timestamp`/`now` with bind parameters
- replace Instant with HelsinkiDateTime
- replace `LocalDate.now()` with EvakaClock. Note that `Localdate.now()` without parameters always returns current date *according to server timezone (= probably UTC)*, which can be wrong
- there are some remaining `HelsinkiDateTime.now` and `LocalDate.now` calls which still need to be replaced, but this pull request is already a bit too big
- in many cases a function is actually just interested in today/now and doesn't need the whole clock. I didn't want to spend time thinking on every single case in this pull request, so these can be refactored and simplified later

This pull request doesn't yet guarantee reliable tests, because it just makes it possible to use mocked clock and have it work (mostly) correctly. The next step is to (usually) replace `RealEvakaClock` in integration tests, and (usually) pass a mocked time in E2E tests.

### DIFF
--- a/service/src/main/kotlin/fi/espoo/evaka/shared/async/AsyncJobQueries.kt
+++ b/service/src/main/kotlin/fi/espoo/evaka/shared/async/AsyncJobQueries.kt
@@ -51,7 +51,7 @@ fun <T : AsyncJobPayload> Database.Transaction.claimJob(now: HelsinkiDateTime, j
 WITH claimed_job AS (
   SELECT id
   FROM async_job
-  WHERE run_at < :now
+  WHERE run_at <= :now
   AND retry_count > 0
   AND completed_at IS NULL
   AND type = ANY(:jobTypes)
--- a/service/src/main/kotlin/fi/espoo/evaka/shared/async/AsyncJobQueries.kt
+++ b/service/src/main/kotlin/fi/espoo/evaka/shared/async/AsyncJobQueries.kt
@@ -44,14 +44,14 @@ AND type = ANY(:jobTypes)
         .mapTo<Int>()
         .one()
 
-fun <T : AsyncJobPayload> Database.Transaction.claimJob(jobTypes: Collection<AsyncJobType<out T>>): ClaimedJobRef<out T>? =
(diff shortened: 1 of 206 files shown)


service/src/main/kotlin/fi/espoo/evaka/shared/async/AsyncJobQueries.kt
```diff
@@ -44,14 +44,14 @@ AND type = ANY(:jobTypes)
...
-fun <T : AsyncJobPayload> Database.Transaction.claimJob(jobTypes: Collection<AsyncJobType<out T>>): ClaimedJobRef<out T>? =
+fun <T : AsyncJobPayload> Database.Transaction.claimJob(now: HelsinkiDateTime, jobTypes: Collection<AsyncJobType<out T>>): ClaimedJobRef<out T>? =
     createUpdate(
         // language=SQL
         """
 WITH claimed_job AS (
   SELECT id
   FROM async_job
-  WHERE run_at < now()
+  WHERE run_at <= :now
```
reynders, line 54: Is this functional change intended?

reynders (approved):

service/src/main/kotlin/fi/espoo/evaka/shared/async/AsyncJobQueries.kt
Gekkio: Yes, that's why it's in a separate commit: https://github.com/espoon-voltti/evaka/pull/2857/commits/a179e77d211ca0d3199636652b0ae39c6d548841

PR #2860, opened 2022-09-07
Find questions using section and question name instead of indexes

zch:
Find questions using section and question names instead of indexes as finding by index broke down completely when sections were dynamically hidden by previous answers.

### DIFF
--- a/frontend/src/employee-frontend/components/vasu/sections/DynamicSections.tsx
+++ b/frontend/src/employee-frontend/components/vasu/sections/DynamicSections.tsx
@@ -9,12 +9,15 @@ import styled, { css } from 'styled-components'
 
 import {
   CheckboxQuestion,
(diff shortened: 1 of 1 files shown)


jaahon (approved):
This obviously breaks if there are duplicate section question name combos. It's probably not a very likely scenario but could there be a more reliable way to find the correct questions like uuids? Even just dynamically generated ones only used in front end would be enough. I think merging this is fine as first aid though.

zch:
> This obviously breaks if there are duplicate section question name combos. It's probably not a very likely scenario but could there be a more reliable way to find the correct questions like uuids? Even just dynamically generated ones only used in front end would be enough. I think merging this is fine as first aid though.

Yep, I had the same thoughts. In practice I think this will be enough, but UUIDs for the questions would be great.

PR #2861, opened 2022-09-07
Fix footer

reynders:

### DIFF
--- a/frontend/src/citizen-frontend/messages/MessagesPage.tsx
+++ b/frontend/src/citizen-frontend/messages/MessagesPage.tsx
@@ -18,7 +18,6 @@ import { TabletAndDesktop } from 'lib-components/layout/responsive-layout'
 import { defaultMargins, Gap } from 'lib-components/white-space'
 
 import { useTranslation } from '../localization'
-import { headerHeightDesktop } from '../navigation/const'
 
 import EmptyThreadView from './EmptyThreadView'
 import MessageEditor from './MessageEditor'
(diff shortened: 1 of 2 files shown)


frontend/src/citizen-frontend/navigation/MobileNav.tsx
```diff
@@ -142,7 +142,7 @@ const BottomBar = styled.nav`
   height: ${mobileBottomNavHeight}px;
   width: 100%;
   padding: ${defaultMargins.xs};
-  position: sticky;
+  position: fixed;
```
jaahon, line 145: Using `position: fixed` means that 60px of content on every page is hidden behind the bottom bar. The problem in messages page is that the `FullHeightContainer` container in `MessagesPage.tsx` has a fixed height so the bottom bar is positioned after container but the actual contents inside the container overflow and are shown under the bottom bar. It seems that removing `height: calc(100vh - ${headerHeightDesktop}px);` fixes it and doesn't have an effect on anything else. It must have been put there for some reason at some point but whatever the reason was it doesn't seem to work anymore.

jaahon (changes requested):

frontend/src/citizen-frontend/messages/MessagesPage.tsx
```diff
@@ -27,9 +25,7 @@ import ThreadView from './ThreadView'
 import { getReceivers, sendMessage } from './api'
 import { MessageContext } from './state'
 
-const FullHeightContainer = styled(Container)`
-  height: calc(100vh - ${headerHeightDesktop}px);
-`
```
akheron, line 32: It's not a "full height container" anymore, so you should change the name. Or just use a plain `<div>` instead.

frontend/src/citizen-frontend/messages/MessagesPage.tsx
```diff
@@ -78,7 +72,7 @@ export default React.memo(function MessagesPage() {
   )
 
   return (
-    <FullHeightContainer>
+    <div>
```
jaahon, line 75: You should still use `lib-components/layout/Container` here.

jaahon (approved):

PR #2862, opened 2022-09-07
Remove bindMap

Gekkio:
#### Summary

`bindMap` is problematic and was only left for backwards compatibility when the query API wrapper was created:

- in most cases using repeated bind calls is simpler and less code
- `bindMap` puts all arguments in one untyped `Map<String, Any>`, so type information is lost and bindings don't necessarily work (e.g. when nulls and/or empty lists are involved, where the type can't be figured out from the value alone)

Instead of `bindMap`, use repeated bind calls, or a list of `Binding` values. `Binding.of` captures full type information so it doesn't have the same disadvantage as `bindMap`. With these changes some of the extra processing around bindings (e.g. `toTypedArray()` or converting enums to strings) becomes unnecessary since type information is not lost.


service/src/main/kotlin/fi/espoo/evaka/shared/dev/DataInitializers.kt
```diff
@@ -439,16 +431,12 @@ fun Database.Transaction.insertTestPlacement(
...
-                "unitId" to unitId,
-                "type" to type.toString(),
-                "startDate" to startDate,
-                "endDate" to endDate
-            )
-        )
+        .bind("id", id)
+        .bind("childId", childId)
+        .bind("unitId", unitId)
+        .bind("type", type.toString())
```
jaahon, line 437: Is there a reason you left this `.toString()` in? Couldn't `PlacementType` implement `DatabaseEnum`?

service/src/main/kotlin/fi/espoo/evaka/shared/dev/DataInitializers.kt
```diff
@@ -663,20 +643,16 @@ fun Database.Transaction.insertTestPlacementPlan(
...
-                "preschoolDaycareStartDate" to preschoolDaycareStartDate,
-                "preschoolDaycareEndDate" to preschoolDaycareEndDate,
-                "updated" to updated,
-                "deleted" to deleted
-            )
-        )
+        .bind("id", id)
+        .bind("applicationId", applicationId)
+        .bind("unitId", unitId)
+        .bind("type", type.toString())
```
jaahon, line 649: Same thing here.

jaahon (approved):

service/src/main/kotlin/fi/espoo/evaka/shared/dev/DataInitializers.kt
Gekkio: I just missed this :sweat_smile:

PR #2868, opened 2022-09-09
Remove wrong info text for now

reynders:
#### Summary
- The vasu email notification seems to be missing functionality. Remove related info text until the functionality really exists
Screenshot 2022-09-09 at 8 51 04


akheron:
It should probably still say something like `Suunnitelma julkaistaan huoltajien nähtäväksi eVakassa.`

PR #2876, opened 2022-09-12
Configurable requested daycare start upper limit

jaahon:
#### Summary

Unlike any other configurable thing I left in a default. Is that a good idea?


service/src/main/kotlin/fi/espoo/evaka/shared/FeatureConfig.kt
```diff
@@ -86,4 +86,7 @@ data class FeatureConfig(
 
     /** Controls whether permission to share is required for curriculum documents */
     val curriculumDocumentPermissionToShareRequired: Boolean,
+
+    /** Optionally change the number of days citizens can move daycare start. Default is 14 days or 2 weeks. */
+    val requestedStartUpperLimit: Int = 14
```
akheron, line 91: I think having a default is a bad idea. Many of these config variables could have a default, but none have.

service/src/main/kotlin/fi/espoo/evaka/decision/Decision.kt
```diff
@@ -30,23 +31,51 @@ data class Decision(
...
+    val endDate: LocalDate,
+    val unit: DecisionUnit,
+    val applicationId: ApplicationId,
+    val childId: ChildId,
+    val childName: String,
+    val sentDate: LocalDate,
+    val status: DecisionStatus,
+    val validRequestedStartDatePeriod: FiniteDateRange
+)
+
```
akheron, line 78: We have talked about minimizing data classes that are almost identical. How about having something like `{ decision: Decision, validRequestedStartDatePeriod: FiniteDaterange }` to avoid copy-pasting all those fields?

An even better option would be to just add `validRequestedStartDatePeriod` to `Decision`, but that would need bigger changes to propagate `featureConfig` to every place where `Decision` instances are created.

service/src/main/kotlin/fi/espoo/evaka/shared/FeatureConfig.kt
akheron: I meant that the default should be removed 🙈

jaahon: I know, I thought I did 🙈 😂

service/src/main/kotlin/fi/espoo/evaka/shared/FeatureConfig.kt
```diff
@@ -86,4 +86,7 @@ data class FeatureConfig(
 
     /** Controls whether permission to share is required for curriculum documents */
     val curriculumDocumentPermissionToShareRequired: Boolean,
+
+    /** Optionally change the number of days citizens can move daycare start */
```
akheron, line 90: One final nitpick: It's not "optional" to change this anymore, like the comment says.

akheron (approved):

service/src/main/kotlin/fi/espoo/evaka/shared/FeatureConfig.kt
jaahon: Right, thanks 😄

PR #2878, opened 2022-09-12
Use index-compatible range operator in condition

Gekkio:
#### Summary

`lower(somerange)` can't use a gist index

### DIFF
--- a/service/src/main/kotlin/fi/espoo/evaka/invoicing/data/FeeDecisionQueries.kt
+++ b/service/src/main/kotlin/fi/espoo/evaka/invoicing/data/FeeDecisionQueries.kt
@@ -326,7 +326,10 @@ fun Database.Read.searchFeeDecisions(
         if (numberParamsRaw.isNotEmpty()) numberQuery else null,
         if (searchTextWithoutNumbers.isNotBlank()) freeTextQuery else null,
         if ((startDate != null || endDate != null) && !searchByStartDate) "daterange(:start_date, :end_date, '[]') && valid_during" else null,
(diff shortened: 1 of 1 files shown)


service/src/main/kotlin/fi/espoo/evaka/invoicing/data/FeeDecisionQueries.kt
```diff
@@ -326,7 +326,10 @@ fun Database.Read.searchFeeDecisions(
         if (numberParamsRaw.isNotEmpty()) numberQuery else null,
         if (searchTextWithoutNumbers.isNotBlank()) freeTextQuery else null,
         if ((startDate != null || endDate != null) && !searchByStartDate) "daterange(:start_date, :end_date, '[]') && valid_during" else null,
-        if ((startDate != null || endDate != null) && searchByStartDate) "daterange(:start_date, :end_date, '[]') @> lower(valid_during)" else null,
+        if ((startDate != null || endDate != null) && searchByStartDate)
+        // valid_during overlaps but does not extend to the left of search range
+        // = start date of valid_during is included in the search range
+            "(valid_during && daterange(:start_date, :end_date, '[]') valid_during &> daterange(:start_date, :end_date, '[]'))" else null,
```
jaahon, line 332: This is missing an `AND`.

Gekkio: :facepalm: 

(also looks like we have 0 tests for this filter)

akheron (approved):
Doesn't work if `startDate == null`, but that problem has always been there 🤷

PR #2883, opened 2022-09-13
Disallow non-negative amounts in invoice rows

akheron:

### DIFF
--- a/service/src/main/kotlin/fi/espoo/evaka/invoicing/service/InvoiceService.kt
+++ b/service/src/main/kotlin/fi/espoo/evaka/invoicing/service/InvoiceService.kt
@@ -73,6 +73,10 @@ class InvoiceService(
         val original = tx.getInvoice(uuid)
             ?: throw BadRequest("No original found for invoice with given ID ($uuid)")
 
+        if (invoice.rows.any { it.amount <= 0 }) {
+            throw BadRequest("Invoice rows amounts must be positive")
+        }
+
         val updated = when (original.status) {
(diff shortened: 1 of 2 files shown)


service/src/main/kotlin/fi/espoo/evaka/invoicing/service/InvoiceService.kt
```diff
@@ -73,6 +73,10 @@ class InvoiceService(
         val original = tx.getInvoice(uuid)
             ?: throw BadRequest("No original found for invoice with given ID ($uuid)")
 
+        if (invoice.rows.any { it.amount < 0 }) {
+            throw BadRequest("Invoice rows cannot have negative amounts")
+        }
```
jaahon, line 78: I don't think `0` really makes sense either.

jaahon (approved):

service/src/main/kotlin/fi/espoo/evaka/invoicing/service/InvoiceService.kt
akheron: My thinking behind allowing `0` was that maybe they want to have "comment rows" or such 🤔

akheron: Although then it would be possible to use `1` as amount and `0` as price, too.

akheron: Fixed 👍

PR #2901, opened 2022-09-15
EVAKA-4481 Associate messages with children

zch:
Help the employees identify who a message is regarding by allowing the guardian to indicate who they are messaging about.

Citizen composing a message:
image

Unit supervisor's inbox:
image

### DIFF
--- a/frontend/src/citizen-frontend/messages/MessageEditor.tsx
+++ b/frontend/src/citizen-frontend/messages/MessageEditor.tsx
@@ -2,21 +2,32 @@
 //
 // SPDX-License-Identifier: LGPL-2.1-or-later
 
-import React, { useCallback, useState } from 'react'
+import React, {
+  useCallback,
+  useContext,
+  useEffect,
+  useMemo,
+  useState
+} from 'react'
 import FocusLock from 'react-focus-lock'
 import styled from 'styled-components'
 
+import { renderResult } from 'citizen-frontend/async-rendering'
+import { ChildrenContext } from 'citizen-frontend/children/state'
 import { Result } from 'lib-common/api'
 import {
   CitizenMessageBody,
-  MessageAccount
+  GetReceiversResponse
 } from 'lib-common/generated/api-types/messaging'
+import { formatPreferredName } from 'lib-common/names'
+import { SelectionChip } from 'lib-components/atoms/Chip'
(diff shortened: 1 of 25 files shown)


frontend/src/citizen-frontend/messages/MessageEditor.tsx
```diff
@@ -105,6 +141,41 @@ export default React.memo(function MessageEditor({
               />
             </label>
 
+            <Gap size="s" />
+            <label>
+              <Bold>{i18n.messages.messageEditor.children}</Bold>
+              {renderResult(children, (children) => (
+                <FixedSpaceFlexWrap horizontalSpacing="xs">
+                  {childIds &&
+                    childIds.length > 1 &&
```
jaahon, line 150: This condition should be outside the label as well.

jaahon (commented):
When citizens reply to a message sent to them there's no child selection and the employees do not see whose parent is sending the message.

Also if the citizen's children are in multiple daycares and a message is simultaneously sent to groups in different units all of the children are shown in the respective group's accounts. It might not be a huge issue but it still leaks personal data that the recipients probably should not get.

zch:
> When citizens reply to a message sent to them there's no child selection and the employees do not see whose parent is sending the message.

Children are now automatically added to the messages created by staff so the names of the children involved are shown to the staff members when they read the messages, but not to the parent (yet. Not sure if necessary?).

> Also if the citizen's children are in multiple daycares and a message is simultaneously sent to groups in different units all of the children are shown in the respective group's accounts. It might not be a huge issue but it still leaks personal data that the recipients probably should not get.

Interesting point, have to think about this a bit. I'm trying to think of relevant cases where I, as a parent, would want to send the same message to different daycares regarding different children. 🤔

jaahon:
> > Also if the citizen's children are in multiple daycares and a message is simultaneously sent to groups in different units all of the children are shown in the respective group's accounts. It might not be a huge issue but it still leaks personal data that the recipients probably should not get.
> 
> Interesting point, have to think about this a bit. I'm trying to think of relevant cases where I, as a parent, would want to send the same message to different daycares regarding different children. 🤔

It's probably not likely nor intended but it's still possible. Just a concern I had.

jaahon (approved):
Looks good to me

PR #2908, opened 2022-09-19
Send financial decision sfi messages to restricted info recipient

reynders:
#### Summary
If financial decision recipient has restricted info enabled, send the message via suomi.fi. If the recipient has the sfi e-channel enabled he will receive the message immediately. If instead the paper channel is enabled, the decision is sent to financial handers who will then manually send it to the proper address

### DIFF
--- a/service/src/main/kotlin/fi/espoo/evaka/invoicing/domain/FeeDecisions.kt
+++ b/service/src/main/kotlin/fi/espoo/evaka/invoicing/domain/FeeDecisions.kt
@@ -187,12 +187,15 @@ data class FeeDecisionDetailed(
 
     val requiresManualSending
         get(): Boolean {
+            if (decisionType !== FeeDecisionType.NORMAL || headOfFamily.forceManualFeeDecisions) {
+                return true
+            }
+
             // Restricted will be sent to allow fast receiving via suomi.fi e-channel.
             if (headOfFamily.restrictedDetailsEnabled) {
                 return false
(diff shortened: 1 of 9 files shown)


service/src/main/kotlin/fi/espoo/evaka/invoicing/domain/FeeDecisions.kt
```diff
@@ -187,7 +187,10 @@ data class FeeDecisionDetailed(
 
     val requiresManualSending
         get(): Boolean {
-            if (decisionType !== FeeDecisionType.NORMAL || headOfFamily.forceManualFeeDecisions) {
+            // Restricted will be sent to allow fast receiving via suomi.fi e-channel.
+            if (headOfFamily.restrictedDetailsEnabled) {
+                return false
+            } else if (decisionType !== FeeDecisionType.NORMAL || headOfFamily.forceManualFeeDecisions) {
```
jaahon, line 193: It's not this part that needs to be changed. Decisions where `decisionType !== FeeDecisionType.NORMAL || headOfFamily.forceManualFeeDecisions` is `true` should still be manually sent. It's the next condition that needs to take into account that people with `restrictedDetailsEnabled` have no address data.

jaahon (changes requested):

service/src/main/kotlin/fi/espoo/evaka/invoicing/domain/FeeDecisions.kt
reynders: Good catch!

jaahon (approved):

PR #2927, opened 2022-09-22
Fix navigation between months in the unit monthly view

zch:
A previous change broke the navigation and caused the user to only be able to navigate one week at a time even when in the monthly view. Now fixed and will navigate one whole month in the month view and one week at a time in the week view.

### DIFF
--- a/frontend/src/employee-frontend/components/unit/TabCalendar.tsx
+++ b/frontend/src/employee-frontend/components/unit/TabCalendar.tsx
@@ -2,7 +2,14 @@
 //
 // SPDX-License-Identifier: LGPL-2.1-or-later
 
-import React, { useContext, useEffect, useMemo, useRef, useState } from 'react'
+import React, {
+  useCallback,
+  useContext,
+  useEffect,
+  useMemo,
+  useRef,
+  useState
+} from 'react'
 import styled from 'styled-components'
 
 import { UserContext } from 'employee-frontend/state/user'
@@ -279,7 +286,6 @@ export default React.memo(function TabCalendar() {
           )}
 
         <StickyTopBar>
-          <div data-qa-week-range={week.dateRange.toString()} />
           <FixedSpaceRow spacing="s" alignItems="center">
             <GroupSelectorWrapper>
               <GroupSelector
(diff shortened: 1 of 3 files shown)


frontend/src/employee-frontend/components/unit/TabCalendar.tsx
```diff
@@ -406,3 +405,61 @@ const Caretakers = React.memo(function Caretakers({
...
+  const subUnitOfTime = useCallback(
+    (date: LocalDate) =>
+      mode === 'week' ? date.subDays(7) : date.subMonths(1),
+    [mode]
+  )
+
+  return (
+    <>
+      <ColoredH3 noMargin>
+        {startDate.format('dd.MM.')}–{endDate.format('dd.MM.yyyy')}
```
jaahon, line 442: Would this be even better if it was just eg. "September 2022" in month mode instead of the date range?

frontend/src/employee-frontend/components/unit/TabCalendar.tsx
```diff
@@ -279,7 +286,16 @@ export default React.memo(function TabCalendar() {
...
+          <div
+            data-qa-date-range={
+              mode === 'week'
+                ? week.dateRange.toString()
+                : new FiniteDateRange(
+                    selectedDate.startOfMonth(),
+                    selectedDate.lastDayOfMonth()
+                  ).toString()
+            }
+          />
```
jaahon, line 298: I think this could be moved inside `ActiveDateRangeSelector` component so that you don't have to form the actual date range twice.

jaahon (approved):
It works well, just have two minor comments.

frontend/src/employee-frontend/components/unit/TabCalendar.tsx
zch: From a readability point of view, yes. But I tried this and it makes it awful to navigate more than one month: The month names are so different in length that the arrows jump _a lot_ making it impossible to click the same arrow twice and easy to go e.g. forward 1 and back again by mistake because the position of the arrows changed underneath the mouse cursor.

Thought about moving the arrows before the month name, but then it is more unclear what the arrows refer to. I think leaving the dates is the better option after all.

frontend/src/employee-frontend/components/unit/TabCalendar.tsx
zch: Good idea, done.

frontend/src/employee-frontend/components/unit/TabCalendar.tsx
jaahon: Oh yeah, I had already forgotten about that. Some dates are longer than others, so they kind of have the same problem 😅. Could a `MM.yyyy` or `M.yyyy` formatted date work?

PR #2930, opened 2022-09-23
Citizen scroll page to top on navigation

jaahon:
#### Summary

In citizen frontend scroll every page to the top on navigation. Navigating back behaves similarly because there's no simple way to make it work like in a traditional web page.

Also:
- Fix messages reply button focus outline
- Remove back button in child page

### DIFF
--- a/frontend/src/citizen-frontend/App.tsx
+++ b/frontend/src/citizen-frontend/App.tsx
@@ -3,12 +3,11 @@
 // SPDX-License-Identifier: LGPL-2.1-or-later
 
 import { ErrorBoundary } from '@sentry/react'
-import React, { ReactNode, useCallback, useContext, useRef } from 'react'
+import React, { ReactNode, useCallback, useContext } from 'react'
 import { Route, BrowserRouter, Navigate, Routes } from 'react-router-dom'
 import { ThemeProvider } from 'styled-components'
 import styled from 'styled-components'
 
-import { scrollElementToPos } from 'lib-common/utils/scrolling'
 import SkipToContent from 'lib-components/atoms/buttons/SkipToContent'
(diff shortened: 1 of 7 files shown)


frontend/src/citizen-frontend/ScrollToTop.tsx
```diff
@@ -0,0 +1,23 @@
...
+import { useLocation } from 'react-router-dom'
+
+import { scrollToPos } from 'lib-common/utils/scrolling'
+
+export default React.memo(function ScrollToTop() {
+  const location = useLocation()
+
+  useLayoutEffect(() => {
+    // /messages/:id already scrolls to the lastest message in a thread on navigation
+    if ((location.pathname.split('/messages')[1]?.length ?? 0) > 0) {
```
akheron, line 15: This is ugly 😬

frontend/src/citizen-frontend/App.tsx
```diff
@@ -94,33 +94,17 @@ const Content = React.memo(function Content() {
...
-        top: 0,
-        left: 0,
-        behavior: 'auto'
-      }),
-    []
-  )
-
   return (
     <FullPageContainer>
+      <ScrollToTop />
```
akheron, line 99: This might require the fewest keystrokes, but the condition whether to scroll or not is hidden inside the component.

Maybe `<Route path="..." element={<ScrollToTop><MapPage /><ScrollToTop>} />` would be a good middle ground?

jaahon: I agree that it's ugly, but in my opinion it's better that it's enabled everywhere by default. But I guess it's very likely that if a new route is added it's wrapped in `<ScrollToTop>` like other routes.

akheron (approved):

PR #2935, opened 2022-09-26
Messaging speedups

akheron:
#### Summary

Create a new cache table that makes fast SQL queries possible. Write those fast SQL queries and use them instead of the previous mega query.

These optimizations only apply to citizen and employee (personal and group) inboxes. The employee sent folder is not affected, as it doesn't seem to be a performance bottleneck, and it currently works very differently to how it _should_ work.


service/src/main/resources/db/migration/V272__message_thread_participant.sql
```diff
@@ -0,0 +1,18 @@
...
+    last_message_timestamp  timestamp with time zone NOT NULL,  -- timestamp of last sent or received message
+    last_received_timestamp timestamp with time zone,           -- timestamp of last received message
+    last_sent_timestamp     timestamp with time zone,           -- timestamp of last sent message
+
+    CONSTRAINT uniq$thread_participant UNIQUE (thread_id, participant_id)
+);
+
+CREATE INDEX idx$thread_participant_message ON message_thread_participant (participant_id, last_message_timestamp);
+CREATE INDEX idx$thread_participant_received ON message_thread_participant (participant_id, last_received_timestamp) WHERE last_received_timestamp IS NOT NULL;
+CREATE INDEX idx$thread_participant_sent ON message_thread_participant (participant_id, last_sent_timestamp) WHERE last_sent_timestamp IS NOT NULL;
```
akheron, line 17: The `last_sent_timestamp` and the `_sent` indexes are not used currently. I decided to add them however to make future changes easier (and migrations faster). There's a chance that they are left unused, but let's try to remember to remove then if that happens. I feel that removing would be a faster operation than adding them later.

service/src/integrationTest/kotlin/fi/espoo/evaka/messaging/MessageQueriesTest.kt
```diff
@@ -144,37 +145,36 @@ class MessageQueriesTest : PureJdbiTest(resetDbBeforeEach = true) {
...
-                    1, true
-                )
+                it.getThreads(person2Account, 10, 1)
             }.data.flatMap { it.messages.mapNotNull { m -> m.readAt } }.size
         )
 
         // when employee gets a reply
         db.transaction {
-            val recipients = listOf(employee1Account)
+            // TODO: Replying to a thread should be MessageService's job
```
akheron, line 154: TODOs in tests will be addressed in the next PR 😉

service/src/main/resources/db/migration/V272__message_thread_participant.sql
```diff
@@ -0,0 +1,17 @@
+CREATE TABLE message_thread_participant
+(
+    id                      uuid PRIMARY KEY         NOT NULL DEFAULT ext.uuid_generate_v1mc(),
+    created                 timestamp with time zone NOT NULL DEFAULT now(),
+    updated                 timestamp with time zone NOT NULL DEFAULT now(),
+    thread_id               uuid                     NOT NULL REFERENCES message_thread (id) ON DELETE CASCADE,
+    participant_id          uuid                     NOT NULL REFERENCES message_account (id) ON DELETE CASCADE,
+    last_message_timestamp  timestamp with time zone NOT NULL,  -- timestamp of last sent or received message
+    last_received_timestamp timestamp with time zone,           -- timestamp of last received message
+    last_sent_timestamp     timestamp with time zone,           -- timestamp of last sent message
```
akheron, line 10: `last_sent_timestamp` is currently used, but might be used in the future. I thought it's better to compute it in this data migration rather than have more slow data migrations in the future.

service/src/main/kotlin/fi/espoo/evaka/messaging/MessageQueries.kt
```diff
@@ -206,133 +244,161 @@ data class ReceivedMessageResultItem(
...
-        FROM attachment att WHERE att.message_content_id = msg.content_id
-    ) AS attachments
-    FROM threads t
-    JOIN participated_messages msg ON msg.thread_id = t.id
-    ORDER BY t.last_message DESC, msg.sent_at ASC
-    """.trimIndent()
-
-    return createQuery(sql)
+    coalesce((
+        SELECT json_agg(json_build_object(
```
Gekkio, line 273: `json` -> `jsonb`

akheron: 👍

PR #2948, opened 2022-09-29
EVAKA-4532 foster parents UI

jaahon:
#### Summary

Add foster parents and children sections to person and child pages. Hide the sections behind a feature flag for now.

Also:
- Replace deprecated CollapsibleSection in PersonProfile.
- Do not always show date error when input is empty (it would show errors even when it was ok that a date is null)


zch (commented):
Implementation looks good, but I'd like to see some tests.

zch (approved):

PR #2949, opened 2022-09-29
EVAKA-4532 treat foster parents like guardians

jaahon:
#### Summary

Gives foster parents access to all citizen features apart from child income. Also removes the ability to send messages for fridge parents (they shouldn't have it anyway).

### DIFF
--- a/service/src/main/kotlin/fi/espoo/evaka/messaging/MessageQueries.kt
+++ b/service/src/main/kotlin/fi/espoo/evaka/messaging/MessageQueries.kt
@@ -497,24 +497,6 @@ WITH backup_care_placements AS (
         SELECT 1 FROM daycare u
         WHERE p.unit_id = u.id AND 'MESSAGING' = ANY(u.enabled_pilot_features)
     )
-
-    UNION
-
-    SELECT p.id, p.unit_id, p.child_id, p.group_id
-    FROM fridge_child fg
-    JOIN backup_care p ON fg.child_id = p.child_id AND daterange(p.start_date, p.end_date, '[]') @> :today
-    WHERE daterange(fg.start_date, fg.end_date, '[]') @> :today
-    AND fg.head_of_child = (SELECT person_id AS id FROM message_account WHERE id = :accountId)
-    AND fg.conflict = false
-    AND NOT EXISTS (
-        SELECT 1 FROM messaging_blocklist b
-        WHERE b.child_id = p.child_id
-        AND b.blocked_recipient = fg.head_of_child
-    )
-    AND EXISTS (
-        SELECT 1 FROM daycare u
-        WHERE p.unit_id = u.id AND 'MESSAGING' = ANY(u.enabled_pilot_features)
-    )
 ), placements AS (
     SELECT p.id, p.unit_id, p.child_id
     FROM guardian g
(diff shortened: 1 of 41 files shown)


service/src/main/kotlin/fi/espoo/evaka/shared/security/Action.kt
```diff
@@ -218,37 +218,37 @@ sealed interface Action {
...
-            READ_VASU_DOCUMENT_SUMMARIES(IsCitizen(allowWeakLogin = false).guardianOfChild()),
-            CREATE_ABSENCE(IsCitizen(allowWeakLogin = true).guardianOfChild()),
+            READ_PLACEMENT_STATUS_BY_APPLICATION_TYPE(IsCitizen(allowWeakLogin = false).guardianOfChild(), IsCitizen(allowWeakLogin = false).fosterParentOfChild()),
+            READ_DUPLICATE_APPLICATIONS(IsCitizen(allowWeakLogin = false).guardianOfChild(), IsCitizen(allowWeakLogin = false).fosterParentOfChild()),
+            READ_VASU_DOCUMENT_SUMMARIES(IsCitizen(allowWeakLogin = false).guardianOfChild(), IsCitizen(allowWeakLogin = false).fosterParentOfChild()),
+            CREATE_ABSENCE(IsCitizen(allowWeakLogin = true).guardianOfChild(), IsCitizen(allowWeakLogin = true).fosterParentOfChild()),
 
-            CREATE_HOLIDAY_ABSENCE(IsCitizen(allowWeakLogin = true).guardianOfChild()),
-            CREATE_RESERVATION(IsCitizen(allowWeakLogin = true).guardianOfChild()),
+            CREATE_HOLIDAY_ABSENCE(IsCitizen(allowWeakLogin = true).guardianOfChild(), IsCitizen(allowWeakLogin = true).fosterParentOfChild()),
```
reynders, line 234: Maybe worth having a combined guardianOrFosterParentOfChild function to shorten this?

reynders (approved):

service/src/main/kotlin/fi/espoo/evaka/messaging/MessageQueries.kt
```diff
@@ -771,6 +733,13 @@ fun Database.Read.getReceiversForNewMessage(
             FROM guardian g
             LEFT JOIN messaging_blocklist bl ON g.guardian_id = bl.blocked_recipient AND c.child_id = bl.child_id
             WHERE g.child_id = c.child_id AND bl.id IS NULL
+
+            UNION ALL
+
+            SELECT 1
+            FROM foster_parent fp
```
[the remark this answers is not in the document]
jaahon: Yes, definitely. Good catch!

service/src/main/kotlin/fi/espoo/evaka/messaging/MessageQueries.kt
```diff
@@ -848,6 +817,18 @@ WHERE NOT EXISTS (
     WHERE bl.child_id = c.child_id
     AND bl.blocked_recipient = g.guardian_id
 )
+
+UNION
+
+SELECT DISTINCT acc.id as account_id, c.child_id
+FROM children c
+JOIN foster_parent fp ON fp.child_id = c.child_id
```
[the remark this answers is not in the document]
jaahon: There is one, although it's in an odd place (the row below this one) 😅

service/src/main/kotlin/fi/espoo/evaka/shared/security/Action.kt
jaahon: Yeah, good idea, since we have no need to authorize only foster parents.

jaahon: Actually now that I think of it I think it's better that the rules are separate so that we don't have to create multiple variations if there is ever a need to separate them.

PR #2950, opened 2022-09-30
Improve audit logging - part 2

Gekkio:
#### Summary

- include more context in audit events
- improve event naming


service/src/main/kotlin/fi/espoo/evaka/attendance/RealtimeStaffAttendanceController.kt
```diff
@@ -173,12 +176,13 @@ class RealtimeStaffAttendanceController(
                             occupancyCoefficientSeven
                         )
                     }
+                    Pair(staffAttendanceIds, externalStaffAttendanceIds)
```
zch, line 179: Nit: Why not return a list here, since the `Pair` is deconstructed only to later be put in a list?

zch (approved):
Looks great 🏅

PR #2956, opened 2022-10-03
Fix keycloak fonts

patari:
#### Summary

Fixed error "Refused to load the stylesheet 'https://fonts.googleapis.com/css?family=Montserrat:200,300,400,700|Open+Sans:300,400,600,700' because it violates the following Content Security Policy directive".

Also fixes CSS file 404 error.

### DIFF
--- a/proxy/files/etc/nginx/conf.d/nginx.conf.template
+++ b/proxy/files/etc/nginx/conf.d/nginx.conf.template
@@ -149,7 +149,7 @@ server {
   add_header Report-To '{"group","csp-endpoint","max_age":31536000,"endpoints":[{"url":"https://$host/api/csp"}]}';
 
   set $loadFailedInlineScript "'sha256-z1vaAvxob9VDuw7klCB049Y2Xr6lf7KjhDrsLvsvcPU='"; # SHA256 calculated from the script contents
-  set $contentSecurityPolicyBase "block-all-mixed-content; upgrade-insecure-requests; default-src 'self'; font-src 'self' data:; style-src 'self' 'unsafe-inline'; img-src 'self' data: https://a.tile.openstreetmap.org https://b.tile.openstreetmap.org https://c.tile.openstreetmap.org; connect-src 'self' https://o318158.ingest.sentry.io https://api.digitransit.fi; object-src 'none'; report-uri /api/csp; report-to csp-endpoint";
+  set $contentSecurityPolicyBase "block-all-mixed-content; upgrade-insecure-requests; default-src 'self'; font-src 'self' https://fonts.gstatic.com data:; style-src 'self' https://fonts.googleapis.com 'unsafe-inline'; img-src 'self' data: https://a.tile.openstreetmap.org https://b.tile.openstreetmap.org https://c.tile.openstreetmap.org; connect-src 'self' https://o318158.ingest.sentry.io https://api.digitransit.fi; object-src 'none'; report-uri /api/csp; report-to csp-endpoint";
   add_header Content-Security-Policy "${contentSecurityPolicyBase}; script-src 'self' ${loadFailedInlineScript}; frame-ancestors 'none'; form-action 'self';";
 
   # Tracing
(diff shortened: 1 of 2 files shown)


proxy/files/etc/nginx/conf.d/nginx.conf.template
```diff
@@ -149,7 +149,7 @@ server {
   add_header Report-To '{"group","csp-endpoint","max_age":31536000,"endpoints":[{"url":"https://$host/api/csp"}]}';
 
   set $loadFailedInlineScript "'sha256-z1vaAvxob9VDuw7klCB049Y2Xr6lf7KjhDrsLvsvcPU='"; # SHA256 calculated from the script contents
-  set $contentSecurityPolicyBase "block-all-mixed-content; upgrade-insecure-requests; default-src 'self'; font-src 'self' data:; style-src 'self' 'unsafe-inline'; img-src 'self' data: https://a.tile.openstreetmap.org https://b.tile.openstreetmap.org https://c.tile.openstreetmap.org; connect-src 'self' https://o318158.ingest.sentry.io https://api.digitransit.fi; object-src 'none'; report-uri /api/csp; report-to csp-endpoint";
+  set $contentSecurityPolicyBase "block-all-mixed-content; upgrade-insecure-requests; default-src 'self'; font-src 'self' https://fonts.gstatic.com data:; style-src 'self' https://fonts.googleapis.com 'unsafe-inline'; img-src 'self' data: https://a.tile.openstreetmap.org https://b.tile.openstreetmap.org https://c.tile.openstreetmap.org; connect-src 'self' https://o318158.ingest.sentry.io https://api.digitransit.fi; object-src 'none'; report-uri /api/csp; report-to csp-endpoint";
```
akheron, line 152: This change affects all proxied content, not just keycloak. Line 344 would be the correct place.

BTW, wou should can start the proxy with `KEYCLOAK_CONFIGURATION` env variable to use a custom keycloak config. Espoo infra has this in terraform:

```
  environment = {
    ...
    KEYCLOAK_CONFIGURATION    = "include keycloak.conf;"
  }
```

and `keycloak.conf` contains nginx configuration that's specific to keycloak.

patari: Why shouldn't we share config between installations? Seems extra work for everybody

patari: > This change affects all proxied content, not just keycloak. Line 344 would be the correct place.

Fixed.

akheron: @ilja-git might know why the keycloak config override exists.

patari: https://github.com/espoon-voltti/evaka/pull/2956/commits/44bac9883e75587ec1c75bed6f47c8b155328cfd doesn't work:
```
Ignoring duplicate Content-Security-Policy directive 'font-src'.
Ignoring duplicate Content-Security-Policy directive 'style-src'.
```

Header is: `block-all-mixed-content; upgrade-insecure-requests; default-src 'self'; font-src 'self' data:; style-src 'self' 'unsafe-inline'; img-src 'self' data: https://a.tile.openstreetmap.org https://b.tile.openstreetmap.org https://c.tile.openstreetmap.org; connect-src 'self' https://o318158.ingest.sentry.io https://api.digitransit.fi; object-src 'none'; report-uri /api/csp; report-to csp-endpoint; script-src 'self' 'unsafe-inline' 'unsafe-eval'; font-src 'self' https://fonts.gstatic.com data:; style-src 'self' https://fonts.googleapis.com 'unsafe-inline'; frame-ancestors 'self'; form-action 'self' https://testi.apro.tunnistus.fi/idp/profile/SAML2/POST/SSO https://testi.apro.tunnistus.fi/idp/profile/SAML2/POST/SLO`..

ilja-git: There is no issue to put it on base variable. I see more issues that it is Google fonts that is allowed.

patari: It's coming from here: https://github.com/espoon-voltti/evaka/blob/36c9ba6faf38a17047f48a98752c15ce39fbd763/keycloak/theme/evaka/login/template.ftl#L9

Maybe quick fix to revert https://github.com/espoon-voltti/evaka/commit/44bac9883e75587ec1c75bed6f47c8b155328cfd and a long term plan to bundle fonts to theme (like frontends)?

ilja-git: Sounds good to me.

PR #2960, opened 2022-10-04
Add missing sortings

reynders:
#### Summary
- Sort voucher value unit report by area and unit name
- Sort groups on employee mobile attendance UI


service/src/main/kotlin/fi/espoo/evaka/attendance/MobileUnitController.kt
```diff
@@ -37,10 +37,13 @@ class MobileUnitController(private val accessControl: AccessControl) {
...
         return db.connect { dbc ->
             dbc.read { tx ->
-                tx.fetchUnitInfo(
+                val unsorted = tx.fetchUnitInfo(
                     unitId,
                     clock.today(),
                 )
+                unsorted.copy(
+                    groups = unsorted.groups.sortedBy { it.name }
+                )
```
akheron, line 46: You should add `ORDER BY` to the SQL query, or at least move this `sortedBy` call inside the `fetchUnitInfo` function. This is the only place where the function is used, so there are no other users who would expect a different ordering.

akheron (changes requested):

akheron (approved):

PR #2965, opened 2022-10-05
Fix children not being invoiced when they are split between parents

jaahon:
#### Summary

Invoice generation assumed that families with children that are split between two fridge parents would still have two fee decisions, one for each parent. This caused some of the children not getting an invoice if their parent was not the head of family on a fee decision.

### DIFF
--- a/service/src/main/kotlin/fi/espoo/evaka/invoicing/service/InvoiceGenerator.kt
+++ b/service/src/main/kotlin/fi/espoo/evaka/invoicing/service/InvoiceGenerator.kt
@@ -38,6 +38,7 @@ import fi.espoo.evaka.shared.domain.asDistinctPeriods
 import fi.espoo.evaka.shared.domain.europeHelsinki
 import fi.espoo.evaka.shared.domain.mergePeriods
 import fi.espoo.evaka.shared.domain.operationalDays
+import org.jdbi.v3.core.mapper.Nested
 import org.springframework.stereotype.Component
 import java.time.Duration
 import java.time.LocalDate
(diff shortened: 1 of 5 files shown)


service/src/main/kotlin/fi/espoo/evaka/invoicing/service/InvoiceGenerator.kt
```diff
@@ -330,33 +334,51 @@ fun Database.Read.getAbsenceStubs(spanningRange: DateRange, categories: Collecti
...
+"""
     )
         .bind("period", spanningPeriod)
-        .bind("invoicedTypes", PlacementType.invoiced())
-        .map { row -> row.mapColumn<ChildId>("child_id") to row.mapRow<PlacementRow>() }
-        .groupBy { (childId) -> childId }
-        .mapValues { it.value.map { (_, placements) -> placements } }
+        .bind("types", arrayOf(PlacementType.TEMPORARY_DAYCARE, PlacementType.TEMPORARY_DAYCARE_PART_DAY))
+        .map { rv -> rv.mapColumn<DateRange>("date_range") to rv.mapRow<PlacementStub>() }
+        .groupBy { it.second.child.id }
```
akheron, line 381: It seems to me that the only difference between these two SQL queries is the value passed to `invoicedTypes`/`types`, so they should be merged I think.

service/src/main/kotlin/fi/espoo/evaka/placement/PlacementType.kt
```diff
@@ -49,6 +49,8 @@ enum class PlacementType : DatabaseEnum {
     override val sqlType: String = "placement_type"
 
     companion object {
-        fun invoiced() = values().filter { it.isInvoiced() }
+        fun invoiced() = values().filter { it.isInvoiced() }.filterNot { temporary().contains(it) }
+
+        fun temporary() = listOf(TEMPORARY_DAYCARE, TEMPORARY_DAYCARE_PART_DAY)
```
akheron, line 54: This could also be just `val temporary = listOf(...)`

akheron (approved):

PR #2974, opened 2022-10-07
Show backup care child multiplier in employee group view

reynders:
#### Summary
Show backup care child multiplier on employee group view


service/src/main/kotlin/fi/espoo/evaka/placement/PlacementService.kt
```diff
@@ -314,14 +314,14 @@ fun Database.Read.getUnitChildrenCapacities(
                 WHEN extract(YEARS FROM age(:date, ch.date_of_birth)) < 3 THEN coalesce(sno.occupancy_coefficient_under_3y, default_sno.occupancy_coefficient_under_3y)
                 ELSE coalesce(sno.occupancy_coefficient, default_sno.occupancy_coefficient, 1)
             END) AS service_need_factor
-        FROM placement pl
+        FROM realized_placement_all(:date) pl
```
jaahon, line 317: I think you should use `realized_placement_one` instead of `realized_placement_all` since you only want one value for each child.

jaahon (approved):

PR #2980, opened 2022-10-10
Hide decisions waiting mailing from citizen

jaahon:
#### Summary

Citizens would previously see decisions waiting mailing in their decisions page. Accepting or rejecting those decisions would still fail. Hide them until they are marked as sent.


zch (approved):
LGTM

PR #2987, opened 2022-10-11
Make sure financial decisions before 1.9.2019 are not sent to varda

reynders:


service/src/main/kotlin/fi/espoo/evaka/varda/VardaQueries.kt
```diff
@@ -267,6 +267,7 @@ WITH child_fees AS (
         JOIN daycare d ON d.id = p.unit_id
         JOIN child_fees ON p.child_id = child_fees.child_id
             AND child_fees.valid_during && daterange(sn.start_date, sn.end_date, '[]')
+            AND sn.end_date >= '2019-09-01'
```
jaahon, line 270: Not directly related to this PR only, but I think it would be a good idea to extract all these dates into variables with names that explain their meaning a bit.

jaahon (approved):

PR #2989, opened 2022-10-12
Add voucher value decision difference sort

patari:
#### Summary

Filter and data are now in the same order.


frontend/src/employee-frontend/components/voucher-value-decisions/VoucherValueDecisionDifferenceIcon.tsx
```diff
@@ -10,27 +10,58 @@ import { VoucherValueDecisionDifference } from 'lib-common/generated/api-types/i
...
-    >
+    <Tooltip key={props.item.value} tooltip={props.item.label} position="top">
       <RoundIcon
-        content={i18n.valueDecisions.table.difference.valueShort[props.value]}
+        content={props.item.labelShort}
         color={theme.colors.main.m1}
         size="L"
-        onClick={
-          props.toggle !== undefined ? props.toggle(props.value) : undefined
+        onClick={(_) =>
```
jaahon, line 61: You can just completely omit the parameter if it's not used
```suggestion
        onClick={() =>
```

jaahon (approved):

PR #2990, opened 2022-10-12
Add fee decision difference filter

patari:
#### Summary

https://github.com/espoon-voltti/evaka/pull/2961 & https://github.com/espoon-voltti/evaka/pull/2989 for fee decisions.

### DIFF
--- a/service/src/main/kotlin/fi/espoo/evaka/invoicing/domain/FeeDecisions.kt
+++ b/service/src/main/kotlin/fi/espoo/evaka/invoicing/domain/FeeDecisions.kt
@@ -5,6 +5,7 @@
 package fi.espoo.evaka.invoicing.domain
 
 import com.fasterxml.jackson.annotation.JsonProperty
+import fi.espoo.evaka.ConstList
 import fi.espoo.evaka.placement.PlacementType
 import fi.espoo.evaka.shared.ChildId
 import fi.espoo.evaka.shared.DaycareId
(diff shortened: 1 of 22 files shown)


service/src/main/kotlin/fi/espoo/evaka/invoicing/domain/FeeDecisions.kt
```diff
@@ -148,6 +140,42 @@ enum class FeeDecisionType {
...
+                return if (GUARDIANS.contentEquals(d1, d2)) emptySet() else setOf(GUARDIANS)
+            }
+            return values().filterNot { it.contentEquals(d1, d2) }.toSet()
+        }
+
+        private fun <T> childrenEquals(d1: FeeDecision, d2: FeeDecision, fn: (child: FeeDecisionChild) -> T): Boolean {
+            val map1 = d1.children.associateBy({ it.child.id }, fn)
+            val map2 = d2.children.associateBy({ it.child.id }, fn)
+            // equals only with matching keys
+            return map1.all { (key, value) -> map2.getOrDefault(key, value) == value }
```
jaahon, line 174: I think this should be just a normal equality check, ie. `map1 == map2`.

patari: That would mean every child change on decision there would be difference with income, placement, service need, sibling discount and fee alterations. Because of that it was said "in the spec" that it should show only `CHILDREN` difference when a child is added/removed (if there are no other changes of course, well sibling discount can be different by just child added/removed, but it already works with current implementation).

jaahon: Ah yes, now I get it!

jaahon (approved):

patari:
Rebased to fix conflict in Filters.tsx (it was only a small import conflict)

PR #3006, opened 2022-10-14
Upgrade ktlint and kotlinter

akheron:


Gekkio (approved):
No more trailing commas :cry:

PR #3010, opened 2022-10-14
Use DocumentLang enum everywhere

akheron:


jaahon (approved):

service/src/main/kotlin/fi/espoo/evaka/decision/DecisionQueries.kt
```diff
@@ -280,7 +281,7 @@ fun Database.Read.getDecisionLanguage(decisionId: DecisionId): String {
         """.trimIndent().trimIndent()
     )
         .bind("id", decisionId)
-        .mapTo<String>()
+        .mapTo<DocumentLang>()
```
Gekkio, line 284: Isn't `daycare.language` in lowercase? Does this `mapTo` still work even though the enum variants are in uppercase?

PR #3018, opened 2022-10-18
Add income statement search filter: placement valid date

patari:
#### Summary

Adds new filter to income statement search.


service/src/main/kotlin/fi/espoo/evaka/incomestatement/IncomeStatementQueries.kt
```diff
@@ -535,6 +535,7 @@ WHERE handler_id IS NULL
 AND (cardinality(:areas) = 0 OR ca.short_name = ANY(:areas))
 AND (cardinality(:providerTypes) = 0 OR d.provider_type = ANY(:providerTypes::unit_provider_type[]))
 AND daterange(:sentStartDate, :sentEndDate, '[]') @> i.created::date
+AND (:placementValidDate IS NULL OR daterange(p.start_date, p.end_date, '[]') @> :placementValidDate)
```
jaahon, line 538: This condition is always true when `p.start_date` and `p.end_date` are `null` (ie. when a child has no placement), which means that the condition does not work as expected if any of the children don't have a placement.

patari: Fixed

jaahon (approved):

PR #3022, opened 2022-10-19
Add voucher value decision search filters

patari:
#### Summary

Same filters as in fee decision search.

### DIFF
--- a/service/src/main/kotlin/fi/espoo/evaka/invoicing/data/VoucherValueDecisionQueries.kt
+++ b/service/src/main/kotlin/fi/espoo/evaka/invoicing/data/VoucherValueDecisionQueries.kt
@@ -306,11 +306,8 @@ fun Database.Read.searchValueDecisions(
             VoucherValueDecisionSortParam.STATUS -> "decision.status"
         }
 
-    val retroactiveOnly =
-        distinctiveParams.contains(VoucherValueDecisionDistinctiveParams.RETROACTIVE)
-
     val params =
-        listOfNotNull(
+        listOf(
             Binding.of("page", page),
             Binding.of("pageSize", pageSize),
             Binding.of("status", status),
@@ -322,7 +319,7 @@ fun Database.Read.searchValueDecisions(
             Binding.of("financeDecisionHandlerId", financeDecisionHandlerId),
             Binding.of("difference", difference),
             Binding.of("firstPlacementStartDate", evakaClock.now().toLocalDate().withDayOfMonth(1)),
-            if (retroactiveOnly) Binding.of("now", evakaClock.now()) else null
+            Binding.of("now", evakaClock.now())
         )
 
     val (freeTextQuery, freeTextParams) =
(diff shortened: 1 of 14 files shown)


service/src/main/kotlin/fi/espoo/evaka/invoicing/data/VoucherValueDecisionQueries.kt
```diff
@@ -305,23 +306,34 @@ fun Database.Read.searchValueDecisions(
...
             Binding.of("areas", areas),
             Binding.of("unit", unit),
+            Binding.of("postOffice", postOffice),
             Binding.of("start_date", startDate),
             Binding.of("end_date", endDate),
             Binding.of("financeDecisionHandlerId", financeDecisionHandlerId),
             Binding.of("difference", difference),
-            Binding.of("firstPlacementStartDate", evakaClock.now().toLocalDate().withDayOfMonth(1))
+            Binding.of("firstPlacementStartDate", evakaClock.now().toLocalDate().withDayOfMonth(1)),
+            if (retroactiveOnly) Binding.of("now", evakaClock.now()) else null
```
jaahon, line 325: This binding doesn't have to be included conditionally, there are other bindings already that are not always used in the query.

jaahon (commented):
The way that the `UNCONFIRMED_HOURS` filter works now is a bit different compared to fee decisions. With fee decisions it's a property of the decision so the filter works the same way even if a child has their service need updates. In theory draft decisions should stay in sync with service need so the filter should work roughly the same for drafts but the behaviour is very different for sent and annulled decisions. I think both of the implementations should behave the same.

service/src/main/kotlin/fi/espoo/evaka/invoicing/data/VoucherValueDecisionQueries.kt
patari: Yeah, should we add `service_need_missing` to `voucher_value_decision`-table?

patari:
> The way that the `UNCONFIRMED_HOURS` filter works now is a bit different compared to fee decisions. With fee decisions it's a property of the decision so the filter works the same way even if a child has their service need updates. In theory draft decisions should stay in sync with service need so the filter should work roughly the same for drafts but the behaviour is very different for sent and annulled decisions. I think both of the implementations should behave the same.

Yeah, should we add `service_need_missing` to `voucher_value_decision`-table?

jaahon:
> > The way that the `UNCONFIRMED_HOURS` filter works now is a bit different compared to fee decisions. With fee decisions it's a property of the decision so the filter works the same way even if a child has their service need updates. In theory draft decisions should stay in sync with service need so the filter should work roughly the same for drafts but the behaviour is very different for sent and annulled decisions. I think both of the implementations should behave the same.
> 
> Yeah, should we add `service_need_missing` to `voucher_value_decision`-table?

I think that would be the ideal solution.

jaahon (approved):

PR #3023, opened 2022-10-20
EVAKA-4471 mass messaging

jaahon:
#### Summary

* Refactor message recipients queries to work with municipal accounts.
* Let unit supervisors message all of their units simultaneously.
* Use the new TreeDropdown component in employee message editor.
* Create a new `MESSAGING` user role.
* Add a municipal messaging account type that can be used by users with `ADMIN` or `MESSAGING` roles.


service/src/main/kotlin/fi/espoo/evaka/messaging/MessageQueries.kt
```diff
@@ -801,32 +818,98 @@ fun Database.Read.getReceiversForNewMessage(
...
+        .map { (group, children) ->
+            val (groupId, groupName) = group
+            MessageReceiver(
+                id = groupId,
+                type = MessageRecipientType.GROUP,
+                name = groupName,
                 receivers =
-                    receiverChildren.map { child ->
+                    children.map {
                         MessageReceiver(
```
reynders, line 904: Should these be sorted by child's name?

service/src/main/kotlin/fi/espoo/evaka/messaging/MessageQueries.kt
```diff
@@ -801,32 +818,98 @@ fun Database.Read.getReceiversForNewMessage(
...
+                val areas = receivers.groupBy { it.areaId to it.areaName }
+                val accountReceivers =
+                    areas.map { (area, units) ->
+                        val (areaId, areaName) = area
+                        MessageReceiver(
+                            id = areaId,
+                            type = MessageRecipientType.AREA,
+                            name = areaName,
+                            receivers =
+                                units.map { unit ->
```
reynders, line 877: Should these be sorted by unit's name?

frontend/src/lib-components/employee/messages/MessageEditor.tsx
```diff
@@ -217,39 +199,20 @@ export default React.memo(function MessageEditor({
...
       setMessage((old) => ({
         ...old,
-        recipientIds: selected.map((s) => s.selectorId),
-        recipientNames: selected.map((s) => s.name)
+        recipientIds: selectedReceivers.map((s) => s.key),
+        recipientNames: selectedReceivers.map((s) => s.text)
       }))
     },
-    [receiverTree]
+    [selectedReceivers]
```
akheron, line 219: Both `useEffect`s here should be removed as per [You Might Not Need an Effect](https://beta.reactjs.org/learn/you-might-not-need-an-effect).

When `useEffect` is used to adjust state based on user input (message content or recipient changes in these cases), it would be better to do the state update in the event handler itself. Using `useEffect` causes unnecessary rerenders.

service/src/main/resources/db/migration/R__message_account_name_view.sql
```diff
@@ -1,14 +1,16 @@
...
+    type,
     account_name
 ) AS (
 SELECT
     acc.id,
+    acc.type,
     COALESCE(
         nullif(concat_ws(' - ', d.name, dg.name), ''),
         nullif(concat_ws(' ', p.last_name, p.first_name), ''),
-        concat_ws(' ', e.last_name, coalesce(e.preferred_first_name, e.first_name))
```
akheron, line 11: Instead of making changes to `message_account_name_view`, you should probably make the changes to `message_account_view` and use it instead. In my experience, it's quite a bit faster and should be used everywhere. I just didn't have time to change all places to use it back then.

service/src/main/resources/db/migration/V281__municipal_message_account.sql
```diff
@@ -0,0 +1,15 @@
...
+    CASE
+        WHEN employee_id IS NOT NULL THEN 'PERSONAL'::message_account_type
+        WHEN daycare_group_id IS NOT NULL THEN 'GROUP'::message_account_type
+        WHEN person_id IS NOT NULL THEN 'CITIZEN'::message_account_type
+        ELSE 'MUNICIPAL'::message_account_type
+    END
+) STORED;
+CREATE UNIQUE INDEX only_one_municipality_account ON message_account (type) WHERE type = 'MUNICIPAL';
+
+INSERT INTO message_account (employee_id, daycare_group_id, person_id) VALUES (NULL, NULL, NULL);
```
akheron, line 15: Does this imply that the municipal message users have a shared account that's not tied to any employee account?

service/src/main/kotlin/fi/espoo/evaka/messaging/Message.kt
```diff
@@ -59,16 +58,15 @@ enum class MessageType {
...
 
 data class MessageReceiver(
-    val childId: ChildId,
-    val childFirstName: String,
-    val childLastName: String,
-    val childDateOfBirth: LocalDate
+    val id: Id<*>,
+    val type: MessageRecipientType,
+    val name: String,
+    val receivers: List<MessageReceiver>
```
akheron, line 69: Did you try implementing this as `sealed class` with 3 cases instead? I think it would be simpler to understand, plus more type safe because you could use the correct `id` type instead of `Id<*>`.

Furthermore, another commit in this PR seems to add a 4th case...

service/src/main/resources/db/migration/V281__municipal_message_account.sql
jaahon: Yeah, that's the idea. So they share the sent message box automatically. I guess the problem with this approach now that I think of it is that drafts are shared as well.

service/src/main/kotlin/fi/espoo/evaka/messaging/Message.kt
jaahon: Are there any upsides besides being able to distinguish between the id types? Using sealed classes also means that code generation doesn't work. It just didn't seem like sealed classes bring a lot of value.

akheron: It's both safer and in my opinion also cleaner. You could model the hierarchy correctly on the type level, so that unit has groups under it and groups have children under it. `Id<*>` allows to put _any_ ID in it, so this doesn't really prevent anything and isn't much better than `UUID`.

I didn't realize that code generation doesn't work 🤔 But adding those types manually to the frontend shouldn't be hard. The correct fix would be to make code generation work instead 😄

jaahon: Modeling the hierarchy doesn't really give anything though since the front end doesn't really care about the hierarchy. This just feels like extra work for no benefit.

akheron: I had really hard time understanding the diff first, before it occurred to me that this generates the hierarchy. So I think it could make the code cleaner.

akheron (approved):

PR #3024, opened 2022-10-20
Add db constraints

reynders:
#### Summary
-Make daycare round_the_clock non nullable, and set all null values to false
-Add check for voucher value decision voucher_value is greater or equal than co_payment


service/src/main/resources/db/migration/V279__voucher_value_vs_co_payment_check.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE voucher_value_decision ADD CONSTRAINT check$voucher_value_gte_co_payment CHECK ( voucher_value >= voucher_value_decision.co_payment );
```
akheron, line 1:
```suggestion
ALTER TABLE voucher_value_decision ADD CONSTRAINT check$voucher_value_gte_co_payment CHECK (voucher_value >= co_payment);
```

akheron (approved):

PR #3026, opened 2022-10-21
Do not send search requests when date filters are inverted

jaahon:


Gekkio (commented):
Since date ranges are inclusive, shouldn't we use `.isEqualOrAfter` so `[start, end]` (= one date) works?

Gekkio (approved):

PR #3033, opened 2022-10-24
Use capacity 1.0 for child without placement

reynders:


service/src/integrationTest/kotlin/fi/espoo/evaka/occupancy/RealtimeOccupancyTest.kt
```diff
@@ -236,6 +236,30 @@ class RealtimeOccupancyTest : FullApplicationTest(resetDbBeforeEach = true) {
...
+                                .save()
+                    }
+        }
+
+        val result = getRealtimeOccupancy()
+        val occupancies = result.occupancySeries
+        assertEquals(2, occupancies.size)
+
+        occupancies.get(0)
+                ?.also { assertEquals(1.0, it.childCapacity) }
```
akheron, line 260: Wouldn't `occupancies.get(0)?.also { ... }` possibly let a bug through? This would be better:
```
assertEquals(1.0, occupancies[0].childCapacity)
```

akheron (approved):

PR #3035, opened 2022-10-24
Add assistance decision report to special education teacher

patari:
#### Summary

Special education teacher should see if unit supervisor has made the decision.


Gekkio:
Based on discussions with Miia, this conflicts with Espoo requirements and since Espoo is the default (at least for now), this access control change should be done in customizations.

patari:
Funny, Heli said that this should be same for every city.

PR #3036, opened 2022-10-24
Add absences to attendance reservation report

patari:
#### Summary

Also fixes empty columns if child's placement starts in the middle of given period (or is in backup care).

### DIFF
--- a/service/src/main/kotlin/fi/espoo/evaka/reports/AttendanceReservationReportController.kt
+++ b/service/src/main/kotlin/fi/espoo/evaka/reports/AttendanceReservationReportController.kt
@@ -212,7 +212,7 @@ private fun Database.Read.getAttendanceReservationReportByChild(
         dates AS (SELECT generate_series::date AS date FROM generate_series(:start, :end, interval '1 day')),
         children AS (
           SELECT
-            CASE WHEN bc.id IS NOT NULL THEN bc.group_id ELSE dgp.daycare_group_id END AS group_id,
+            g.id AS group_id, g.name AS group_name,
             date, p.id, p.last_name, p.first_name, bc.id IS NOT NULL AS is_backup_care
           FROM dates date
           JOIN placement pl ON date BETWEEN pl.start_date AND pl.end_date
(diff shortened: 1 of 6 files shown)


service/src/main/kotlin/fi/espoo/evaka/reports/AttendanceReservationReportController.kt
```diff
@@ -204,27 +206,25 @@ private fun Database.Read.getAttendanceReservationReportByChild(
 ): List<AttendanceReservationReportByChildRow> {
     val sql =
         """
-        WITH reservations AS (
+        WITH
+        dates AS (SELECT generate_series::date AS date FROM generate_series(:start, :end, interval '1 day')),
```
Gekkio, line 210: `generate_series` is probably going to cause performance problems...yes, we use it sometimes but probably shouldn't.

Here's a quick test of this report query in our production with same parameters in both cases.

Old:

```
 Nested Loop Left Join  (cost=18.72..4936.87 rows=1 width=122) (actual time=0.269..31.996 rows=82 loops=1)
...
 Planning Time: 5.996 ms
 Execution Time: 32.197 ms
```

New:

```
 Nested Loop Left Join  (cost=90368.96..100221.02 rows=1 width=146) (actual time=1249.822..1290.971 rows=264 loops=1)
...
 Planning Time: 40.757 ms
 Execution Time: 1291.270 ms
```

~1s execution time isn't terrible, but I fear it might get out of hand, especially since the date range is freely selectable and greatly affects the cost of generate_series.

I don't have any immediate good solutions, but here are two thoughts:

- add empty rows/columns in the query in Kotlin code or frontend instead. We had performance problems with one other report at some point and an important part of the fix was to fetch only the real data (using ranges) and "fill in the blanks" in frontend
- look into multiranges or do some stuff in Kotlin code, instead of "looping over" every date in the query range

Or, we can merge now and fix later if/when problems appear :man_shrugging:

patari: Refactored query a bit to get more readable query plan, but no performance gain unfortunately.

PR #3037, opened 2022-10-24
EVAKA-4488 Implement deletion of messages

zch:
Adds rudimentary folder support for messaging and uses a special folder named 'ARCHIVE' to hold deleted/archived messages.

Employees can view their archived messages, but as for now, citizens cannot view archived (deleted) messages.


service/src/main/kotlin/fi/espoo/evaka/pis/service/MergeService.kt
```diff
@@ -84,9 +84,13 @@ class MergeService(private val asyncJobRunner: AsyncJobRunner<AsyncJob>) {
...
             UPDATE message_recipients SET recipient_id = (SELECT id FROM message_account WHERE person_id = :id_master) WHERE recipient_id = (SELECT id FROM message_account WHERE person_id = :id_duplicate);
             UPDATE message_draft SET account_id = (SELECT id FROM message_account WHERE person_id = :id_master) WHERE account_id = (SELECT id FROM message_account WHERE person_id = :id_duplicate);
+            
+            UPDATE message_thread_participant mtp SET folder_id = (SELECT id FROM message_thread_folder WHERE owner_id = mtp.participant_id AND name = 'ARCHIVE') WHERE mtp.folder_id = (SELECT id FROM message_thread_folder WHERE owner_id = mtp.participant_id AND name = 'ARCHIVE');
+            DELETE FROM message_thread_folder WHERE owner_id = (SELECT id FROM message_account WHERE person_id = :id_master) AND name = 'ARCHIVE';
+            UPDATE message_thread_folder SET owner_id = (SELECT id FROM message_account WHERE person_id = :id_master) WHERE owner_id = (SELECT id FROM message_account WHERE person_id = :id_duplicate);
```
akheron, line 90: These queries do this:
- The first `UPDATE` doesn't have any condition that would mention `:id_master` or `:id_duplicate`, so it acts on all data 😬 
- The `DELETE` deletes _master's_ archive folder 😬 
- The second update moves duplicate's folders to be owned by master.

I think the correct thing to do would be:
- `UPDATE` threads that are in duplicate's ARCHIVE folder to be in master's archive folder (i.e. modify the first `UPDATE` a bit)
- Update duplicate's folders other than ARCHIVE to be owned by master

There should be no need to `DELETE` anything, as that's handled by cascade when the `message_account` is deleted.

service/src/main/kotlin/fi/espoo/evaka/messaging/MessageQueries.kt
```diff
@@ -325,14 +373,64 @@ FROM message_thread_participant tp
...
+            .bind("pageSize", pageSize)
+            .bind("offset", (page - 1) * pageSize)
+            .mapToPaged<ReceivedThread>(pageSize)
+
+    val messagesByThread = getThreadMessages(accountId, threads.data.map { it.id })
+    return combineThreadsAndMessages(accountId, threads, messagesByThread)
+}
+
+/** Return all threads in a specific folder in the account */
+fun Database.Read.getThreadsInFolder(
```
akheron, line 392: Why not generalize `getThreads` and `getReceivedThreads` to always work on folders, `null` meaning inbox? I don't see a need to add a new query at all.

service/src/main/kotlin/fi/espoo/evaka/messaging/MessageQueries.kt
```diff
@@ -325,14 +373,64 @@ FROM message_thread_participant tp
...
+    accountId: MessageAccountId,
+    pageSize: Int,
+    page: Int,
+    folderName: String
```
akheron, line 396: If you searched by folder ID instead of folder name, you wouldn't need to join on the folder, so querying for `NULL` would be easier. I think it would be better if folder id would be passed all the way from the frontend, and the folder name would be used for viewing purposes only.

service/src/main/kotlin/fi/espoo/evaka/messaging/MessageQueries.kt
```diff
@@ -103,6 +105,52 @@ WHERE rec.message_id = msg.id
...
+    return this.createUpdate(
+            "UPDATE message_thread_participant SET folder_id = :archiveFolderId WHERE thread_id = :threadId AND participant_id = :accountId"
+        )
+        .bind("accountId", accountId)
+        .bind("threadId", threadId)
+        .bind("archiveFolderId", archiveFolderId)
+        .execute()
+}
+
+fun Database.Transaction.moveThreadToFolderByName(
```
akheron, line 130: This function seems to be unused

service/src/main/resources/db/migration/V281__message_thread_folder.sql
```diff
@@ -0,0 +1,19 @@
...
+    id                      uuid PRIMARY KEY         NOT NULL DEFAULT ext.uuid_generate_v1mc(),
+    created                 timestamp with time zone NOT NULL DEFAULT now(),
+    updated                 timestamp with time zone NOT NULL DEFAULT now(),
+    name                    text                     NOT NULL DEFAULT '',
+    owner_id                uuid                     NOT NULL REFERENCES message_account (id) ON DELETE CASCADE,
+
+    CONSTRAINT uniq$thread_folder UNIQUE (name, owner_id)
+);
+
+CREATE INDEX idx$thread_folder_owner ON message_thread_folder (id, owner_id);
```
akheron, line 12: This index seems useless and should be removed. `id` implicitly has an index because it's the primary key, so this would only be marginally useful for queries that return `id`, `owner_id` pairs.

service/src/main/resources/db/migration/V281__message_thread_folder.sql
```diff
@@ -0,0 +1,19 @@
...
+CREATE INDEX idx$thread_folder_name ON message_thread_folder (owner_id, name);
```
akheron, line 13: If you change folder matching to be based on id instead of name, you should remove this index and change the unique constraint above to have the column order as `(owner_id, name)`. The unique constraint is backed by an index, and that index would facilitate queries by `owner_id`, and you wouldn't need a separate index for that purpose.

service/src/main/resources/db/migration/V281__message_thread_folder.sql
```diff
@@ -0,0 +1,19 @@
...
+CREATE TRIGGER set_timestamp BEFORE UPDATE ON message_thread_folder FOR EACH ROW EXECUTE PROCEDURE trigger_refresh_updated();
+
+ALTER TABLE message_thread_participant
+    ADD COLUMN folder_id uuid REFERENCES message_thread_folder (id) ON DELETE CASCADE;
+
+CREATE INDEX idx$thread_participant_folder ON message_thread_participant (participant_id, folder_id, last_message_timestamp);
```
akheron, line 19: If you change all message thread querying to include the `folder_id` (which can be NULL), then you should modify the existing indices instead of adding this new index. The existing ones are:

```
create index idx$thread_participant_message
    on message_thread_participant (participant_id, last_message_timestamp);

create index idx$thread_participant_received
    on message_thread_participant (participant_id, last_received_timestamp)
    where (last_received_timestamp IS NOT NULL);
```

Just drop those and re-create them as `(participant_id, folder_id, last_*_timestamp)`.

akheron (changes requested):

service/src/main/kotlin/fi/espoo/evaka/pis/service/MergeService.kt
zch: Great catches! Did some last minute updates to this and might have been too quick on the trigger 😬 

Updated the PR as per your suggestions.

service/src/main/kotlin/fi/espoo/evaka/messaging/MessageQueries.kt
```diff
@@ -339,14 +367,22 @@ JOIN message_thread t on t.id = tp.thread_id
 WHERE
     tp.participant_id = :accountId AND
     tp.last_received_timestamp IS NOT NULL AND
-    NOT t.is_copy
+    NOT t.is_copy AND
+    ${
+        if (folderId === null) {
```
akheron, line 372: Does `===` work in Kotlin? 🤔 😄

akheron (approved):

service/src/main/kotlin/fi/espoo/evaka/messaging/MessageQueries.kt
akheron: > Note that there's no point in optimizing your code when comparing to null explicitly: a == null will be automatically translated to a === null

akheron: TIL: https://kotlinlang.org/docs/equality.html

> Referential equality is checked by the === operation and its negated counterpart !==. a === b evaluates to true if and only if a and b point to the same object.

> Note that there's no point in optimizing your code when comparing to null explicitly: a == null will be automatically translated to a === null

PR #3038, opened 2022-10-25
Fix finance decision search query filters

reynders:
#### Summary
- If both area filter and starting child filter was on, the generated SQL was invalid
- Add unit tests for both fee decision search and voucher value decision search

### DIFF
--- a/service/src/integrationTest/kotlin/fi/espoo/evaka/invoicing/data/FeeDecisionSearchTest.kt
+++ b/service/src/integrationTest/kotlin/fi/espoo/evaka/invoicing/data/FeeDecisionSearchTest.kt
@@ -216,13 +216,16 @@ class FeeDecisionSearchTest : PureJdbiTest(resetDbBeforeEach = true) {
 
     @Test
     fun `starting children filter search`() {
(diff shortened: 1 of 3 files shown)


service/src/integrationTest/kotlin/fi/espoo/evaka/invoicing/data/FeeDecisionSearchTest.kt
```diff
@@ -212,6 +214,50 @@ class FeeDecisionSearchTest : PureJdbiTest(resetDbBeforeEach = true) {
...
+                )
+                .size
+        )
+
+        db.transaction { tx ->
+            tx.insertTestPlacement(
+                DevPlacement(
+                    childId = testChild_3.id,
+                    unitId = testDaycare.id,
+                    startDate = LocalDate.now(),
```
Gekkio, line 249: Suggestion: don't use `LocalDate.now()`, because it adds an unnecessary random element to the test (we don't put random values to all String/Int/Long/Boolean fields either). If the *actual* date value is not important, prefer a constant `val today = LocalDate.of(something)` and use that. If we truly need "today's date" and the test depends on the time it's executed, refactor the underlying code to use mock clock instead. This applies to all LocalDate.now() calls that have been added.

Gekkio (approved):

PR #3052, opened 2022-10-28
Make editing of dates with a keyboard work smoothly

zch:


jaahon (approved):
Works very nicely now!

PR #3069, opened 2022-11-07
Change the staff attendance cleanup to end days at 20:00

zch:
... and fallback to 24:00

### DIFF
--- a/service/src/main/kotlin/fi/espoo/evaka/attendance/StaffAttendanceQueries.kt
+++ b/service/src/main/kotlin/fi/espoo/evaka/attendance/StaffAttendanceQueries.kt
@@ -456,12 +456,12 @@ FROM missing_planned_departures WHERE realtime.id = missing_planned_departures.i
         .bind("now", now)
         .execute()
 
-    val defaultDepartureTime = now.minusDays(1).withTime(LocalTime.of(18, 0))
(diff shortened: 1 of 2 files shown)


jaahon (commented):
This query can still fail if there's is an arrival after 20:00. This case should be handled in some way (either ignored or with a departure time of 24:00) since there's nothing stopping users making these kinds of arrivals either by mistake or on purpose.

service/src/main/kotlin/fi/espoo/evaka/attendance/StaffAttendanceQueries.kt
```diff
@@ -456,15 +456,15 @@ FROM missing_planned_departures WHERE realtime.id = missing_planned_departures.i
         .bind("now", now)
         .execute()
 
-    val defaultDepartureTime = now.minusDays(1).withTime(LocalTime.of(18, 0))
+    val defaultDepartureTime = now.minusDays(1).withTime(LocalTime.of(20, 0))
     createUpdate(
             // language=SQL
             """
 UPDATE staff_attendance_realtime a
 SET departed = :defaultDepartureTime
```
jaahon, line 464: Instead of basically duplicating the updates you could also replace this with `SET departed = CASE WHEN arrived < :defaultDepartureTime THEN :defaultDepartureTime ELSE :startOfDay END` which would also be simpler.

PR #3074, opened 2022-11-08
Undo sent messages

jaahon:
#### Summary

Add a 11 second window for undoing messages to employees. Undoing a message that is not a reply to an existing thread converts the message back into a draft.

Also,
- Save employee selected message box state in url
- Store employee selected thread in url
- Open newly created message on send
- Refactor citizen calendar notifications and reload notifications into a component that supports any kind of notifications


akheron:
Conflicts 😰

frontend/src/employee-frontend/components/messages/MessagesPage.tsx
```diff
@@ -52,24 +60,37 @@ export default React.memo(function MessagesPage() {
...
+    const firstGroupAccount = groupAccounts[0]
+    if (firstGroupAccount) {
+      return selectAccount({
+        view: groupMessageBoxes[0],
+        account: firstGroupAccount.account,
+        unitId: firstGroupAccount.daycareGroup.unitId
       })
     }
-  }, [accounts, setSelectedAccount, selectedAccount])
+  }, [accounts, groupAccounts, municipalAccount, personalAccount, selectAccount, selectedAccount])
```
akheron, line 96: Would it be cleaner if this effect lived in the context provider instead of inside this component? 🤔 

Also: Would it be possible to use `renderResult` a few lines down where the JSX is rendered?

frontend/src/employee-frontend/components/messages/ThreadListContainer.tsx
```diff
@@ -117,88 +137,6 @@ export default React.memo(function ThreadListContainer({
...
-            ],
-            readAt: message.readAt,
-            content: message.content,
-            attachments: message.attachments,
-            recipientNames: message.recipientNames
-          }
-        ]
-      }))
-  )
-
```
akheron, line 201: Also add `useMemo` for all the mapping stuff below?

service/src/main/kotlin/fi/espoo/evaka/messaging/MessageQueries.kt
```diff
@@ -389,7 +401,8 @@ WHERE
         } else {
             "tp.folder_id = :folderId"
         }
-    }
+    } AND
+    EXISTS (SELECT 1 FROM message m WHERE m.thread_id = t.id AND m.sent_at < :undoThreshold)
```
akheron, line 405: Have you checked this change with EXPLAIN in prod? It seems to me that this has the potential to slow down the query by a lot.

akheron: If there is a significant performance hit, it could possibly be mitigated by setting `thread.last_received_timestamp` to be send time + 11 seconds, and then instead filter here that `thread.last_received_timestamp <= :now`, but not sure how hard that is in the sending and undoing code.

service/src/main/kotlin/fi/espoo/evaka/messaging/MessageQueries.kt
```diff
@@ -334,22 +343,25 @@ SELECT
 FROM message_thread_participant tp
 JOIN message_thread t on t.id = tp.thread_id
 WHERE tp.participant_id = :accountId AND tp.folder_id IS NULL
+AND EXISTS (SELECT 1 FROM message m WHERE m.thread_id = t.id AND (m.sender_id = :accountId OR m.sent_at < :undoThreshold))
```
akheron, line 346: The same possible performance hit exists here, too.

service/src/main/kotlin/fi/espoo/evaka/messaging/MessageController.kt
```diff
@@ -488,6 +498,30 @@ class MessageController(
...
+    ): MessageDraftId? {
+        return db.connect { dbc ->
+                requireMessageAccountAccess(dbc, user, clock, accountId)
+                dbc.transaction {
+                    when {
+                        messageId != null ->
+                            it.undoMessageReply(clock.now(), accountId, messageId).let { null }
+                        contentId != null ->
+                            it.undoMessageAndAllThreads(clock.now(), accountId, contentId)
+                        else -> throw BadRequest("Either messageId or contentId is required")
```
akheron, line 518: It would be nice to adjust typings in a way that you can't give both `messageId` and `contentId`, or add separate routes for canceling a reply or a new thread, or at least add runtime validation checking that both are not passed.

service/src/main/kotlin/fi/espoo/evaka/messaging/MessageController.kt
```diff
@@ -282,66 +284,74 @@ class MessageController(
...
+                            accountIdsToChildIds = messageAccountIdsToChildIds,
+                            municipalAccountName = featureConfig.municipalMessageAccountName
+                        )
+                    if (body.draftId != null)
+                        tx.deleteDraft(accountId = accountId, draftId = body.draftId)
+                    messageContentId
+                }
             }
-        }.also { Audit.MessagingNewMessageWrite.log(targetId = accountId, objectId = it) }
+            .also { Audit.MessagingNewMessageWrite.log(targetId = accountId, objectId = it) }
```
akheron, line 354: Is the whole hunk just a formatting change, or are there some real changes I can't spot? I thought reformatting wouldn't be a problem now that we have `ktfmt` 😂

frontend/src/employee-frontend/components/messages/UndoMessageNotification.tsx
```diff
@@ -0,0 +1,103 @@
...
+  cancelableMessage
+}: {
+  cancelableMessage: CancelableMessage
+}) {
+  const theme = useTheme()
+  const { i18n } = useTranslation()
+  const { refreshMessages, setCancelableMessage, selectThread } =
+    useContext(MessageContext)
+  const [secondsLeft, setSecondsLeft] = useState<number>(
+    getSecondsLeft(cancelableMessage.sentAt)
```
akheron: Nitpick: Use `() => getSecondsLeft(...)` to avoid useless computing of the initial value on each render.

frontend/src/employee-frontend/components/messages/MessageContext.tsx
```diff
@@ -501,8 +512,20 @@ export const MessageContextProvider = React.memo(
...
+            setCancelableMessage({
+              accountId: params.accountId,
+              messageId: result.value.message.id,
+              sentAt: HelsinkiDateTime.now()
+            })
+          }
+          return result
+        }),
+      setReplyResponse
+    )
```
akheron, line 553: It would be so cool if we had a chained async calls abstraction in the frontend, because this is quite ugly and has been used in many places 😞

service/src/main/kotlin/fi/espoo/evaka/messaging/MessageQueries.kt
akheron: Hmm, now that I think of it, how well do the the `message_thread_participant.last_*_timestamp` columns now work with this change? If a reply is sent and then undone, does the thread show as the newest thread in the receiver's inbox?

service/src/main/kotlin/fi/espoo/evaka/messaging/MessageController.kt
jaahon: The result of `db.conntect` is returned and the audit call if wrapped in a `also` block because of that. That's why the indentation is changed. I guess github doesn't notice that. I think ktfmt's formatting is a bit ugly here, but it should be consistent.

service/src/main/kotlin/fi/espoo/evaka/messaging/MessageController.kt
jaahon: So which solution would be preferable in your opinion? Separate routes? I don't have strong opinions on any of the options.

frontend/src/employee-frontend/components/messages/MessagesPage.tsx
jaahon: Do you mean this part?

```
{showEditor && accounts.isSuccess && receivers && selectedAccount && (
  <MessageEditor ... />
)}
```

jaahon: Do you mean this part?

```
{showEditor && accounts.isSuccess && receivers && selectedAccount && (
  <MessageEditor ... />
)}

File: service/src/main/kotlin/fi/espoo/evaka/messaging/MessageQueries.kt (hunk 3 above)
Comment (jaahon) [reply service/src/main/kotlin/fi/espoo/evaka/messaging/MessageQueries.kt:405]: I completely forgot about `message_thread_participant`. I'll need to fix that.

When it comes to performance I'm not sure how to test this beforehand. Querying a random person's messages seems pretty questionable.

Comment (jaahon) [reply service/src/main/kotlin/fi/espoo/evaka/messaging/MessageQueries.kt:405]: I've now added a function that resets `message_thread_participant` to its previous state when a message is undone.

File: service/src/main/kotlin/fi/espoo/evaka/messaging/MessageController.kt (hunk 5 above)
Comment (jaahon) [reply service/src/main/kotlin/fi/espoo/evaka/messaging/MessageController.kt]: I've split the route into two.

File: service/src/main/kotlin/fi/espoo/evaka/messaging/MessageQueries.kt (hunk 4 above)
Comment (jaahon) [reply service/src/main/kotlin/fi/espoo/evaka/messaging/MessageQueries.kt:346]: I compared this and the original version of the queries with a few random people with received messages in prod. It seems like the new query is a few milliseconds slower than the original.

File: service/src/main/kotlin/fi/espoo/evaka/messaging/MessageQueries.kt (hunk 3 above)
Comment (akheron) [reply service/src/main/kotlin/fi/espoo/evaka/messaging/MessageQueries.kt:405]: > When it comes to performance I'm not sure how to test this beforehand. Querying a random person's messages seems pretty questionable.

`EXPLAIN ANALYZE` in the production database. It doesn't show you the query result, only the query plan and timing info.

File: service/src/main/kotlin/fi/espoo/evaka/messaging/MessageQueries.kt [hunk 9]
```diff
@@ -1100,3 +1119,168 @@ fun Database.Read.getArchiveFolderId(accountId: MessageAccountId): MessageThread
...
+FROM (
+    SELECT MAX(sent_at) AS sent_at
+    FROM message m
+    WHERE m.thread_id = :threadId AND m.sender_id = :senderId
+) sent_message
+JOIN (
+    SELECT MAX(sent_at) AS sent_at
+    FROM message m
+    WHERE m.thread_id = :threadId
+) any_message ON true
```
akheron, line 1257: The logic here doesn't work if there are messages in the thread that are not visible to `senderId`. You have to filter using the `message_recipients` table, too.

This made me think, would it be better to only update `last_sent_timestamp` and `last_message_timestamp` only for the sender when a message is sent, and then add an async job to update the `last_message_timestamp` for the recipients after the undo period has expired. Then the thread would be bumped to the top of inbox only after the recipient actually sees the message.

The async job could include the sent message id and only do its job if that message still exist, i.e. sending it hasn't been undone.

akheron: Furthermore, I would personally find the query better in this form:

```
UPDATE ... 
SET
    x = (SELECT MAX(...) ...),
    y = (SELECT MAX(...) ...)
WHERE z
```

This would be easier for a human to parse (in my opinion) and as a bonus, it would be an error if the subqueries returned more than one row. This would minimize the risk that someone in the future changes the query so that it would affect the wrong set of rows.

If you decide to keep it such that it still has a `FROM` clause, I'd like to point out that `FROM a JOIN b ON true` is equivalent to `FROM a, b`.

akheron (approved):
🥇

PR #3077, opened 2022-11-09
Manage DatePicker state without useEffect

akheron:
#### Summary

This fixes a problem where DatePicker flashes validation errors briefly before useEffect fires. And is better in general.

Also simplify assistance need decision form state management. The check for missing start date is redundant because the types ensure it's always set.


zch (approved):
Nice!

PR #3079, opened 2022-11-09
Use direct route calls instead of HTTP in some integration tests

Gekkio:

### DIFF
--- a/service/src/integrationTest/kotlin/fi/espoo/evaka/pis/controller/ParentshipControllerIntegrationTest.kt
+++ b/service/src/integrationTest/kotlin/fi/espoo/evaka/pis/controller/ParentshipControllerIntegrationTest.kt
@@ -15,7 +15,6 @@ import fi.espoo.evaka.shared.auth.AuthenticatedUser
 import fi.espoo.evaka.shared.auth.CitizenAuthLevel
 import fi.espoo.evaka.shared.auth.UserRole
 import fi.espoo.evaka.shared.auth.insertDaycareAclRow
-import fi.espoo.evaka.shared.db.Database
 import fi.espoo.evaka.shared.dev.DevEmployee
 import fi.espoo.evaka.shared.dev.DevPlacement
 import fi.espoo.evaka.shared.dev.insertTestEmployee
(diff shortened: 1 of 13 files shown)


service/src/integrationTest/kotlin/fi/espoo/evaka/pis/controller/ParentshipControllerIntegrationTest.kt
```diff
@@ -90,10 +89,10 @@ class ParentshipControllerIntegrationTest : FullApplicationTest(resetDbBeforeEac
         val endDate = startDate.plusDays(200)
         val reqBody =
             ParentshipController.ParentshipRequest(parent.id, child.id, startDate, endDate)
-        controller.createParentship(Database(jdbi), user, clock, reqBody)
+        controller.createParentship(dbInstance(), user, clock, reqBody)
```
akheron, line 92: Should `controller` be renamed as `parentshipController`? That has been the naming convention in all other test classes.

service/src/integrationTest/kotlin/fi/espoo/evaka/pis/controller/PartnershipsControllerIntegrationTest.kt
```diff
@@ -106,9 +105,9 @@ class PartnershipsControllerIntegrationTest : FullApplicationTest(resetDbBeforeE
         val endDate = startDate.plusDays(200)
         val reqBody =
             PartnershipsController.PartnershipRequest(person.id, partner.id, startDate, endDate)
-        controller.createPartnership(Database(jdbi), user, clock, reqBody)
+        controller.createPartnership(dbInstance(), user, clock, reqBody)
```
akheron, line 108: Should `controller` be renamed as `partnershipController`?

akheron (approved):

service/src/integrationTest/kotlin/fi/espoo/evaka/pis/controller/ParentshipControllerIntegrationTest.kt
Gekkio: I think `controller` is fine in this case, because the test is literally a test for that specific controller (it's even in the class name)

PR #3085, opened 2022-11-14
Fix personal message accounts

patari:
#### Summary

Special education teacher and early childhood secretary should be able to send messages with their own name (like unit supervisor).

Also makes it possible to configure who is able to see every group from unit by refactoring message_account_access_view to actions.


service/src/main/kotlin/fi/espoo/evaka/shared/security/Action.kt
```diff
@@ -1176,9 +1138,21 @@ sealed interface Action {
...
+        ACCESS(
+            IsEmployee.hasPersonalMessageAccount(),
+            HasUnitRole(UNIT_SUPERVISOR, SPECIAL_EDUCATION_TEACHER).hasDaycareMessageAccount(),
+            IsEmployee.hasDaycareGroupMessageAccount(),
+            IsEmployee.hasMunicipalMessageAccount(),
+            IsMobile(requirePinLogin = true).hasPersonalMessageAccount(),
+            IsMobile(requirePinLogin = true)
+                .hasDaycareMessageAccount(UNIT_SUPERVISOR, SPECIAL_EDUCATION_TEACHER),
+            IsMobile(requirePinLogin = true).hasDaycareGroupMessageAccount(),
+            IsMobile(requirePinLogin = true).hasMunicipalMessageAccount(),
```
jaahon, line 1153: I don't think this rule makes much sense so I think it should be removed. It's not even practically possible to send municipal messages from employee mobile.

jaahon (commented):
Everything looked ok to me 🤔

jaahon (approved):

PR #3101, opened 2022-11-16
Require groupId only for staff attendances that are present in group

jaahon:
#### Summary

Change realtime staff attendances to not require `groupId` if the attendance is of type `TRAINING` or `OTHER_WORK`.
Also fix `<StaffAttendanceDetailsModal />` to no longer show `→ - →` for each planned attendance that is not on the shown date.


akheron:
Discussed on Slack: The typings in the modal's code can probably be enhanced, and @jaahon will try that.

akheron (approved):

PR #3105, opened 2022-11-17
Make admin permissions explicit

Gekkio:
#### Summary

Some actions were skipped:

- anything only used by citizens
- anything only used by mobile devices
- actions which are meant to be allowed to a specific employee only


jaahon (approved):
💪

PR #3108, opened 2022-11-18
Fix using the `DatePicker` calendar dialog with mouse on Safari

akheron:
#### Summary

Remove `tabindex="-1"` from `<main>`. This has the effect that `<main>` won't receive focus events, and thus fixes the premature closing of the <DatePicker> calendar dialog on Safari.

Removen `tabindex` from `<main>` shouldn't break anything, as `tabindex="-1"` means that the element can be focused programmatically or by clicking, but not by keyboard. I don't see why this would be required. See the [MDN article on `tabindex`](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/tabindex).

Also remove some dead code.

### DIFF
--- a/frontend/src/citizen-frontend/utils.ts
+++ b/frontend/src/citizen-frontend/utils.ts
@@ -1,8 +1,0 @@
-// SPDX-FileCopyrightText: 2017-2022 City of Espoo
-//
-// SPDX-License-Identifier: LGPL-2.1-or-later
-
-import { scrollElementToPos } from 'lib-common/utils/scrolling'
-
-export const scrollMainToPos = (opts: ScrollToOptions) =>
-  scrollElementToPos(document.getElementById('main'), opts)
--- a/frontend/src/lib-components/atoms/Main.tsx
+++ b/frontend/src/lib-components/atoms/Main.tsx
@@ -14,9 +14,5 @@ export default React.memo(function Loader({
 }: {
   children: React.ReactNode
 }) {
-  return (
-    <MainContainer tabIndex={-1} id="main">
-      {children}
-    </MainContainer>
-  )
+  return <MainContainer id="main">{children}</MainContainer>
 })


zch (approved):

jolle:
The `tabindex="-1"` was added for skip to content links, as some olders browsers may only change the focus visually without actually changing the keyboard focus.

akheron:
@jolle The skip to content link is rendered outside of `<main>` via `React.createPortal`, and the actual scrolling is done with anchors and appending `#foo` to the URL (the scrolling part has probably changed since your time in the project). Do you still think that `tabindex="-1"` is needed? 🤔

jolle:
@akheron I don't see any changes to the [SkipToContent](https://github.com/espoon-voltti/evaka/blob/master/frontend/src/lib-components/atoms/buttons/SkipToContent.tsx) component since I added it, so I don't think there's been new changes to its behavior. There was [a bug in Chrome](https://bugs.chromium.org/p/chromium/issues/detail?id=454172) that didn't properly modify the keyboard focus with anchor links on non-focusable elements, but this has been fixed for many years now. Apparently, IE exhibits similar behavior, but I don't think this project works on IE anyway. The `tabindex="-1"` trick for the main element is included in various guides ([1](https://accessibility.oit.ncsu.edu/it-accessibility-at-nc-state/developers/accessibility-handbook/mouse-and-keyboard-events/skip-to-main-content/#:~:text=You%20should%20include%20a%20tabindex%3D%E2%80%9D%2D1%E2%80%B3), [2](https://css-tricks.com/how-to-create-a-skip-to-content-link/#comment-1755109), [3](https://medium.com/@belwerks/a-longer-note-on-skip-links-and-tabindex-dd0f1e1482b4#efee)), but as mentioned it's probably due to old browser bugs, so I think the skip to content links will work fine in most cases even without the tabindex.

PR #3112, opened 2022-11-21
Cancel all placements after an inserted CLUB placement

zch:
#### Summary
Earlier all placements were inserted such that any existing placement would be trimmed from the ends, or split into two to make room for the new placement. This is still the case for all other kinds of placements, but CLUB placements only trim existing placements from the end (if they start before the new CLUB placement) and cancel any placement (or part of a placement) that would occur after the new CLUB placement has ended.

### DIFF
--- a/service/src/main/kotlin/fi/espoo/evaka/placement/PlacementPlanService.kt
+++ b/service/src/main/kotlin/fi/espoo/evaka/placement/PlacementPlanService.kt
@@ -249,7 +249,8 @@ class PlacementPlanService(private val asyncJobRunner: AsyncJobRunner<AsyncJob>,
             childId = childId,
             unitId = unitId,
             placementTypePeriods = placementTypePeriods,
-            serviceNeed = serviceNeed
+            serviceNeed = serviceNeed,
+            cancelPlacementsAfterClub = true
         )
         val timeline = Timeline.of(placementTypePeriods.map { it.first })
         asyncJobRunner.plan(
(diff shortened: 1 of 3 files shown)


service/src/main/kotlin/fi/espoo/evaka/placement/PlacementPlanService.kt
```diff
@@ -241,15 +241,16 @@ class PlacementPlanService(private val asyncJobRunner: AsyncJobRunner<AsyncJob>,
         extent: PlacementPlanExtent
     ) {
         val childId = application.childId
-        val placementTypePeriods = getPlacementTypePeriods(tx, childId, unitId, type, extent)
+        var placementTypePeriods = getPlacementTypePeriods(tx, childId, unitId, type, extent)
```
jaahon, line 244: Accidental change? 😄

service/src/main/kotlin/fi/espoo/evaka/placement/PlacementService.kt
```diff
@@ -314,6 +320,36 @@ private fun Database.Transaction.clearOldPlacements(
...
+                        PlacementType.PREPARATORY_DAYCARE,
+                        PlacementType.PRESCHOOL,
+                        PlacementType.PRESCHOOL_DAYCARE
+                    )
+                    .contains(it.type)
+            }
+            .map {
+                FiniteDateRange(
+                    it.startDate.withMonth(8).withDayOfMonth(1),
+                    it.endDate.withMonth(7).withDayOfMonth(31)
```
jaahon, line 342: Doesn't this work in an unexpected way if `startDate` is before 1st of August or `endDate` is after 31st of July?

zch: 👍 I reworked it a bit from another angle, which I think should take care of the issue.

jaahon (approved):

PR #3116, opened 2022-11-22
Search person by email

reynders:


jaahon (commented):
This doesn't work on employee search, because it doesn't use `freeTextSearchQuery`, it just uses `freeTextSearchQueryForColumns`. In any case I think this would not be a good change because it would mean that every place where `freeTextSearchQuery` is used would use the new email search as well, even when it's not needed.

The simplest change that would enable the intended feature is to change `listOf("first_name", "last_name")` in EmployeeQueries.kt:260 to `listOf("first_name", "last_name", "email")`.

PR #3123, opened 2022-11-25
Make sure to only suggest actually reservable days

zch:
#### Summary
The suggested first reservable date in the reservation modal suggested the first day of a holiday period in the following case:
- A holiday period is defined to start at a date later than today
- The deadline for reservations during the holiday period was earlier than today
- A reservation has been made for the kids, starting today and ending at a date after the end of the holiday period

---> The reservation modal suggested the first date of the holiday as the next reservable date.

This is now fixed by checking that the suggestion is an actual reservable day, which holidays are not.

### DIFF
--- a/frontend/src/citizen-frontend/calendar/CalendarPage.tsx
+++ b/frontend/src/citizen-frontend/calendar/CalendarPage.tsx
@@ -40,7 +40,6 @@ import NonReservableDaysWarningModal from './NonReservableDaysWarningModal'
 import ReservationModal from './ReservationModal'
 import { getCalendarEvents, getReservations } from './api'
 import FixedPeriodSelectionModal from './holiday-modal/FixedPeriodSelectionModal'
-import { getEarliestReservableDate } from './utils'
 
 async function getReservationsDefaultRange(): Promise<
   Result<ReservationsResponse>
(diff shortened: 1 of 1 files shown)


frontend/src/citizen-frontend/calendar/CalendarPage.tsx
```diff
@@ -125,10 +125,15 @@ const CalendarPage = React.memo(function CalendarPage() {
...
       )
+      const allReservableDateRanges = Object.keys(
+        data.value.reservableDays
+      ).flatMap((childId) => data.value.reservableDays[childId])
+
       // First reservable day that has no reservations
       const firstReservableEmptyDate = data.value.dailyData.find(
         (day) =>
           earliestReservableDate?.isEqualOrBefore(day.date) &&
+          allReservableDateRanges.find((range) => range.includes(day.date)) &&
```
jaahon, line 136: With this change we could just not calculate `earliestReservableDate`. I don't think it has any function since it's calculated from `data.value.reservableDays` anyway.

zch: Good point!

jaahon (approved):

PR #3134, opened 2022-11-30
Add preschool club placement type

patari:
#### Summary

In Tampere preschool "connected daycare" can be either daycare or club with different service need options.

This is final part of preschool application customizations for Tampere.


service/src/main/kotlin/fi/espoo/evaka/decision/DecisionService.kt
```diff
@@ -354,7 +356,8 @@ class DecisionService(
                     DecisionType.DAYCARE,
                     DecisionType.DAYCARE_PART_TIME -> "Varhaiskasvatuspäätös"
                     DecisionType.PRESCHOOL -> "Esiopetuspäätös"
-                    DecisionType.PRESCHOOL_DAYCARE -> "Liittyvä_varhaiskasvatuspäätös"
+                    DecisionType.PRESCHOOL_DAYCARE,
+                    DecisionType.PRESCHOOL_CLUB -> "Liittyvä_varhaiskasvatuspäätös"
```
jaahon, line 360: Shouldn't these two have a different file name?

jaahon (commented):
You should add `PRESCHOOL_CLUB` to the list in PlacementService.kt:336.

The `mapRequestedPlacementType` function in ApplicationQueries.kt will never return `PRESCHOOL_CLUB`, which in practice means that applications with preschool club appear as preschool daycare applications on applications lists in different parts of the employee frontend. Is that a problem?

patari:
> You should add `PRESCHOOL_CLUB` to the list in PlacementService.kt:336.
> 
> The `mapRequestedPlacementType` function in ApplicationQueries.kt will never return `PRESCHOOL_CLUB`, which in practice means that applications with preschool club appear as preschool daycare applications on applications lists in different parts of the employee frontend. Is that a problem?

:facepalm: somehow missed these two.

Second one is difficult, should PlacementType come from service need option (like in `PlacementPlanService.kt#derivePlacementType`?

jaahon:
> Second one is difficult, should PlacementType come from service need option (like in `PlacementPlanService.kt#derivePlacementType`?

Yeah, maybe. That's what I would do probably. I guess `validPlacementType` could be persisted in the `ServiceNeedOption` in the application form.

jaahon (approved):
Good work

PR #3135, opened 2022-11-30
Show all child daily attendances and reservations in the desktop UI

reynders:

### DIFF
--- a/service/src/main/kotlin/fi/espoo/evaka/reservations/AttendanceReservationController.kt
+++ b/service/src/main/kotlin/fi/espoo/evaka/reservations/AttendanceReservationController.kt
@@ -23,6 +23,7 @@ import fi.espoo.evaka.shared.domain.FiniteDateRange
 import fi.espoo.evaka.shared.domain.NotFound
 import fi.espoo.evaka.shared.security.AccessControl
 import fi.espoo.evaka.shared.security.Action
+import java.lang.Integer.max
 import java.time.LocalDate
 import org.jdbi.v3.core.mapper.Nested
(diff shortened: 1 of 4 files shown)


service/src/main/kotlin/fi/espoo/evaka/reservations/AttendanceReservationController.kt
```diff
@@ -403,30 +404,31 @@ private fun toChildDayRows(
...
+                            0
+                        }
+                    }
+                )
+
             UnitAttendanceReservations.ChildDailyRecords(
                 child = child.child,
                 dailyData =
-                    listOfNotNull(
+                    (0 until maxRecordCountOnAnyDay).mapNotNull { rowIndex ->
```
akheron, line 426: `associateBy` doesn't return `null`, so this should probably be just `map`?

akheron: I think `0..maxRecordCountOnAnyDay` would be more idomatic Kotlin than using `until`.

akheron (commented):
Zero e2e tests 😱

akheron:
Seems to work well 🏅

service/src/main/kotlin/fi/espoo/evaka/reservations/AttendanceReservationController.kt
zch: `0..maxRecordCountOnAnyDay` is inclusive, while `0 until maxRecordCountOnAnyDay` is exclusive. I find using `until` to be cleaner in this case.

akheron (approved):

PR #3139, opened 2022-12-02
[EVAKA-4604] Allow citizen to choose secondary recipients when sending a new message

Gekkio:
#### Summary

- change the UI so that children are selected first and then receivers
- when selecting multiple children, only receivers that are related to *all selected children* are valid
- allow citizen to choose secondary recipients: if the citizen is a guardian, this means other guardians, and if the citizen is a foster parent, this means other foster parents. A foster parent can never select a guardian, and vice versa.

### DIFF
--- a/frontend/src/lib-components/molecules/ToggleableRecipient.tsx
+++ b/frontend/src/lib-components/molecules/ToggleableRecipient.tsx
@@ -1,0 +1,77 @@
+{
+  /*
+SPDX-FileCopyrightText: 2017-2022 City of Espoo
+
+SPDX-License-Identifier: LGPL-2.1-or-later
+*/
+}
+
+import { FontAwesomeIcon } from '@fortawesome/react-fontawesome'
+import { faTimes } from 'Icons'
+import React from 'react'
+import styled from 'styled-components'
+
+import { UUID } from 'lib-common/types'
+
+import { fontWeights } from '../typography'
+import { defaultMargins } from '../white-space'
+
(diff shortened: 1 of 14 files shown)


frontend/src/lib-components/molecules/ToggleableRecipient.tsx
```diff
@@ -0,0 +1,77 @@
+{
+  /*
+SPDX-FileCopyrightText: 2017-2022 City of Espoo
+
+SPDX-License-Identifier: LGPL-2.1-or-later
+*/
+}
```
akheron, line 7: Nitpick: A different style of comments is used for copyright in other files

frontend/src/lib-components/molecules/ToggleableRecipient.tsx
```diff
@@ -0,0 +1,77 @@
...
+  const onClick = toggleable
+    ? () => onToggleRecipient(id, !selected)
+    : undefined
+
+  return (
+    <Recipient
+      onClick={onClick}
+      selected={selected}
+      toggleable={toggleable}
+      aria-disabled={!toggleable}
```
akheron, line 63: `aria-disabled` is not needed when it has the same value as `disabled`

jaahon (commented):
The UI allows you to pick a secondary recipient that is not the guardian of all the children, but the backend rejects the message. Other than that everything looks good to me.

jaahon (approved):

PR #3148, opened 2022-12-07
Make sure both requests are ready before rendering them

reynders:
#### Summary
- Because the list of guardians is a combination on the results of two different backend requests, sometimes the rendering rendered a combination of old stale data for the unfinished part and new data from the competed part of the list. This caused error in the console stating that keys of a list are not unique. Ensure that both requests are done before calculating the result
- Might fix: Flaky E2E: adding and removing evaka rights from guardians works


akheron:
Line 98:
```
<Table data-qa="table-of-guardians" data-loading={isReloading}>
```

Wouldn't it be more correct to check for `data-loading` in the test than to show an empty table when one of the requests is reloading?

akheron:
I think this problem should be fixable solely in the frontend without adding a new backend endpoint.

Option 1 (better): You could combine the two requests in the frontend api level, something like this:

```
const getGuardiansAndBlockedGuardians = () => Promise.all([
  getGuardians(),
  getBlockedGuardians()
]).then(([guardians, blockedGuardians]) => ({ guardians, blockedGuardians }))
```

Option 2 (worse): Keep 2 api functions and `combine()`, but remove `renderResult` and use `<UnwrapResult>` instead. It doesn't render stale data:
```
<UnwrapResult result={allGuardians}>
  {(guardians) => (
    /* jsx */
  )}
</UnwrapResult>
```

akheron (approved):

PR #3150, opened 2022-12-08
RTK Query PoC

akheron:


akheron:
RTK Query doesn't seem to work so well for our needs. See #3160 for an alternative.

PR #3158, opened 2022-12-13
Fix missing group placements min date

jaahon:
#### Summary

Missing group placements showed placements from before 1.3.2020 after my previous refactoring.


service/src/integrationTest/kotlin/fi/espoo/evaka/placement/PlacementServiceIntegrationTest.kt
```diff
@@ -1299,4 +1299,46 @@ class PlacementServiceIntegrationTest : FullApplicationTest(resetDbBeforeEach =
             placements
         )
     }
+
+    @Test
+    fun `only missing group placements before evaka launch are included`() {
```
Gekkio, line 1304: "*before* evaka launch"? :thinking:

Gekkio (approved):

PR #3159, opened 2022-12-13
Implement local dev login for suomi.fi

Gekkio:
#### Summary

This makes it possible to log in as different citizens in local development and E2E tests.
Also, refactor apigw to depend less on global config so tests can pass values to functions instead of mutating globals and using hacks.


reynders (approved):
Mob reviewed

PR #3161, opened 2022-12-13
Spread message notifications execution time

reynders:

### DIFF
--- a/service/src/integrationTest/kotlin/fi/espoo/evaka/messaging/MessageNotificationEmailServiceIntegrationTest.kt
+++ b/service/src/integrationTest/kotlin/fi/espoo/evaka/messaging/MessageNotificationEmailServiceIntegrationTest.kt
@@ -106,6 +106,8 @@ class MessageNotificationEmailServiceIntegrationTest :
         }
     }
 
+    val SPREAD_MESSAGE_NOTIFICATION_SECONDS: Long = 60 * 60 * 24
+
     @Test
     fun `notifications are sent to citizens`() {
         val clock = MockEvakaClock(HelsinkiDateTime.now())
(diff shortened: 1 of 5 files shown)


service/src/integrationTest/kotlin/fi/espoo/evaka/messaging/MessageNotificationEmailServiceIntegrationTest.kt
```diff
@@ -106,6 +106,8 @@ class MessageNotificationEmailServiceIntegrationTest :
         }
     }
 
+    val SPREAD_MESSAGE_NOTIFICATION_SECONDS: Long = 60 * 60 * 24
```
Gekkio, line 109: Instead of duplicating the constant here, where it can get out of sync and lead to integration test flakiness, you could have the value only in the MessageService companion object, and access it in both the test and real implementation as `MessageService.SPREAD_MESSAGE_NOTIFICATION_SECONDS`

service/src/main/kotlin/fi/espoo/evaka/shared/async/AsyncJobRunner.kt
```diff
@@ -118,6 +118,11 @@ class AsyncJobRunner<T : AsyncJobPayload>(
         tx.afterCommit(wakeUpHook)
     }
 
+    fun plan(tx: Database.Transaction, jobs: List<JobParams<*>>) {
```
Gekkio, line 121: This should use the `T` type parameter instead of a wildcard: `JobParams<T>`
The wildcard accepts *any* type of async job, even jobs that this particular async job runner can't execute. In this kind of scenario the insert to database succeeds (because it doesn't care) but the wake up hook breaks...not a critical problem since the jobs will eventually be executed by the correct runner, but this can cause unexpected delays and possible test flakiness.

jaahon (approved):

PR #3171, opened 2022-12-16
Placement count report csv decimal separator fix

tmkrepo:
Replace decimal separator `.` to `,` in calculated CSV report values to enable easier Excel processing for FI users.

### DIFF
--- a/frontend/src/employee-frontend/components/reports/PlacementCount.tsx
+++ b/frontend/src/employee-frontend/components/reports/PlacementCount.tsx
@@ -158,9 +158,8 @@ export default React.memo(function PlacementCount() {
         ...areaResult.daycareResults.map((daycareResult) => {
           return {
             ...daycareResult,
-            calculatedPlacements: daycareResult.calculatedPlacements
-              .toString()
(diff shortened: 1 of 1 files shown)


frontend/src/employee-frontend/components/reports/PlacementCount.tsx
```diff
@@ -158,6 +158,9 @@ export default React.memo(function PlacementCount() {
         ...areaResult.daycareResults.map((daycareResult) => {
           return {
             ...daycareResult,
+            calculatedPlacements: daycareResult.calculatedPlacements
```
reynders, line 161: I suggest using calculatedPlacements.toLocaleString('fi-FI') instead "1.23" -> "1,23"

reynders (changes requested):

frontend/src/employee-frontend/components/reports/PlacementCount.tsx
tmkrepo: Switched to locale strings.

reynders (approved):

PR #3184, opened 2022-12-20
Add E2E assertion helpers

akheron:
#### Summary

Drop `textContent`, use `innerText` instead. It's counterproductive to have two ways to do almost the same thing, and `innerText` is more useful for E2E purposes, because it returns what the user actually sees.

Add `Element.assertContentEquals`, `Element.assertAttribute` and `TextInput.assertValue` helpers and use them in some places.

Also rename `ElementCollection.elem` to `only`, to better describe what it does: focuses on the only element of a collection.

### DIFF
--- a/frontend/src/e2e-test/pages/admin/application-details-page.ts
+++ b/frontend/src/e2e-test/pages/admin/application-details-page.ts
@@ -56,8 +56,7 @@ export default class ApplicationDetailsPage {
   async assertNote(index: number, note: string) {
     await waitUntilEqual(
       () =>
-        this.#notes.nth(index).findByDataQa('application-note-content')
-          .textContent,
+        this.#notes.nth(index).findByDataQa('application-note-content').text,
       note
     )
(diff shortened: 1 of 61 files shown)


reynders (approved):

Gekkio (commented):
Suggestions:

- `assertContentEquals` -> `assertTextEquals`
- add `assertText(assertion: (text: string) => boolean)` to handle other use cases (e.g. includes, startsWith, etc...)

akheron:
> Suggestions:
> 
> * `assertContentEquals` -> `assertTextEquals`
> * add `assertText(assertion: (text: string) => boolean)` to handle other use cases (e.g. includes, startsWith, etc...)

Done.

Also renamed:
- `innerText` -> `text`
- `assertAttribute` -> `assertAttributeEquals`
- `assertValue` -> `assertValueEquals`

This makes everything nice and consistent!

PR #3200, opened 2022-12-23
Add permitted actions support for service need

patari:
#### Summary

Replace hard coded roles from frontend to permitted actions returned from backend.

### DIFF
--- a/frontend/src/employee-frontend/api/child/placements.ts
+++ b/frontend/src/employee-frontend/api/child/placements.ts
@@ -43,18 +43,19 @@ export async function createPlacement(
 
 export async function getPlacements(
   childId: UUID
-): Promise<Result<PlacementResponse[]>> {
+): Promise<Result<PlacementResponse>> {
   const config = {
     params: {
       childId
     }
   }
   return client
-    .get<JsonOf<PlacementResponse[]>>('/placements', config)
+    .get<JsonOf<PlacementResponse>>('/placements', config)
     .then((res) => res.data)
-    .then((data) =>
-      data.map(({ data: p, permittedActions }) => ({
-        data: {
+    .then((data) => {
+      return {
+        ...data,
+        placements: data.placements.map((p) => ({
           ...p,
           child: {
             ...p.child,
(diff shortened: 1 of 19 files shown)


Gekkio (commented):
Looks good! :+1: 

One suggestion: you could try to reduce the amount of nested extra wrappers by pulling permitted actions to one or more top-level maps. This is the basic idea:

```kotlin
data class PlacementResponse(
  val data: DaycarePlacementWithDetails, // contains service needs
  val permittedActions: Set<Action.Placement>,
  val serviceNeedPermittedActions: Map<ServiceNeedId, Action.ServiceNeed>,
)
```

or even

```kotlin
data class TopLevelResponse(
  val placements: List<DaycarePlacementWithDetails>, // each contains service needs
  val placementPermittedActions: Map<PlacementId, Set<Action.Placement>>,
  val serviceNeedPermittedActions: Map<ServiceNeedId, Set<Action.ServiceNeed>>,
)
```

In the frontend we then need to pass these maps around and handle nulls because there's no compile-time guarantee the map will contain a non-null entry for each id. But this should not be a terrible price to pay. This is already used in for example here: https://github.com/espoon-voltti/evaka/blob/d9f41c2594f15b4a24715c01bb2b6092859510b2/service/src/main/kotlin/fi/espoo/evaka/daycare/controllers/DaycareController.kt#L686

patari:
> Looks good! +1
> 
> One suggestion: you could try to reduce the amount of nested extra wrappers by pulling permitted actions to one or more top-level maps. This is the basic idea:
> 
> ```kotlin
> data class PlacementResponse(
>   val data: DaycarePlacementWithDetails, // contains service needs
>   val permittedActions: Set<Action.Placement>,
>   val serviceNeedPermittedActions: Map<ServiceNeedId, Action.ServiceNeed>,
> )
> ```
> 
> or even
> 
> ```kotlin
> data class TopLevelResponse(
>   val placements: List<DaycarePlacementWithDetails>, // each contains service needs
>   val placementPermittedActions: Map<PlacementId, Set<Action.Placement>>,
>   val serviceNeedPermittedActions: Map<ServiceNeedId, Set<Action.ServiceNeed>>,
> )
> ```
> 
> In the frontend we then need to pass these maps around and handle nulls because there's no compile-time guarantee the map will contain a non-null entry for each id. But this should not be a terrible price to pay. This is already used in for example here:
> 
> https://github.com/espoon-voltti/evaka/blob/d9f41c2594f15b4a24715c01bb2b6092859510b2/service/src/main/kotlin/fi/espoo/evaka/daycare/controllers/DaycareController.kt#L686

:+1: done.

jaahon (approved):

PR #3215, opened 2023-01-02
Fix preschool club application filter

patari:
#### Summary

Also hides preparatory filters if preparatory is disabled with feature flag.

### DIFF
--- a/service/src/integrationTest/kotlin/fi/espoo/evaka/application/GetApplicationSummaryIntegrationTests.kt
+++ b/service/src/integrationTest/kotlin/fi/espoo/evaka/application/GetApplicationSummaryIntegrationTests.kt
@@ -4,30 +4,44 @@
 
 package fi.espoo.evaka.application
 
-import com.github.kittinunf.fuel.core.extensions.jsonBody
-import com.github.kittinunf.fuel.jackson.responseObject
 import fi.espoo.evaka.FullApplicationTest
 import fi.espoo.evaka.application.persistence.daycare.DaycareFormV0
 import fi.espoo.evaka.insertGeneralTestFixtures
+import fi.espoo.evaka.placement.PlacementType
+import fi.espoo.evaka.shared.DaycareId
 import fi.espoo.evaka.shared.Paged
 import fi.espoo.evaka.shared.auth.AuthenticatedUser
 import fi.espoo.evaka.shared.auth.CitizenAuthLevel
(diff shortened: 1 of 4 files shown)


service/src/integrationTest/kotlin/fi/espoo/evaka/application/GetApplicationSummaryIntegrationTests.kt
```diff
@@ -42,64 +56,180 @@ class GetApplicationSummaryIntegrationTests : FullApplicationTest(resetDbBeforeE
...
+            .containsExactlyInAnyOrder(
+                Tuple(ApplicationType.PRESCHOOL, null),
+                Tuple(ApplicationType.PRESCHOOL, PlacementType.PRESCHOOL_DAYCARE),
+                Tuple(ApplicationType.PRESCHOOL, PlacementType.PRESCHOOL_CLUB)
+            )
     }
 
+    private fun getApplicationSummaries(
+        user: AuthenticatedUser,
+        clock: EvakaClock,
```
jaahon, line 163: Does `clock` need to be a parameter if the function is always called with the same value? I guess the same applies to the `user` parameter.

service/src/main/kotlin/fi/espoo/evaka/application/ApplicationForm.kt
```diff
@@ -523,4 +523,15 @@ data class ServiceNeedOption(
...
+        fun of(option: fi.espoo.evaka.serviceneed.ServiceNeedOption) =
+            ServiceNeedOption(
+                option.id,
+                option.nameFi,
+                option.nameSv,
+                option.nameEn,
+                option.validPlacementType
+            )
+    }
+}
```
jaahon, line 537: Since this function is only used in tests I think a function defined in a test file would be a more appropriate place for this.

service/src/integrationTest/kotlin/fi/espoo/evaka/application/GetApplicationSummaryIntegrationTests.kt
patari: Fixed!

service/src/main/kotlin/fi/espoo/evaka/application/ApplicationForm.kt
patari: Updated!

PR #3216, opened 2023-01-02
VEO should see people from assistance need applications only

reynders:


Gekkio (approved):
My concerns about these changes:

- Performance impact? We're adding another join and extra conditions to *every permission check involving a child in some way*, even if the check has nothing to do with VEOs
- Customization / hard-coding: we're now hard-coding even more stuff in the acl view, when IMHO it would be better to slowly move away from it. Both the location (= acl view) and the data (= `SPECIAL_EDUCATION_TEACHER`) are very rigid, so customization is impossible and we're digging a deeper hole into hard-coded roles
- E2E tests are a very heavy hammer for testing permissions. We already have quick(er) integration tests for various other things related to permissions

Summary: I think we should move away from acl views instead of using them even more.

We can move forward with these changes since they work, and I just predict we'll need to come back to this soon(ish).

PR #3221, opened 2023-01-03
Fix citizen decisions order

patari:
#### Summary

E.g. show preschool decision before preschool daycare because former must be accepted before latter.

### DIFF
--- a/frontend/src/citizen-frontend/decisions/decisions-page/Decisions.tsx
+++ b/frontend/src/citizen-frontend/decisions/decisions-page/Decisions.tsx
@@ -69,7 +69,7 @@ export default React.memo(function Decisions() {
                 (decision) =>
                   'decisionMade' in decision
                     ? decision.decisionMade?.formatIso()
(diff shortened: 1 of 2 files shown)


frontend/src/citizen-frontend/decisions/decisions-page/Decisions.tsx
```diff
@@ -69,7 +69,7 @@ export default React.memo(function Decisions() {
                 (decision) =>
                   'decisionMade' in decision
                     ? decision.decisionMade?.formatIso()
-                    : decision.sentDate.formatIso()
+                    : [decision.sentDate.formatIso(), decision.type]
```
akheron, line 72: What happens if there are mixed strings and arrays here? Should the `decisionMade` timestamp also be wrapped in an array for the sorting to work?

akheron: Never mind, seems to work correctly 👍

PR #3224, opened 2023-01-03
Don't allow fees due to surplus attendance grow too big

zch:
The maximum fee for contract days should not grow bigger than the max fee for the default placement (full days). This is currently in place only for PRESCHOOL_DAYCARE and PREPARATORY_DAYCARE placements.


akheron (changes requested):
Otherwise looks great 🏅

akheron (approved):

PR #3236, opened 2023-01-10
Fix message visibility during undo period

akheron:
#### Background

When an employee sends a message, there's a period during which the message sending can be undone. This was implemented by adjusting message state in the db in an async job that was scheduled to run after the undo period expires.

#### The Problem

This PR fixes two problems:
- The citizen saw unread message count bubble as soon as the employee hit the send button even _before the undo time expired_
- If the message was a reply, the citizen saw that the thread contained unread messages, and saw the message by opening the thread after the undo period expired. _The thread moved to the top of the list only after the async job was run._

After this PR, both the unread count bubble and the message (new message or reply) are visible only after the async job has run.

#### Tech

Keep `message`'s `sent_at` timestamp as `NULL` to mark that the message should not be seen by the receiver(s). The `created` timestamp is used instead of the `sent_at` timestamp when the _sender_ is looking at their own sent messages. The async job sets `sent_at` as well as updating the thread timestamps.

### DIFF
--- a/service/src/main/kotlin/fi/espoo/evaka/messaging/MessageService.kt
+++ b/service/src/main/kotlin/fi/espoo/evaka/messaging/MessageService.kt
@@ -26,29 +26,51 @@ class MessageService(
     private val notificationEmailService: MessageNotificationEmailService
 ) {
     init {
-        asyncJobRunner.registerHandler {
-            db: Database.Connection,
-            _: EvakaClock,
-            msg: AsyncJob.MarkMessageAsSent ->
-            db.transaction {
-                it.upsertReceiverThreadParticipants(msg.threadId, msg.recipientIds, msg.sentAt)
-                it.markMessagesAsSent(msg.messageIds, msg.sentAt)
-            }
-        }
-
-        // TODO: Not used anymore, remove after the above job has been deployed
-        asyncJobRunner.registerHandler {
-            db: Database.Connection,
-            _: EvakaClock,
-            msg: AsyncJob.UpdateMessageThreadRecipients ->
-            db.transaction {
-                it.upsertReceiverThreadParticipants(msg.threadId, msg.recipientIds, msg.sentAt)
-            }
-        }
+        asyncJobRunner.registerHandler(::handleMarkMessageAsSent)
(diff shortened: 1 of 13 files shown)


service/src/main/kotlin/fi/espoo/evaka/messaging/MessageService.kt
```diff
@@ -26,6 +26,17 @@ class MessageService(
...
     init {
+        asyncJobRunner.registerHandler {
+            db: Database.Connection,
+            _: EvakaClock,
+            msg: UrgentAsyncJob.MarkMessageAsSent ->
+            db.transaction {
+                it.upsertReceiverThreadParticipants(msg.threadId, msg.recipientIds, msg.sentAt)
+                it.markMessagesAsSent(msg.messageIds, msg.sentAt)
+            }
+        }
```
akheron, line 37: We need to add one more field to the job payload. Is it better to add a new `AsyncJob` or add a nullable parameter to the existing one? 🤷

service/src/main/kotlin/fi/espoo/evaka/shared/async/AsyncJob.kt
```diff
@@ -79,6 +79,15 @@ sealed interface UrgentAsyncJob : AsyncJobPayload {
     ) : UrgentAsyncJob {
         override val user: AuthenticatedUser? = null
     }
+
+    data class MarkMessageAsSent(
```
akheron, line 83: One field need to be added to the job payload. Is it better to add a new AsyncJob or add a nullable parameter to the existing one? 🤷

service/src/main/kotlin/fi/espoo/evaka/shared/async/AsyncJob.kt
```diff
@@ -219,6 +219,15 @@ sealed interface AsyncJob : AsyncJobPayload {
         override val user: AuthenticatedUser? = null
     }
 
+    data class MarkMessageAsSent(
+        val threadId: MessageThreadId,
+        val recipientIds: Set<MessageAccountId>,
+        val messageIds: Set<MessageId>,
+        val sentAt: HelsinkiDateTime
```
akheron, line 226: Use `messageContentId` and `sentAt` instead

PR #3239, opened 2023-01-11
Avoid double arrival/departure bookings

akheron:
#### Summary

Refresh the unit's attendance data each time a child page is opened in employee mobile. This way it should be much more unlikely that two employees mark the same child as arrived or departed.


akheron:
It seems that the actual problem is the network, not multiple devices not being in sync.

PR #3249, opened 2023-01-13
Replace weak custom Profile types with validation code

Gekkio:
#### Summary

The custom types didn't provide any type safety since we just did a type assertion and assumed the types were correct. Embrace the unknown typing instead, and validate/cast the data.


akheron (approved):
Not an expert in this, but LGTM

PR #3253, opened 2023-01-16
Allow arrival after a departure

reynders:
#### Summary
Allow arrival after a non work departure.

Example: If employee departs 1400, allow arriving 1500. This leaves a gap in the attendance data between 14-15

### DIFF
--- a/service/src/main/kotlin/fi/espoo/evaka/attendance/StaffAttendanceQueries.kt
+++ b/service/src/main/kotlin/fi/espoo/evaka/attendance/StaffAttendanceQueries.kt
@@ -12,7 +12,6 @@ import fi.espoo.evaka.shared.StaffAttendanceId
 import fi.espoo.evaka.shared.db.Database
 import fi.espoo.evaka.shared.db.mapColumn
 import fi.espoo.evaka.shared.db.mapRow
-import fi.espoo.evaka.shared.domain.EvakaClock
 import fi.espoo.evaka.shared.domain.FiniteDateRange
 import fi.espoo.evaka.shared.domain.HelsinkiDateTime
 import fi.espoo.evaka.shared.domain.HelsinkiDateTimeRange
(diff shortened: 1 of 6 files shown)


service/src/main/kotlin/fi/espoo/evaka/attendance/StaffAttendanceQueries.kt
```diff
@@ -565,6 +566,26 @@ FROM staff_attendance_realtime WHERE employee_id = :employeeId AND departed IS N
...
+    employeeId: EmployeeId,
+    clock: EvakaClock
+): StaffAttendance? =
+    createQuery(
+            """
+SELECT id, employee_id, group_id, arrived, departed, occupancy_coefficient, type
+FROM staff_attendance_realtime 
+WHERE employee_id = :employeeId 
+    AND departed IS NOT NULL
+    AND DATE(departed) = :today
```
akheron, line 579: This probably doesn't work around midnight because `departed` is in UTC and `:today` is in Helsinki timezone.

akheron: Discussed in tech coffee break: This also doesn't use indexes very well, so it will become slow when an employee has lots of attendances.

service/src/main/kotlin/fi/espoo/evaka/attendance/StaffAttendanceQueries.kt
```diff
@@ -565,6 +566,26 @@ FROM staff_attendance_realtime WHERE employee_id = :employeeId AND departed IS N
...
+fun Database.Read.getLatestDepartureToday(
+    employeeId: EmployeeId,
+    clock: EvakaClock
+): StaffAttendance? =
+    createQuery(
+            """
+SELECT id, employee_id, group_id, arrived, departed, occupancy_coefficient, type
+FROM staff_attendance_realtime 
+WHERE employee_id = :employeeId 
+    AND departed IS NOT NULL
```
akheron, line 578: NULL check is not needed

service/src/main/kotlin/fi/espoo/evaka/attendance/StaffAttendanceQueries.kt
reynders: Agreed to leave the index out

akheron (approved):

PR #3255, opened 2023-01-16
 Disallow marking arrival on non-operational days

Gekkio:


frontend/src/e2e-test/specs/6_mobile/realtime-staff-attendances.spec.ts
```diff
@@ -126,6 +129,22 @@ describe('Realtime staff attendance page', () => {
     await staffAttendancePage.assertPresentStaffCount(0)
   })
 
+  test('New staff member cannot be added on a non-operational day', async () => {
```
akheron, line 132: I think `Staff member cannot be marked as arrived on a non-operational day` or something like that would be more correct. Externals are "added", regular staff is "marked as arrived".

akheron (approved):

PR #3261, opened 2023-01-17
EVAKA-4694 service worker messaging: send a message

zch:
#### Summary
This is the first part of the service worker messaging functionality. It implements the following:
- Messaging access for service workers, with a shared message account
- Messaging access for citizens without actively placed children, IF an application is sent AND any message has been received
- Sending messages to citizens with active applications as a service worker

### DIFF
--- a/service/src/main/kotlin/fi/espoo/evaka/messaging/MessageQueries.kt
+++ b/service/src/main/kotlin/fi/espoo/evaka/messaging/MessageQueries.kt
@@ -18,6 +18,7 @@ import fi.espoo.evaka.shared.MessageRecipientId
 import fi.espoo.evaka.shared.MessageThreadFolderId
 import fi.espoo.evaka.shared.MessageThreadId
 import fi.espoo.evaka.shared.Paged
+import fi.espoo.evaka.shared.PersonId
 import fi.espoo.evaka.shared.db.Database
 import fi.espoo.evaka.shared.db.mapColumn
 import fi.espoo.evaka.shared.domain.BadRequest
@@ -962,6 +963,13 @@ data class MunicipalMessageReceiversResult(
     val unitName: String
 )
 
+data class ServiceWorkerMessageReceiversResult(
+    val accountId: MessageAccountId,
+    val personId: PersonId,
+    val firstName: String,
+    val lastName: String
+)
+
 fun Database.Read.getReceiversForNewMessage(
     idFilter: AccessControlFilter<MessageAccountId>,
     today: LocalDate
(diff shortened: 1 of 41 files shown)


service/src/main/resources/db/migration/V294__service_worker_message_account.sql
```diff
@@ -0,0 +1,13 @@
+ALTER TABLE message_account DROP COLUMN type CASCADE;
+ALTER TABLE message_account ADD COLUMN type message_account_type;
```
akheron, line 2: Is dropping and re-adding actually needed? The column already exists and is the correct type.

service/src/main/resources/db/migration/V294__service_worker_message_account.sql
```diff
@@ -0,0 +1,13 @@
...
+
+CREATE UNIQUE INDEX only_one_service_worker_account ON message_account (type) WHERE type = 'SERVICE_WORKER';
+
+UPDATE message_account SET type = CASE
+    WHEN employee_id IS NOT NULL THEN 'PERSONAL'::message_account_type
+    WHEN daycare_group_id IS NOT NULL THEN 'GROUP'::message_account_type
+    WHEN person_id IS NOT NULL THEN 'CITIZEN'::message_account_type
+    ELSE 'MUNICIPAL'::message_account_type
+END;
```
akheron, line 11: If you don't drop the `type` column, you don't need this `UPDATE` either.

service/src/main/kotlin/fi/espoo/evaka/messaging/MessageQueries.kt
```diff
@@ -1053,7 +1091,37 @@ fun Database.Read.getReceiversForNewMessage(
...
+        createQuery(
+                """
+        WITH accounts AS (
+            SELECT id, type, daycare_group_id, employee_id, person_id FROM message_account
+            WHERE id = ANY(:accountIds) AND type = 'SERVICE_WORKER'::message_account_type
+        )
+        SELECT acc.id AS account_id, p.id as person_id, p.first_name, p.last_name 
+        FROM accounts acc, person p
+        JOIN application app ON p.id = app.guardian_id
+        WHERE app.status = 'SENT'
```
akheron, line 1104: This query is probably going to be slow. There's no suitable index on `application.status`, so it will probably be a seq scan on either `person` or `application`.

Is this a temporary query or will it be there also when the feature is ready?

akheron:
Idea: Would it be more meaningful to add `municipalAccountName` or `serviceWorkerAccountName` in the Kotlin side, instead of binding it to many sql queries? Not sure exactly what this would look like, but it might make the code cleaner.

service/src/main/resources/db/migration/V294__service_worker_message_account.sql
akheron: It's needed because it seems that `GENERATED ALWAYS` cannot be dropped.

akheron (approved):

PR #3264, opened 2023-01-18
Do not require setting attendance for weekend if sibling is in 24/7 daycare

reynders:


akheron (approved):
1 comment, otherwise looks good 👍

akheron (approved):

PR #3266, opened 2023-01-19
Hide some application info from other guardian

reynders:

### DIFF
--- a/frontend/src/e2e-test/specs/0_citizen/citizen-daycare-application.spec.ts
+++ b/frontend/src/e2e-test/specs/0_citizen/citizen-daycare-application.spec.ts
@@ -242,6 +242,8 @@ describe('Citizen daycare applications', () => {
     const applicationReadView = await applications.viewApplication(
       applicationId
     )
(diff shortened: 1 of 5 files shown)


frontend/src/e2e-test/specs/0_citizen/citizen-daycare-application.spec.ts
```diff
@@ -236,8 +236,12 @@ describe('Citizen daycare applications', () => {
...
-    await new CitizenApplicationsPage(
-      otherGuardianPage
-    ).assertApplicationExists(applicationId)
+
+    const applications = new CitizenApplicationsPage(otherGuardianPage)
+    await applications.assertApplicationExists(applicationId)
+    const applicationReadView = await applications.viewApplication(
+      applicationId
+    )
+    await applicationReadView.contactInfoSection.waitUntilHidden()
```
akheron, line 245: Should there first be a check that "something" is visible, and only after that a check that the contact info section is hidden?

PR #3267, opened 2023-01-19
EVAKA-4694 Prefill message and recipient selector

zch:
#### Summary
Prefills the recipient and subject line for messages sent by service workers. Also removes some potentially slow queries introduced in the previous PR, as they are now unnecessary.

### DIFF
--- a/frontend/src/employee-frontend/components/messages/MessagesPage.tsx
+++ b/frontend/src/employee-frontend/components/messages/MessagesPage.tsx
@@ -11,6 +11,7 @@ import {
 } from 'lib-common/generated/api-types/messaging'
 import HelsinkiDateTime from 'lib-common/helsinki-date-time'
 import { UUID } from 'lib-common/types'
+import { useApiState } from 'lib-common/utils/useRestApi'
 import MessageEditor from 'lib-components/employee/messages/MessageEditor'
 import Container from 'lib-components/layout/Container'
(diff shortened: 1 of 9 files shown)


frontend/src/employee-frontend/components/messages/MessagesPage.tsx
```diff
@@ -148,6 +158,25 @@ export default React.memo(function MessagesPage({
...
+        accountId: selectedAccount?.account.id,
+        receivers: [
+          {
+            id: person.id,
+            name: formatPersonName(person, i18n, true),
+            type: 'CITIZEN'
+          }
+        ]
+      }
+    ] as MessageReceiversResponse[]
```
akheron, line 177: `foo as MessageReceiversResponse[]` is not typesafe, because `as` doesn't do any type checking. Instead, it just trusts that you are right.

It would be better to annotate the function return type instead:

```
const getReceivers = useCallback((): MessageReceiversResponse[] | undefined => {
  ...
```

If you do this, you might notice that the returned value is actually not correct...

akheron:
Is there validation in the backend side that the sender is allowed to send a message to this person?

zch:
> Is there validation in the backend side that the sender is allowed to send a message to this person?

Added a note for implementing this to our board.

akheron (approved):

PR #3270, opened 2023-01-20
Changed placement plan draft creation to use correct preferred start date from application

tmkrepo:
- Preschool start date was used for both preschool and preschool connected daycare despite the application form having separate preferred start dates for both
- Earlier functionality relegated to fallback if no preferred start date for connected daycare is found

### DIFF
--- a/service/src/main/kotlin/fi/espoo/evaka/placement/PlacementPlanService.kt
+++ b/service/src/main/kotlin/fi/espoo/evaka/placement/PlacementPlanService.kt
@@ -89,7 +89,7 @@ class PlacementPlanService(private val asyncJobRunner: AsyncJobRunner<AsyncJob>,
                         PlacementType.PRESCHOOL_DAYCARE,
                         PlacementType.PREPARATORY_DAYCARE ->
                             FiniteDateRange(
-                                startDate,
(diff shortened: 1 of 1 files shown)


service/src/main/kotlin/fi/espoo/evaka/placement/PlacementPlanService.kt
```diff
@@ -89,7 +89,10 @@ class PlacementPlanService(private val asyncJobRunner: AsyncJobRunner<AsyncJob>,
                         PlacementType.PRESCHOOL_DAYCARE,
                         PlacementType.PREPARATORY_DAYCARE ->
                             FiniteDateRange(
-                                startDate,
+                                maxOf(
+                                    minStartDate,
+                                    form.preferences.connectedDaycarePreferredStartDate ?: startDate
+                                ),
```
akheron, line 95: Shouldn't this be applied also to `PRESCHOOL_CLUB` below?

tmkrepo: Yes, I verified the occurrence of the problem for preschool club and applying the same logic fixes the start date

akheron (approved):

PR #3279, opened 2023-01-24
SAML nameIdFormat and userId configurable in AD SAML messages for apigw

mrosin:
#### Summary
Turku uses ADFS instead of AzureAD. Hardcoded userId (objectidentifier claim) is not available in ADFS. NameIdFormat by default is unspecified in ADFS. Therefor made both configurable through environment variables AD_USER_ID_KEY, AD_NAME_ID_FORMAT.

Turku ADFS encrypts SAML responses. Added environment variable to be able to define passport to decrypt responses, AD_DECRYPT_ASSERTIONS

Old hardcoded values are used as defaults. Should not require any changes from other cities.


Gekkio (approved):
:gem: :tada:

PR #3292, opened 2023-01-27
EVAKA-4694 new message creates application note

zch:
#### Summary
An application note is created whenever a service worker sends a message relating to the application. The note contains a link to the message thread, as well as the contents of the message that was sent.

Deeplinking to a certain message thread didn't work earlier, and had to be fixed. This was done by adding a special message box named "thread", containing (i.e. loading) only the thread that is requested in the URI. This was the least invasive way to get this working - a complete refactoring, which might have resulted in a better solution, would carry more risk of introducing new bugs.


service/src/main/kotlin/fi/espoo/evaka/messaging/MessageController.kt
```diff
@@ -345,6 +370,14 @@ class MessageController(
...
                     if (body.draftId != null)
                         tx.deleteDraft(accountId = accountId, draftId = body.draftId)
+                    if (body.relatedApplicationId != null) {
+                        tx.createApplicationNote(
+                            applicationId = body.relatedApplicationId,
+                            content = body.content,
+                            createdBy = user.evakaUserId,
+                            messageThreadId = threadIds.first()
+                        )
+                    }
```
akheron, line 380: If the message sending is undone, the application note should also be deleted.

akheron (changes requested):

service/src/main/kotlin/fi/espoo/evaka/application/notes/ApplicationNoteQueries.kt
```diff
@@ -127,3 +127,12 @@ fun Database.Transaction.deleteApplicationNote(id: ApplicationNoteId) {
 
     createUpdate(sql).bind("id", id).execute()
 }
+
+fun Database.Transaction.deleteApplicationNotesLinkedToMessages(
+    messageContentIds: List<MessageContentId>
```
akheron, line 132: I would make the parameter `Set<MessageContentId>` because there can be duplicates (in theory).

service/src/main/kotlin/fi/espoo/evaka/application/notes/ApplicationNoteQueries.kt
```diff
@@ -17,8 +18,10 @@ fun Database.Read.getApplicationNotes(applicationId: ApplicationId): List<Applic
 SELECT 
     n.id, n.application_id, n.content, 
     n.created, n.created_by, (SELECT name FROM evaka_user WHERE id = n.created_by) AS created_by_name,
-    n.updated, n.updated_by, (SELECT name FROM evaka_user WHERE id = n.updated_by) AS updated_by_name
+    n.updated, n.updated_by, (SELECT name FROM evaka_user WHERE id = n.updated_by) AS updated_by_name,
+    n.message_content_id, m.thread_id as message_thread_id
 FROM application_note n
+LEFT JOIN message m ON m.content_id = n.message_content_id
```
akheron, line 24: This might (in theory) have the effect that duplicate note rows are returned, if this join matches multiple `message` rows. Not sure if this is a problem in practise, because messages related to application are always sent to a single recipient in a single thread, so there's only one `message`.

akheron (approved):
Looks good, a few (theoretical) comments 😅

service/src/main/kotlin/fi/espoo/evaka/application/notes/ApplicationNoteQueries.kt
zch: Should never happen™ 😅

But if it does, I'm sure we'll get a bug report that multiple copies of the same note are visible in the UI.

PR #3302, opened 2023-01-31
EVAKA-4694 Citizen can reply to service worker messages

zch:
Replies to service worker messages create notes with the contents of the reply in the application


akheron (approved):
🏅

PR #3306, opened 2023-02-02
Manual duplication report

tmkrepo:
Adds a report for identifying cases where manual duplication of child instances is required:
- lists children that have 2 simultaneous accepted decisions for preschool and supplementary early childhood education **in different units**
- report lists necessary information for manually verifying the need for duplication:
  - child name and link to detailed information
  - date of birth
  - supplementary ECE unit
  - supplementary ECE service need option
  - supplementary ECE duration
  - preschool unit
  - preschool duration


frontend/src/employee-frontend/components/reports/ManualDuplicationReport.tsx
```diff
@@ -0,0 +1,182 @@
...
+  )
+
+  return (
+    <Container>
+      <ReturnButton label={i18n.common.goBack} />
+      <ContentArea opaque>
+        <Title size={1}>{i18n.reports.manualDuplication.title}</Title>
+        {rowsResult.isLoading && <Loader />}
+        {rowsResult.isFailure && <span>{i18n.common.loadingFailed}</span>}
+        {rowsResult.isSuccess && (
```
akheron, line 89: `renderResult()` provides better user experience than `Result.getOrElse()` and removes the need for manual `isLoading`/`isFailure`/`isSuccess` handling.

akheron (changes requested):

service/src/main/kotlin/fi/espoo/evaka/reports/ManualDuplicationReportController.kt
```diff
@@ -0,0 +1,108 @@
...
+                    it.getManualDuplicationReportRows()
+                }
+            }
+            .also { Audit.ManualDuplicationReportRead.log() }
+    }
+
+    private fun Database.Read.getManualDuplicationReportRows(): List<ManualDuplicationReportRow> {
+        val sql =
+            """
+select supp_a.id                                       as supplementary_application_id,
```
akheron, line 46: "Connected daycare" is the term that's widely used in the code base. It would be better to use that instead of "supplementary".

akheron (approved):

PR #3307, opened 2023-02-02
Fix preschool/preparatory invoice capping with contract surplus days

akheron:
#### Summary

Also cap preschool/preparatory max price with `useContractDaysAsDailyFeeDivisor = false`


zch (approved):
Looks great 🥇

PR #3317, opened 2023-02-06
EVAKA-4694 Disable the new message button for Service workers

zch:


service/src/main/kotlin/fi/espoo/evaka/pis/SystemController.kt
```diff
@@ -184,7 +184,9 @@ class SystemController(
                                 createDraftInvoices =
                                     permittedGlobalActions.contains(
                                         Action.Global.CREATE_DRAFT_INVOICES
-                                    )
+                                    ),
+                                composeNewMessage =
+                                    !user.globalRoles.contains(UserRole.SERVICE_WORKER)
```
akheron, line 189: Discussed face-to-face: This is not good, will be done in another way

akheron (approved):

PR #3344, opened 2023-02-13
Espoo customizations

ilja-git:


zch (approved):

Gekkio (changes requested):
`espoo-logo.html` should be re-fetched from S3, because it was scaled down to a smaller size more than a week ago (original consumed 50 MB memory uncompressed)

PR #3355, opened 2023-02-15
Improve manual duplication report

tmkrepo:
Adds a new filter to the manual duplication report for selecting one of two view modes:
- show only cases where the person has already been duplicated
- show only cases without a duplicated person

### DIFF
--- a/service/src/main/kotlin/fi/espoo/evaka/reports/ManualDuplicationReportController.kt
+++ b/service/src/main/kotlin/fi/espoo/evaka/reports/ManualDuplicationReportController.kt
@@ -25,7 +25,7 @@ class ManualDuplicationReportController(private val accessControl: AccessControl
         db: Database,
         user: AuthenticatedUser,
         clock: EvakaClock,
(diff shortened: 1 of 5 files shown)


service/src/main/kotlin/fi/espoo/evaka/reports/ManualDuplicationReportController.kt
```diff
@@ -23,7 +24,8 @@ class ManualDuplicationReportController(private val accessControl: AccessControl
     fun getPlacementCountReport(
         db: Database,
         user: AuthenticatedUser,
-        clock: EvakaClock
+        clock: EvakaClock,
+        @RequestParam("toggleDuplicatedCaseVisibility") toggleDuplicatedCaseVisibility: Boolean?
```
zch, line 28: The word "toggle" implies that the state of this is stored somewhere. The actual result of this doesn't seem to be very suitable to modeling by a boolean though 🤔 

Could the parameter be an enum with `DUPLICATED` and `NONDUPLICATED` or something?

zch (approved):

PR #3370, opened 2023-02-20
Attendance reservation report improvements

reynders:
#### Summary
- Resolution 30min -> 15min
- Scroll the report to 6AM

### DIFF
--- a/frontend/src/employee-frontend/components/reports/AttendanceReservation.tsx
+++ b/frontend/src/employee-frontend/components/reports/AttendanceReservation.tsx
@@ -2,7 +2,7 @@
 //
 // SPDX-License-Identifier: LGPL-2.1-or-later
 
-import React, { useState } from 'react'
+import React, { RefObject, useEffect, useRef, useState } from 'react'
 import styled from 'styled-components'
 
 import { getDaycareGroups, getDaycares } from 'employee-frontend/api/unit'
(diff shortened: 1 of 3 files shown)


service/src/main/kotlin/fi/espoo/evaka/reports/AttendanceReservationReportController.kt
```diff
@@ -179,15 +179,15 @@ private fun Database.Read.getAttendanceReservationReport(
...
         CROSS JOIN groups g
         LEFT JOIN reservations r ON
           (r.group_id IS NULL AND g.id IS NULL OR r.group_id = g.id)
           AND
           dateTime BETWEEN
-          r.date + r.start_time - interval '30 minutes' + interval '1 second'
+          r.date + r.start_time - interval '15 minutes' + interval '1 second'
           AND
-          r.date + r.end_time + interval '30 minutes' - interval '1 second'
+          r.date + r.end_time + interval '15 minutes' - interval '1 second'
```
akheron, line 190: Using `<` and `>` instead of `BETWEEN` would make the `+/- interval '1 second'` unnecessary

frontend/src/employee-frontend/components/reports/AttendanceReservation.tsx
```diff
@@ -81,6 +81,13 @@ export default React.memo(function AttendanceReservation() {
     [unitId, filters]
   )
 
+  useEffect(() => {
+    const elements = document.getElementsByClassName('time-row-05:30')
+    if (elements && elements.length > 0) {
+      elements[0].scrollIntoView()
+    }
+  }, [report])
+
```
akheron, line 90: `useRef` is better than finding an element by using a fixed class name.

Our own `scrollRefIntoView` is better than builtin `scrollIntoView`, because it does nothing in e2e tests, uses smooth scrolling, waits for the element to be in the dom, etc.

```
const autoScrollRef = useRef<HTMLTableRowElement>(null)
useEffect(() => {
  scrollRefIntoView(autoScrollRef)
}, [])

...

<Tr ref={autoScrollRef}>
```

akheron (approved):

PR #3372, opened 2023-02-21
Configure access logging in code

Gekkio:
#### Summary

logback-access-spring-boot-starter doesn't support Spring Boot 3.0, but we can replace this fragile dependency with ~50 lines of boilerplate that has more compile-time safety than our previous XML-based configuration.

Also adds userId/userIdHash to service access logs now that we can access those with very little effort.


service/src/main/kotlin/fi/espoo/evaka/EspooConfig.kt
```diff
@@ -151,18 +158,64 @@ class EspooConfig {
...
+    LogbackValve().apply {
+        filename = "we-use-programmatic-config.xml"
+
+        val staticFields =
+            arrayOf(
+                "type" to "app-requests-received",
+                "version" to 1,
+                "appBuild" to System.getenv("APP_BUILD"),
+                "appCommit" to System.getenv("APP_COMMIT"),
+                "volttiEnv" to System.getenv("VOLTTI_ENV"),
```
akheron, line 180: This should be just `"env"`?

service/src/main/kotlin/fi/espoo/evaka/EspooConfig.kt
```diff
@@ -151,18 +158,64 @@ class EspooConfig {
...
+                "hostIp" to System.getenv("HOST_IP")
+            )
```
akheron, line 182: Is `"appName"` missing ?

akheron (approved):

PR #3381, opened 2023-02-24
Add vasu copy to duplicates

patari:

### DIFF
--- a/service/src/integrationTest/kotlin/fi/espoo/evaka/vasu/VasuIntegrationTest.kt
+++ b/service/src/integrationTest/kotlin/fi/espoo/evaka/vasu/VasuIntegrationTest.kt
@@ -9,6 +9,7 @@ import fi.espoo.evaka.insertGeneralTestFixtures
 import fi.espoo.evaka.pis.service.insertGuardian
 import fi.espoo.evaka.placement.checkAndCreateGroupPlacement
 import fi.espoo.evaka.shared.ChildId
+import fi.espoo.evaka.shared.PersonId
 import fi.espoo.evaka.shared.VasuDocumentId
 import fi.espoo.evaka.shared.VasuTemplateId
 import fi.espoo.evaka.shared.auth.AuthenticatedUser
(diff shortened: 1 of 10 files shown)


service/src/integrationTest/kotlin/fi/espoo/evaka/vasu/VasuIntegrationTest.kt
```diff
@@ -338,6 +341,15 @@ class VasuIntegrationTest : FullApplicationTest(resetDbBeforeEach = true) {
                 CurriculumType.PRESCHOOL -> ChildLanguage("kiina", "kiina")
             }
 
+        db.transaction { tx ->
```
zch, line 344: Should there be an assertion somewhere as well?

service/src/main/kotlin/fi/espoo/evaka/vasu/VasuController.kt
```diff
@@ -389,12 +393,46 @@ class VasuController(
...
+                if (events.contains(MOVED_TO_CLOSED)) {
+                    val duplicates = tx.listPersonByDuplicateOf(document.basics.child.id)
+                    if (duplicates.isNotEmpty()) {
+                        val child = tx.getChild(document.basics.child.id)!!
+                        val template = tx.getVasuTemplate(document.templateId)!!
+                        val allEvents = document.events + addedEvents
+                        duplicates.forEach { duplicate ->
+                            val duplicateChild =
+                                tx.getChild(duplicate.id)
+                                    ?: tx.createChild(child.copy(id = duplicate.id))
```
zch, line 414: I'm probably missing something here, but why do we need to create a `child` for a duplicated person here? Are these not created when the person is duplicated?

Or is this just being defensive for a potential case of bad data?

service/src/integrationTest/kotlin/fi/espoo/evaka/vasu/VasuIntegrationTest.kt
patari: Added!

service/src/main/kotlin/fi/espoo/evaka/vasu/VasuController.kt
patari: The latter.

zch (approved):

PR #3393, opened 2023-02-28
Schedule patu report sending for 06:00 every day

zch:


service/src/main/kotlin/fi/espoo/evaka/shared/job/JobSchedule.kt
```diff
@@ -112,6 +112,11 @@ data class ScheduledJobSettings(
                         enabled = false,
                         schedule = JobSchedule.cron("0 0 18 * * 0") // Sunday 18:00
                     )
+                ScheduledJob.SendPatuReport ->
+                    ScheduledJobSettings(
+                        enabled = false,
+                        schedule = JobSchedule.daily(LocalTime.of(3, 0))
```
huhtari, line 118: New spec is 06:00

reynders (approved):

PR #3399, opened 2023-03-01
Citizen messaging frontend refactor

akheron:
#### Summary

- Use react-query
- Use `Result` properly
- Make the context smaller and better
- Use `renderResult` better to show spinners when and where they belong


zch (approved):
🏅

PR #3416, opened 2023-03-06
Move Digitransit API calls to server and add API key

zch:
- Move Digitransit API calls to apigw
- Use API key from the environment (Added to infra in https://github.com/espoon-voltti/evaka-infra/pull/657)
- Refactor to call mocked digitransit API (in dev-api) when run locally or in e2e tests

### DIFF
--- a/apigw/src/enduser/app.ts
+++ b/apigw/src/enduser/app.ts
@@ -25,6 +25,7 @@ import session, {
 } from '../shared/session'
 import publicRoutes from './publicRoutes'
 import routes from './routes'
+import mapRoutes from './mapRoutes'
 import authStatus from './routes/auth-status'
 import { cacheControl } from '../shared/middleware/cache-control'
 import { RedisClient } from 'redis'
@@ -69,6 +70,7 @@ export default function enduserGwApp(config: Config, redisClient: RedisClient) {
     const router = Router()
 
     router.use(publicRoutes)
+    router.use(mapRoutes)
     const suomifiSamlConfig = createSuomiFiSamlConfig(config.sfi, redisClient)
     router.use(
       createSamlRouter(config, {
--- a/frontend/src/citizen-frontend/map/api.ts
+++ b/frontend/src/citizen-frontend/map/api.ts
@@ -10,7 +10,6 @@ import { ApplicationType } from 'lib-common/generated/api-types/application'
 import { PublicUnit } from 'lib-common/generated/api-types/daycare'
 import { Coordinate } from 'lib-common/generated/api-types/shared'
 import { JsonOf } from 'lib-common/json'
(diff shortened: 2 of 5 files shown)


apigw/src/enduser/maproutes.ts
```diff
@@ -0,0 +1,39 @@
...
+
+const router = Router()
+
+router.get('/map-api/autocomplete', (req: Request, res) => {
+  axios
+    .get('https://api.digitransit.fi/geocoding/v1/autocomplete', {
+      params: req.query,
+      headers: {
+        'digitransit-subscription-key':
+          process.env.EVAKA_DIGITRANSIT_API_KEY ?? ''
```
Gekkio, line 16: Environment variables should be read exactly once during startup in config.ts

apigw/src/enduser/maproutes.ts
```diff
@@ -0,0 +1,39 @@
+// SPDX-FileCopyrightText: 2017-2023 City of Espoo
+//
+// SPDX-License-Identifier: LGPL-2.1-or-later
+
+import express, { Router, Request } from 'express'
+import axios from 'axios'
+
+const router = Router()
+
+router.get('/map-api/autocomplete', (req: Request, res) => {
```
Gekkio, line 10: Suggestion: use a proxy based on `express-http-proxy` instead of custom endpoints. You can add the extra header easily using `proxyReqOptDecorator`

These custom solutions throw away all response headers and the post endpoint does unnecessary extra work just to parse the proxied body into memory

apigw/src/enduser/app.ts
```diff
@@ -25,6 +25,7 @@ import session, {
 } from '../shared/session'
 import publicRoutes from './publicRoutes'
 import routes from './routes'
+import maproutes from './maproutes'
```
Gekkio, line 28: Tiny nitpick: why not `map-routes` or `mapRoutes`? This is the third naming convention in the same import list

Gekkio (changes requested):

frontend/src/citizen-frontend/map/api.ts
```diff
@@ -45,9 +44,7 @@ type AutocompleteResponse = {
 export const autocompleteAddress = async (
   text: string
 ): Promise<MapAddress[]> => {
-  const url = isAutomatedTest
-    ? '/api/internal/dev-api/digitransit/autocomplete'
-    : 'https://api.digitransit.fi/geocoding/v1/autocomplete'
+  const url = '/api/internal/dev-api/digitransit/autocomplete'
```
Gekkio, line 47: This url looks wrong :thinking:

Gekkio (approved):

PR #3451, opened 2023-03-22
Allow employee to edit language at home info

zch:
- Move the `language_at_home` and  `language_at_home_details` columns from the `person` table to the `child` table
- API changed to reflect the above
- Move the language at home fields to the additional details section in the UI


Gekkio (approved):
:gem:

PR #3454, opened 2023-03-23
Add duplicated child identifier to citizen frontend

patari:
#### Summary

Parent should be able to distinct duplicate.


frontend/src/citizen-frontend/utils/duplicated-child-utils.ts
```diff
@@ -0,0 +1,23 @@
...
+import { Translations } from 'citizen-frontend/localization'
+import { Child } from 'lib-common/generated/api-types/children'
+import { ReservationChild } from 'lib-common/generated/api-types/reservations'
+
+export const toDuplicatedChildIds = (
+  children: Array<Child | ReservationChild>
+) =>
+  children
+    .filter((child) => child.duplicateOf !== null)
+    .flatMap((child) => [child.id, child.duplicateOf as string])
```
akheron, line 14: Just FYI, you can use `flatMap` for filtering in one call. The types will also be correct automatically:

```suggestion
  children
    .flatMap((child) => child.duplicateOf !== null ? [child.id, child.duplicateOf] : [])
```

frontend/src/citizen-frontend/children/ChildHeader.tsx
```diff
@@ -55,6 +55,8 @@ export default React.memo(function ChildHeader({
           {`${firstName} ${lastName}`}
         </H1>
         {group && <Title>{group.name}</Title>}
+        <br />
+        {unit && <span>{unit.name}</span>}
```
akheron, line 59: This will affect all users and all municipalities, not just duplicates. I think it's OK because the group name is shown anyways, but maybe we should check with if the UX people want to say something about how it looks.

frontend/src/citizen-frontend/navigation/MobileNav.tsx
```diff
@@ -309,6 +315,7 @@ const ChildrenMenu = React.memo(function ChildrenMenu({
             data-qa={`children-menu-${child.id}`}
             to={`/children/${child.id}`}
             onClick={closeMenu}
+            alignRight
```
akheron, line 318: This also affects everyone. But I think this is also a good change because all the other dropdowns are aligned to the right anyway.

patari: I think texts are already aligned right in mobile so no changes here, i hope.

service/src/main/kotlin/fi/espoo/evaka/reservations/ReservationControllerCitizen.kt
```diff
@@ -366,8 +371,15 @@ LEFT JOIN (
     ) p
     GROUP BY p.child_id
 ) p ON p.child_id = c.child_id
+LEFT JOIN LATERAL (
+  SELECT child_id, type
+  FROM placement
+  WHERE child_id = ch.id AND :today <= end_date
+  ORDER BY start_date
+  LIMIT 1
+) upcoming_pl ON true
```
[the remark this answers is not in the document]
patari: :crossed_fingers:

service/src/main/kotlin/fi/espoo/evaka/children/Children.kt
```diff
@@ -4,22 +4,35 @@
...
     val hasPedagogicalDocuments: Boolean,
     val hasCurriculums: Boolean
-)
+) {
+    @Deprecated(
+        message = "For UI backward compatibility",
+        replaceWith = ReplaceWith("upcomingPlacementType != null")
+    )
+    fun getHasUpcomingPlacements() = upcomingPlacementType != null
+}
```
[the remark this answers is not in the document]
patari: Well, it's not needed if we don't want uninterrupted deployment (property is included in the response even though its not in the codegen)

patari:
Fixed! Also rebased to include latest owasp suppressions.

PR #3455, opened 2023-03-23
Update Koski date range handling

Gekkio:
#### Summary

- stop adjusting disability ranges. The adjustments added an unnecessary extra confusion, because the data visible in eVaka didn't match exactly what was sent
- trim all sent date ranges to the spanning range of placements. We already fetch from the db only date ranges that *overlap* with the spanning range, but this alone doesn't guarantee we won't try to send ranges that continue beyond the study right


akheron (approved):
Would this change require re-sending all data to Koski?

PR #3472, opened 2023-03-27
Ensure that latest guardians are retrieved from VTJ for decisions

reynders:


akheron:
I suggest this be closed. MUTPA should keep the data up to date.

reynders:
In MUTPa we trust

PR #3498, opened 2023-04-04
Reservations not required without service need

Joosakur:
If child has no service need (e.g. they are only in preschool) then there is no need to do reservations.

In these cases:

1. Do not send the missing reservations notification email to guardians
2. Do not show reservation missing warnings on the citizen calendar
3. Do not show missing holiday reservation warnings on the group's calendar


reynders (approved):

service/src/main/kotlin/fi/espoo/evaka/reservations/MissingReservationsReminders.kt
```diff
@@ -112,6 +112,7 @@ FROM (
     FROM generate_series(${bind(range.start)}, ${bind(range.end)}, '1 day') t
     JOIN placement p ON daterange(p.start_date, p.end_date, '[]') @> t::date
     JOIN daycare d ON p.unit_id = d.id
+    JOIN service_need sn ON p.id = sn.placement_id AND daterange(sn.start_date, sn.end_date, '[]') @> t::date
```
Joosakur, line 115: exist

PR #3501, opened 2023-04-05
Fix application club placement filter

reynders:
#### Summary
- Fix the (K) application filter to show applications for children that are or have been in club care


service/src/main/kotlin/fi/espoo/evaka/application/ApplicationQueries.kt
```diff
@@ -447,6 +446,11 @@ fun Database.Read.fetchApplicationSummaries(
             WHERE application_id IS NOT NULL
             GROUP by application_id
         ) attachments ON a.id = attachments.application_id
+        LEFT JOIN LATERAL (
+            SELECT a.id, EXISTS( SELECT 1 
+            FROM placement p 
+            WHERE p.child_id = a.child_id AND p.type = 'CLUB'::placement_type AND p.start_date <= :today) AS was_on_club_care
+        ) club_care ON a.id = club_care.id
```
Gekkio, line 453: This join condition is not needed since this is a lateral join:

- use `ON true` since we're doing a lateral join
- drop a.id from columns since it's not needed

Gekkio (approved):
I think we should monitor the performance impact after merging this, since it's a quite bad-performing query already and adding a new lateral join *might* have a big effect on performance...or not :sweat_smile:

PR #3502, opened 2023-04-05
Send only one notification per outdated income, not per child

reynders:
#### Summary
Only send one notification per guardian for outdated income instead of one notification per child

### DIFF
--- a/service/src/main/kotlin/fi/espoo/evaka/invoicing/service/IncomeQueries.kt
+++ b/service/src/main/kotlin/fi/espoo/evaka/invoicing/service/IncomeQueries.kt
@@ -46,7 +46,7 @@ fun Database.Read.expiringIncomes(
     val existingRecentIncomeNotificationQuery =
         """
     SELECT 1 FROM income_notification 
-    WHERE receiver_id = pl.guardian_id AND notification_type = :notificationType 
+    WHERE receiver_id = expiring_income.person_id AND notification_type = :notificationType 
         AND created > :today - INTERVAL '1 month'
     """
             .trimIndent()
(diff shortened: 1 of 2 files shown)


service/src/main/kotlin/fi/espoo/evaka/invoicing/service/IncomeQueries.kt
```diff
@@ -68,15 +68,15 @@ WITH latest_income AS (
     WHERE :checkForExpirationRange @> i.valid_to
      AND daterange(pl.start_date, pl.end_date, '[]') @> (i.valid_to + INTERVAL '1 day')::date
 )
-SELECT guardian_id, valid_to AS expiration_date
+SELECT person_id AS guardian_id, valid_to AS expiration_date
```
Joosakur, line 71: Possible alternative solution might have been to just change this to
```
SELECT DISTINCT ON(guardian_id)
guardian_id, valid_to AS expiration_date
...
ORDER BY guardian_id, valid_to
```

Joosakur:
> #### Summary
> When application is waiting for acceptance / rejection of a placement proposal, the related decisions cannot be accepted / rejected for the guardian by the employees

PR description has a copy-paste error I guess

service/src/main/kotlin/fi/espoo/evaka/invoicing/service/IncomeQueries.kt
reynders: If no other reason for this suggested change I'm willing to go with current as I like to explicitly do the distinct selection where it happens now

Joosakur (approved):

PR #3506, opened 2023-04-06
Move sending PRESCHOOL_CLUB placements to Koski instead of Varda

tmkrepo:
Removes Varda mapping, adds Koski mapping.

### DIFF
--- a/service/src/integrationTest/kotlin/fi/espoo/evaka/koski/KoskiPayloadIntegrationTest.kt
+++ b/service/src/integrationTest/kotlin/fi/espoo/evaka/koski/KoskiPayloadIntegrationTest.kt
@@ -846,4 +846,102 @@ class KoskiPayloadIntegrationTest : FullApplicationTest(resetDbBeforeEach = true
         """
         JSONAssert.assertEquals(expected, stored.payload, JSONCompareMode.STRICT)
     }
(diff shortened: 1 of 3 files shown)


service/src/integrationTest/kotlin/fi/espoo/evaka/koski/KoskiPayloadIntegrationTest.kt
```diff
@@ -846,4 +846,102 @@ class KoskiPayloadIntegrationTest : FullApplicationTest(resetDbBeforeEach = true
         """
         JSONAssert.assertEquals(expected, stored.payload, JSONCompareMode.STRICT)
     }
+    @Test
+    fun `simple preschool club placement in 2019`() {
```
Gekkio, line 850: I think this test should be implemented as a normal integration test in KoskiIntegrationTest.kt, instead of a "raw payload" test. These raw tests are verbose and brittle, and they only exist as a sanity check to guarantee we don't break the JSON format accidentally.

tmkrepo: Test moved and simplified.

Gekkio (approved):

PR #3508, opened 2023-04-06
do not send reservation reminder email if children have no service needs

Joosakur:


service/src/main/kotlin/fi/espoo/evaka/reservations/MissingReservationsReminders.kt
```diff
@@ -116,6 +116,11 @@ FROM (
     WHERE date_part('isodow', t::date) = ANY(d.operation_days)
     AND h.date IS NULL
     AND 'RESERVATIONS' = ANY(d.enabled_pilot_features)
+    AND EXISTS (
+        SELECT 1
+        FROM service_need sn
+        WHERE sn.placement_id = p.id AND daterange(sn.start_date, sn.end_date, '[]') @> t::date
```
akheron, line 122: This is the perfect `WHERE` condition for this query, because an index exists for it.

akheron (approved):

PR #3529, opened 2023-04-14
Attendance reservation report v2

Joosakur:
Use v2 report, misc fixes


Joosakur:
I'm unsure whether to keep the new queries in AttendanceReservationReportController.kt private or should they be moved somewhere and made reusable. On one hand they are quite generic, but the exact set of retrieved fields is currently optimized for this report.

reynders (approved):

PR #3531, opened 2023-04-14
Hide the original child of a duplication case from Koski integration

tmkrepo:
- filter to remove persons that have at least one duplicate from Koski study right selection
- integration test to check that a study right is sent for the duplicate but not the original


service/src/main/resources/db/migration/R__koski_views.sql
```diff
@@ -32,6 +32,8 @@ TABLE (
         AND placement.type IN ('PRESCHOOL', 'PRESCHOOL_DAYCARE', 'PRESCHOOL_CLUB', 'PREPARATORY', 'PREPARATORY_DAYCARE')
         WINDOW child AS (PARTITION BY child_id)
     ) p
+    JOIN person pr ON p.child_id = pr.id
```
Gekkio, line 35: Is this join necessary? We're joining based on `pr.id` and then later only using `pr.id` which is the same as `p.child_id` without any joins.

tmkrepo: Removed join

PR #3537, opened 2023-04-17
sent applications can be cancelled but not deleted

Joosakur:


service/src/integrationTest/kotlin/fi/espoo/evaka/application/ApplicationControllerCitizenIntegrationTest.kt
```diff
@@ -0,0 +1,141 @@
...
+import org.junit.jupiter.api.BeforeEach
+import org.junit.jupiter.api.assertThrows
+import org.springframework.beans.factory.annotation.Autowired
+
+class ApplicationControllerCitizenIntegrationTest : FullApplicationTest(resetDbBeforeEach = true) {
+    @Autowired private lateinit var applicationControllerCitizen: ApplicationControllerCitizen
+
+    @Autowired private lateinit var stateService: ApplicationStateService
+
+    private val clock = MockEvakaClock(HelsinkiDateTime.now())
```
Gekkio, line 38: Mock clock + now is not a very useful combo, because the test still depends on the time it's executed. If the precise timestamp is not important in the test, just pick one timestamp and hard-code it here.

service/src/main/kotlin/fi/espoo/evaka/application/ApplicationControllerCitizen.kt
```diff
@@ -347,7 +347,7 @@ class ApplicationControllerCitizen(
     }
 
     @DeleteMapping("/applications/{applicationId}")
-    fun deleteUnprocessedApplication(
+    fun cancelUnprocessedApplication(
```
Gekkio, line 350: The action name is still `Action.Citizen.Application.DELETE`. Changing it is technically a breaking change, but I don't think there are any customizations that break if we rename it to match the name of the actual operation.

service/src/integrationTest/kotlin/fi/espoo/evaka/application/ApplicationControllerCitizenIntegrationTest.kt
```diff
@@ -0,0 +1,141 @@
...
+
+    @BeforeEach
+    fun beforeEach() {
+        db.transaction { tx ->
+            tx.insertTestPerson(testAdult_1)
```
Gekkio, line 43: This is much better than using `insertGeneralTestFixtures` but this test still depends on various global fixture definitions. Does it *really* depend on the actual values, or could we just use minimal `DevFoo(...)` data here?

Gekkio (approved):
The actual change looks good! :+1: Just some minor nits in comments

service/src/integrationTest/kotlin/fi/espoo/evaka/application/ApplicationControllerCitizenIntegrationTest.kt
Joosakur: I think issue was inserting the application form which assumed to have those specific people. Long term fix could be to add a proper `insertTestApplication` function and deprecate/remove others. Currently we have many functions related to that, but none are really good. The one I ended up using even had a nameclash with the normal insert so I had to rename it in the import. I'll create a ticket to inbox.

PR #3539, opened 2023-04-18
Log digitransit API errors

Gekkio:
#### Summary

This forces us to buffer all responses in apigw memory, but it seems like an acceptable tradeoff

### DIFF
--- a/apigw/src/enduser/mapRoutes.ts
+++ b/apigw/src/enduser/mapRoutes.ts
@@ -10,6 +10,7 @@ import {
 } from '../shared/config'
 import expressHttpProxy from 'express-http-proxy'
 import { createProxy } from '../shared/proxy-utils'
+import { logError, logWarn } from '../shared/logging'
 
 const router = Router()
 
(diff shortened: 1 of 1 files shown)


apigw/src/enduser/mapRoutes.ts
```diff
@@ -28,6 +29,43 @@ function createDigitransitProxy(path: string) {
           : {})
       }
       return proxyReqOpts
+    },
+    userResDecorator: (proxyRes, proxyResData) => {
+      function parseBody(): unknown {
+        if (!Buffer.isBuffer(proxyResData)) {
+          return undefined
+        }
+        const body = proxyResData as Buffer
```
akheron, line 38: `Buffer.isBuffer()` is a type guard, so `as Buffer` should not be needed 🤔

Gekkio: IDEA shows a warning, but TS is happy :thinking:

Joosakur (approved):

PR #3540, opened 2023-04-18
improve vasu followup question formatting

Joosakur:
Screenshot 2023-04-18 at 12 09 21


Gekkio (approved):
:+1: 
I like the changed `followupEntry` function that returns a sensible object (vs for example making separate `followupEntryText` + `followupEntryDate` functions, which is a style we've sometimes unfortunately used)

PR #3544, opened 2023-04-18
only show active units on backup-care unit dropdown

Joosakur:

### DIFF
--- a/service/src/main/kotlin/fi/espoo/evaka/daycare/controllers/LocationController.kt
+++ b/service/src/main/kotlin/fi/espoo/evaka/daycare/controllers/LocationController.kt
@@ -69,10 +69,11 @@ class LocationController {
     fun getUnits(
         db: Database,
         @RequestParam type: UnitTypeFilter,
-        @RequestParam area: String?
+        @RequestParam(required = false) area: String?,
+        @RequestParam(required = false) date: LocalDate?
     ): List<UnitStub> {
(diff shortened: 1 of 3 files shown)


service/src/main/kotlin/fi/espoo/evaka/daycare/controllers/LocationController.kt
```diff
@@ -128,11 +130,13 @@ AND (:type = 'ALL' OR
     (:type = 'DAYCARE' AND '{CENTRE, FAMILY, GROUP_FAMILY}' && unit.type) OR
     (:type = 'PRESCHOOL' AND '{PRESCHOOL, PREPARATORY_EDUCATION}' && unit.type)
 )
+AND (:date::date IS NULL OR daterange(unit.opening_date, unit.closing_date, '[]') @> :date)
```
Gekkio, line 133: `IS NULL OR` for conditional predicates is nowadays discouraged because a generic query plan can't assume either case and will always choose the worst case scenario. But there's another in the same query anyway and in this case the performance impact is small, so it's ok

Gekkio (approved):

service/src/main/kotlin/fi/espoo/evaka/daycare/controllers/LocationController.kt
Joosakur: what's the current way? just conditionally adding the predicate via string templating?

PR #3547, opened 2023-04-19
EVAKA-4809: remove redis VTJ cache

Joosakur:


service/build.gradle.kts
```diff
@@ -92,7 +92,6 @@ dependencies {
     implementation("com.zaxxer:HikariCP")
     implementation("org.flywaydb:flyway-core")
     implementation("org.postgresql:postgresql")
-    implementation("redis.clients:jedis")
```
Gekkio, line 95: This should also be removed from evaka-bom

Gekkio (approved):

PR #3553, opened 2023-04-20
Keep holiday period reservations when creating/updating daily service times

akheron:
#### Summary

When daily service times are adjusted, all reservations are cleared from under the new daily service times.

After this change, reservations inside holiday periods are kept.

### DIFF
--- a/service/src/main/kotlin/fi/espoo/evaka/reservations/ReservationQueries.kt
+++ b/service/src/main/kotlin/fi/espoo/evaka/reservations/ReservationQueries.kt
@@ -95,9 +95,17 @@ fun Database.Transaction.clearOldReservations(
     return batch.executeAndReturn().mapTo<AttendanceReservationId>().toList()
 }
 
-fun Database.Transaction.clearReservationsForRange(childId: ChildId, range: DateRange): Int {
+fun Database.Transaction.clearReservationsForRangeExceptInHolidayPeriod(
+    childId: ChildId,
+    range: DateRange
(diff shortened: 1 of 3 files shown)


reynders (approved):

service/src/main/kotlin/fi/espoo/evaka/reservations/ReservationQueries.kt
```diff
@@ -95,9 +95,17 @@ fun Database.Transaction.clearOldReservations(
...
+    childId: ChildId,
+    range: DateRange
+): Int {
     return this.createUpdate(
-            "DELETE FROM attendance_reservation WHERE child_id = :childId AND between_start_and_end(:range, date)"
+            """
+            DELETE FROM attendance_reservation
+            WHERE child_id = :childId
+            AND between_start_and_end(:range, date)
+            AND NOT EXISTS (SELECT 1 FROM holiday_period hp WHERE period @> date)
```
Joosakur, line 107: nitpick: i'd like to see aliases, now period @> date is ambiguous unless you remember which of them is part of which table

Joosakur (approved):

service/src/main/kotlin/fi/espoo/evaka/reservations/ReservationQueries.kt
akheron: I had them first in this exact spot, but it looked so ugly 😄

PR #3565, opened 2023-04-25
tsconfig changes

akheron:
#### Summary

- Upgrade typescript and typescript-eslint
- Enable `isolatedModules`
- Enable `verbatimModuleSyntax` and update all type-only imports


akheron:
This breaks webpack in a weird way, and I'm not interested in trying to fix it. There are no real advantages to enabling these tsconfig flags.

PR #3581, opened 2023-05-03
Add more Espoo BI CSV routes/tables

Gekkio:

### DIFF
--- a/service/src/main/kotlin/fi/espoo/evaka/espoobi/EspooBiPoc.kt
+++ b/service/src/main/kotlin/fi/espoo/evaka/espoobi/EspooBiPoc.kt
@@ -4,17 +4,54 @@
 
 package fi.espoo.evaka.espoobi
 
+import fi.espoo.evaka.shared.auth.AuthenticatedUser
+import fi.espoo.evaka.shared.db.Database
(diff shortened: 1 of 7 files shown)


reynders (approved):

service/src/main/kotlin/fi/espoo/evaka/espoobi/EspooBiPoc.kt
```diff
@@ -4,17 +4,117 @@
...
+    val getGroupPlacements =
+        streamingCsvRoute<BiGroupPlacement> {
+            sql(
+                """
+SELECT id, updated, daycare_placement_id AS placement, daycare_group_id AS "group", start_date, end_date
+FROM daycare_group_placement
+
+UNION ALL
+
+SELECT id, updated, id AS placement, group_id AS "group", start_date, end_date
```
Joosakur, line 81: id AS placement?

Gekkio: It's a bit confusing, but a single row in our `backup_care` table is *both* a "placement" and "group placement" in the exported CSV model -> same id for both.

PR #3585, opened 2023-05-04
Update placement unit name visibility

patari:
#### Summary

Adds placement unit name to ~~unit supervisor and special education teacher~~ all unit roles for all placements (not just own units).


akheron (changes requested):
Does this really need 3 states, or could everyone with access to child's page see the name of the unit?

Gekkio (approved):

patari:
> > Does this really need 3 states
> 
> According to spec, yes it does.

Spec changed, unit name is now visible for all unit roles.

akheron (approved):

PR #3587, opened 2023-05-04
Fix finance decision generation for families with multisegment partnerships

vesahonkasaari:
When a family had partnerships that consisted of multiple segments in the fridge_partner table, the partner's children would be included multiple times in the family composition causing the finance decision generation to eventually fail. This would be case for example for parents that separated but later moved back together. Fix this to include each partner's children only once


akheron:
`findFamiliesByChild` potentially needs to be fixed, too.

reynders:
Duplicate of https://github.com/espoon-voltti/evaka/pull/3918

PR #3588, opened 2023-05-04
Daycare operation times

tmkrepo:
__NOTE!__ this PR effectively changes `daycare.operation_days` column to read-only and __will break external (i.e. not in this repo) scripts trying to insert operation days directly__ (notably non-core dev-data scripts)

Includes:
#### DB:
- new daycare column `operation_times` that uses existing custom type `timerange` to persist unit operation times to the second
- recreated daycare column `operation_days` as a generated column from information stored in `operation_times`
- compatibility migration for existing operation days to be depicted as corresponding operation times (default operation day is assumed to be 00:00 - 23:59 for compatibility reasons, these should be updated manually after installation according to city specific schedules)
#### Backend:
- backend support for column changes in the daycare controllers
- integration test changes
- JDBI configurations
#### Frontend
- UI for showing and editing operation times to the minute
  - uses existing component `TimeRangeInput` and adds some compatibility code to work better with unit details form and associated validation structures
- e2e test changes
  - new tests for operation times and their UI
  - lots of small changes to move test initialization from editing `operation_days` to editing `operation_times`
#### Other:
- core local test data changes to work with the new daycare structures


service/src/main/kotlin/fi/espoo/evaka/shared/db/JdbiExtensions.kt
```diff
@@ -161,7 +159,7 @@ val dateRangeColumnMapper = PgObjectColumnMapper {
 }
 
 val timeRangeColumnMapper = PgObjectColumnMapper {
-    assert(it.type == "timerange")
+    assert(it.type == "timerange" || it.type == "timerange_non_nullable_range")
```
tmkrepo, line 162: This is a weird workaround to an issue where integration tests read array columns as `timerage_non_nullable_range` but running local env live mapped to `timerange`. Any ideas how to actually fix this?

Gekkio: This looks ok to me and it's actually the original type that is wrong... `timerange` in the database allows nulls in both start and end, but this doesn't map correctly to our Kotlin `TimeRange` that doesn't allow nulls.

frontend/src/employee-frontend/components/unit/unit-details/UnitEditor.tsx
```diff
@@ -81,6 +87,7 @@ interface FormData {
   ophUnitOid: string
   ophOrganizerOid: string
   operationDays: DayOfWeek[]
+  operationTimes: EditableTimeRange[]
```
Gekkio, line 90: We now have two sources of truth in this form. Is there really a need to retain the old `operationDays` array if `operationTimes` is an array with guaranteed 7 nullable elements?

frontend/src/employee-frontend/components/unit/unit-details/UnitEditor.tsx
```diff
@@ -1013,6 +1103,64 @@ export default function UnitEditor(props: Props): JSX.Element {
           onChange={(roundTheClock) => updateForm({ roundTheClock })}
         />
       </FormPart>
+      <FormPart>
+        <div>{showRequired(i18n.unitEditor.label.operationDays)}</div>
+        <FixedSpaceColumn spacing="xs">
+          {([1, 2, 3, 4, 5, 6, 7] as const).map((day) => {
```
Gekkio, line 1109: Using just one source of truth would probably reduce this to `form.operationTimes.map((day, index) => ...`

Gekkio (commented):
Looks ok and I have just two concerns:

- This is yet another place infected with the use of 23:59 magic values. Humans usually have a mental model about time ranges where the end time is exclusive, so some place closing at midnight closes at 00:00, not 23:59, and 24h opening hours are 00:00-00:00, not 00:00-23:59 with one magical "lost minute". But we already have this hack in multiple places and have no systematic plan or way to improve things, so this is acceptable for now
- There's a lot of changes to the unit editor form, which move us further away from using our new form library when we should be refactoring old code to use it instead of reinventing yet another way of doing things. But changing the unit editor form to use the library is a large task and is probably better to do separately without complicating this pull request, so this is acceptable, even though technical debt is added

frontend/src/employee-frontend/components/unit/unit-details/UnitEditor.tsx
tmkrepo: Removed and refactored to dynamically produce days from times.

frontend/src/employee-frontend/components/unit/unit-details/UnitEditor.tsx
tmkrepo: Simplified to produce operation days from times.

Gekkio (approved):

PR #3603, opened 2023-05-10
Fix time validation

reynders:
#### Summary
Validate arrival time on frontend. Previously it was possible to send invalid time strings like 9:00 (should have been 09:00)

### DIFF
--- a/frontend/src/employee-frontend/components/unit/unit-reservations/attendanceEditState.spec.ts
+++ b/frontend/src/employee-frontend/components/unit/unit-reservations/attendanceEditState.spec.ts
@@ -209,14 +209,14 @@ describe('validateEditState', () => {
         id: null,
         type: 'PRESENT',
         groupId: 'group1',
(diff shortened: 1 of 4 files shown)


frontend/src/employee-mobile-frontend/child-attendance/actions/MarkPresent.tsx
```diff
@@ -68,7 +68,8 @@ export default React.memo(function MarkPresent() {
     return attendances.length > 0 ? attendances[0].departed : null
   }, [attendanceStatuses, childId])
 
-  const isValidTime = useCallback(() => {
+  const isValidTimeCheck = useCallback(() => {
+    if (!isValidTime(time)) return false
```
akheron, line 72: `isValidTime()` should be moved to `LocalTime.tryParse`, otherwise this same bug may appear somewhere else, too.

akheron (approved):

PR #3612, opened 2023-05-16
Fix AbsenceModal validation

akheron:
#### Summary

Use the new form framework instead of handcrafted artisanal validation.


akheron:
This will be a part of a bigger PR

PR #3643, opened 2023-05-23
Child documents: more basic functionality

Joosakur:
#### Summary

- deleting child document drafts
- add and display `published_at` time
- admin can unpublish child documents
- bug fix: should not try to save content of published document (caused http 400 in background)
- fix i18n

### DIFF
--- a/service/src/main/kotlin/fi/espoo/evaka/document/childdocument/ChildDocumentController.kt
+++ b/service/src/main/kotlin/fi/espoo/evaka/document/childdocument/ChildDocumentController.kt
@@ -78,21 +78,37 @@ class ChildDocumentController(private val accessControl: AccessControl) {
         user: AuthenticatedUser,
         clock: EvakaClock,
         @PathVariable documentId: ChildDocumentId
-    ): ChildDocumentDetails {
+    ): ChildDocumentWithPermittedActions {
         return db.connect { dbc ->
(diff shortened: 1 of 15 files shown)


service/src/main/kotlin/fi/espoo/evaka/document/childdocument/ChildDocumentController.kt
```diff
@@ -78,21 +78,37 @@ class ChildDocumentController(private val accessControl: AccessControl) {
...
+                    tx,
+                    user,
+                    clock,
+                    Action.ChildDocument.READ,
+                    documentId
+                )
+
+                val document =
+                    tx.getChildDocument(documentId)?.also {
+                        Audit.ChildDocumentRead.log(targetId = documentId)
```
akheron, line 94: Audit logging should happen outside of the transaction

akheron (changes requested):

reynders (approved):

service/src/main/kotlin/fi/espoo/evaka/document/childdocument/ChildDocumentController.kt
Joosakur: fixed 👍

akheron (approved):

PR #3683, opened 2023-05-29
Upgrade redis and connect-redis, replace redis-mock

akheron:


apigw/src/shared/saml/passport-saml-cache-redis.ts
```diff
@@ -33,41 +32,35 @@ export default function redisCacheProvider(
...
-      ),
-    removeAsync: (key) =>
-      fromCallback((callback) =>
-        client.del(`${keyPrefix}${key}`, (err, count) => {
-          if (err) return callback(err, null)
-          if (count === 0) return callback(null, null)
-          return callback(null, key)
-        })
-      )
+      } catch (err: unknown) {
```
Gekkio, line 55: Why is the error caught and discarded here? :thinking:

akheron: Seems I misunderstood the callback in the original code. Fixed.

Gekkio (approved):

PR #3696, opened 2023-06-02
Fee generator v2

Joosakur:
#### Summary

This PR introduces a rewritten FeeDecisionGenerator v2.

There's two new environment variables to control which code is ran:

- `evaka.fee_decision.generator.v1_enabled` (default `true`)
- `evaka.fee_decision.generator.v2_enabled` (default `false`)

If BOTH are set to true, then both generators are ran, but the v2 generator will use the new duplicated `fee_decision_v2` and `fee_decision_child_v2 tables`.

Tests currently have both set to true in `application-integration-test.yaml`, so both codes are ran but the test data and assertions only apply to v1. To test v2 set `v1_enabled` to `false`.

### Other changes

- Added test cases
- Disabled automatic async job running in e2e tests and added manual requests which correctly pass the mocked running time
- use mocked time in fee decision integration tests (overrides similar change which was merged to master while this has been open


service/src/main/kotlin/fi/espoo/evaka/invoicing/service/FinanceDecisionGenerator.kt
```diff
@@ -496,3 +544,26 @@ internal fun addECHAFeeAlterations(
...
+    tx: Database.Read,
+    adultId: PersonId
+): Set<PersonId> {
+    val partners = tx.getPartnersForPerson(adultId, false).map { it.person.id }
+    val children = tx.getParentships(headOfChildId = adultId, childId = null).map { it.childId }
+    val otherParents =
+        children.flatMap { child ->
+            tx.getParentships(headOfChildId = null, childId = child).map { it.headOfChildId }
+        }
+    return (partners + otherParents + adultId).toSet()
```
Joosakur, line 558: check fee decisions?

service/src/main/kotlin/fi/espoo/evaka/invoicing/service/FinanceDecisionGenerator.kt
```diff
@@ -565,5 +574,50 @@ internal fun getAllPossiblyAffectedAdultsByChild(
...
+        .set()
+
+private fun Database.Read.getChildrenFromFeeDecisions(personId: PersonId) =
+    createQuery<Any> {
+            sql(
+                """
+        SELECT fdc.child_id 
+        FROM fee_decision fd
+        JOIN fee_decision_child fdc ON fd.id = fdc.fee_decision_id
+        WHERE fd.head_of_family_id = ${bind(personId)} OR fd.partner_id = ${bind(personId)} AND fd.status <> 'DRAFT'
```
akheron: This `AND/OR` requires parentheses

akheron (changes requested):

service/src/main/kotlin/fi/espoo/evaka/invoicing/service/FinanceDecisionGenerator.kt
Joosakur: yes, we noticed same 😓 already fixed!

akheron (approved):

PR #3715, opened 2023-06-06
Intermittent shift care

tmkrepo:
Refactor shift care from a boolean to consist of 3 separate states
- `NONE` matches previous value `false`
- `INTERMITTENT` is the new value describing occasional separately agreed shift care
- `FULL` matches previous value `true`

### DB
- shift care data model change and migration
### Backend
- enum api type
- compatibility changes
  - Varda-integration will consider only `FULL` to be reportable shift care
  - `INTERMITTENT` cannot be automatically assigned through the application process
  - reports that account shift care service needs consider only `FULL` to be shift care
  - integration test and test fixture changes
### Frontend
- enum selection UI in child service need creation and editing
  - new UI behind a feature flag, functionality as before by default  
- e2e test for feature flagged UI


service/src/main/resources/db/migration/V329__service_need_shift_care_enum.sql
```diff
@@ -0,0 +1,22 @@
+CREATE TYPE shift_care_type AS ENUM (
+    'FULL',
+    'INTERMITTENT',
+    'NONE'
+);
+
+ALTER TABLE service_need
+    ADD COLUMN shift_care_temp shift_care_type NOT NULL DEFAULT 'NONE'::shift_care_type;
+
+ALTER TABLE service_need DISABLE TRIGGER set_timestamp;
```
tmkrepo, line 10: Trigger is disabled to prevent launching unnecessary Varda-updates for all children in shift care. Is there any risk for launching other automated operations based on a detected service need change?

Gekkio (approved):

PR #3719, opened 2023-06-07
move child documents into same section with vasu

Joosakur:

### DIFF
--- a/frontend/src/employee-frontend/components/child-information/ChildDocumentsSection.tsx
+++ b/frontend/src/employee-frontend/components/child-information/ChildDocumentsSection.tsx
@@ -38,7 +38,7 @@ export default React.memo(function ChildDocumentsSection({
             placement.daycare.enabledPilotFeatures.includes('VASU_AND_PEDADOC')
           )
         )
-        .getOrElse(true),
+        .getOrElse(false),
     [permittedActions, placements]
   )
 


frontend/src/employee-frontend/components/ChildInformation.tsx
```diff
@@ -146,7 +147,7 @@ const layouts: Layouts<typeof components> = {
     { component: 'placements', open: false },
     { component: 'backup-care', open: false },
     { component: 'daily-service-times', open: false },
-    { component: 'vasuAndLeops' as keyof typeof components, open: false },
+    { component: 'childDocuments' as keyof typeof components, open: false },
```
akheron, line 150: Is `as keyof typeof components` needed at all? Other entries don't have it.

In any case, `as const` would be the most type safe option. `as` lets you state false things as facts and is a source of bugs!

frontend/src/employee-frontend/components/child-information/ChildDocumentsSection.tsx
```diff
@@ -0,0 +1,73 @@
...
+  const hasVasuPermission = useMemo(
+    () =>
+      permittedActions.has('READ_VASU_DOCUMENT') &&
+      placements
+        .map((ps) =>
+          ps.placements.some((placement) =>
+            placement.daycare.enabledPilotFeatures.includes('VASU_AND_PEDADOC')
+          )
+        )
+        .getOrElse(true),
```
akheron, line 41: `true` doesn't sound right here 🤔

PR #3724, opened 2023-06-08
Prioritize backup care for daily operation time resolution

tmkrepo:
Fixes issue where regular placement unit times were used to validate citizen reservations even during a backup care period.

### DIFF
--- a/service/src/main/kotlin/fi/espoo/evaka/reservations/ReservationControllerCitizen.kt
+++ b/service/src/main/kotlin/fi/espoo/evaka/reservations/ReservationControllerCitizen.kt
@@ -285,6 +285,8 @@ private fun placementDay(
     val operationDays =
         backupPlacements?.find { it.range.includes(date) }?.operationDays ?: placement.operationDays
 
+    // Backup placements take precedence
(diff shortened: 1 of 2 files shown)


service/src/main/kotlin/fi/espoo/evaka/reservations/ReservationControllerCitizen.kt
```diff
@@ -285,6 +285,8 @@ private fun placementDay(
     val operationDays =
         backupPlacements?.find { it.range.includes(date) }?.operationDays ?: placement.operationDays
 
+    // Backup placements take precedence
+    val operationTime = (backupPlacements?.find { it.range.includes(date) }?.operationTimes ?: placement.operationTimes)[date.dayOfWeek.value - 1]
```
akheron: `backupPlacements?.find { it.range.includes(date) }` is used twice and should be extracted as a variable

akheron (changes requested):

service/src/main/kotlin/fi/espoo/evaka/reservations/ReservationControllerCitizen.kt
tmkrepo: Restructured accordingly.

akheron (approved):

PR #3759, opened 2023-06-19
New assistance model (MVP)

Gekkio:
#### Summary

- old model is still the default
- backend env flag for switching to the new model
- if new model is enabled, it's only used in the basic CRUD, not yet in Koski/Varda/reports/etc...
- overlapping rows are not forbidden and validation / error handling of these is not yet done in the UI correctly


service/src/main/kotlin/fi/espoo/evaka/assistance/AssistanceController.kt
```diff
@@ -0,0 +1,801 @@
...
+            dbc.read {
+                accessControl.requirePermissionFor(
+                    it,
+                    user,
+                    clock,
+                    Action.Global.READ_ASSISTANCE_ACTION_OPTIONS
+                )
+            }
+            assistanceActionService.getAssistanceActionOptions(dbc)
+        }
```
reynders, line 499: log missing

reynders (changes requested):

reynders (approved):

PR #3770, opened 2023-06-22
New assistance model - part 2

Gekkio:
#### Summary

- add support for new assistance model in almost all places
- assistance needs and actions report still uses the old model, because it needs a different approach with the new model
- integration / E2E tests and local dev use new model, but in production environments the old model is still the default


Gekkio:
Reviewed in huddle

PR #3808, opened 2023-07-04
Evaka-4448: new espoo logos

huhtari:
Use latest logos in keycloak too. Note: switches Espoo logo to SVG format.


Gekkio (approved):
Logo sizes change too but these can be changed later if necessary

PR #3836, opened 2023-07-12
Feature child discussions

martinLeht:
#### Summary
Implemented new child discussion section under child documents collapsible (in child information), which can be enabled through feature flag childDiscussion. Includes frontend changes for the new fields and respectively new endpoints in the backend to create, update and fetch discussion data for child. Also added db migration script to create the new table.

Testing this feature requires the following:
- childDiscussion feature flag enabled
- Child needs a placement and a group placement
- User has to be admin, staff, unit supervisor or special education teacher
- Action rules needs to be in place: CREATE_CHILD_DISCUSSION, READ_CHILD_DISCUSSION, UPDATE_CHILD_DISCUSSION


frontend/src/employee-frontend/components/child-information/ChildDiscussions.tsx
```diff
@@ -0,0 +1,224 @@
...
+        <Title size={4}>{i18n.childInformation.childDiscussion.title}</Title>
+        {!editing && (
+          <RequireRole
+            oneOf={[
+              'UNIT_SUPERVISOR',
+              'ADMIN',
+              'STAFF',
+              'SPECIAL_EDUCATION_TEACHER'
+            ]}
+          >
```
akheron, line 141: Hard-coded role lists should not be used in new frontend code, and you should use the `permittedAction` mechanism instead. The `/children/{childId}` route returns actions that are permitted on the given child, and you should check here whether it contains the `CREATE_CHILD_DISCUSSION` action.

frontend/src/employee-frontend/components/child-information/ChildDiscussions.tsx
```diff
@@ -0,0 +1,224 @@
...
+
+  const childDiscussionData = useQueryResult(childDiscussionQuery(childId))
+
+  useEffect(() => {
+    if (!childDiscussionData.isSuccess) return
+
+    setOfferedDate(childDiscussionData.value.offeredDate)
+    setHeldDate(childDiscussionData.value.heldDate)
+    setCounselingDate(childDiscussionData.value.counselingDate)
+  }, [childDiscussionData])
```
akheron, line 73: Don't use `useEffect` for initializing component state. Instead, create an outer component that does the query and uses `renderResult()` to render an inner component that is passed the initial data as props and renders the view/edit stuff. It will probably also simplify the view/edit mode handling.

See https://react.dev/learn/you-might-not-need-an-effect for more info about avoiding `useEffect`.

akheron: You might also benefit from using eVaka's "form framework" which has some learning curve but greatly reduces the need for manual state management and validation. See [here](https://github.com/espoon-voltti/evaka/blob/master/frontend/src/citizen-frontend/personal-details/PersonalDetails.tsx#L312-L328) for an example.

frontend/src/e2e-test/specs/5_employee/child-discussion.spec.ts
```diff
@@ -0,0 +1,162 @@
...
+      startDate: daycarePlacementFixture.startDate,
+      endDate: daycarePlacementFixture.endDate
+    })
+    .save()
+})
+
+describe('Child Information - Child discussion section', () => {
+  let section: ChildDocumentsSection
+  beforeEach(async () => {
+    page = await Page.open({
```
akheron, line 69: Because these tests do stuff with dates, you should consider using `mockedTime` to set the "current time" for the test. This is not strictly necessary in this case because there's no computations or limitations regarding the dates, but it's a good practice nevertheless to make sure that these tests don't break some day in the future when the date is different than it is today.

akheron (changes requested):

frontend/src/e2e-test/specs/5_employee/child-discussion.spec.ts
martinLeht: Good point, I will make the change to leverage `mockedTime` instead.

frontend/src/employee-frontend/components/child-information/ChildDiscussions.tsx
martinLeht: Interesting to use an outer component, sounds like a good approach.

I'll add the outer component for initialization and check out the form framework.

frontend/src/employee-frontend/components/child-information/ChildDiscussions.tsx
martinLeht: I see, well that sounds like a better way. I saw this hard-coded way somewhere and just used it here also. I'll add the changes to use permittedAction to check the required actions.

frontend/src/employee-frontend/components/child-information/ChildDiscussions.tsx
```diff
@@ -0,0 +1,338 @@
...
+  reloadDiscussionData
+}: {
+  discussions: ChildDiscussionWithPermittedActions[]
+  reloadDiscussionData: () => void
+}) {
+  return (
+    <>
+      {discussions.map((discussion, index) => (
+        <Fragment key={discussion.data.id}>
+          <ChildDiscussionRow
```
akheron, line 180: This doesn't look optimal. Maybe it should be a table like the other sections 🤔 

Screenshot 2023-07-26 at 8 02 14

akheron (changes requested):
Few minor nitpicks, otherwise looks good!

service/src/main/kotlin/fi/espoo/evaka/childdiscussion/ChildDiscussionController.kt
```diff
@@ -0,0 +1,140 @@
...
+        db.connect { dbc ->
+                dbc.transaction { tx ->
+                    accessControl.requirePermissionFor(
+                        tx,
+                        user,
+                        clock,
+                        Action.ChildDiscussion.UPDATE,
+                        discussionId
+                    )
+                    tx.getChildDiscussionById(discussionId)
```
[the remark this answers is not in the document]
martinLeht: Perfect, I did not notice that - made the fix.

frontend/src/e2e-test/utils/index.ts
```diff
@@ -75,6 +75,16 @@ export async function waitUntilDefined<T>(
   )
 }
 
+export async function waitUntilUndefined<T>(f: () => Promise<T>): Promise<T> {
```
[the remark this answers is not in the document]
martinLeht: Well spotted! I forgot to remove this after I realized that I did not need it - now its removed.

frontend/src/employee-frontend/components/child-information/ChildDiscussions.tsx
martinLeht: Good point, now I refactored it to be a table.

akheron (approved):

PR #3850, opened 2023-07-17
Finance view preschool club filter for fee decisions

tmkrepo:
Adds a feature flagged finance view filter option for fee decisions with a child that has placement type `PRESCHOOL_CLUB`.

### DIFF
--- a/frontend/src/lib-customizations/types.d.ts
+++ b/frontend/src/lib-customizations/types.d.ts
@@ -40,15 +40,15 @@ declare global {
   }
 
   // eslint-disable-next-line @typescript-eslint/no-empty-interface
-  interface EvakaWindowConfig { }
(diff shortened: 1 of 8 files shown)


frontend/src/lib-customizations/types.d.ts
```diff
@@ -222,6 +222,11 @@ interface BaseFeatureFlags {
      * Enable support for intermittent shift care
      */
     intermittentShiftCare?: boolean
+
+    /**
+     * Enable support for filtering fee decisions by preschool club placement type
+     */
+    feeDecisionPreschoolClubFilter?: boolean
```
akheron, line 229: Is `experimental` the correct place for this flag? 🤔

akheron (approved):
Otherwise looks good 👍

PR #3863, opened 2023-07-21
Family daycare meal report

tmkrepo:
Adds a new report that counts the attendances of children in family daycare at internally specified meal times and groups them by unit and care area.

Includes:
- report frontend with csv report download
- report backend
- integration tests to verify key features


service/src/main/kotlin/fi/espoo/evaka/reports/FamilyDaycareMealReport.kt
```diff
@@ -0,0 +1,253 @@
...
+
+    data class MealReportConfig(
+        val breakfastTime: MealTimeRange,
+        val lunchTime: MealTimeRange,
+        val snackTime: MealTimeRange
+    )
+    data class MealTimeRange(
+        @ForceCodeGenType(String::class) val startTime: LocalTime,
+        @ForceCodeGenType(String::class) val endTime: LocalTime
+    )
```
akheron, line 252: There's also a reusable `TimeRange` type in `shared/domain/Time.kt`. Furthermore, the `@ForceCodeGenType` annotations are not necessary because this type doesn't end up in generated files, and `LocalTime` works with codegen anyway.

akheron (approved):

PR #3875, opened 2023-07-24
Configure dependabot package groups

akheron:
Packages in the same group will be updated in a single PR.

Dependabot announcement: https://github.blog/changelog/2023-06-30-grouped-version-updates-for-dependabot-public-beta/


patari (approved):
Good stuff. I think react+react-dom (and maybe their types) should also be a group.

PR #3907, opened 2023-08-01
Citizen email notification settings

akheron:
#### Summary

The UI is still behind a feature flag.

Screenshot 2023-08-01 at 11 58 49


Joosakur (approved):
👏  Just wondering if the email type should be enum instead of text in the db too?

reynders (approved):

PR #3908, opened 2023-08-01
Income effects NOT_AVAILABLE and INCOMPLETE are equal

reynders:
If there is a SENT decision and DRAFT diffs only by income effect NOT_AVAILABLE vs INCOMPLETE, consider it equal and thus do not generate new decision

BLOCKED: Current generator modifies SENT decisions. If the decision change detection logic is changed like in this PR, this causes conflicts with existing SENT data

Waiting for the new generator which does not modify SENT decisions.

### DIFF
--- a/service/src/integrationTest/kotlin/fi/espoo/evaka/invoicing/service/FeeDecisionGeneratorIntegrationTest.kt
+++ b/service/src/integrationTest/kotlin/fi/espoo/evaka/invoicing/service/FeeDecisionGeneratorIntegrationTest.kt
@@ -40,6 +40,7 @@ import fi.espoo.evaka.shared.DaycareId
 import fi.espoo.evaka.shared.EvakaUserId
 import fi.espoo.evaka.shared.FeeAlterationId
 import fi.espoo.evaka.shared.FeeDecisionId
+import fi.espoo.evaka.shared.IncomeId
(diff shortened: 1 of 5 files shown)


service/src/integrationTest/kotlin/fi/espoo/evaka/invoicing/service/FeeDecisionGeneratorIntegrationTest.kt
```diff
@@ -2507,6 +2508,52 @@ class FeeDecisionGeneratorIntegrationTest : FullApplicationTest(resetDbBeforeEac
...
+                RealEvakaClock(),
+                testAdult_1.id,
+                incomePeriod.start
+            )
+        }
+
+        // No new DRAFT is generated because the only diff was head of family income type change
+        // NOT_AVAILABLE -> INCOMPLETE
+        assertEquals(1, getAllFeeDecisions().size)
+    }
```
akheron, line 2556: I would like to see more tests. The case where a child has incomes is very rare, so this is not a very realistic test.

More interesting tests would be:
* the income of a single parent changes
* the income of one adult changes in a two-adult family

reynders:
Implemented with the new fee decision generator

PR #3912, opened 2023-08-02
Only placement unit change does not generate new fee decision

reynders:
BLOCKED: Current generator modifies SENT decisions. If the decision change detection logic is changed like in this PR, this causes conflicts with existing SENT data (see the failing test).

- Waiting for the new generator which does not modify SENT decisions.
- TODO: voucher value unit change SHOULD generate a new decision, so undo those changes


reynders:
Implemented with the new fee decision generator

PR #3915, opened 2023-08-03
add remaining fields to assistance need preschool decision pdf

Joosakur:
#### Release notes
Lisäyksiä esiopetuksen tuentarpeen päätös-pdf:ään


akheron (approved):
I guess sv translations are missing because we don't have them?

Joosakur:
> I guess sv translations are missing because we don't have them?

aye, and everything else is also just a best guess for now, there's no example decision anywhere to use as a spec

PR #3950, opened 2023-08-10
duplicate fee decision tables for v2 generator testing

Joosakur:


service/src/main/resources/db/migration/V337__duplicate_fee_decision_tables.sql
```diff
@@ -0,0 +1,85 @@
+CREATE TABLE fee_decision_v2 AS SELECT * FROM fee_decision;
```
Joosakur, line 1: is it fine to copy the data too or should this have something like `WHERE false`?

PR #3964, opened 2023-08-14
New role for outsourced finance admins and permission based filtering in finance page tabs

martinLeht:
#### Summary
Implemented new role for outsourced finance admins/manager and permission based filtering in finance page tabs.


service/src/main/kotlin/fi/espoo/evaka/shared/security/actionrule/HasGlobalRole.kt
```diff
@@ -162,4 +165,26 @@ WHERE pd.provider_type = 'PRIVATE_SERVICE_VOUCHER'
...
+            sql(
+                """
+SELECT id
+FROM daycare
+WHERE provider_type = ANY(${bind(providerTypes)})
+${
+    if (careTypes.isNotEmpty()) "AND (${careTypes.mapIndexedNotNull{ i, value -> 
+            if (i == 0) "${"'$value'"} = ANY(daycare.type)"
+            else " OR ${"'$value'"} = ANY(daycare.type)"
+        }.joinToString(" ")})"
```
Joosakur, line 183: I think this could be much simpler. Something like
`careTypes.map { "$it" = ANY(daycare.type)" }.joinToString(" OR ")`

Joosakur: actually concatenating OR statements isnt even necessary. Isn't this what we want:
```
anyarray && anyarray → boolean

Do the arrays overlap, that is, have any elements in common?
```
So simply `${bind(careTypes)} && daycare.type`

frontend/src/lib-customizations/defaults/employee/i18n/fi.tsx
```diff
@@ -4188,6 +4188,7 @@ export const fi = {
       MESSAGING: 'Viestintä',
       REPORT_VIEWER: 'Raportointi',
       FINANCE_ADMIN: 'Talous',
+      FINANCE_STAFF: 'Talous työntekijä',
```
Joosakur, line 4191: varmaan yhdyssana?

frontend/src/employee-frontend/components/FinancePage.tsx
```diff
@@ -18,90 +20,163 @@ import InvoicesPage from './invoices/InvoicesPage'
...
-          link: '/finance/value-decisions',
-          label: i18n.header.valueDecisions
-        },
-        {
-          id: 'invoices',
-          link: '/finance/invoices',
-          label: i18n.header.invoices
-        },
-        featureFlags.experimental?.voucherUnitPayments
+        user?.permittedGlobalActions.includes('SEARCH_FEE_DECISIONS')
```
Joosakur, line 36: nullability chaining looks weird here. if user really can be null then `user?.permittedGlobalActions` can also be null so there should be more question marks I think

frontend/src/employee-frontend/components/FinancePage.tsx
```diff
@@ -18,90 +20,163 @@ import InvoicesPage from './invoices/InvoicesPage'
...
+      'invoices',
+      requireOneOfPermittedActionsForPage(
+        {
+          path: 'invoices',
+          title: i18n.titles.invoices,
+          Page: InvoicesPage
+        },
+        'SEARCH_INVOICES'
+      )
+    ],
```
Joosakur, line 146: duplicate

frontend/src/employee-frontend/components/FinancePage.tsx
```diff
@@ -18,90 +20,163 @@ import InvoicesPage from './invoices/InvoicesPage'
...
-              </EmployeeRoute>
-            }
-          />
+      )
+    } else {
+      return null
+    }
+  }
+
+  const permittedRoutes = new Map([
```
Joosakur, line 102: This thing feels overly complicated. 

- I'm not sure if conditionally including routes is even necessary, since even if the user manages to get there with direct URL, it wont show them anything they don't have access to.
- It looks like the `tabs` and `permitted` routes need to be in sync (e.g. doesn't make sense to render a tab if its associated route isn't permitted) so I think these two lists should be merged so you also don't need to duplicate the role checking

Joosakur (changes requested):

frontend/src/employee-frontend/components/FinancePage.tsx
martinLeht: I agree with the merging. I was not sure if it would be considered a problem or not, if the user may access the route with direct URL, so i decided to filter them also. I shall make the fix so that only the tabs are filtered according to permissions.

frontend/src/employee-frontend/components/FinancePage.tsx
martinLeht: Good point, I'll add the null check.

Joosakur (approved):

PR #4027, opened 2023-08-25
Future preschoolers report

msavolainen-gofore:
#### Future preschoolers report for automatic placement tool

Report to download 3 different csv files; next years preschool aged children, all daycare groups from municipal units with preschool and all daycare groups from private units with preschool. All needed data cannot be obtained from current evaka data model so some enrichment is needed from end user.


service/src/main/kotlin/fi/espoo/evaka/shared/Id.kt
```diff
@@ -45,6 +45,7 @@ sealed interface DatabaseTable {
     sealed class DailyServicesTimeNotification : DatabaseTable
     sealed class DaycareAssistance : DatabaseTable
     sealed class DaycareCaretaker : DatabaseTable
+    sealed class DaycareGroup : DatabaseTable
```
akheron, line 48: This already exists as `Group`.

service/src/main/kotlin/fi/espoo/evaka/shared/Id.kt
```diff
@@ -163,6 +164,8 @@ typealias DaycareAssistanceId = Id<DatabaseTable.DaycareAssistance>
 
 typealias DaycareCaretakerId = Id<DatabaseTable.DaycareCaretaker>
 
+typealias DaycareGroupId = Id<DatabaseTable.DaycareGroup>
+
```
akheron, line 168: This already exists as `GroupId`

service/src/main/kotlin/fi/espoo/evaka/reports/FuturePreschoolersReport.kt
```diff
@@ -0,0 +1,196 @@
...
+    CASE WHEN sn.shift_care = 'FULL'::shift_care_type OR sn.shift_care = 'INTERMITTENT'::shift_care_type THEN true ELSE false END AS shiftCare,
+    NULL AS languageEmphasisGroup, -- cannot be obtained from current evaka dataset
+    CASE WHEN sno.name_fi like 'Kaksivuotinen%' THEN true ELSE false END AS twoYearPreschool
+FROM person p
+JOIN placement pl ON pl.child_id = p.id
+JOIN daycare d ON d.id = pl.unit_id
+LEFT JOIN (
+    SELECT id, last_name, first_name, street_address, postal_code, post_office, phone, email
+    FROM person
+) g1 ON g1.id = (SELECT head_of_child FROM fridge_child WHERE child_id = p.id)
```
akheron, line 107: Head of child (päämies) is not the same as guardian. If you want guardian (huoltaja), use the `guardian` table. Also consider whether you may want foster parents (sijaisvanhempi) in the `foster_parent` table.

If head of child is what you want, note that it's timeseries data, so you should take the one where the date of interest falls in the `start_date` - `end_date` range.

service/src/main/kotlin/fi/espoo/evaka/reports/FuturePreschoolersReport.kt
```diff
@@ -0,0 +1,196 @@
...
+LEFT JOIN (
+    SELECT id, last_name, first_name, street_address, postal_code, post_office, phone, email
+    FROM person
+) g2 ON g2.id = (SELECT person_id FROM fridge_partner WHERE partnership_id =
+    (SELECT partnership_id FROM fridge_partner WHERE person_id = g1.id) AND person_id != g1.id)
```
akheron, line 112: Fridge partner (jääkaappipuoliso) is also not the same as guardian.

service/src/main/kotlin/fi/espoo/evaka/reports/FuturePreschoolersReport.kt
```diff
@@ -0,0 +1,196 @@
...
+LEFT JOIN service_need sn ON sn.placement_id = p.id and sn.start_date < current_date AND sn.end_date >= current_date
```
akheron, line 113: You should not use `current_date`, but pass the date as a bind parameter. This makes testing easier and allows using mock clock for example.

service/src/main/kotlin/fi/espoo/evaka/reports/FuturePreschoolersReport.kt
```diff
@@ -0,0 +1,196 @@
...
+    createQuery<DatabaseTable> {
+            sql(
+                """
+SELECT dg.id, 
+    d.name AS unitName, 
+    dg.name AS groupName, 
+    d.street_address AS address, 
+    d.postal_code AS postalCode, 
+    d.post_office as postOffice, 
+    (SELECT count(employee_id) FROM daycare_group_acl WHERE daycare_group_id = dg.id) * 7 AS groupSize,
```
akheron, line 137: There's also `daycare_caretaker` table which you might want to use instead of `daycare_group_acl`.

service/src/main/kotlin/fi/espoo/evaka/reports/FuturePreschoolersReport.kt
```diff
@@ -0,0 +1,196 @@
...
+    g2.last_name AS guardian2LastName, 
+    g2.first_name AS guardian2FirstName, 
+    g2.street_address AS guardian2Address,
+    g2.postal_code AS guardian2PostalCode,
+    upper(g2.post_office) AS guardian2PostOffice, 
+    g2.phone AS guardian2Phone,
+    g2.email AS guardian2Email,
+    CASE WHEN sn.shift_care = 'FULL'::shift_care_type OR sn.shift_care = 'INTERMITTENT'::shift_care_type THEN true ELSE false END AS shiftCare,
+    NULL AS languageEmphasisGroup, -- cannot be obtained from current evaka dataset
+    CASE WHEN sno.name_fi like 'Kaksivuotinen%' THEN true ELSE false END AS twoYearPreschool
```
akheron, line 100: Style note: We usually use `snake_case` names for SQL output columns. JDBI maps them correctly to `snakeCase` properties in the mapped data class.

frontend/src/employee-frontend/components/reports/FuturePreschoolersReport.tsx
```diff
@@ -0,0 +1,202 @@
...
+
+  const municipalGroupRows: PreschoolGroupsReportRow[] = useMemo(
+    () => municipalRows.getOrElse([]),
+    [municipalRows]
+  )
+
+  const privateVoucherGroupRows: PreschoolGroupsReportRow[] = useMemo(
+    () => voucherRows.getOrElse([]),
+    [voucherRows]
+  )
```
akheron, line 71: You should use either `useQueryResult` (newest hot) or at least `useApiState` (a bit older convention) to make the backend requests. You should also use `renderResult()` instead of `.isLoading` etc. to unwrap the `Result` and show a spinner while data is loading.

Unfortunately, most other reports are not good examples in this regard because they're such old code, but you should be able to find plenty of examples in other parts of the code and among more "modern" reports, too.

akheron (changes requested):

service/src/main/kotlin/fi/espoo/evaka/reports/FuturePreschoolersReport.kt
```diff
@@ -0,0 +1,216 @@
...
+JOIN daycare d ON d.id = pl.unit_id
+LEFT JOIN (
+    SELECT id, last_name, first_name, street_address, postal_code, post_office, phone, email
+    FROM person
+) g1 ON g1.id = (
+    SELECT guardian_id AS id FROM guardian WHERE child_id = p.id
+    UNION
+    SELECT parent_id AS id FROM foster_parent WHERE child_id = p.id AND valid_during @> :today
+    EXCEPT 
+    SELECT guardian_id AS id FROM guardian_blocklist where child_id = p.id
```
akheron, line 113: `guardian_blocklist` is only used to prevent automatically adding VTJ guardians to the `guardian` table, so this `EXCEPT` is not needed.

service/src/main/kotlin/fi/espoo/evaka/reports/FuturePreschoolersReport.kt
```diff
@@ -0,0 +1,216 @@
...
+    )
+LEFT JOIN (
+    SELECT id, last_name, first_name, street_address, postal_code, post_office, phone, email
+    FROM person
+) g2 ON g2.id = (
+    SELECT guardian_id AS id FROM guardian WHERE child_id = p.id
+    UNION
+    SELECT parent_id AS id FROM foster_parent WHERE child_id = p.id AND valid_during @> :today
+    EXCEPT 
+    SELECT guardian_id AS id FROM guardian_blocklist where child_id = p.id
```
akheron, line 124: Same here.

akheron (approved):
One small thing, otherwise looks good 👍

PR #4029, opened 2023-08-25
Add service need option fee table

patari:
#### Summary

Allows to override default finance decision fees.


akheron:
LGTM.

@Joosakur maybe you want to check this out too?

Joosakur (commented):
Looks good to me too, but there's one test which fails if `v1_enabled` is set to false so that it actually tests v2. Not immediately sure why this test is affected. 🤔 

Screenshot 2023-08-28 at 14 28 08

patari:
> Looks good to me too, but there's one test which fails if `v1_enabled` is set to false so that it actually tests v2. Not immediately sure why this test is affected. 🤔
> 
> Screenshot 2023-08-28 at 14 28 08

I think its related to https://github.com/espoon-voltti/evaka/pull/4022

patari:
I guess V2 tests are run but not asserted when both v1 and v2 are enabled

Joosakur (approved):
ah, that test is broken in master too, I'll fix that

Joosakur:
> I guess V2 tests are run but not asserted when both v1 and v2 are enabled

yeah, I couldn't think of easy way to run tests with two different configs. hopefully we can soon remove v1 altogether

PR #4030, opened 2023-08-25
Redesign job scheduler to support multiple job types

Gekkio:
#### Summary

We can't easily have duplicate `ScheduledJobRunner` instances for city-specific jobs, but instead need to supply *all possible jobs* to a single instance. Redesign how jobs are declared and configured, and pass a list of JobSchedule objects to `ScheduledJobRunner` which may come from different sources.


reynders (approved):
Mob review

PR #4051, opened 2023-08-30
Fix preschool club invoicing

patari:
#### Summary

August is half fee and attendance is always full month.

### DIFF
--- a/service/src/main/kotlin/fi/espoo/evaka/invoicing/service/DraftInvoiceGenerator.kt
+++ b/service/src/main/kotlin/fi/espoo/evaka/invoicing/service/DraftInvoiceGenerator.kt
@@ -62,6 +62,11 @@ object DefaultInvoiceGenerationLogic : InvoiceGenerationLogicChooser {
     ): InvoiceGenerationLogic = InvoiceGenerationLogic.Default
 }
 
+private val PRESCHOOL_CLUB_HALF_FEE_MONTHS = setOf(Month.AUGUST)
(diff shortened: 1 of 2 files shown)


service/src/main/kotlin/fi/espoo/evaka/invoicing/service/DraftInvoiceGenerator.kt
```diff
@@ -278,11 +292,31 @@ class DraftInvoiceGenerator(
...
+                                                            feeAlteration ->
+                                                            currentFee +
+                                                                feeAlterationEffect(
+                                                                    currentFee,
+                                                                    feeAlteration.type,
+                                                                    feeAlteration.amount,
+                                                                    feeAlteration.isAbsolute
+                                                                )
+                                                        }
+                                                    }
```
akheron, line 317: I don't understand why fee alterations are summed here, and there seems to be no test for this case

patari: `finalFee = fee + feeAlterations` and if we just use `calculateHalfFee(part.finalFee)` here and absolute fee alterations are used, alterations are also halved which is probably wrong.

patari: Oh and I added a test.

akheron (approved):

PR #4069, opened 2023-09-05
RangeSet/RangeMap data structures + necessary refactoring and abstractions

Gekkio:


reynders (approved):
Mob review

service/src/main/kotlin/fi/espoo/evaka/shared/data/RangeBasedMap.kt
```diff
@@ -0,0 +1,152 @@
...
+                            null
+                        }
+                        else -> {
+                            val subNew = newRange.subtract(overlap)
+                            val subOld = oldRange.subtract(overlap)
+
+                            // only one left range can be non-null, because otherwise there would
+                            // be more overlap
+                            subNew.left?.let { append(it to newValue) }
+                            subOld.left?.let { append(it to oldValue) }
```
Joosakur, line 123: I might consider adding some invariant/sanity checks whenever they are cheap to fail visibly 
`if (subNew.left != null && subOld.left != null) throw Error("Bug: unexpected error")`

Joosakur: We could even add `fun assertInvariants()` to RangeBasedSet and RangeBasedMap and call it on constructor to make sure that ranges don't overlap etc

service/src/main/kotlin/fi/espoo/evaka/shared/data/RangeBasedSet.kt
```diff
@@ -0,0 +1,101 @@
...
+                a.intersection(b)?.let { ranges.add(it) }
+                if (a.end <= b.end) {
+                    a = if (iterA.hasNext()) iterA.next() else null
+                } else {
+                    b = if (iterB.hasNext()) iterB.next() else null
+                }
+            }
+            return ranges
+        }
+    }
```
Joosakur, line 100: this was impossible for me to follow without explanation

Gekkio: Tidied up slightly and added a reference link for inspiration

PR #4086, opened 2023-09-08
Add attachments for fee alterations

tmkrepo:
Allows 0-n attachments per fee alteration for admin and finance admin users.

Includes:
- fee alteration id column for attachment table
- fee alteration logic for attachment configurations
- employee UI for fee alteration attachment viewing, uploading and deletion
- integration test additions
- e2e test additions


service/src/main/kotlin/fi/espoo/evaka/attachment/AttachmentsController.kt
```diff
@@ -378,6 +417,11 @@ class AttachmentsController(
                         )
                     is AttachmentParent.MessageDraft,
                     is AttachmentParent.MessageContent -> 0
+                    is AttachmentParent.FeeAlteration ->
+                        it.userFeeAlterationAttachmentCount(
+                            attachTo.feeAlterationId,
+                            user.evakaUserId
+                        )
```
akheron, line 424: This could be just `Int.MAX_VALUE` because citizens don't upload fee alteration attachments. The signature of `checkAttachmentCount` tells that this is only ever called for citizen users.

The same is true for income, pedagogical document, message draft and message content, but for some reason they have been implemented in a different way 🤷. So you could refactor it like this:

```
                    is AttachmentParent.Income,
                    is AttachmentParent.PedagogicalDocument,
                    is AttachmentParent.MessageDraft,
                    is AttachmentParent.MessageContent,
                    is AttachmentParent.FeeAlteration ->
                        Int.MAX_VALUE
```

tmkrepo: Refactored and the useless counter function removed.

PR #4099, opened 2023-09-11
regenerate finance decisions nightly

Joosakur:


service/src/main/kotlin/fi/espoo/evaka/invoicing/service/FinanceDecisionGenerator.kt
```diff
@@ -51,6 +54,45 @@ class FinanceDecisionGenerator(
...
+WITH ids AS (
+    SELECT head_of_child AS head_of_family_id
+    FROM fridge_child
+
+    UNION
+
+    SELECT head_of_family_id
+    FROM fee_decision
+)
+INSERT INTO async_job(type, payload, retry_count, retry_interval)
```
Gekkio, line 72: This will break if the GenerateFinanceDecisions async job is ever changed. I think it would be better to plan with the Kotlin API *but* in this case we're talking about a very large amount of rows and AsyncJobRunner doesn't use batch inserts internally so it would come with a performance cost :shrug:

Gekkio (approved):

PR #4103, opened 2023-09-12
Move evaka-bom and service-lib under service

Gekkio:
#### Summary

- move directories and adjust build
- the `service` Docker container is still built at repository root, because it needs `frontend/src/lib-common/generated` for the API types check


patari (approved):
I think here is wrong path (should be `from(files("$rootDir/../gradle/libs.versions.toml"))`?): https://github.com/espoon-voltti/evaka/blob/890c1ca6d8ae2b41285bdd2eda09b936f7fbfebb/service/evaka-bom/settings.gradle.kts#L8

I have no idea why both paths work for your build, but old path breaks our build. And this fixes it (I don't know if it is the correct fix): https://github.com/Tampere/trevaka/commit/46598f8acebc5c8f1339981b78b4546c74c0d603

PR #4110, opened 2023-09-13
Hide delete button in thread list on touch devices

Wnt:
## Before this change
Opening a thread required two taps on iPhone because `:hover` styles made delete button to appear. More info about this browser quirk: https://stackoverflow.com/a/50500578
## After this change
Delete button in thread list is always hidden on touch devices. Touch device detection is done with media query `@media (pointer: coarse)`

Deleting thread is still possible in thread view.


Gekkio (approved):
TIL: pointer: coarse/fine :sweat_smile:

PR #4118, opened 2023-09-15
refactor: combine application and application_form tables

msavolainen-gofore:
- move document from application_form to application table
- modify document operations to affect application table
- drop application_form table
- add document index to application table


service/src/main/resources/db/migration/V355__application_document_refactoring.sql
```diff
@@ -0,0 +1,13 @@
...
+
+ALTER TABLE application
+    ADD COLUMN document     jsonb default '{}'                     NOT NULL,
+    ADD COLUMN form_modified timestamp with time zone default now() NOT NULL;
+
+UPDATE application a SET
+    document = (SELECT document FROM application_form af WHERE a.id = af.application_id AND af.latest = true),
+    form_modified = (SELECT created FROM application_form af WHERE a.id = af.application_id AND af.latest = true);
+
+DROP TABLE application_form;
```
akheron, line 11: It might be good to keep this table (temporarily) to make it easier to revert the change if something breaks.

service/src/main/kotlin/fi/espoo/evaka/application/ApplicationQueries.kt
```diff
@@ -983,16 +970,7 @@ fun Database.Transaction.updateForm(
...
-WITH old_revisions AS (
-    UPDATE application_form SET latest = FALSE
-    WHERE application_id = :applicationId
-    RETURNING revision
-)
-INSERT INTO application_form (application_id, document, revision, latest)
-VALUES (:applicationId, :document, (SELECT coalesce(max(revision) + 1, 1) FROM old_revisions), TRUE)
-        """
-            .trimIndent()
+        "UPDATE application SET document = :document, form_modified = now() WHERE id = :applicationId;"
```
akheron, line 973: We usually pass the current timestamp as a parameter from outside instead of using `now()` or similar database functions.

In this case using `now()` might be justified, though, because it's anyway used for `created` and `updated` timestamps, and there are ~5 places that would need to be changed.

service/src/main/resources/db/migration/V355__application_document_refactoring.sql
```diff
@@ -0,0 +1,13 @@
+DROP VIEW IF EXISTS application_view;
+
+ALTER TABLE application
+    ADD COLUMN document     jsonb default '{}'                     NOT NULL,
+    ADD COLUMN form_modified timestamp with time zone default now() NOT NULL;
```
akheron, line 5: Are you sure it's the best to have defaults and `NOT NULL` here? Wouldn't it be more "correct" have `document = NULL, form_modified = NULL` when the form has not been saved yet 🤔

service/src/main/kotlin/fi/espoo/evaka/application/ApplicationQueries.kt
Gekkio: I think we *should* start passing these kind of things as parameters, especially in cases where the information is actually displayed to the user.

However, in this case `formModified` from the view doesn't seem to be used anywhere, so I'd just get rid of the whole column.

msavolainen-gofore: I didn't want to remove `formModified` as the field was already a part of application view. However I added timestamp as parameter.

msavolainen-gofore: removed `formModified`

msavolainen-gofore:
@Gekkio The `form_modified` timestamp is actually used in UI so I reverted todays changes related to removing that field.

PR #4120, opened 2023-09-15
Pass Content-Disposition header in nginx response, fixes #2412

### LINKED ISSUE #2412: Message attachments does not open correctly
If a message contains an attachment with type docx, eVaka provides a file with no file name extension or type data (see attached picture). This causes that on windows machines you have to rename the file before it can be opened. And with iPhones it can not be opened at all. Tried with Safari and Chrome.

image

Wnt:
## Before this change
Service generated Content-disposition header was not present attachment responses.
## After this change
Content-disposition header generated in `DocumentService` `X-Accel-Redirect` response is passed on to the response response nginx creates.

### DIFF
--- a/proxy/files/etc/nginx/conf.d/nginx.conf.template
+++ b/proxy/files/etc/nginx/conf.d/nginx.conf.template
@@ -336,6 +336,9 @@ server {
   location ~* ^/internal_redirect/(.*) {
     internal;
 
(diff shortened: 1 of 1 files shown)


Wnt:
Looks like nginx already internally passes the `Content-Disposition` header value to the response, so this change is not needed after all.

PR #4122, opened 2023-09-15
Staff occupancy effect selection for external staff

tmkrepo:
Adds checkboxes to external staff member addition forms for selecting whether added person has any effect on staff occupancy (mobile & desktop). Previously externals were assumed to always have staff occupancy effect.

Includes:
- UI modifications to add checkboxes
- translation modifications to remove references to assumed occupancy effect
- API changes to allow relaying the user's selection
- desktop UI logic to implicitly interpret attendance specific occupancy effect from earlier attendance markings
- test additions

### DIFF
--- a/frontend/src/employee-frontend/components/unit/unit-reservations/StaffAttendanceTable.tsx
+++ b/frontend/src/employee-frontend/components/unit/unit-reservations/StaffAttendanceTable.tsx
@@ -313,6 +313,9 @@ const StaffAttendanceModal = React.memo(function StaffAttendanceModal({
     <StaffAttendanceDetailsModal
       isExternal={true}
       onSave={onSaveExternal}
+      previousStaffOccupancyEffect={
+        detailsModalConfig.attendances[0]?.occupancyCoefficient > 0 ?? true
+      }
       validate={externalAttendanceValidator(detailsModalConfig)}
       name={detailsModalConfig.name}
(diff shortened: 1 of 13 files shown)


frontend/src/employee-frontend/components/unit/unit-reservations/StaffAttendanceTable.tsx
```diff
@@ -313,6 +313,9 @@ const StaffAttendanceModal = React.memo(function StaffAttendanceModal({
     <StaffAttendanceDetailsModal
       isExternal={true}
       onSave={onSaveExternal}
+      previousStaffOccupancyEffect={
+        detailsModalConfig.attendances[0]?.occupancyCoefficient > 0 ?? true
```
Joosakur, line 317: I think this has a bug because `undefined > 0` is `false` and therefore `?? true` is never executed

Joosakur: otherwise LGTM

tmkrepo: Default value assessment fixed. Idea is to avoid using potentially variable coefficient values (e.g. 7) in UI code.

Joosakur (approved):

PR #4134, opened 2023-09-19
Track user agent and last seen info of mobile devices

Gekkio:


service/src/main/kotlin/fi/espoo/evaka/pis/SystemController.kt
```diff
@@ -218,15 +219,20 @@ class SystemController(
             .also { Audit.EmployeeGetOrCreate.log(targetId = id) }
     }
 
-    @GetMapping("/system/mobile-devices/{id}")
```
Wnt, line 221: Will it cause problems for existing sessions that this GET endpoint is removed?

Gekkio: Yes, during deployment if any request goes through the combination of old gw + new service, or new service + old gw.
It's possible to avoid this by splitting the change into 2-3 separate pull requests and deploying each one separately, but we rarely do these since it's difficult to coordinate this with so many production environments.

akheron (approved):

PR #4154, opened 2023-09-22
Hojks e2e tests

Joosakur:

### DIFF
--- a/frontend/src/e2e-test/specs/0_citizen/citizen-child-documents.spec.ts
+++ b/frontend/src/e2e-test/specs/0_citizen/citizen-child-documents.spec.ts
@@ -112,7 +112,6 @@ beforeEach(async () => {
 
 describe('Citizen child documents listing page', () => {
   test('Published hojks is in the list', async () => {
-    await page.reload()
     await header.openChildPage(child.id)
     const childPage = new CitizenChildPage(page)
     await childPage.openCollapsible('vasu')
@@ -124,7 +123,6 @@ describe('Citizen child documents listing page', () => {
   })
 
   test('Published pedagogical report is in the list', async () => {
-    await page.reload()
     await header.openChildPage(child.id)
     const childPage = new CitizenChildPage(page)
     await childPage.openCollapsible('vasu')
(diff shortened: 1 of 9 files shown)


frontend/src/e2e-test/specs/0_citizen/citizen-child-documents.spec.ts
```diff
@@ -75,13 +111,25 @@ beforeEach(async () => {
 })
 
 describe('Citizen child documents listing page', () => {
-  test('Published document is in the list', async () => {
+  test('Published hojks is in the list', async () => {
+    await page.reload()
```
akheron, line 115: Why do both tests have `page.reload()`? I don't see a reason. All the test data has been set up before the page is opened.

In any case, it's better to avoid reloading the page because it takes some time. The usual reason is that we want to reflect some new data that was inserted after the page was initially loaded, but that can usually be worked around.

frontend/src/e2e-test/specs/5_employee/child-documents.spec.ts
```diff
@@ -89,36 +90,110 @@ describe('Employee - Child documents', () => {
...
+    await childDocument.previewButton.click()
+    await childDocument.publish()
+
+    // Assert status and publish time
+    await childDocument.returnButton.click()
+    childDocumentsSection = await childInformationPage.openCollapsible(
+      'childDocuments'
+    )
+    await waitUntilEqual(childDocumentsSection.childDocumentsCount, 1)
+    row = childDocumentsSection.childDocuments(0)
```
akheron, line 143: Not relevant for this PR, but I think having  a `childDocumentsSection.childDocumentRows` locator would be a better abstraction. You could then just have:

```
await childDocumentsSection.childDocumentRows.assertCount(1)
row = childDocumentsSection.childDocumentRows.nth(0)
```

akheron (approved):

frontend/src/e2e-test/specs/0_citizen/citizen-child-documents.spec.ts
Joosakur: uhm, maybe some copy-paste-mistake

PR #4164, opened 2023-09-25
Add download (print) button to citizen child document view

Joosakur:

### DIFF
--- a/frontend/src/lib-components/layout/flex-helpers.ts
+++ b/frontend/src/lib-components/layout/flex-helpers.ts
@@ -5,6 +5,7 @@
 import { Property } from 'csstype'
 import styled from 'styled-components'
 
+import { tabletMin } from '../breakpoints'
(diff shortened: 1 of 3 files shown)


frontend/src/lib-components/layout/flex-helpers.ts
```diff
@@ -14,6 +15,7 @@ interface FixedSpaceRowProps {
   alignItems?: Property.AlignItems
   fullWidth?: boolean
   maxWidth?: string
+  mobileMargin?: boolean
```
akheron, line 18: I don't think it's good to add one more styling prop just for one use case to a very very widely used component that already has 7 styling props 🤔. It's possible to do this instead in the usage site:

```
const CustomRow = styled(FixedSpaceRow)`
  @media (max-width: ${tabletMin}) {
    margin-right: ${defaultMargins.s};
  }
`
```

Joosakur: yeah you're probably right, fixed 👍

PR #4170, opened 2023-09-26
add StaticTextDisplayQuestion to child documents

Joosakur:

### DIFF
--- a/service/src/main/kotlin/fi/espoo/evaka/document/childdocument/ChildDocument.kt
+++ b/service/src/main/kotlin/fi/espoo/evaka/document/childdocument/ChildDocument.kt
@@ -75,10 +75,10 @@ sealed class AnsweredQuestion<Answer>(val type: QuestionType) {
     @JsonTypeName("STATIC_TEXT_DISPLAY")
     data class StaticTextDisplayAnswer(
         override val questionId: String,
-        override val answer: String?
-    ) : AnsweredQuestion<String?>(QuestionType.STATIC_TEXT_DISPLAY) {
+        override val answer: Nothing?
+    ) : AnsweredQuestion<Nothing?>(QuestionType.STATIC_TEXT_DISPLAY) {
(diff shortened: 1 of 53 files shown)


service/src/main/kotlin/fi/espoo/evaka/document/childdocument/ChildDocument.kt
```diff
@@ -71,6 +71,16 @@ sealed class AnsweredQuestion<Answer>(val type: QuestionType) {
             return answer == null || question.options.any { it.id == answer }
         }
     }
+
+    @JsonTypeName("STATIC_TEXT_DISPLAY")
+    data class StaticTextDisplayAnswer(
+        override val questionId: String,
+        override val answer: String?
+    ) : AnsweredQuestion<String?>(QuestionType.STATIC_TEXT_DISPLAY) {
```
akheron, line 79: Perhaps the type of `answer` could be `Unit`. Not sure how it serializes to JSON, but it might become `null`.

Joosakur: I was considering that, but I feel it increases complexity by raising too many unnecessary questions:

- how does it (de)serialize to json in postgres?
- how does it (de)serialize to json in api?
- if I add Unit to tsTypings in codegen, should it be null, undefined, void or what, and is the answer going to be same in other use cases which might use Unit later on?

akheron: `Nothing?` seems to work without having these problems. `Nothing` has no instances, so it neatly maps to `never` in the TS typings (in codegen, add `java.lang.Void` => `never`). The only possible value you can pass to `Nothing?` is therefore `null`, and serializing it to postgres and/or json is well defined.

akheron (approved):

PR #4173, opened 2023-09-27
Use dependabot for gradle dependencies

akheron:


patari (approved):
Wait for https://github.com/espoon-voltti/evaka/pull/4172

PR #4217, opened 2023-10-04
Fix applications list start date when changed from placement draft

patari:

### DIFF
--- a/frontend/src/e2e-test/pages/employee/placement-draft-page.ts
+++ b/frontend/src/e2e-test/pages/employee/placement-draft-page.ts
@@ -2,9 +2,10 @@
 //
 // SPDX-License-Identifier: LGPL-2.1-or-later
 
+import LocalDate from 'lib-common/local-date'
 import { UUID } from 'lib-common/types'
 
-import { Combobox, Page } from '../../utils/page'
+import { Combobox, DatePickerDeprecated, Page } from '../../utils/page'
 
 export class PlacementDraftPage {
   constructor(private page: Page) {}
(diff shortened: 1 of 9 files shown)


frontend/src/e2e-test/pages/employee/placement-draft-page.ts
```diff
@@ -13,6 +14,13 @@ export class PlacementDraftPage {
     '[data-qa="restricted-details-warning"]'
   )
 
+  setStartDate = async (date: LocalDate) => {
```
Gekkio, line 17: Unnecessary wrapper, expose a public field instead: `readonly datePicker = this.page.findByDataQa('start-date')`

service/src/main/kotlin/fi/espoo/evaka/application/ApplicationQueries.kt
```diff
@@ -538,6 +539,7 @@ fun Database.Read.fetchApplicationSummaries(
                                         row.mapColumn("unit_reject_other_reason")
                                 )
                             },
+                    placementProposalStartDate = row.mapColumn("placement_plan_start_date"),
```
Gekkio, line 542: I think this should be called `placementPlanStartDate` because proposal = a plan that has been sent to the unit, but this information is shown regardless of this state whenever a plan exists (and yes, the existing `placementProposalUnitName` field is also wrong).

frontend/src/e2e-test/pages/employee/placement-draft-page.ts
patari: Fixed

service/src/main/kotlin/fi/espoo/evaka/application/ApplicationQueries.kt
patari: Fixed both

Gekkio (approved):

PR #4234, opened 2023-10-09
Fix assistance acl

reynders:
#### Summary
In daycare
- STAFF, SUPERVISOR sees only current and future assistance info for child in same unit
- VEO sees also past info for child in same unit
- Admin sees all

In preschool same rules, and
- VEO sees only preschool period assistance info, not earlier

### DIFF
--- a/service/src/main/kotlin/fi/espoo/evaka/assistance/AssistanceQueries.kt
+++ b/service/src/main/kotlin/fi/espoo/evaka/assistance/AssistanceQueries.kt
@@ -43,8 +43,7 @@ WHERE child_id = ${bind(childId)} AND ${predicate(filter.forTable("assistance_fa
                     .trimIndent()
             )
         }
-        .mapTo<AssistanceFactor>()
-        .toList()
+        .toList<AssistanceFactor>()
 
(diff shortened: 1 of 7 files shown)


service/src/main/kotlin/fi/espoo/evaka/assistance/AssistanceQueries.kt
```diff
@@ -266,3 +316,45 @@ fun Database.Transaction.deleteOtherAssistanceMeasure(id: OtherAssistanceMeasure
...
+LEFT JOIN assistance_action_option_ref aaor ON aaor.action_id = aa.id
+LEFT JOIN assistance_action_option aao ON aao.id = aaor.option_id
+WHERE aa.child_id = ${bind(childId)} AND ${predicate(filter.forTable("aa"))}
+GROUP BY aa.id, child_id, start_date, end_date, other_action
+ORDER BY start_date DESC     
+            """
+                    .trimIndent()
+            )
+        }
+        .mapTo<AssistanceAction>()
```
Gekkio, line 359: remove mapTo and use `.toList<AssistanceAction>()`

Gekkio (approved):

PR #4240, opened 2023-10-10
Fix remaining group notes queries

Gekkio:
#### Summary

Looks like I missed these in the first PR :(

### DIFF
--- a/frontend/src/employee-mobile-frontend/child-attendance/actions/MarkDeparted.tsx
+++ b/frontend/src/employee-mobile-frontend/child-attendance/actions/MarkDeparted.tsx
@@ -10,7 +10,12 @@ import { formatTime, isValidTime } from 'lib-common/date'
 import { AttendanceTimes } from 'lib-common/generated/api-types/attendance'
 import LocalDate from 'lib-common/local-date'
 import LocalTime from 'lib-common/local-time'
-import { useMutationResult, useQuery, useQueryResult } from 'lib-common/query'
+import {
+  queryOrDefault,
+  useMutationResult,
+  useQuery,
+  useQueryResult
+} from 'lib-common/query'
 import useNonNullableParams from 'lib-common/useNonNullableParams'
 import { mockNow } from 'lib-common/utils/helpers'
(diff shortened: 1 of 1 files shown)


frontend/src/employee-mobile-frontend/child-attendance/ChildListItem.tsx
```diff
@@ -115,9 +115,10 @@ export default React.memo(function ChildListItem({
   }>()
   const { selectedGroupId } = useSelectedGroup()
 
-  const { data: groupNotes } = useQuery(groupNotesQuery(child.groupId ?? ''), {
-    enabled: child.groupId != null
+  const rawGroupNotes = useQueryResult(groupNotesQuery(child.groupId ?? ''), {
+    enabled: !!child.groupId
   })
+  const groupNotes = child.groupId ? rawGroupNotes.getOrElse([]) : []
```
akheron, line 121: It seems to me that this jus a longer way to do this:
```
  const { data: groupNotes = [] } = useQuery(groupNotesQuery(child.groupId ?? ''), {
    enabled: !!child.groupId
  })
```

When query is disabled or fetching, `useQuery` returns `undefined` `data` and the default `[]` kicks in.

frontend/src/employee-mobile-frontend/child-attendance/actions/MarkDeparted.tsx
```diff
@@ -101,7 +100,11 @@ export default React.memo(function MarkDeparted() {
     AbsenceTypeWithNoAbsence | undefined
   >(undefined)
 
-  const groupNotes = useQueryResult(groupNotesQuery(groupId))
+  const groupId = child.map(({ groupId }) => groupId).getOrElse(null)
+  const rawGroupNotes = useQueryResult(groupNotesQuery(groupId ?? ''), {
+    enabled: !!groupId
+  })
+  const groupNotes = groupId ? rawGroupNotes : Success.of([])
```
akheron, line 107: This place would benefit from something like `constantQuery([])`, be we don't have that yet.

PR #4241, opened 2023-10-11
fix unread pedagogical document without attachment

Joosakur:
#### Summary
docs without attachment cannot be marked as read so exclude them from unread count


akheron:
Would it be better to mark documents without attachment as read in the frontend when the list is opened? 🤔

PR #4248, opened 2023-10-12
Add flyway additional locations support

patari:
#### Summary

eVaka doesn't have UI to modify every database so we have been using flyway after migrate scripts to populate the data. This has been working great: every environment we have has the exactly same data and it is well documented and versioned through git.

We are going to use service for multiple municipalities, so we would like to have different after migrate scripts for every municipality. This little change adds support for it.

Example configuration with application.yml:

```
evaka:
  database:
    flyway:
      additional_locations: db/data/tampere
```

### DIFF
--- a/service/src/main/kotlin/fi/espoo/evaka/EvakaEnv.kt
+++ b/service/src/main/kotlin/fi/espoo/evaka/EvakaEnv.kt
@@ -124,6 +124,7 @@ data class DatabaseEnv(
     val password: Sensitive<String>,
     val flywayUsername: String,
     val flywayPassword: Sensitive<String>,
+    val flywayAdditionalLocations: List<String>?,
     val flywayIgnoreFutureMigrations: Boolean,
     val leakDetectionThreshold: Long,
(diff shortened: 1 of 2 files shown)


Gekkio (approved):
Seems good :+1: 
One idea: would it be simpler to specify all locations in the config and not just *additional locations*. So `flywayLocations: List<String>` + sensible default for it? This would mean anybody adding extra would need to include `"db/migration"` too, but it could actually be good to be explicit like that.

:shrug: Either way is fine by me, so you can choose

patari:
> Seems good 👍 One idea: would it be simpler to specify all locations in the config and not just _additional locations_. So `flywayLocations: List<String>` + sensible default for it? This would mean anybody adding extra would need to include `"db/migration"` too, but it could actually be good to be explicit like that.
> 
> 🤷 Either way is fine by me, so you can choose

:+1: now the configuration is:

```
evaka:
  database:
    flyway:
      locations: db/migration,db/data/tampere
```

PR #4249, opened 2023-10-12
Add frontend build context support

patari:
#### Summary

To use `frontend/Dockerfile` we first need to copy frontend customization files to correct location and then use `docker build`. This changes Dockerfile so that we can provide customization files with just docker (which is great because we can just use `docker compose up` and everything works).

Downside to this (for Espoo) is that `docker build -t evaka/frontend` doesn't anymore work and requires additional command line parameter: `--build-context customizations=src/lib-customizations/espoo`. This can be fixed with high-level docker bake file, but I left that out of this commit because bake files aren't currently used.

Note: I think this requires a little change to `voltti-actions` but I don't have write access to it so here it goes:
```
diff --git a/docker-build-registry/action.yml b/docker-build-registry/action.yml
index 6143fcc..bb28892 100644
--- a/docker-build-registry/action.yml
+++ b/docker-build-registry/action.yml
@@ -24,6 +24,10 @@ inputs:
     description: "Docker build-args. See docker/build-push-action build-args"
     required: false
     default: ""
+  build-contexts:
+    description: "Docker build-contexts. See docker/build-push-action build-contexts"
+    required: false
+    default: ""
   target:
     description: "Docker target"
     default: null
@@ -106,6 +110,7 @@ runs:
         target: ${{ inputs.target }}
         file: ${{ inputs.dockerfile }}
         build-args: ${{ inputs.build-args }}
+        build-contexts: ${{ inputs.build-contexts }}
         cache-from: |
           ${{ inputs.cache_from != '' && format('type=registry,ref={0}', inputs.cache_from) || '' }}
           ${{ format('type=registry,ref={0}', steps.image.outputs.cache) }}
```

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -190,7 +190,7 @@ jobs:
             ICONS=pro
             SENTRY_AUTH_TOKEN=${{ secrets.SENTRY_AUTH_TOKEN }}
           build-contexts:
-            customizations=src/lib-customizations/espoo
+            customizations=frontend/src/lib-customizations/espoo
 
       - name: Docker cleanup
         if: always()
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -107,6 +107,8 @@ jobs:
             build=${{ github.run_number }}
             commit=${{ github.event.pull_request.head.sha || github.sha }}
             ICONS=free
+          build-contexts:
+            customizations=frontend/src/lib-customizations/espoo
 
       - name: Build frontend image builder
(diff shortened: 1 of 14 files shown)


patari:
Tampere commit: https://github.com/Tampere/trevaka/commit/fe757a69e115fa516d83c1a0da7df3ce8b6ab591 (no docker compose changes here because I need to change some other stuff first, but the change is just adding `additional_contexts` like in this PR).

ilja-git (approved):

ilja-git:
https://github.com/espoon-voltti/voltti-actions/pull/44/files

ilja-git (approved):

PR #4250, opened 2023-10-12
Expanded ECES access to reports, groups, placements and service needs

tmkrepo:
Grant ECES permission to:
- add and update placements
- remove future placements
- add, update and remove service needs
- add, update and remove daycare groups
- view placement guarantee report

Changes OK for Tampere, Turku
ECES role not in use in Espoo, Oulu

### DIFF
--- a/frontend/src/employee-frontend/components/child-information/Placements.tsx
+++ b/frontend/src/employee-frontend/components/child-information/Placements.tsx
@@ -60,7 +60,8 @@ export default React.memo(function Placements({ id, startOpen }: Props) {
               'SERVICE_WORKER',
               'UNIT_SUPERVISOR',
               'FINANCE_ADMIN',
-              'ADMIN'
+              'ADMIN',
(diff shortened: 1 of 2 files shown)


frontend/src/employee-frontend/components/child-information/Placements.tsx
```diff
@@ -60,7 +60,8 @@ export default React.memo(function Placements({ id, startOpen }: Props) {
               'SERVICE_WORKER',
               'UNIT_SUPERVISOR',
               'FINANCE_ADMIN',
-              'ADMIN'
+              'ADMIN',
+              'EARLY_CHILDHOOD_EDUCATION_SECRETARY'
```
akheron, line 64: It should be possible to replace this with a check against `permittedActions` from `unitQuery`, but perhaps it's not relevant for this PR.

akheron (approved):

PR #4254, opened 2023-10-13
Add `queryOrDefault` helper

akheron:

### DIFF
--- a/frontend/src/employee-mobile-frontend/messages/state.tsx
+++ b/frontend/src/employee-mobile-frontend/messages/state.tsx
@@ -86,16 +86,25 @@ const markMatchingThreadRead = (
 export const MessageContextProvider = React.memo(
   function MessageContextProvider({ children }: { children: JSX.Element }) {
     const { unitInfoResponse } = useContext(UnitContext)
+
     const { user } = useContext(UserContext)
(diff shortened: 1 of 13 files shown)


frontend/src/employee-mobile-frontend/messages/state.tsx
```diff
@@ -92,9 +93,10 @@ export const MessageContextProvider = React.memo(
 
     const { selectedGroupId } = useSelectedGroup()
 
-    const accounts = useQueryResult(messagingAccountsQuery(unitId ?? ''), {
-      enabled: unitId !== undefined && pinLoginActive
-    })
+    const accounts = useQueryResult(
+      queryOrDefault(messagingAccountsQuery, [])(unitId),
+      { enabled: pinLoginActive }
+    )
```
akheron, line 99: Check how this behaves when `enabled` changes from true to false

akheron: Cached data is returned. The correct fix would be to include `employeeId` in the query's cache key.

akheron: Fixed by using `employeeId` in the account query's cache key

PR #4274, opened 2023-10-17
Upgrade apigw's node to fix vulnerabilities

akheron:

### DIFF
--- a/apigw/Dockerfile
+++ b/apigw/Dockerfile
@@ -2,7 +2,7 @@
 #
 # SPDX-License-Identifier: LGPL-2.1-or-later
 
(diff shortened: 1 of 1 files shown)


patari (approved):

akheron:
> Not related to this PR but I think
> 
> ```
> ARG BASE_IMAGE=node:18.18.2-buster-slim
> FROM "${BASE_IMAGE}" AS base
> ```
> 
> should just be `FROM node:18.18.2-buster-slim AS base`. Then dependabot should work (which is already configured for this path).
> 
> I have no idea why we need ARG for base image version.

Fixed in https://github.com/espoon-voltti/evaka/pull/4275

PR #4283, opened 2023-10-18
Titania port

msavolainen-gofore:

### DIFF
--- a/service/src/integrationTest/kotlin/fi/espoo/evaka/titania/TitaniaControllerTest.kt
+++ b/service/src/integrationTest/kotlin/fi/espoo/evaka/titania/TitaniaControllerTest.kt
@@ -1,0 +1,111 @@
+// SPDX-FileCopyrightText: 2021-2022 City of Tampere
+//
+// SPDX-License-Identifier: LGPL-2.1-or-later
+
+package fi.espoo.evaka.titania
+
+import com.github.kittinunf.fuel.core.extensions.jsonBody
+import fi.espoo.evaka.FullApplicationTest
+import fi.espoo.evaka.shared.auth.AuthenticatedUser
+import fi.espoo.evaka.shared.auth.asUser
+import java.nio.charset.StandardCharsets
+import org.assertj.core.api.Assertions.assertThat
+import org.junit.jupiter.api.Test
+import org.skyscreamer.jsonassert.JSONAssert.assertEquals
+import org.skyscreamer.jsonassert.JSONCompareMode
+import org.springframework.core.io.ClassPathResource
+import org.springframework.util.StreamUtils
+
+internal class TitaniaControllerTest : FullApplicationTest(resetDbBeforeEach = true) {
+
+    @Test
+    fun `put working time events with system user should respond 403`() {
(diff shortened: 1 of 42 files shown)


service/src/main/kotlin/fi/espoo/evaka/titania/Titania.kt
```diff
@@ -0,0 +1,248 @@
...
+
+enum class TitaniaPayrollItemUnit {
+    /** Minuutit */
+    MINUTE,
+
+    /** Kappaleet */
+    QUANTITY
+}
+
+data class UpdateWorkingTimeEventsRequest(
```
Wnt, line 125: Add `@ExcludeCodeGen` annotation to all route payload / return types to not have them included in `frontend/src/lib-common/generated/api-types/`
```suggestion
@ExcludeCodeGen
data class UpdateWorkingTimeEventsRequest(
```

Gekkio (commented):
One general comment: I think `internal` never useful, and things should generally be just public (no visibility keywords) or `private` if you really want to hide something.

service/src/integrationTest/kotlin/fi/espoo/evaka/titania/TitaniaControllerTest.kt
```diff
@@ -0,0 +1,213 @@
...
+import org.mockito.kotlin.verifyNoInteractions
+import org.mockito.kotlin.whenever
+import org.skyscreamer.jsonassert.JSONAssert.assertEquals
+import org.skyscreamer.jsonassert.JSONCompareMode
+import org.springframework.boot.test.mock.mockito.MockBean
+import org.springframework.core.io.ClassPathResource
+import org.springframework.http.HttpStatus
+import org.springframework.util.StreamUtils
+
+internal class TitaniaControllerTest : FullApplicationTest(resetDbBeforeEach = true) {
```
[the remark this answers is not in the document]
vesahonkasaari: Original TitaniaControllerTest was entirely replaced with tests which call the HTTP endpoints to verify the status code handling

service/src/main/kotlin/fi/espoo/evaka/EvakaEnv.kt
```diff
@@ -98,7 +99,8 @@ data class EvakaEnv(
                 feeDecisionGeneratorV2Enabled =
                     env.lookup("evaka.fee_decision.generator.v2_enabled") ?: false,
                 voucherValueDecisionGeneratorV2Enabled =
-                    env.lookup("evaka.voucher_value_decision.generator.v2_enabled") ?: false
+                    env.lookup("evaka.voucher_value_decision.generator.v2_enabled") ?: false,
+                titaniaEnabled = env.lookup("evaka.titania_enabled") ?: false
```
Gekkio, line 103: This should probably be removed since it doesn't seem to be now used anywhere (?)

vesahonkasaari: Good catch, fixed.

Gekkio (approved):

PR #4287, opened 2023-10-19
Paged infinite query

akheron:
#### Summary

Add a better API for implementing infinite scrolling with eVaka's pagination conventions.

### DIFF
--- a/frontend/src/citizen-frontend/messages/state.tsx
+++ b/frontend/src/citizen-frontend/messages/state.tsx
@@ -2,7 +2,6 @@
 //
 // SPDX-License-Identifier: LGPL-2.1-or-later
 
-import uniqBy from 'lodash/uniqBy'
 import React, {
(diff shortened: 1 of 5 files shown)


frontend/src/citizen-frontend/messages/state.tsx
```diff
@@ -67,21 +65,18 @@ export const isRegularThread = (
 ): thread is CitizenMessageThread.Regular => thread.type === 'Regular'
 
 const markMessagesReadByThreadId = (
-  threads: CitizenMessageThread[],
+  thread: CitizenMessageThread,
   threadId: UUID
-): CitizenMessageThread[] =>
-  threads.filter(isRegularThread).map(
```
akheron, line 73: There's a bug here that removes all the redacted threads from react-query cache when a regular thread is marked as read. It's fixed by this PR.

PR #4294, opened 2023-10-19
Add flags to employee message editor

Wnt:
## Before this change
It was not possible to send message threads marked as sensitive.
## After this change
Employee front-end gains support for sending message threads marked as sensitive. This feature (displaying the sensitivity checkbox) is behind a feature flag `sensitiveMessaging` which is set false in all Espoo production and staging environments, but defaults to true in order to have it enabled during localhost development.
sensitive messaging

### DIFF
--- a/frontend/src/lib-customizations/defaults/components/i18n/fi.tsx
+++ b/frontend/src/lib-customizations/defaults/components/i18n/fi.tsx
@@ -118,7 +118,7 @@ const components: Translations = {
       bulletin: 'Tiedote (ei voi vastata)'
     },
     flags: {
-      heading: 'Viestin lisämekinnät',
+      heading: 'Viestin lisämerkinnät',
       urgent: {
         info: 'Lähetä viesti kiireellisenä vain, jos haluat että huoltaja lukee sen työpäivän aikana.',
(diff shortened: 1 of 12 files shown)


frontend/src/lib-customizations/defaults/components/i18n/fi.tsx
```diff
@@ -116,10 +117,18 @@ const components: Translations = {
       message: 'Viesti',
       bulletin: 'Tiedote (ei voi vastata)'
     },
-    urgent: {
-      heading: 'Merkitse kiireelliseksi',
-      info: 'Lähetä viesti kiireellisenä vain, jos haluat että huoltaja lukee sen työpäivän aikana.',
-      label: 'Kiireellinen'
+    flags: {
+      heading: 'Viestin lisämekinnät',
```
akheron, line 121:
```suggestion
      heading: 'Viestin lisämerkinnät',
```

akheron (approved):

frontend/src/lib-customizations/defaults/components/i18n/fi.tsx
Wnt: done

PR #4308, opened 2023-10-23
Message undo optimization

Joosakur:
#### Summary

- Add `on delete cascade`s to simplify deletion and possibly improve performance
- Avoid fetching a long list of IDs and then later deleting based on passing it back
- Introduce assumptions to improve performance (e.g. `limit 1` vs `distinct`)

### DIFF
--- a/service/src/main/kotlin/fi/espoo/evaka/messaging/MessageQueries.kt
+++ b/service/src/main/kotlin/fi/espoo/evaka/messaging/MessageQueries.kt
@@ -1458,7 +1458,7 @@ fun Database.Transaction.undoMessageReply(
         )
     }
     lockMessageContentForUpdate(undoTarget.contentId)
-    this.deleteMessages(undoTarget.contentId)
+    this.deleteMessages(undoTarget.contentId, deleteThreads = false)
     this.resetSenderThreadParticipants(undoTarget.threadId, accountId)
 }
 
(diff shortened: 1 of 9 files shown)


akheron:
The only thing I'm thinking is whether the thread cleanup should be done immediately after in an async job, or is it better to have it in a scheduled job 🤔

Joosakur:
> The only thing I'm thinking is whether the thread cleanup should be done immediately after in an async job, or is it better to have it in a scheduled job 🤔

Yeah I was thinking that too. I think nightly causes less total load since the query anyway targets all orphan threads and I also did not find any issues in the UI due to the leftovers, so maybe there's no advantage in cleaning them up immediately.

service/src/main/kotlin/fi/espoo/evaka/messaging/MessageQueries.kt
```diff
@@ -1519,6 +1521,22 @@ fun Database.Transaction.deleteMessages(contentId: MessageContentId) {
...
+            sql(
+                """
+            DELETE FROM message_thread
+            WHERE id IN (
+                SELECT mt.id
+                FROM message_thread mt
+                JOIN message m on mt.id = m.thread_id
+                WHERE m.content_id = ${bind(contentId)}
+                FOR UPDATE 
+            )
```
akheron, line 1534: There exists a shorter syntax for the same thing:

```
DELETE FROM message_thread mt
USING message m
WHERE mt.id = m.thread_id AND m.content_id = ${bind(contentId)}
```

Also not sure why you added `FOR UPDATE` here 🤔

akheron: ... and even if you prefer the longer syntax, you can select directly from `message`:

```
DELETE FROM message_thread
WHERE id IN (
    SELECT thread_id
    FROM message m
    WHERE content_id = ${bind(contentId)}
)
```

akheron (commented):
Otherwise LGTM 👍

service/src/main/kotlin/fi/espoo/evaka/messaging/MessageQueries.kt
Joosakur: 💡

akheron (approved):

PR #4310, opened 2023-10-23
Fix a bug in the voucher value decision report

vesahonkasaari:
When a voucher value decision's validity has been first reduced, and then it's completely annulled later in the same month, the period for which the decision's validity was reduced would not be refunded properly. Fix the "validity updated after last freeze" check to also include annulled decisions. It should be safe to do so, since we know that that the validity can't be changed after the decision has been annulled, so the annulment must have happened after the validity update, which has happened after the last freeze, and thus the annulled decision is in scope for corrections


Joosakur (approved):

Joosakur:
Should be fine 👍

PR #4313, opened 2023-10-24
warn citizen of marking multiple absences for locked period

Joosakur:
#### Summary

Screenshot 2023-10-24 at 12 29 45


frontend/src/lib-customizations/defaults/citizen/i18n/sv.tsx
```diff
@@ -368,6 +368,9 @@ const sv: Translations = {
       selectedChildren: 'Utvalda barn',
       selectChildrenInfo:
         'Anmäl endast heldagsfrånvaro. Kortare frånvaro kan anmälas genom reservationer.',
+      lockedAbsencesWarningTitle: 'Frånvaro i flera dagar',
+      lockedAbsencesWarningText:
+        'Du håller på att markera frånvaro för dagar ,där anmälningstiden har gått ut. Detta kan endast återkallas genom att vara i kontakt med personalen.',
```
akheron, line 373:
```suggestion
        'Du håller på att markera frånvaro för flera dagar, där anmälningstiden har gått ut. Detta kan endast återkallas genom att vara i kontakt med personalen.',
```

Comma in wrong place and missing `flera`.

frontend/src/lib-customizations/defaults/citizen/i18n/sv.tsx
```diff
@@ -368,6 +368,9 @@ const sv: Translations = {
       selectedChildren: 'Utvalda barn',
       selectChildrenInfo:
         'Anmäl endast heldagsfrånvaro. Kortare frånvaro kan anmälas genom reservationer.',
+      lockedAbsencesWarningTitle: 'Frånvaro i flera dagar',
```
akheron, line 371:
```suggestion
      lockedAbsencesWarningTitle: 'Frånvaro för flera dagar',
```
Not sure about `för` vs `i` vs `på` here, but I think that at least this text and the longer text should use the same word 🤷 

Perhaps these translations should be checked anyway with someone who actually knows the language.

frontend/src/citizen-frontend/calendar/AbsenceModal.tsx
```diff
@@ -119,6 +120,13 @@ export default React.memo(function AbsenceModal({
     i18n
   )
 
+  const absencesWarning =
+    range.isValid() &&
+    range
+      .value()
+      .complement(reservationsResponse.reservableRange)
```
akheron, line 127: It should be enough to check that `range.start < reservableRange.start`. If you wan to make it fancier, you could also use `!range.intersection(reservationsResponse).isEqual(range)` to avoid adding the `complement()` method.

akheron (changes requested):

frontend/src/lib-customizations/defaults/citizen/i18n/sv.tsx
Joosakur: the text came from svebi 😅

frontend/src/citizen-frontend/calendar/AbsenceModal.tsx
Joosakur: Hmm I don't think so. We need to know if absence period contains more than one nonreservable day. Neither of those answer that AFAIK.

akheron (approved):

akheron (approved):

PR #4314, opened 2023-10-24
Sensitive messages citizen views

Wnt:
## Before this change
Display of redacted versions on sensitive message threads was not implemented.
## After this change
When citizen is logged-in with weak authentication they see a prompt to login with strong authentication when trying to view message threads marked as sensitive.
image

After strong authentication the lock icons are removed from sensitive threads and instead their title is appended with "(Sensitive message thread)"
image

This change also adds end-to-end tests for sending and viewing sensitive messages.

### DIFF
--- a/frontend/src/lib-components/messages/MessageEditor.tsx
+++ b/frontend/src/lib-components/messages/MessageEditor.tsx
@@ -378,7 +378,7 @@ export default React.memo(function MessageEditor({
   const sensitiveCheckbox = (
     <FixedSpaceRow spacing="xs" alignItems="center">
       <Checkbox
-        data-qa="checkbox-senstitive"
+        data-qa="checkbox-sensitive"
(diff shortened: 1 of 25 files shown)


Wnt:
WIP - fixing few things found out during UX walktrough

frontend/src/lib-components/messages/MessageEditor.tsx
```diff
@@ -420,20 +420,27 @@ export default React.memo(function MessageEditor({
         />
       </>
     )
+  const FlagsInfoContent = () => {
```
akheron, line 423: You should not define a React component inside another component. React will see it as a new component every time and will not be able to reconcile it optimally. You should move the component to the top level, pass the `message` as a prop, and use `useTranslation()` to get `i18n`.

Wnt: Moved outside!

PR #4322, opened 2023-10-25
Improve access control performance - part 2

Gekkio:
#### Summary

Old implementation had this query structure:

```
SELECT ...
FROM foo
JOIN (
  SELECT ...
  
  UNION ALL
  
  SELECT ...
  
  UNION ALL
  
  SELECT ...
) acl
```

This didn't generate good query plans when checking for permissions for a single target (= the common case). The union queries effectively fetched acl information for *all possible targets* for the current user, regardless of which targets we're interested in.

New implementation has this query structure:

```
SELECT ...
FROM foo
JOIN (
  SELECT ...
) acl

UNION ALL

SELECT ...
FROM foo
JOIN (
  SELECT ...
) acl

UNION ALL

SELECT ...
FROM foo
JOIN (
  SELECT ...
) acl
```

This can generate an optimal query plan, where the database effectively does three separate checks, but each of them is optimal for a single target. The repetition in the query is minimized by abstracting it away and requiring rules to provide a *minimal query* (= just target id + child id pairs).

The new implementation is currently only used by one vasu rule, so we can check its impact first in production.

### DIFF
--- a/service/src/integrationTest/kotlin/fi/espoo/evaka/shared/security/ChildDailyNoteAccessControlTest.kt
+++ b/service/src/integrationTest/kotlin/fi/espoo/evaka/shared/security/ChildDailyNoteAccessControlTest.kt
@@ -1,0 +1,87 @@
+// SPDX-FileCopyrightText: 2017-2023 City of Espoo
+//
+// SPDX-License-Identifier: LGPL-2.1-or-later
+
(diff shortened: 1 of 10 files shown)


akheron (approved):
So this didn't change all permission checks to use the new model. Is that coming later still?

PR #4324, opened 2023-10-25
Pass a database connection as a parameter to the payment integration

vesahonkasaari:
This is needed by the Turku payment integration to rummage the DB for stuff which is not suitable for being included in the Payment objects

### DIFF
--- a/service/src/main/kotlin/fi/espoo/evaka/invoicing/domain/Payments.kt
+++ b/service/src/main/kotlin/fi/espoo/evaka/invoicing/domain/Payments.kt
@@ -32,10 +32,10 @@ interface PaymentIntegrationClient {
         val failed: List<Payment> = listOf()
     )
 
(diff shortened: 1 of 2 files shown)


service/src/main/kotlin/fi/espoo/evaka/invoicing/domain/Payments.kt
```diff
@@ -32,10 +32,10 @@ interface PaymentIntegrationClient {
         val failed: List<Payment> = listOf()
     )
 
-    fun send(payments: List<Payment>): SendResult
+    fun send(payments: List<Payment>, tx: Database.Transaction): SendResult
```
patari, line 35:
```suggestion
    fun send(payments: List<Payment>, tx: Database.Read): SendResult
```

If there is no need to run updates to database (`Transaction` extends `Read`).

patari (approved):

PR #4344, opened 2023-10-30
Add mobile staff attendance edit

patari:
#### Summary

New feature under experimental feature flag for testing.

### DIFF
--- a/frontend/src/e2e-test/pages/mobile/staff-page.ts
+++ b/frontend/src/e2e-test/pages/mobile/staff-page.ts
@@ -3,10 +3,12 @@
 // SPDX-License-Identifier: LGPL-2.1-or-later
 
 import { StaffAttendanceType } from 'lib-common/generated/api-types/attendance'
+import { UUID } from 'lib-common/types'
 
 import { DaycareGroup } from '../../dev-api/types'
 import { waitUntilEqual } from '../../utils'
 import {
+  AsyncButton,
   Checkbox,
   Combobox,
   Element,
@@ -95,7 +97,11 @@ export default class StaffPage {
 }
 
 export class StaffAttendancePage {
-  constructor(private readonly page: Page) {}
+  editButton: Element
+
+  constructor(private readonly page: Page) {
+    this.editButton = this.page.findByDataQa('edit')
+  }
 
   #tabs = {
     present: this.page.findByDataQa('present-tab'),
@@ -216,10 +222,15 @@ export class StaffAttendancePage {
     await this.anyMemberPage.back.click()
   }
 
-  async assertEmployeeAttendanceTimes(index: number, expected: string) {
-    await this.staffMemberPage.attendanceTimes
-      .nth(index)
-      .assertTextEquals(expected)
+  async assertEmployeeAttendances(expectedArray: string[]) {
+    const attendances = this.staffMemberPage.attendanceTimes
+    await attendances.assertCount(expectedArray.length)
+    return Promise.all(
+      expectedArray.map(async (expected, index) => {
+        const attendance = attendances.nth(index)
+        await attendance.assertTextEquals(expected)
+      })
+    )
   }
 
   async assertExternalStaffArrivalTime(expected: string) {
@@ -335,5 +346,48 @@ export class StaffAttendancePage {
       () => this.staffDeparturePage.departureTypeCheckbox(type).visible,
       visible
     )
+  }
+}
+
+export class StaffAttendanceEditPage {
+  private page: Page
+  addLink: Element
+  pinInput: TextInput
+  cancelButton: Element
+  saveButton: AsyncButton
+
+  constructor(page: Page) {
+    this.page = page
+    this.addLink = page.findByDataQa('add')
+    this.pinInput = new TextInput(page.findByDataQa('pin-input'))
+    this.cancelButton = page.findByDataQa('cancel')
+    this.saveButton = new AsyncButton(page.findByDataQa('save'))
+  }
+
+  async selectGroup(index: number, groupId: UUID) {
+    const select = new Select(this.page.findAllByDataQa('group').nth(index))
+    await select.selectOption({ value: groupId })
+  }
+
+  async selectType(index: number, type: StaffAttendanceType) {
+    const select = new Select(this.page.findAllByDataQa('type').nth(index))
+    await select.selectOption({ value: type })
+  }
+
+  async fillArrived(index: number, time: string) {
+    const field = new TextInput(this.page.findAllByDataQa('arrived').nth(index))
+    await field.fill(time)
+  }
+
+  async fillDeparted(index: number, time: string) {
+    const field = new TextInput(
+      this.page.findAllByDataQa('departed').nth(index)
+    )
+    await field.fill(time)
+  }
+
+  async remove(index: number) {
+    const icon = this.page.findAllByDataQa('remove').nth(index)
+    await icon.click()
   }
 }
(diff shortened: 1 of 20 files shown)


service/src/main/kotlin/fi/espoo/evaka/attendance/MobileRealtimeStaffAttendanceController.kt
```diff
@@ -193,6 +198,84 @@ class MobileRealtimeStaffAttendanceController(private val ac: AccessControl) {
...
+                            attendanceId = attendance.id,
+                            employeeId = body.employeeId,
+                            groupId = attendance.groupId,
+                            arrivalTime = attendance.arrived,
+                            departureTime = attendance.departed,
+                            occupancyCoefficient = occupancyCoefficients[attendance.groupId]
+                                    ?: occupancyCoefficientZero,
+                            type = attendance.type
+                        )
+                    }
```
akheron, line 272: It would be better to delete all entries that intersect with `body.date` and then insert fresh rows from `body.rows`. Otherwise the order of upserts will affect whether some changed are allowed.

For example, the initial rows can be:

| id | arrived | departed |
| --- | --- | --- |
| 1 |08:00 |12:00 |
| 2 | 14:00|18:00 |

The user submits:

| id | arrived | departed |
| --- | --- | --- |
| 1 |08:00 |15:00 |
| 2 | 16:00|18:00 |

If the upsert order is id 1, then id 2, the first upsert will conflict with the existing row 2 in the database.

frontend/src/e2e-test/pages/mobile/staff-page.ts
```diff
@@ -337,3 +348,46 @@ export class StaffAttendancePage {
...
+  cancelButton: Element
+  saveButton: AsyncButton
+
+  constructor(page: Page) {
+    this.page = page
+    this.addLink = page.findByDataQa('add')
+    this.pinInput = new TextInput(page.findByDataQa('pin-input'))
+    this.cancelButton = page.findByDataQa('cancel')
+    this.saveButton = new AsyncButton(page.findByDataQa('save'))
+  }
```
akheron, line 365: New property initialisation style 👍

frontend/src/e2e-test/specs/6_mobile/realtime-staff-attendances.spec.ts
```diff
@@ -664,3 +667,187 @@ describe('Realtime staff attendance page', () => {
...
+    await editPage.fillArrived(0, newArrivalTime)
+    await editPage.fillDeparted(0, newDepartureTime)
+    await editPage.pinInput.fill(pin)
+    await editPage.saveButton.click()
+
+    await staffAttendancePage.assertEmployeeStatus('Poissa')
+    await staffAttendancePage.assertEmployeeAttendances([
+      `Ylityö ${newArrivalTime}–${newDepartureTime}`
+    ])
+    // TODO: assert group
```
akheron, line 788: Is TODO here for a reason?

frontend/src/employee-mobile-frontend/staff-attendance/StaffAttendanceEditPage.tsx
```diff
@@ -0,0 +1,423 @@
...
+  const { groupRoute } = useSelectedGroup()
+  const { i18n, lang } = useTranslation()
+  const navigate = useNavigate()
+  const [date] = useState(LocalDate.todayInHelsinkiTz())
+  const form = useForm(
+    staffAttendancesForm,
+    () => initialFormState(i18n, date, groups, staffMember.attendances),
+    i18n.attendances.staff.validationErrors
+  )
+  const { pinCode, rows } = useFormFields(form)
```
akheron, line 223: Use `const boundRows = useFormElems(rows)` here to avoid `useFormElem` + if + throw in the `StaffAttendanceRowEditor` component. The prop's type will be `bind: BoundForm<typeof staffAttendanceRow>`.

frontend/src/employee-mobile-frontend/staff-attendance/StaffAttendanceEditPage.tsx
```diff
@@ -0,0 +1,423 @@
...
+    <FixedSpaceRow alignItems="baseline" flexWrap="wrap">
+      {featureFlags.staffAttendanceTypes && (
+        <div>
+          <SelectF bind={type} data-qa="type" />
+          {type.validationError() && (
+            <InputFieldUnderRow className={classNames('warning')}>
+              {type.inputInfo()?.text}
+              <StatusIcon status="warning" />
+            </InputFieldUnderRow>
+          )}
```
akheron, line 383: I wonder if it would be good to add validation error showing directly to `Select`/`SelectF`, because many other form components already can do it. It beats the idea of the form framework a bit if individual components don't display their errors and you have to remember to do it in the parent.

akheron:
I tested the functionality and have some more comments:

* As stated previously, the functionality has to work even if the municipality doesn't have staff attendance types enabled
* Frontend validation is lacking:
    * Overlapping times should be disallowed
    * End time is not allowed to be in the future (when editing today's entries)
* If an entry's start time is yesterday or before, the start time should not be editable
* The layout doesn't match the designs
    * Group selection should be like it's implemented on the desktop: group name is shown as text on its own line, and it changes to a dropdown if it's clicked
    * Attendance type dropdown is hidden if attendance types are not enabled
    * The pin code query should be in a different view that is shown when the save button is pressed
* The back button near the top of the page doesn't work optimally if you visit the editing view and save and visit it again multiple times.

frontend/src/employee-mobile-frontend/staff-attendance/StaffMemberPage.tsx
```diff
@@ -86,6 +90,19 @@ export default React.memo(function StaffMemberPage() {
                         <span>
                           {arrived.toLocalTime().format()}–
                           {departed?.toLocalTime().format() ?? ''}
+                          {featureFlags.employeeMobileStaffAttendanceEdit &&
```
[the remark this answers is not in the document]
patari: Fixed

service/src/main/kotlin/fi/espoo/evaka/attendance/MobileRealtimeStaffAttendanceController.kt
patari: Fixed

frontend/src/e2e-test/specs/6_mobile/realtime-staff-attendances.spec.ts
patari: Fixed

frontend/src/employee-mobile-frontend/staff-attendance/StaffAttendanceEditPage.tsx
patari: Fixed

frontend/src/employee-mobile-frontend/staff-attendance/StaffMemberPage.tsx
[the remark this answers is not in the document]
akheron: Now the button is visible if `staffAttendanceTypes === false && employeeMobileStaffAttendanceEdit === false`. You should add a check for the `employeeMobileStaffAttendanceEdit` feature flag in the second button.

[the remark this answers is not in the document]
patari: Fixed

akheron (approved):

PR #4347, opened 2023-10-30
Assistance info ACL changes

reynders:
#### Summary
-Instead of VEO seeing all assistance info, only show own units info
-Allow service worker to also read assistance factor

### DIFF
--- a/service/src/main/kotlin/fi/espoo/evaka/shared/security/Action.kt
+++ b/service/src/main/kotlin/fi/espoo/evaka/shared/security/Action.kt
@@ -696,7 +696,7 @@ sealed interface Action {
             HasUnitRole(SPECIAL_EDUCATION_TEACHER)
                 .inPlacementUnitOfChildOfAssistanceNeedDecision(false),
             HasGlobalRole(DIRECTOR).andAssistanceNeedDecisionHasBeenSent(),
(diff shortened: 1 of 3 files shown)


service/src/main/kotlin/fi/espoo/evaka/shared/security/Action.kt
```diff
@@ -675,37 +675,42 @@ sealed interface Action {
...
         ),
         READ(
             HasGlobalRole(ADMIN, SERVICE_WORKER),
-            HasUnitRole(SPECIAL_EDUCATION_TEACHER).inAnyUnit(),
+            HasUnitRole(SPECIAL_EDUCATION_TEACHER)
+                .inPlacementUnitOfChildOfAssistanceNeedDecision(false),
             HasGlobalRole(DIRECTOR).andAssistanceNeedDecisionHasBeenSent(),
-            HasUnitRole(UNIT_SUPERVISOR).inPlacementUnitOfChildOfAssistanceNeedDecision(true),
-            HasUnitRole(STAFF).inPlacementUnitOfChildOfAcceptedAssistanceNeedDecision()
+            HasUnitRole(UNIT_SUPERVISOR).inPlacementUnitOfChildOfAssistanceNeedDecision(false),
```
Gekkio, line 699: This probably should be `true`

reynders: That is true

Gekkio (approved):

PR #4376, opened 2023-11-06
Fix frontend PR conflict

Gekkio:


reynders (approved):

frontend/src/lib-customizations/defaults/employee-mobile-frontend/i18n/fi.ts
```diff
@@ -91,7 +91,8 @@ export const fi = {
     validity: 'Voimassaolo',
     validTo: (date: string) => `Voimassa ${date} saakka`,
     lastName: 'Sukunimi',
-    firstName: 'Etunimi'
+    firstName: 'Etunimi',
+    openExpandingInfo: 'Avaa lisätietokenttä'
```
akheron, line 95: The `openExpandingInfo` translation should not be needed here 🤔

PR #4377, opened 2023-11-06
assistance decision unit can be any non-club unit

Joosakur:


frontend/src/employee-frontend/components/child-information/queries.ts
```diff
@@ -80,7 +80,7 @@ export const queryKeys = createQueryKeys('childInformation', {
     'assistanceNeedPreschoolDecision',
     decisionId
   ],
-  preschoolUnits: () => ['preschoolUnits'],
+  units: () => ['preschoolUnits'],
```
akheron, line 83: The `preschoolUnits` query key should probably be changed too.

PR #4381, opened 2023-11-07
Disallow adding absence on days which already contain attendance records

Wnt:
## Before this change
It was possible to add an absence (e.g. sick leave) for a day which already contains attendance records. This may affect billing in a way that should not be allowed.
## After this change
Creating absence markings for a day with existing attendance record is no more possible via `POST /citizen/absences`. Instead the API will return a HTTP 400 - Invalid Request with error code `ATTENDANCE_ALREADY_EXISTS` and the UI will react to this error by displaying 
> There is already attendance record for the selected days
> It is not possible to add an absence for a child on a day that already has an attendance record.

image


Joosakur:
need to run ktfmt

Joosakur (approved):

PR #4389, opened 2023-11-10
Refactored income coefficient multipliers to be configurable through IncomeCoefficientMultiplierProvider

martinLeht:
#### Summary
Refactored income coefficient multipliers to be configurable through IncomeCoefficientMultiplierProvider


service/src/main/kotlin/fi/espoo/evaka/invoicing/service/IncomeCoefficientMultiplierProvider.kt
```diff
@@ -0,0 +1,26 @@
...
+package fi.espoo.evaka.invoicing.service
+
+import fi.espoo.evaka.invoicing.domain.IncomeCoefficient
+import java.math.BigDecimal
+
+fun interface IncomeCoefficientMultiplierProvider {
+    fun multiplier(coefficient: IncomeCoefficient): BigDecimal
+}
+
+class EspooIncomeCoefficientMultiplierProvider : IncomeCoefficientMultiplierProvider {
```
akheron, line 14: It would be more consistent to move this class under `src/main/kotlin/fi/espoo/evaka/espoo`, because we're slowly migrating all espoo-specific code and configuration there anyway.

martinLeht: Makes sense, I will move it there.

akheron (approved):

PR #4391, opened 2023-11-10
Add reset varda child button to varda errors report

reynders:


frontend/src/e2e-test/specs/5_employee/varda_error_report.spec.ts
```diff
@@ -95,5 +96,8 @@ describe('Varda error report', () => {
 
     await vardaErrorsReportPage.assertErrorRowCount(1)
     await vardaErrorsReportPage.assertErrorsContains(childId, 'test error')
+    await vardaErrorsReportPage.resetChild(childId)
```
Wnt, line 99: It would be cleaner to have to have a separate test for `resetChild` functionality. Or maybe the test needs to be renamed from `Varda errors are shown` into something like `Varda errors are shown and can be reset`

akheron (approved):

PR #4411, opened 2023-11-17
Don't calculate due date for paper transfer applications

akheron:

### DIFF
--- a/service/src/main/kotlin/fi/espoo/evaka/application/ApplicationStateService.kt
+++ b/service/src/main/kotlin/fi/espoo/evaka/application/ApplicationStateService.kt
@@ -936,12 +936,17 @@ class ApplicationStateService(
         setCheckedByAdminToDefault(original.id, updatedForm)
         when (manuallySetDueDate) {
             null ->
-                calculateAndUpdateDueDate(
-                    now.toLocalDate(),
-                    original,
-                    updatedForm.preferences.preferredStartDate,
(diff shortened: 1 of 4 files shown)


service/src/main/kotlin/fi/espoo/evaka/application/ApplicationStateService.kt
```diff
@@ -940,12 +936,17 @@ class ApplicationStateService(
...
-                calculateAndUpdateDueDate(
-                    today,
-                    original,
-                    updatedForm.preferences.preferredStartDate,
-                    updatedForm.preferences.urgent
-                )
+                // We don't want to calculate the due date for applications in the CREATED state.
+                // Whether this is a transfer application affects the calculation of the due date,
+                // but it's only determined when the application is sent.
+                if (original.status == SENT) {
```
Joosakur, line 942: Should this be `!= CREATED`? What if service worker updates it e.g. in WAITING_FOR_PLACEMENT status, should the due date be updated as with old logic?

akheron: There's a check on line 924 that the status is one of CREATED, SENT.

Joosakur: ah 👍

Joosakur (approved):

PR #4427, opened 2023-11-21
Support updating children from VTJ

akheron:
#### Summary

The "update from VTJ" button in the frontend can now update children in addition to guardians.


reynders (approved):
Mob review

PR #4428, opened 2023-11-21
Send outdated income notifications to fridge parents

reynders:
Send outdated income notifications to fridge parents instead of guardians.

### DIFF
--- a/service/src/main/kotlin/fi/espoo/evaka/invoicing/service/IncomeQueries.kt
+++ b/service/src/main/kotlin/fi/espoo/evaka/invoicing/service/IncomeQueries.kt
@@ -34,14 +34,14 @@ enum class IncomeNotificationType {
     EXPIRED_EMAIL
 }
 
-data class GuardianIncomeExpirationDate(val guardianId: PersonId, val expirationDate: LocalDate)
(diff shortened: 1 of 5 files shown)


service/src/main/kotlin/fi/espoo/evaka/invoicing/service/IncomeQueries.kt
```diff
@@ -62,12 +62,25 @@ WITH latest_income AS (
     FROM placement pl
     JOIN service_need sn ON pl.id = sn.placement_id AND daterange(sn.start_date, sn.end_date, '[]') @> :dayAfterExpiration
     JOIN service_need_option sno ON sn.option_id = sno.id AND sno.fee_coefficient > 0
-    JOIN guardian g ON g.child_id = pl.child_id
-    JOIN latest_income i ON i.person_id = g.guardian_id
+    
+    -- head of child
+    LEFT JOIN fridge_child fc_head ON (
```
akheron, line 67: This should be a `JOIN`, not `LEFT JOIN`. If the child doesn't have a head of child, then the query shouldn't return anything.

akheron (approved):

PR #4434, opened 2023-11-22
Hide placement plan date from citizen

akheron:
#### Summary

Refactor all citizen related queries to use the `application` table directly instead of `application_view`. Then fix the start date shown to citizen to never use placement plan's start date.

It was not enough to just remove the placement plan's date from the query, because the decision is created at the same time as the placement plan, so the application's state also needs to be checked to decide which date to use.

While at it, change on action rule to also query `application` directly.


patari:
> the decision is created at the same time as the placement plan

TIL

PR #4439, opened 2023-11-24
Show metadata about partners in timeline

Wnt:
## Before this change
It was hard for people working in billing to figure out where information about partnerships originated from and when the partnerships were added to the system.
## After this change
Any modifications to partnership linking table `fridge_partner` also store creation source (user created / created from application) and modification source (user edited / DVV sync). This data is the displayed in the timeline component for each partner like so:

image
In case a partnership was created because of an application was processed, there is also a link to the application.


service/src/main/kotlin/fi/espoo/evaka/pis/PartnershipQueries.kt
```diff
@@ -116,51 +141,61 @@ fun Database.Transaction.createPartnership(
...
 }
 
-fun Database.Transaction.retryPartnership(id: PartnershipId) {
-    // language=SQL
-    val sql = "UPDATE fridge_partner SET conflict = false WHERE partnership_id = :id"
-
-    createUpdate(sql).bind("id", id).execute()
+fun Database.Transaction.retryPartnership(
+    id: PartnershipId,
+    modifiedById: EvakaUserId?,
```
akheron, line 185: `modifiedById` is always given and shouldn't be nullable to avoid accidental mistakes in the calling side.

service/src/main/kotlin/fi/espoo/evaka/pis/service/PartnershipService.kt
```diff
@@ -38,21 +53,36 @@ class PartnershipService {
...
         } catch (e: Exception) {
             throw mapPSQLException(e)
         }
     }
 
-    fun retryPartnership(tx: Database.Transaction, partnershipId: PartnershipId): Partnership? {
+    fun retryPartnership(
+        tx: Database.Transaction,
+        partnershipId: PartnershipId,
+        modifiedById: EvakaUserId?,
```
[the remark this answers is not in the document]
akheron: Changing some of these may affect the tests, but ensuring the actual application code correctness is IMO more important than the ease of writing tests.

service/src/main/resources/db/migration/V371__partner_modifier_column.sql
```diff
@@ -0,0 +1,15 @@
+CREATE TYPE create_type AS ENUM ('USER', 'APPLICATION');
+CREATE TYPE modify_type AS ENUM ('USER', 'DVV');
+
+ALTER TABLE fridge_partner
+    ADD COLUMN create_type create_type default null,
+    ADD COLUMN created_at timestamp with time zone default null,
```
akheron, line 6: It's better to rename the existing `created` column to `create_at` and drop the default value.

Wnt: I think we should also consider renaming `updated` -> `modified_at` or copying values from updated to modified_at.

PR #4440, opened 2023-11-24
Unit confirmed day reservations for employee mobile

tmkrepo:
A new companion view for the employee mobile attendance list that fetches unit's reservation statistics and child reservation statuses for the period that is non-reservable. 

- new endpoints
   - fetch and calculate unit's reservation statistics for confirmed days per group
   - fetch unit's child reservation statuses for the given examination day
- frontend:
  - splits employee mobile attendance list to allow perusal of reservation statistics for the coming days
    - same group selection used for both views
  - a statistics accordion view than enables user to see daily head counts and calculated occupancy as well as open detailed child reservation status listings for each day

NB! Despite earlier performance concerns, the 'all' group selection is available for the reservation tab as well. Removing it would have introduced more usability issues than it would have alleviated.


service/src/main/kotlin/fi/espoo/evaka/shared/security/Action.kt
```diff
@@ -1944,6 +1944,28 @@ sealed interface Action {
...
+            HasGlobalRole(ADMIN, SERVICE_WORKER),
+            HasUnitRole(
+                    UNIT_SUPERVISOR,
+                    STAFF,
+                    SPECIAL_EDUCATION_TEACHER,
+                    EARLY_CHILDHOOD_EDUCATION_SECRETARY
+                )
+                .inUnit(),
+            IsMobile(requirePinLogin = false).inUnit()
+        ),
```
akheron, line 1968: Is it necessary to add these permissions employee to roles if the enptoints are only used in mobile?

service/src/main/kotlin/fi/espoo/evaka/reservations/ReservationQueries.kt
```diff
@@ -542,3 +546,224 @@ fun Database.Read.getReservationContractDayRanges(
...
+                       AND bc.unit_id <> :unitId
+                       AND
+                      daterange(bc.start_date, bc.end_date, '[]') @>
+                      :examinationDate
+         LEFT JOIN attendance_reservation ar
+                   ON ar.child_id = pcd.child_id
+                       AND ar.date = :examinationDate
+         LEFT JOIN daily_service_time dst
+                   ON pcd.child_id = dst.child_id
+                       AND dst.validity_period @> :examinationDate
```
akheron, line 645: I think this query would be easier to understand if it used the `realized_placement_one` function which already contains the logic that's in `placement_days` CTE in this query.

I'm also pretty sure that all the reservation and absence edge cases cannot be handled in a single SQL query and will need to be done in the Kotlin side instead. The edge cases have been covered in both citizen calendar and unit's week calendar.

service/src/main/kotlin/fi/espoo/evaka/reservations/ReservationQueries.kt
```diff
@@ -542,3 +546,224 @@ fun Database.Read.getReservationContractDayRanges(
...
+                       bc.child_id,
+                       bc.start_date as placement_start,
+                       bc.end_date   as placement_end,
+                       bc.start_date as grouping_start,
+                       bc.end_date   as grouping_end,
+                       bc.group_id   as group_id
+                FROM (SELECT t.date
+                      FROM unnest(:confirmedDays) AS t(date)) day
+                         LEFT JOIN backup_care bc ON daterange(bc.start_date, bc.end_date, '[]') @> day.date
+                    AND bc.unit_id = :unitId)
```
akheron, line 704: This part could probably also use the `realized_placement_one` function. It it cannot be used because of performance problems, this query should at least take some ideas from there, because in the current form it's quite unreadable.

service/src/main/kotlin/fi/espoo/evaka/reservations/ReservationQueries.kt
```diff
@@ -542,3 +546,224 @@ fun Database.Read.getReservationContractDayRanges(
...
+                             AND
+                            sno_default.valid_placement_type =
+                            pl.type
+               LEFT JOIN person p
+                         ON pcd.child_id = p.id
+               LEFT JOIN assistance_factor af
+                         ON pcd.child_id = af.child_id
+                             AND
+                            af.valid_during @> pcd.date) a
+group by a.date, a.group_id
```
akheron, line 761: Overall, the totally different code style in this query compared to other queries in the code base hurts the readability a lot.

service/src/main/kotlin/fi/espoo/evaka/reservations/AttendanceReservationController.kt
```diff
@@ -351,6 +353,245 @@ class AttendanceReservationController(
...
+                                    clock.now(),
+                                    featureConfig.citizenReservationThresholdHours,
+                                )
+                                .minusDays(1)
+                        )
+
+                    val nextConfirmedUnitDays =
+                        nonReservableRange
+                            .dates()
+                            .filter { unitData.operationDays.contains(it.dayOfWeek.value) }
```
akheron, line 560: This is probably where holidays should be taken into account.

service/src/main/kotlin/fi/espoo/evaka/reservations/AttendanceReservationController.kt
```diff
@@ -351,6 +353,245 @@ class AttendanceReservationController(
...
+                                            else Reservation.NoTimes
+                                        }
+
+                                ChildReservationInfo(
+                                    reservations = res,
+                                    absent = basicInfoItem.absenceType != null,
+                                    groupId = basicInfoItem.groupId,
+                                    childId = basicInfoItem.childId,
+                                    outOnBackupPlacement = basicInfoItem.backupUnitId != null,
+                                    dailyServiceTimes =
```
akheron, line 456: You should at least use a separate `tx.getDailyServiceTimesForChildren()` function call to get daily service times and not have the 50 lines of logic duplicated here.

service/src/main/kotlin/fi/espoo/evaka/shared/security/Action.kt
tmkrepo: Desktop employee permissions removed.

service/src/main/kotlin/fi/espoo/evaka/reservations/AttendanceReservationController.kt
tmkrepo: Holidays removed from date selection using `getUnitOperationalDayData` from desktop calendar code. As I understand holiday periods should not affect the day selection?

service/src/main/kotlin/fi/espoo/evaka/reservations/AttendanceReservationController.kt
tmkrepo: Daily service time logic separated from the main query as proposed.

service/src/main/kotlin/fi/espoo/evaka/reservations/ReservationQueries.kt
tmkrepo: I refactored the query to utilize `realized_placement_one` and it simplified the query strucure significantly, although it also introduced new feature specific incoming/outgoing backup care deduction logic. The addition of term breaks also added some logic (and duplicated logic at that), but the end result is still more readable than the original.

I tried to figure a way of doing the statistics gathering on Kotlin side but apart from the duplicated logic it seemed even more complicated than the group by query.

service/src/main/kotlin/fi/espoo/evaka/reservations/ReservationQueries.kt
tmkrepo: Query refactored around suggested db function.

service/src/main/kotlin/fi/espoo/evaka/reservations/ReservationQueries.kt
tmkrepo: Query refactored, but I wasn't really able to figure out a Kotlin side solution that was "better" than the group by summation. Admittedly the deduction logic is perilously complicated but I tried to address this in integration testing.

service/src/main/kotlin/fi/espoo/evaka/reservations/ReservationQueries.kt
```diff
@@ -543,3 +546,138 @@ fun Database.Read.getReservationContractDayRanges(
...
+)
+
+fun Database.Read.getReservationStatisticsForUnit(
+    confirmedDays: List<LocalDate>,
+    unitId: DaycareId
+): Map<LocalDate, List<GroupReservationStatisticsRow>> {
+    return createQuery(
+            """
+select date,
+       count(1) FILTER ( WHERE NOT a.child_in_unit ) AS absent,
```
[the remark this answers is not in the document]
tmkrepo: Fixed and added integration test data

akheron (approved):

PR #4491, opened 2023-12-07
Cancel all active transfer applications on placement termination

reynders:


service/src/main/kotlin/fi/espoo/evaka/placement/PlacementControllerCitizen.kt
```diff
@@ -146,10 +146,7 @@ class PlacementControllerCitizen(
                         }
 
                         val cancelableTransferApplicationIds =
-                            tx.cancelAllActiveTransferApplicationsAfterDate(
```
akheron, line 149: Should this function be deleted too if it's no longer used?

reynders: Actually it was renamed and refactored, but might be hard to see from this diff

akheron (approved):

PR #4507, opened 2023-12-13
Add initial calendar event time reservation backend

patari:


service/src/main/resources/db/migration/V373__calendar_event_time_reservation.sql
```diff
@@ -0,0 +1,45 @@
+CREATE TABLE calendar_event_time
+(
+    id                uuid PRIMARY KEY                  DEFAULT ext.uuid_generate_v1mc(),
+    created           timestamp with time zone NOT NULL DEFAULT now(),
```
Wnt, line 4: I think this should be rename to `created_at`. AFAIK these new updated & created timestamp field names were agreed during some tech discussions:
image

service/src/main/resources/db/migration/V373__calendar_event_time_reservation.sql
```diff
@@ -0,0 +1,45 @@
...
+    start_time        time without time zone   NOT NULL,
+    end_time          time without time zone   NOT NULL,
+    child_id          uuid REFERENCES child (id)
+);
+
+CREATE TRIGGER set_timestamp
+    BEFORE UPDATE
+    ON calendar_event_time
+    FOR EACH ROW
+EXECUTE PROCEDURE trigger_refresh_updated();
```
Wnt, line 19: If you want to comply with the new naming pattern, I think you need to create a new procedure that updates `updated_at` column

service/src/main/resources/db/migration/V373__calendar_event_time_reservation.sql
```diff
@@ -0,0 +1,45 @@
...
+    id                uuid PRIMARY KEY                  DEFAULT ext.uuid_generate_v1mc(),
+    created           timestamp with time zone NOT NULL DEFAULT now(),
+    updated           timestamp with time zone NOT NULL DEFAULT now(),
+    modified_at       timestamp with time zone NOT NULL,
+    modified_by       uuid                     NOT NULL REFERENCES evaka_user (id),
+    calendar_event_id uuid                     NOT NULL REFERENCES calendar_event (id) ON DELETE CASCADE,
+    date              date                     NOT NULL,
+    start_time        time without time zone   NOT NULL,
+    end_time          time without time zone   NOT NULL,
+    child_id          uuid REFERENCES child (id)
```
Wnt, line 12: Should we limit on DB level that there can only be one unique `calendar_event_id` + `child_id` combination?

service/src/main/resources/db/migration/V373__calendar_event_time_reservation.sql
patari: Fixed

service/src/main/resources/db/migration/V373__calendar_event_time_reservation.sql
patari: No, there should be possibility to add multiple time reservations for child, e.g. one for each guardian.

service/src/main/resources/db/migration/V373__calendar_event_time_reservation.sql
```diff
@@ -0,0 +1,57 @@
...
+);
+
+CREATE OR REPLACE FUNCTION public.trigger_refresh_updated_at() RETURNS trigger
+    LANGUAGE plpgsql
+AS $$
+BEGIN
+    IF NEW.updated_at = OLD.updated_at THEN
+        NEW.updated_at = NOW();
+    END IF;
```
akheron, line 22: One more small thing: This function should always set `updated_at` and not allow it to be updated from the "outside". That's the whole point of `updated_at`.

akheron (approved):

service/src/main/resources/db/migration/V373__calendar_event_time_reservation.sql
akheron: It seems that `trigger_refresh_updated` is in `R__triggers.sql` so the new function should also be added there I think.

patari: > One more small thing: This function should always set `updated_at` and not allow it to be updated from the "outside". That's the whole point of `updated_at`.

Fixed.



> It seems that `trigger_refresh_updated` is in `R__triggers.sql` so the new function should also be added there I think.

Not possible, because repeatable migrations are executed after all pending migrations (`trigger_refresh_updated` is also at V1.sql).

akheron: Ah, true.

PR #4538, opened 2023-12-20
Cache JWT token to session

akheron:
#### Summary

A significant share of apigw's processing time seems to be used for computing JWT tokens. To fix this, compute JWT token once every 10 minutes and cache it with the session.

Alternative approach: #4549


akheron:
Superseded by https://github.com/espoon-voltti/evaka/pull/4611

PR #4541, opened 2023-12-20
Reoccurring absences

Wnt:
## Before this change
It was only possible to create re-occurring attendance reservations from employee's week calendar view.
## After this change
It's also possible to create re-occurring absence records.

image

Daily repetition

image

Weekly repeating reservations

image

Irregular reservations

If the service need is of type "contract days", the absence records are created as `PLANNED_ABSENCE` for days which aren't yet locked. Otherwise the absences are created as `OTHER_ABSENCE`


frontend/src/e2e-test/specs/5_employee/reservation-calendar.spec.ts
```diff
@@ -498,6 +498,24 @@ describe('Unit group calendar', () => {
...
+    await calendarPage.selectMode('week')
+    reservationModal = await childReservations.openReservationModal(
+      child1Fixture.id
+    )
+    await reservationModal.addAbsence(mockedToday)
+    await waitUntilEqual(
+      () =>
+        childReservations
+          .childRows(child1Fixture.id)
+          .findAllByDataQa('absence')
```
akheron, line 514: `data-qa` attributes are such a low-level detail that in my opinion they should only be used in the page model and not directly in tests.

akheron (approved):
Otherwise looks good 🏅

PR #4549, opened 2023-12-22
Cache JWT tokens in memory

akheron:
#### Summary

Cache JWT tokens in apigw's memory. Each apigw has their own cache.

Alternative approach: #4538

### DIFF
--- a/apigw/src/shared/auth/jwt.ts
+++ b/apigw/src/shared/auth/jwt.ts
@@ -2,13 +2,14 @@
 //
 // SPDX-License-Identifier: LGPL-2.1-or-later
 
+import tracer from '../../tracer.js'
 import jwt from 'jsonwebtoken'
 import { readFileSync } from 'node:fs'
 import { jwtKid, jwtPrivateKey } from '../config.js'
 
 const privateKey = readFileSync(jwtPrivateKey)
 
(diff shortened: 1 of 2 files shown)


apigw/src/shared/auth/jwt.ts
```diff
@@ -19,10 +20,65 @@ export function createJwt(payload: {
...
+      span?.addTags({ cacheHit: true, cacheSize: cache.size })
+      return cached.token
+    } else {
+      const token = jwt.sign(payload, privateKey, {
+        algorithm: 'RS256',
+        expiresIn: '48h',
+        keyid: jwtKid
+      })
+      cache.set(key, { token, createdAt: now })
+      span?.addTags({ cacheHit: false, cacheSize: cache.size })
```
Wnt, line 40: Alternative way to `CACHE_CLEANUP_INTERVAL` would be a create a timer for each cache entry like:
```
    // Set a timer to remove the cache entry after JWT_LIFETIME
    setTimeout(() => {
        cache.delete(key);
    }, JWT_LIFETIME);
```
this would reduce the code complexity and avoid the need to iterate trough all cache entries.

akheron:
Superseded by #4611

PR #4602, opened 2024-01-09
Document template import/export

Gekkio:


Joosakur (approved):

frontend/src/employee-frontend/components/document-templates/template-editor/DocumentTemplatesPage.tsx
```diff
@@ -95,10 +107,15 @@ const TemplateRow = React.memo(function TemplateRow({
   const { mutateAsync: deleteDocumentTemplate } = useMutationResult(
     deleteDocumentTemplateMutation
   )
+  const onClickExport = useCallback(() => {
+    window.open(
+      exportDocumentTemplateUrl(template.id, sanitizedFilename(template.name))
```
Wnt, line 112: I think it would be better to determine the filename on server-side instead of replying with the filename given in request path

frontend/src/employee-frontend/components/document-templates/template-editor/DocumentTemplatesPage.tsx
```diff
@@ -139,20 +156,31 @@ const TemplateRow = React.memo(function TemplateRow({
             disabled={template.published}
             onClick={() => deleteDocumentTemplate(template.id)}
           />
+          <IconButton
```
Wnt, line 159: I would prefer presenting the download as a link with `target="_blank"`. This would allow the user to use context menu to "Save target as".

PR #4607, opened 2024-01-11
configurable preschool and preparatory times by unit

Joosakur:
#### Summary

Added these time ranges to data model and editable through unit details. The data is not used anywhere yet.


frontend/src/e2e-test/dev-api/fixtures.ts
```diff
@@ -323,6 +323,8 @@ export const clubFixture: Daycare = {
   areaId: careAreaFixture.id,
   name: 'Alkuräjähdyksen kerho',
   type: ['CLUB'],
+  dailyPreschoolTime: null,
+  dailyPreparatoryTime: null,
```
Joosakur, line 327: Not sure if the db constraint should be loosened to always allow non-null values so we wouldn't need this kind of boilerplate to override the defaults whenever overriding the type field?

frontend/src/employee-frontend/components/unit/unit-details/UnitEditor.tsx
```diff
@@ -205,12 +208,16 @@ const emptyTimeRange: JsonOf<TimeRange> = {
...
 const emptyOperationWeek = [null, null, null, null, null, null, null]
 
-type EditableTimeRange = JsonOf<TimeRange> | null
+type EditableTimeRange = JsonOf<TimeRange>
 
 type FormErrorItem = { key: string; text: string }
 
 type UnitEditorErrors = {
-  rangeErrors: RangeValidationResult[]
+  rangeErrors: {
```
Joosakur, line 216: this whole file is starting to be a bit of a mess, could be refactored to use form framework at some point, but maybe not right now

akheron (approved):
The migrations could be combined as one to reduce the amount of "useless" migrations.

PR #4610, opened 2024-01-12
Absence calculation

Joosakur:
#### Summary

Logic documented here https://github.com/espoon-voltti/evaka/wiki/L%C3%A4sn%C3%A4%E2%80%90-ja-poissaolot#poissaolojen-syntyminen-p%C3%A4iv%C3%A4n-l%C3%A4sn%C3%A4oloajasta

### DIFF
--- a/frontend/src/employee-frontend/components/unit/unit-reservations/ChildDateModal.tsx
+++ b/frontend/src/employee-frontend/components/unit/unit-reservations/ChildDateModal.tsx
@@ -372,10 +372,11 @@ export default React.memo(function ChildDateModal({
         {expectedAbsencesEnabled &&
           expectedAbsences.isSuccess &&
           expectedAbsences.value !== null && (
-            <>
+            <FixedSpaceColumn data-qa="absence-warnings">
               {expectedAbsences.value.includes('NONBILLABLE') &&
                 nonBillableAbsence.value() === undefined && (
                   <AbsenceWarning
+                    data-qa="missing-nonbillable-absence"
                     message={
                       i18n.unit.attendanceReservations.childDateModal
                         .missingNonbillableAbsence
@@ -385,6 +386,7 @@ export default React.memo(function ChildDateModal({
               {!expectedAbsences.value.includes('NONBILLABLE') &&
                 nonBillableAbsence.value() !== undefined && (
                   <AbsenceWarning
+                    data-qa="extra-nonbillable-absence"
                     message={
                       i18n.unit.attendanceReservations.childDateModal
                         .extraNonbillableAbsence
@@ -394,6 +396,7 @@ export default React.memo(function ChildDateModal({
               {expectedAbsences.value.includes('BILLABLE') &&
                 billableAbsence.value() === undefined && (
                   <AbsenceWarning
+                    data-qa="missing-billable-absence"
                     message={
                       i18n.unit.attendanceReservations.childDateModal
                         .missingBillableAbsence
@@ -403,13 +406,14 @@ export default React.memo(function ChildDateModal({
               {!expectedAbsences.value.includes('BILLABLE') &&
                 billableAbsence.value() !== undefined && (
                   <AbsenceWarning
+                    data-qa="extra-billable-absence"
                     message={
                       i18n.unit.attendanceReservations.childDateModal
                         .extraBillableAbsence
                     }
                   />
                 )}
-            </>
+            </FixedSpaceColumn>
           )}
       </FixedSpaceColumn>
     </MutateFormModal>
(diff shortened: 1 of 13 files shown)


service/src/main/kotlin/fi/espoo/evaka/attendance/ChildAttendanceQueries.kt
```diff
@@ -73,6 +73,27 @@ fun Database.Read.getChildAttendance(
...
+        WHERE child_id = :childId AND unit_id = :unitId AND date = :date
+        """
+            .trimIndent()
+
+    return createQuery(sql)
+        .bind("childId", childId)
+        .bind("unitId", unitId)
+        .bind("date", date)
+        .toList<ChildAttendance>()
+}
```
Wnt, line 95: Why aren't you using the new style binder here?
e.g.
```kotlin
    return createQuery<ChildAttendance> {
            sql(
                """
        SELECT id, child_id, unit_id, (date + start_time) AT TIME ZONE 'Europe/Helsinki' AS arrived, (date + end_time) AT TIME ZONE 'Europe/Helsinki' AS departed
        FROM child_attendance
        WHERE child_id = ${bind(childId)} AND unit_id = ${bind(unitId)} AND date = ${bind(date)}
        """
            )
        }
        .toList<ChildAttendance>()
}
```

I think it would be a more readable + less code

Joosakur: Hmm yeah, copy pasted this from above, I'll change it 👍

Joosakur: query fully removed for now

frontend/src/employee-frontend/components/unit/unit-reservations/ChildDateModal.tsx
```diff
@@ -207,6 +213,24 @@ export default React.memo(function ChildDateModal({
...
+  const debouncedAbsencesRequest = useDebounce(expectedAbsencesRequest, 1500)
+  const expectedAbsencesEnabled =
+    date.isBefore(LocalDate.todayInHelsinkiTz()) &&
+    attendances.isValid() &&
+    !attendances.value().some((attendance) => attendance.endTime === null)
+  const expectedAbsences = useQueryResult(
+    childDateExpectedAbsencesQuery(debouncedAbsencesRequest),
+    { enabled: expectedAbsencesEnabled }
+  )
+
```
Joosakur, line 233: This still has UX issues when playing around with values, like old cached warnings flashing after entering a new valid value.

frontend/src/employee-frontend/components/unit/unit-reservations/ChildDateModal.tsx
```diff
@@ -207,6 +212,25 @@ export default React.memo(function ChildDateModal({
...
+    date.isBefore(LocalDate.todayInHelsinkiTz()) &&
+    debouncedAttendances !== null
+  const expectedAbsences = useQueryResult(
+    childDateExpectedAbsencesQuery({
+      childId: child.id,
+      date,
+      attendances: debouncedAttendances ?? []
+    }),
+    { enabled: expectedAbsencesEnabled }
+  )
```
akheron, line 232: This would enable you to use `renderResult`, because `expectedAbsences` would resolve to a `Success` of `null` when backend cannot be requested.

```suggestion
  const expectedAbsences = useQueryResult(
    queryOrDefault(
      (attendances: OpenTimeRange[]) =>
        childDateExpectedAbsencesQuery({
          childId: child.id,
          date,
          attendances
        }),
      null
    )(
      date.isBefore(LocalDate.todayInHelsinkiTz()) ? debouncedAttendances : null
    )
  )
```

akheron (approved):
One suggestion for the frontend, otherwise LGTM 🏅

frontend/src/employee-frontend/components/unit/unit-reservations/ChildDateModal.tsx
Joosakur: ah! 💡  I was longing for something like queryOrDefault, good to know it exists 😅

Joosakur: Refactored like this https://github.com/espoon-voltti/evaka/pull/4610/commits/b2362a8f6f459c6fb60b0ae1d271d168c22dced3

I think `renderResult` is not necessarily useful here. The warnings are only soft info, so the user does not care that much whether it is still loading or even failed, and so the spinner would mostly just increase blinking without much value.

PR #4611, opened 2024-01-12
JWT refactor, part 1

akheron:
#### Summary

Send the same user fields in HTTP headers that are already sent in JWT. Read the fields from headers, but fall back to JWT if the headers don't exist. If the headers exist, JWT payload can be empty, but the JWT needs to be valid in either case.

In integration tests, use empty cached JWT and send user fields in headers only.

Also, stop creating sessions where user's type is `ENDUSER`. Always use `CITIZEN_STRONG` instead.

After this has been deployed to production, the fallback of reading user fields from JWT can be removed and apigw can start sending an empty JWT which can be cached for a longer time. The support for for `ENDUSER` user type can also be removed.

### DIFF
--- a/apigw/src/enduser/dev-sfi-auth.ts
+++ b/apigw/src/enduser/dev-sfi-auth.ts
@@ -64,7 +64,7 @@ export function createDevSfiRouter(sessions: Sessions): Router {
       })
       return {
         id: person.id,
-        userType: 'ENDUSER',
+        userType: 'CITIZEN_STRONG',
         globalRoles: [],
         allScopedRoles: []
       }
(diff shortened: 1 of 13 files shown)


akheron (commented):
- Encode the user data in `AuthenticatedUser`'s JSON format to a single header, e.g. `X-EvakaUser`
- Check if the header needs to be base64 encoded
- Parsing the user data in service become a simple json deserialize call
- Clear the header in nginx config to prevent possible security issues

Gekkio (commented):
Looks good :+1: 
It's just missing `X-User` blacklisting in nginx proxy config (+ maybe in apigw `createProxy` as well, where we currently accept *all* headers from the client)

Gekkio (approved):

PR #4640, opened 2024-01-18
new absence calculation on child departure

Joosakur:


service/src/main/kotlin/fi/espoo/evaka/attendance/ChildAttendanceController.kt
```diff
@@ -276,23 +265,27 @@ class ChildAttendanceController(
...
+                        tx.getChildOngoingAttendance(childId, unitId)
+                            ?: throw BadRequest("Cannot depart, has not yet arrived")
+                    val attendanceTimesToday =
+                        tx.getCompletedChildAttendanceTimes(childId, unitId, today) +
+                            ongoingAttendance.toTimeRange(HelsinkiDateTime.of(today, body.departed))
+                    getExpectedAbsenceCategories(
+                        tx = tx,
+                        date = today,
+                        childId = childId,
+                        presenceTimes = attendanceTimesToday
```
akheron, line 278: Not probably a problem for this PR specially, but it seems that we now have two terms that mean the same thing: attendance and presence.

frontend/src/e2e-test/specs/6_mobile/child-attendances.spec.ts
```diff
@@ -509,24 +501,14 @@ describe('Child mobile attendance list', () => {
...
-          LocalDate.of(2022, 1, 20)
-        ),
-        termBreaks: [new FiniteDateRange(today, today)]
-      })
-      .save()
+    // change mocked now to be during term break
+    await page.close()
+    now = HelsinkiDateTime.of(2024, 1, 1, 13, 0, 0)
+    today = now.toLocalDate()
+    page = await Page.open({ mockedTime: now })
```
akheron, line 508: It might be a bit tedious to do, but it would be much better to refactor all tests in this suite to open the page themselves instead of having a few tests close and reopen the page with different parameters.

akheron (approved):

PR #4688, opened 2024-01-30
Add placement types for connected daycare

akheron:
#### Summary

Add two new placement types: `PRESCHOOL_DAYCARE_ONLY`, `PREPARATORY_DAYCARE_ONLY`.


service/src/main/kotlin/fi/espoo/evaka/reservations/AbsenceCalculator.kt
```diff
@@ -153,6 +153,8 @@ fun getExpectedAbsenceCategories(
             }
         PlacementType.DAYCARE,
         PlacementType.DAYCARE_PART_TIME,
+        PlacementType.PRESCHOOL_DAYCARE_ONLY,
```
Joosakur, line 156: This is probably good enough in practice, but if the attendance is during preschool/preparatory time then maybe it shouldn't count? It might be more correct to duplicate te preschool/preparatory branches for these and just omit the nonbillable absence. There the logic is also slightly different in that it allows 15min buffer separately before and after preschool, so attendance of 8:50-13:10 causes billable absence, while in this version it won't.

Anyway, should also be documented in https://github.com/espoon-voltti/evaka/wiki/L%C3%A4sn%C3%A4%E2%80%90-ja-poissaolot#poissaolojen-syntyminen-p%C3%A4iv%C3%A4n-l%C3%A4sn%C3%A4oloajasta

patari (approved):

patari (approved):

Joosakur (approved):

service/src/main/kotlin/fi/espoo/evaka/reservations/AbsenceCalculator.kt
Joosakur: looks good to me!

akheron: Reverted to the old way after discussion

PR #4695, opened 2024-02-01
Replace remaining employee_child_daycare_acl usage in rules

Gekkio:
#### Summary

Replace employee_child_daycare_acl usage with `ruleViaChildAcl`, or manual `cfg.aclQueries` usage.

This *very likely* affects the performance of all changed queries, hopefully positively because `ruleViaChildAcl` tends to give better query plans than using `employee_child_daycare_acl`


akheron (approved):
🏅

PR #4697, opened 2024-02-02
Make it possible to delete draft payments

vesahonkasaari:
Make it possible to delete individual draft payments in the payment UI

### DIFF
--- a/frontend/src/employee-frontend/components/payments/Actions.tsx
+++ b/frontend/src/employee-frontend/components/payments/Actions.tsx
@@ -6,11 +6,13 @@ import React from 'react'
 import styled from 'styled-components'
 
 import { PaymentStatus } from 'lib-common/generated/api-types/invoicing'
+import AsyncButton from 'lib-components/atoms/buttons/AsyncButton'
 import Button from 'lib-components/atoms/buttons/Button'
 import { fontWeights } from 'lib-components/typography'
 import { Gap } from 'lib-components/white-space'
 import colors from 'lib-customizations/common'
 
+import { deletePayments } from '../../api/invoicing'
 import { useTranslation } from '../../state/i18n'
 import StickyActionBar from '../common/StickyActionBar'
 
@@ -24,12 +26,14 @@ const CheckedRowsInfo = styled.div`
 
 type Props = {
   actions: PaymentsActions
+  reloadPayments: () => void
   status: PaymentStatus
   checkedPayments: Record<string, true>
 }
 
 const Actions = React.memo(function Actions({
   actions,
+  reloadPayments,
   status,
   checkedPayments
 }: Props) {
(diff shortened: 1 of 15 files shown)


service/src/main/kotlin/fi/espoo/evaka/invoicing/data/PaymentQueries.kt
```diff
@@ -130,6 +131,17 @@ fun Database.Read.getMaxPaymentNumber(): Long {
     return createQuery("SELECT max(number) FROM payment").exactlyOneOrNull<Long>() ?: 0
 }
 
+fun Database.Transaction.deleteDraftPayments(draftIds: List<PaymentId>) {
+    if (draftIds.isEmpty()) return
+
+    createUpdate(
+            "DELETE FROM payment WHERE status = :status::payment_status AND id = ANY(:draftIds)"
+        )
+        .bind("status", PaymentStatus.DRAFT.toString())
```
akheron, line 140: `.toString()` shouldn't be needed here. It would also be better to use the `createUpdate<Any> { sql("""...""") }` syntax in new code.

service/src/integrationTest/kotlin/fi/espoo/evaka/invoicing/PaymentsIntegrationTest.kt
```diff
@@ -366,6 +396,21 @@ class PaymentsIntegrationTest : FullApplicationTest(resetDbBeforeEach = true) {
...
+        paymentIds: List<PaymentId>,
+        expectedStatus: Int = 200
+    ) {
+        val (_, response, _) =
+            http
+                .post("/payments/delete-drafts")
+                .asUser(financeUser)
+                .withMockedTime(HelsinkiDateTime.of(today, LocalTime.of(10, 0)))
+                .jsonBody(jsonMapper.writeValueAsString(paymentIds))
+                .response()
```
akheron, line 410: Integration tests should use direct controller calls instead of http requests. It's probably not a big refactor to add `@Autowired` controller as a member property to the class and change all http calls to just call the contoroller's functions directly. There are plenty of other test classes that do this already, if you want examples.

frontend/src/employee-frontend/components/payments/Actions.tsx
```diff
@@ -46,6 +60,19 @@ const Actions = React.memo(function Actions({
...
+      <AsyncButton
+        text={i18n.payments.buttons.deletePayment(checkedIds.length)}
+        disabled={checkedIds.length === 0}
+        onClick={() => deletePayments(checkedIds)}
+        onSuccess={() => {
+          setError(undefined)
+          actions.clearChecked()
+          reloadPayments()
+        }}
+        onFailure={() => setError(i18n.common.error.unknown)}
```
akheron, line 72: Showing a generic error message to the user is not needed IMO. There will be a red cross ❌ indicating an error in the button, which gives the same amount of information to the user as the generic error.

akheron (approved):

PR #4712, opened 2024-02-05
Log new employee creation in the Titania integration

vesahonkasaari:
The Titania integration was not logging new employee creation properly, so fix that


patari:
What if transaction fails? Logging should be after the transaction commits.

vesahonkasaari:
Good point about the transaction, fixed now

PR #4714, opened 2024-02-05
Feature preschool terms

martinLeht:
#### Summary
Created new admin section to view preschool terms.


frontend/src/employee-frontend/api/applications.ts
```diff
@@ -411,6 +411,7 @@ export async function getClubTerms(): Promise<Result<ClubTerm[]>> {
   }
 }
 
+// TODO: refactor below function calls to leverage getPreschoolTermsV2()
```
Joosakur, line 414: Maybe this could be renamed as e.g. `getPreschoolTermsResult` so the new one wouldn't need that `V2` suffix, which sounds more like calling a different API endpoint altogether. This could also internally use the other function so there's less duplication:
```
export async function getPreschoolTermsResult(): Promise<Result<PreschoolTerm[]>> {
  return getPreschoolTerms()
    .then(result => Success.of(result))
    .catch(e => Failure.fromError(e))
}
```

Joosakur (approved):

frontend/src/employee-frontend/api/applications.ts
martinLeht: Very good point, I will make this change.

frontend/src/employee-frontend/api/applications.ts
```diff
@@ -413,11 +413,16 @@ export async function getClubTerms(): Promise<Result<ClubTerm[]>> {
...
-    )
-    return Success.of(result.data.map(deserializePreschoolTerm))
+    const result = await getPreschoolTermsResult()
+    return Success.of(result)
   } catch (e) {
     return Failure.fromError(e)
   }
 }
+
+// For query client use (to be refactored and remove V2 from name)
```
Joosakur, line 423: comment is slightly outdated

Joosakur (approved):

frontend/src/employee-frontend/api/applications.ts
martinLeht: Sorry, I removed it now.

PR #4720, opened 2024-02-06
`LocalHm` data types

akheron:


akheron:
This was a bad idea. In people's heads, time ranges are exclusive-ended, and treating them as inclusive-ended causes more problems that not.

PR #4721, opened 2024-02-07
Fix cache busting in Dockerfiles

akheron:
The `CACHE_BUST` arg must come *after* the `FROM` line to effectively bust the cache, see
https://docs.docker.com/engine/reference/builder/#understand-how-arg-and-from-interact.

### DIFF
--- a/frontend/Dockerfile
+++ b/frontend/Dockerfile
@@ -3,10 +3,10 @@
 #
 # SPDX-License-Identifier: LGPL-2.1-or-later
 
-ARG CACHE_BUST
 ARG NGINX_VERSION=1.25.1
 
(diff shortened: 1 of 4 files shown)


frontend/Dockerfile
```diff
@@ -3,10 +3,10 @@
 #
 # SPDX-License-Identifier: LGPL-2.1-or-later
 
-ARG CACHE_BUST
 ARG NGINX_VERSION=1.25.1
 
 FROM node:20.11.0-bookworm-slim AS builder
+ARG CACHE_BUST
```
patari, line 9: Add also after line 48

patari (approved):

reynders (approved):

PR #4725, opened 2024-02-07
mobile bugfixes

Joosakur:
#### Summary
- Return all attendances that overlap current day, because frontend needs them for validations
- Instead move the 30min logic inside status calculation function (if last attendance started yesterday and ended more than 30min ago then child is seen as COMING instead of DEPARTED
- After marking a full day absence it was not possible to cancel it, because the button reused "return-to-coming" endpoint which no longer deletes absences. Created a new endpoint specific for cancelling today's full day absence.
- Fixed bug in merging over night attendances
- Fixed bug in displaying 'yesterday' label before arrival time
- Fixed bug where full day absence could be marked although child had been present today after overnight absence

### DIFF
--- a/service/src/main/kotlin/fi/espoo/evaka/attendance/ChildAttendanceController.kt
+++ b/service/src/main/kotlin/fi/espoo/evaka/attendance/ChildAttendanceController.kt
@@ -9,6 +9,7 @@ import fi.espoo.evaka.ForceCodeGenType
 import fi.espoo.evaka.absence.AbsenceCategory
 import fi.espoo.evaka.absence.AbsenceType
 import fi.espoo.evaka.absence.AbsenceUpsert
+import fi.espoo.evaka.absence.getAbsencesOfChildByDate
 import fi.espoo.evaka.absence.insertAbsences
 import fi.espoo.evaka.absence.setChildDateAbsences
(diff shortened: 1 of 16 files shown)


service/src/main/kotlin/fi/espoo/evaka/attendance/ChildAttendanceController.kt
```diff
@@ -464,6 +468,40 @@ class ChildAttendanceController(
...
+                    tx.getAbsencesOfChildByDate(childId, today).map { it.category }.toSet()
+                val hasFullDayAbsence = placementType.absenceCategories() == absenceCategories
+                if (hasFullDayAbsence) {
+                    tx.deleteAbsencesByDate(childId, today)
+                } else {
+                    throw Conflict("Cannot cancel full day absence, child is not fully absent")
+                }
+            }
+        }
+        Audit.ChildAttendancesFullDayAbsenceDelete.log(targetId = childId, objectId = unitId)
```
Gekkio, line 502: This audit log is not very useful right now, and should probably include ids of the deleted absence

Gekkio (approved):
A bit tricky to review since there's so many changes unrelated to each other in the same commit, but seems good :+1:

PR #4728, opened 2024-02-09
Feature preschool term create

martinLeht:
#### Summary
Implemented support to create preschool terms


frontend/src/e2e-test/dev-api/fixtures.ts
```diff
@@ -151,6 +158,7 @@ export const nonFullDayTimeRange: TimeRange = {
 }
 
 export const preschoolTermFixture2020: PreschoolTerm = {
+  id: uuidv4(),
```
Joosakur, line 161: This could maybe do with a bit of refactoring. Now we are adding a random id here to satisfy the type, but in the dev-api endpoint the id is ignored and a different one will get generated in the DB. That could be confusing if some test later on tries to use the ID that it reads from the fixture. Minimal fix might be to have dev-api also take in PreschoolTermRequest like the actual controller. 

Next level refactoring could be to create a `DevPreschoolTerm` data class and a new overload for `tx.insert()` which could then be used also in the integration tests in a consistent fashion with most other things.

Great stuff overall! 🏅

martinLeht: Very good point, thanks - I ended up doing the second option of refactoring.

Joosakur: Let's also update the e2e test typings to use `DevPreschoolTerm` (in `dev-api/index.ts` and I guess here too), even though it's at the moment identical with `PreschoolTerm` from typescript's point of view, so there's no visible errors yet. I'm not sure if the dev api types can now be generated after the recent codegen update.. I guess not quite yet, so you'll probably need to add a typing for it yourself in `dev-api/types.ts.`

martinLeht: True, I added the DevPreschoolTerm type to frontend also and that type should now be uniformly in use in both front- and backend tests.

Joosakur (approved):

frontend/src/e2e-test/dev-api/fixtures.ts
Joosakur: 💎

PR #4738, opened 2024-02-12
limit selectable absence types in week calendar

Joosakur:
#### Summary
In week calendar edit modal the available absence types are limited to common cases unless there's an existing absence of a different type. Rare cases should be added through the month calendar which is generally used by less people.

### DIFF
--- a/frontend/src/employee-frontend/components/unit/unit-reservations/ChildDateModal.tsx
+++ b/frontend/src/employee-frontend/components/unit/unit-reservations/ChildDateModal.tsx
@@ -129,6 +129,16 @@ const form = transformed(
   }
 )
 
+const excludedAbsenceTypes: AbsenceType[] = [
+  'FREE_ABSENCE',
+  'PARENTLEAVE',
+  'PLANNED_ABSENCE',
+  'FORCE_MAJEURE'
+]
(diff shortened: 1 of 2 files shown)


frontend/src/employee-frontend/components/unit/unit-reservations/ChildDateModal.tsx
```diff
@@ -143,7 +153,34 @@ export default React.memo(function ChildDateModal({
   const today = LocalDate.todayInHelsinkiTz()
   const editingFuture = date.isAfter(today)
 
-  const absenceOptions = absenceTypes.map((at) => ({
+  const availableBillableAbsences = [...basicAbsenceTypes]
```
Gekkio, line 156: Should these be done inside `useMemo`? Not a *huge* amount of calculation but it's still done on every render

Gekkio (approved):

frontend/src/employee-frontend/components/unit/unit-reservations/ChildDateModal.tsx
Joosakur: Moved the logic to a function that is called from the initial state function and thus executed only once

PR #4745, opened 2024-02-12
respect sfi rate limit and do not generate decision pdf twice

Joosakur:
#### Summary

- Make async jobs more serial so that email and sfi sending jobs are planned after pdf generation
- This way pdf does not need to be generated twice, but sfi job can use the pdf from s3
- Sfi sending now plans AsyncJob.SendMessage instead of calling sfiClient, so that the ratelimit is shared correctly together with other sfi messages
- Remove second argument from sfiClient? It is now always null and apparently should never be used

### DIFF
--- a/service/src/main/kotlin/fi/espoo/evaka/assistanceneed/preschooldecision/AssistanceNeedPreschoolDecisionService.kt
+++ b/service/src/main/kotlin/fi/espoo/evaka/assistanceneed/preschooldecision/AssistanceNeedPreschoolDecisionService.kt
@@ -1,4 +1,4 @@
-// SPDX-FileCopyrightText: 2017-2023 City of Espoo
+// SPDX-FileCopyrightText: 2017-2024 City of Espoo
 //
 // SPDX-License-Identifier: LGPL-2.1-or-later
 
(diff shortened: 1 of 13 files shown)


service/src/main/kotlin/fi/espoo/evaka/assistanceneed/preschooldecision/AssistanceNeedPreschoolDecisionService.kt
```diff
@@ -53,38 +52,77 @@ class AssistanceNeedPreschoolDecisionService(
...
+        val decisionId = msg.decisionId
+        db.transaction { tx ->
+            val decision = tx.getAssistanceNeedPreschoolDecisionById(decisionId)
+            val endDate =
+                tx.getAssistanceNeedPreschoolDecisionsByChildId(decision.child.id)
+                    .find { it.id == decisionId }
+                    ?.validTo
+
+            check(!decision.hasDocument) {
+                "Assistance need preschool decision $decisionId already has a document key"
```
reynders, line 81: Would the be more understandable explanation: "A pdf already exists for the assistance need preschool decision @decisionId"

reynders (approved):

PR #4748, opened 2024-02-13
Add the ability to hide route planner link

cyanidesayonara:
Route planner is not available for treVaka Hämeenkyrö


patari (commented):
Maybe `routeLinkRootUrl` could be changed to optional, like `footerLogo`?

https://github.com/espoon-voltti/evaka/blob/master/frontend/src/lib-customizations/types.d.ts#L78

patari (approved):

akheron (approved):

PR #4751, opened 2024-02-14
Feature preschool term edit and deletion

martinLeht:
#### Summary
Implemented for admins support to edit all and delete future preschool terms.


service/src/main/kotlin/fi/espoo/evaka/daycare/PreschoolTerms.kt
```diff
@@ -58,6 +59,23 @@ fun Database.Read.getPreschoolTerms(): List<PreschoolTerm> {
...
+            finnish_preschool,
+            swedish_preschool,
+            extended_term,
+            application_period,
+            term_breaks
+        FROM preschool_term 
+        WHERE id = :id
+        """
+        )
+        .bind("id", id)
```
Joosakur, line 76: would prefer the new bind syntax in queries

frontend/src/employee-frontend/components/holiday-term-periods/TermsSection.tsx
```diff
@@ -44,53 +103,104 @@ export default React.memo(function TermsSection() {
...
+                        />
+                        {row.finnishPreschool.start.isAfter(
+                          LocalDate.todayInSystemTz()
+                        ) && (
+                          <IconButton
+                            icon={faTrash}
+                            data-qa="btn-delete"
+                            onClick={() => onDeletePreschoolTerm(row.id)}
+                            aria-label={i18n.common.edit}
+                          />
```
Joosakur, line 165: FWW, this thing could be done with much less code using a semi-recent component called ConfirmedMutation. You just pass in the texts and the mutation to be run if confirmed. You don't need to handle any state and modal rendering etc yourself.

frontend/src/employee-frontend/components/holiday-term-periods/TermsSection.tsx
```diff
@@ -26,12 +34,63 @@ export default React.memo(function TermsSection() {
...
+  const onCloseDeletionModal = useCallback(() => {
+    setTermToDelete(undefined)
+    setShowDeletionModal(false)
+  }, [setTermToDelete, setShowDeletionModal])
+
+  const onEditTermHandle = useCallback(
+    (term: PreschoolTerm) => {
+      if (term.finnishPreschool.start.isBefore(LocalDate.todayInSystemTz())) {
+        setTermToEdit(term.id)
+        setShowEditModal(true)
```
Joosakur, line 71: consider if both states are needed or can the modal be shown whenever termToEdit is not null

frontend/src/employee-frontend/components/holiday-term-periods/TermPeriodEditor.tsx
```diff
@@ -26,13 +33,25 @@ export default React.memo(function TermPeriodEditor() {
...
+        {termId === 'new'
+          ? renderResult(preschoolTerms, (preschoolTerms) => (
+              <PreschoolTermForm
+                allTerms={preschoolTerms}
+                onSuccess={navigateToList}
+                onCancel={navigateToList}
+              />
+            ))
+          : renderResult(
+              combine(preschoolTerms, preschoolTerm),
```
Joosakur, line 45: Seems unnecessary to fetch all and also fetch one separately again. We can always find the one that is being edited from the list which we anyway need to fetch, so the whole fetch by id endpoint is unnecessary even on backend.

frontend/src/employee-frontend/components/holiday-term-periods/TermsSection.tsx
```diff
@@ -26,12 +34,63 @@ export default React.memo(function TermsSection() {
...
   const navigateToNewTerm = useCallback(() => {
     navigate('/holiday-periods/preschool-term/new')
   }, [navigate])
 
+  const closeModalAndNavigateToEditTerm = useCallback(() => {
+    if (termToEdit) {
+      navigate(`/holiday-periods/preschool-term/${termToEdit}`)
+    }
+  }, [navigate, termToEdit])
+
```
Joosakur, line 93: On line 97 I'd consider using `AddButtonRow` instead of `AddButton` to keep the alignment more coherent with the other sections on the same page

frontend/src/employee-frontend/components/holiday-term-periods/TermsSection.tsx
martinLeht: I did not realize that this ConfirmedMutation was a thing - great point and it made the implementation much cleaner!

service/src/main/kotlin/fi/espoo/evaka/daycare/PreschoolTerms.kt
martinLeht: I fixed also rest of the preschool term queries to have this syntax.

frontend/src/employee-frontend/components/holiday-term-periods/TermsSection.tsx
martinLeht: Very true, I removed showEditModal state completely.

frontend/src/employee-frontend/components/holiday-term-periods/TermsSection.tsx
martinLeht: I thought about using it, but the UI design in figma had the button on the left compared to the other sections that has the button on the right, so I decided to go with the UI design.

frontend/src/employee-frontend/generated/api-clients/daycare.ts
martinLeht, line 1: These were generated when I ran codegen

frontend/src/employee-mobile-frontend/generated/api-clients/daycare.ts
martinLeht, line 1: These were generated when I ran codegen

frontend/src/employee-frontend/components/holiday-term-periods/TermPeriodEditor.tsx
```diff
@@ -23,11 +26,20 @@ export default React.memo(function TermPeriodEditor() {
     [navigate]
   )
 
+  const findPreschoolTermById = (
+    id: string,
+    allPreschoolTerms: PreschoolTerm[]
+  ) =>
+    termId !== 'new'
+      ? allPreschoolTerms.find((term) => term.id === termId)
+      : undefined
```
Joosakur, line 35: Id parameter is not used. 

If you want to get fancy and memoize the result so that the list won't be searched on every render (not that it takes a lot of power), that could also be done like this:

```
  const term: Result<PreschoolTerm | undefined> = useMemo(() => {
    if (termId === 'new') return Success.of(undefined)

    return preschoolTerms
      .map((allTerms) => allTerms.find((t) => t.id === termId))
      .chain((t) =>
        t ? Success.of(t) : Failure.of({ message: 'Term not found' })
      )
  }, [termId, preschoolTerms])

  return (
    <Container>
      <ContentArea opaque>
        {renderResult(
          combine(preschoolTerms, term),
          ([preschoolTerms, term]) => (
            <PreschoolTermForm
              term={term}
              allTerms={preschoolTerms}
```

This would also handle the error case slightly better where the term is not found from the list by displaying an error instead of passing `undefined` to `PreschoolTermForm` and implicitly switching from edit mode to create mode.

service/src/main/kotlin/fi/espoo/evaka/shared/security/Action.kt
```diff
@@ -1882,6 +1883,16 @@ sealed interface Action {
         override fun toString(): String = "${javaClass.name}.$name"
     }
 
+    enum class PreschoolTerm(
+        override vararg val defaultRules: ScopedActionRule<in PreschoolTermId>
+    ) : ScopedAction<PreschoolTermId> {
+        UPDATE(HasGlobalRole(ADMIN)),
+        DELETE(HasGlobalRole(ADMIN)),
+        READ(HasGlobalRole(ADMIN));
```
Joosakur, line 1891: this single term READ action is probably unused now

frontend/src/employee-mobile-frontend/generated/api-clients/daycare.ts
```diff
@@ -536,6 +536,22 @@ export async function createPreschoolTerm(
 }
 
 
+/**
+* Generated from fi.espoo.evaka.daycare.controllers.TermsController.deletePreschoolTerm
+*/
+export async function deletePreschoolTerm(
```
Joosakur, line 542: This unnecessary duplicate generation under the `employee-mobile-frontend` package could be avoided by specifying the user type more strictly in controller methods:
`user: AuthenticatedUser` --> `user: AuthenticatedUser.Employee`

frontend/src/employee-frontend/components/holiday-term-periods/api.ts
```diff
@@ -98,6 +98,26 @@ export async function getPreschoolTermsResult(): Promise<PreschoolTerm[]> {
     .then((res) => res.data.map(deserializePreschoolTerm))
 }
 
+export async function getPreschoolTerm(termId: UUID): Promise<PreschoolTerm> {
```
Joosakur, line 101: these could now be replaced by the auto-generated ones, but I can also do that together with refactoring other places

Joosakur (approved):
just a couple minor comments, feel free to merge after checking and possibly fixing them

frontend/src/employee-mobile-frontend/generated/api-clients/daycare.ts
martinLeht: Nice, I added this change and regenerated endpoints.

frontend/src/employee-frontend/components/holiday-term-periods/api.ts
martinLeht: Now I removed all the manually made clinet calls and replaced in the queries.ts all the api calls to use api-clients that are generated from codegen. Good to know that the generated clients are now ready for use.

frontend/src/employee-frontend/components/holiday-term-periods/TermPeriodEditor.tsx
martinLeht: Good points, I replaced the block with the snippet you provided, kudos!

PR #4758, opened 2024-02-15
better error message on inactive user login

Joosakur:


frontend/src/lib-components/molecules/modals/LoginErrorModal.tsx
```diff
@@ -26,6 +26,7 @@ export const LoginErrorModal = React.memo(function LoginErrorModal() {
   const [errorModalVisible, setErrorModalVisible] = useState<boolean>(
     queryParams.get(queryParamName) === 'true'
   )
+  const [errorCode] = useState(queryParams.get('errorCode'))
```
akheron, line 29: Why is this stored in component state instead of accessing it directly? There's no `setState` call at all.

akheron (approved):

PR #4761, opened 2024-02-16
Replace almost all manual API clients in citizen-frontend

Gekkio:
#### Summary

Everything is replaced, except attachment uploads and /auth/status


Joosakur (approved):
🏅

PR #4764, opened 2024-02-16
Show used service time and warnings in citizen day view

Wnt:
## Before this change
It was not possible to track service time usage per day in citizen front-end.
## After this change
Citizen day view displays used service time and highlights attendances that exceed planned time.

The new service time usage component is only used when feature flag `timeUsageInfo` is enabled and child's `ServiceNeed` contains a `daycareHoursPerMonth` value.

image

Child departure after reservation end; Both ends of attendance exceed reserved time

07 ei ilmoitusta, ei paikalla joten keskiarvo

No reservation / absence provided -> using monthly service time divided by 21


akheron (approved):

akheron:
Check if this can be put behind the feature flag.

PR #4791, opened 2024-02-20
Deprecate old query APIs

Gekkio:


akheron (approved):
LGTM, but let's try to switch to the new API quicky and re-enable the `allWarningsAsErrors` option.

PR #4795, opened 2024-02-21
Add async job permit mechanism for cluster-wide throttling

Gekkio:
#### Summary

Since we don't actually need to accurately limit cluster-wide *concurrency* of executed jobs, it's enough to throttle *claiming of jobs* and not their execution. This kind of simple per-pool "work permit" (= global lock) held while claiming a job should be enough.


akheron:
Shouldn't the email pool's throttle interval now be changed to `1 / 14` instead of the current `2 / 14`?

akheron (approved):

PR #4799, opened 2024-02-22
Feature/club terms

msavolainen-gofore:
#### Club terms management

Very similar to Preschool terms management


service/src/main/kotlin/fi/espoo/evaka/daycare/controllers/TermsController.kt
```diff
@@ -47,6 +52,76 @@ class TermsController(private val accessControl: AccessControl) {
         return db.connect { dbc -> dbc.read { it.getPreschoolTerms() } }
     }
 
+    @PostMapping("/club-terms")
+    fun createClubTerm(
+        db: Database,
+        user: AuthenticatedUser,
```
Joosakur, line 58: consider using `user: AuthenticatedUser.Employee` --> then these clients won't be generated for employee-mobile-frontend

Joosakur (commented):
Noticed one missing frontend validation. Backend gave error when term breaks were outside the actual term (maybe conflicted with terms breaks of another term). Same issue probably with preschool too. Otherwise all looks good 👍

PR #4819, opened 2024-02-26
Warn citizen about exceeding service need in calendar view month summary

Wnt:
## Before this change
It was hard for citizens to notice when reservations or realized attendance exceeds service need.

## After this change
Monthly summaries display a warning icon when a child has exceeded service need. If reservations exceed service need, the info box is initially open.

image

### DIFF
--- a/frontend/src/citizen-frontend/calendar/MonthElem.tsx
+++ b/frontend/src/citizen-frontend/calendar/MonthElem.tsx
@@ -2,6 +2,8 @@
 //
 // SPDX-License-Identifier: LGPL-2.1-or-later
 
+import { FontAwesomeIcon } from '@fortawesome/react-fontawesome'
+import { fasExclamationTriangle } from 'Icons'
 import React, { useCallback, useState } from 'react'
 import styled, { css } from 'styled-components'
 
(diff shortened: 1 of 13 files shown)


frontend/src/citizen-frontend/calendar/MonthElem.tsx
```diff
@@ -71,28 +74,33 @@ export default React.memo(function MonthElem({
...
-  )
   const displaySummary = featureFlags.timeUsageInfo && childSummaries.length > 0
+  const displayAlert =
+    displaySummary &&
+    childSummaries.some(
+      ({ reservedMinutes, usedServiceMinutes, serviceNeedMinutes }) =>
+        reservedMinutes > serviceNeedMinutes ||
+        usedServiceMinutes > serviceNeedMinutes
+    )
+  const { summaryInfoOpen, toggleSummaryInfo } = useSummaryInfo(childSummaries)
```
akheron, line 85: `displayAlert` could also be returned by the hook. That way, the `usedServiceMinutes > serviceNeedMinutes` logic would be only in one place.

akheron (approved):

PR #4827, opened 2024-02-29
Use range operations for child attendances

akheron:
#### Summary

Add support for mignight endtime for `TimeInterval`, add `overlaps()`, and use it for child attendance stuff.

### DIFF
--- a/service/src/main/kotlin/fi/espoo/evaka/attendance/ChildAttendanceController.kt
+++ b/service/src/main/kotlin/fi/espoo/evaka/attendance/ChildAttendanceController.kt
@@ -32,6 +32,7 @@ import fi.espoo.evaka.shared.domain.Conflict
 import fi.espoo.evaka.shared.domain.EvakaClock
 import fi.espoo.evaka.shared.domain.FiniteDateRange
 import fi.espoo.evaka.shared.domain.HelsinkiDateTime
+import fi.espoo.evaka.shared.domain.TimeInterval
 import fi.espoo.evaka.shared.security.AccessControl
 import fi.espoo.evaka.shared.security.Action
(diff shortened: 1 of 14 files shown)


service/src/main/kotlin/fi/espoo/evaka/attendance/ChildAttendanceController.kt
```diff
@@ -193,6 +194,9 @@ class ChildAttendanceController(
         @PathVariable childId: ChildId,
         @RequestBody body: ArrivalRequest
     ) {
+        if (body.arrived.second != 0 || body.arrived.nano != 0) {
+            throw BadRequest("Arrival time must be on the minute")
+        }
```
Joosakur, line 199: having to check this everywhere feels suboptimal. 🤔 maybe there could at least be a shared extension function `LocalTime.assertMinutePrecision()`?

Joosakur (approved):

service/src/main/kotlin/fi/espoo/evaka/attendance/ChildAttendanceController.kt
akheron: The ultimate fix would be to have `LocalHm` in request and response types which would follow the _parse don't validate_ principle, but I didn't want to do that in this PR.

akheron: ... and the `minute == nano == 0` check is already there as a DB constraint, so these checks could actually just be removed because the frontend doesn't send seconds.

akheron: Checks removed

PR #4831, opened 2024-03-01
Use new sql query api, part 1

akheron:


service/src/main/kotlin/fi/espoo/evaka/reports/AttendanceReservationReportController.kt
```diff
@@ -221,25 +217,21 @@ private fun Database.Read.getServiceNeeds(
...
     pl.child_id,
-    daterange(sn.start_date, sn.end_date, '[]') * daterange(:start, :end, '[]') as range,
+    daterange(sn.start_date, sn.end_date, '[]') * daterange(${bind(start)}, ${bind(end)}, '[]') as range,
     sno.occupancy_coefficient_under_3y AS occupancy_coefficient_under,
     sno.occupancy_coefficient AS occupancy_coefficient_over
 FROM service_need sn
 JOIN placement pl on pl.id = sn.placement_id
 LEFT JOIN service_need_option sno ON sno.id = sn.option_id
-WHERE pl.child_id = ANY(:children) AND daterange(sn.start_date, sn.end_date, '[]') && daterange(:start, :end, '[]');
+WHERE pl.child_id = ANY(${bind(children)}) AND daterange(sn.start_date, sn.end_date, '[]') && daterange(${bind(start)}, ${bind(end)}, '[]');
```
Wnt, line 231: Is it OK that children binding doesn't have the `.toTypedArray()` part in the new version?

akheron: Should be ok. Typed arrays were required way back when the database query abstraction didn't understand lists.

Wnt (approved):

PR #4848, opened 2024-03-05
Fix css problems on different browsers

reynders:
#### Summary
Css layout of having both reservations and attendances fixed on chromium, chrome, firefox and safari

Screenshot 2024-03-05 at 8 15 21


reynders (commented):
Mob review with @Joosakur

PR #4855, opened 2024-03-06
Staff occupancy effect on arrival

Joosakur:

### DIFF
--- a/frontend/src/e2e-test/specs/6_mobile/realtime-staff-attendances.spec.ts
+++ b/frontend/src/e2e-test/specs/6_mobile/realtime-staff-attendances.spec.ts
@@ -155,7 +155,7 @@ describe('Realtime staff attendance page', () => {
     await staffAttendancePage.assertPresentStaffCount(0)
   })
 
-  test('Occupancy effect can not be unchecked if it has been given permanently', async () => {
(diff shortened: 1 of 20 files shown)


frontend/src/e2e-test/specs/6_mobile/realtime-staff-attendances.spec.ts
```diff
@@ -154,6 +155,30 @@ describe('Realtime staff attendance page', () => {
     await staffAttendancePage.assertPresentStaffCount(0)
   })
 
+  test('Occupancy effect can not be unchecked if it has been given permanently', async () => {
```
Gekkio, line 158: I think this test is not very useful, because it only tests that a checkbox did not appear, which could be true for many reasons. It does not test that the system did the right thing (= the correct coefficient was saved in the saved row)

Gekkio (approved):

frontend/src/e2e-test/specs/6_mobile/realtime-staff-attendances.spec.ts
Joosakur: Hmm true, although at least it's checked after the other fields so it cannot be missing simply because the view is still loading. Since the coefficient is not clearly shown on UI that's a bit difficult to assert, the kotlin based e2e test would be really nice here, but I'll try extending the test by checking the 'edit previous attendances' view.

PR #4864, opened 2024-03-08
no closed units in admin mass acl selection

Joosakur:

### DIFF
--- a/frontend/src/employee-frontend/generated/api-clients/daycare.ts
+++ b/frontend/src/employee-frontend/generated/api-clients/daycare.ts
@@ -215,10 +215,18 @@ export async function getDaycare(
 /**
 * Generated from fi.espoo.evaka.daycare.controllers.DaycareController.getDaycares
 */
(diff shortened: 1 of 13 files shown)


frontend/src/employee-frontend/generated/api-clients/daycare.ts
```diff
@@ -215,10 +215,18 @@ export async function getDaycare(
 /**
 * Generated from fi.espoo.evaka.daycare.controllers.DaycareController.getDaycares
 */
-export async function getDaycares(): Promise<Daycare[]> {
+export async function getDaycares(
+  request: {
+    includeClosed?: boolean | null
```
Wnt, line 220: I find it weird that this can be undefined, null or boolean. But I guess that's because of how our codegen works. Or do you see a way how this could be improved?

Joosakur: Yeah. Once we can update spring and default values work the null option could possibly be removed.

Wnt (approved):

PR #4854, opened 2024-03-06
Add scheduled job EndActiveDaycareAssistanceDecisions

patari:
#### Summary

Automatically set end date for daycare assistance decisions from placements.

PR #4865, opened 2024-03-08
Add scheduled job EndActivePreschoolAssistanceDecisions

patari:
#### Summary

https://github.com/espoon-voltti/evaka/pull/4854 for preschool.

### DIFF
--- a/service/src/main/kotlin/fi/espoo/evaka/assistanceneed/preschooldecision/AssistanceNeedPreschoolDecisionQueries.kt
+++ b/service/src/main/kotlin/fi/espoo/evaka/assistanceneed/preschooldecision/AssistanceNeedPreschoolDecisionQueries.kt
@@ -266,7 +266,7 @@ fun Database.Read.getAssistanceNeedPreschoolDecisionsByChildId(
         @Suppress("DEPRECATION")
         createQuery(sql).bind("childId", childId).toList<AssistanceNeedPreschoolDecisionBasics>()
 
-    return fillInValidToForDecisionResults(decisions)
+    return decisions
 }
 
 fun Database.Read.getAssistanceNeedPreschoolDecisionsByChildIdUsingFilter(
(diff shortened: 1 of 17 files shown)


Joosakur:
If we add an end_date field, then we need to also update it whenever a new decision is inserted within the old decisions validity range, and furthermore if that decision then gets cancelled (or its start date somehow changes), we'd need to again recalculate the validity. That's why it seemed easier to only store start points and calculate the endpoint dynamically as next start minus one. Now apparently the requirements have changed and the placement end date should also be taken into account, but maybe that could be part of the same dynamic calculation?

I guess reporting is a place where this strategy might not work as well, so yeah maybe for that we may actually need to store the end date.. But then we'd probably need some function that is executed whenever there's any change in child's decisions or placements, which would recalculate all end dates (kind of like a generated column but over multiple rows and tables).

patari:
> If we add an end_date field, then we need to also update it whenever a new decision is inserted within the old decisions validity range

Yes, this is now implemented.

> and furthermore if that decision then gets cancelled (or its start date somehow changes), we'd need to again recalculate the validity

I don't decisions work like that in general. If decision gets annulled, previous decision should not magically extend its validity.

> I guess reporting is a place where this strategy might not work as well, so yeah maybe for that we may actually need to store the end date.. But then we'd probably need some function that is executed whenever there's any change in child's decisions or placements, which would recalculate all end dates (kind of like a generated column but over multiple rows and tables).

I think added nightly job takes care of this already.

service/src/main/kotlin/fi/espoo/evaka/assistanceneed/preschooldecision/AssistanceNeedPreschoolDecisionQueries.kt
```diff
@@ -372,6 +349,25 @@ fun Database.Transaction.decideAssistanceNeedPreschoolDecision(
...
+    createUpdate {
+            sql(
+                """
+UPDATE assistance_need_preschool_decision
+SET valid_to = ${bind(endDate)}
+WHERE id <> ${bind(excludingId)}
+  AND (valid_to IS NULL OR valid_to > ${bind(endDate)})
+  AND child_id = ${bind(childId)}
+  AND status = 'ACCEPTED'
+"""
```
Joosakur, line 366: Arguably not a very realistic/common case, but currently this fails if you insert a new decision before an old one, as it results in an inverted range. 
Maybe needs `AND valid_from <= ${bind(endDate)}` or some range operator.

service/src/main/resources/db/migration/V388__preschool_assistance_decision_valid_to.sql
```diff
@@ -1,3 +1,24 @@
...
+UPDATE assistance_need_preschool_decision
+SET valid_to = fill_in_valid_to.new_valid_to
+FROM fill_in_valid_to
+WHERE fill_in_valid_to.id = assistance_need_preschool_decision.id;
+
+ALTER TABLE assistance_need_preschool_decision
+    ADD CONSTRAINT check$assistance_need_preschool_decision_no_overlap EXCLUDE USING gist (
+        child_id WITH =,
+        daterange(valid_from, valid_to, '[]') WITH &&
+        ) WHERE (status = 'ACCEPTED');
```
Joosakur, line 24: Current model supports multiple accepted decisions starting on the same date so each city might need to check that they don't have such data in staging/prod or this will explode on deploy. Also might want to add some validation to check that in the controller, so that there's no http500 and alert on insert.

service/src/main/kotlin/fi/espoo/evaka/assistanceneed/preschooldecision/AssistanceNeedPreschoolDecisionQueries.kt
patari: Fixed

service/src/main/resources/db/migration/V388__preschool_assistance_decision_valid_to.sql
patari: > Also might want to add some validation to check that in the controller, so that there's no http500 and alert on insert.

Fixed

Joosakur (approved):
Before merging let's check if some db has multiple decisions with same start date, child and status = 'ACCEPTED'

PR #4880, opened 2024-03-13
upsert/deactivate messaging accounts during mass acl

Joosakur:

### DIFF
--- a/service/src/main/kotlin/fi/espoo/evaka/pis/controllers/EmployeeController.kt
+++ b/service/src/main/kotlin/fi/espoo/evaka/pis/controllers/EmployeeController.kt
@@ -5,6 +5,8 @@
 package fi.espoo.evaka.pis.controllers
 
 import fi.espoo.evaka.Audit
+import fi.espoo.evaka.daycare.deactivatePersonalMessageAccountIfNeeded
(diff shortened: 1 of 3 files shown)


service/src/main/kotlin/fi/espoo/evaka/pis/controllers/EmployeeController.kt
```diff
@@ -157,6 +159,7 @@ class EmployeeController(private val accessControl: AccessControl) {
                 )
 
                 it.upsertEmployeeDaycareRoles(id, body.daycareIds, body.role)
+                it.upsertEmployeeMessageAccount(id)
```
Wnt, line 162: Does this now create personal messaging accounts even when `List<DaycareId>` is empty or when the added role is `STAFF`? Shouldn't messaging accounts be created only for `UNIT_SUPERVISOR` and `SPECIAL_EDUCATION_TEACHER` roles?

Joosakur: Added validation for empty list. Roles are apparently as intended.

akheron (approved):

PR #4914, opened 2024-03-15
HTML-escape event title in calendar notification

Gekkio:

### DIFF
--- a/service/src/main/kotlin/fi/espoo/evaka/emailclient/EvakaEmailMessageProvider.kt
+++ b/service/src/main/kotlin/fi/espoo/evaka/emailclient/EvakaEmailMessageProvider.kt
@@ -15,6 +15,7 @@ import fi.espoo.evaka.shared.domain.FiniteDateRange
 import java.time.format.DateTimeFormatter
 import java.time.format.FormatStyle
 import java.util.Locale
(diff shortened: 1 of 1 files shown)


service/src/main/kotlin/fi/espoo/evaka/emailclient/EvakaEmailMessageProvider.kt
```diff
@@ -526,7 +527,7 @@ $unsubscribeEn
                     if (event.period.end != event.period.start) {
                         period += "-${event.period.end.format(format)}"
                     }
-                    "<li>$period: ${event.title}</li>"
+                    "<li>$period: ${HtmlEscape.escapeHtml5(event.title)}</li>"
```
Joosakur, line 530: wonder if escapeHtml5xml would be better here too?

Joosakur (approved):

service/src/main/kotlin/fi/espoo/evaka/emailclient/EvakaEmailMessageProvider.kt
Gekkio: It probably doesn't matter since we can assume e-mail clients support HTML entities properly

PR #4927, opened 2024-03-19
Vasu&Leops migration part 1

Joosakur:
#### Summary

- Admin can trigger migration of curriculum_documents to document_templates and child_documents
- Migration can be ran repeatedly (although some orphan document_templates may be left as garbage) so that we can iterate the logic after checking results with real data 
- For now only admin sees the created documents and templates so that the results can be validated after each iteration
- Only published vasus&leops in COMPLETED state are migrated so the final migration run should be done during summer after current season has ended

### DIFF
--- a/service/src/main/kotlin/fi/espoo/evaka/document/childdocument/ChildDocumentController.kt
+++ b/service/src/main/kotlin/fi/espoo/evaka/document/childdocument/ChildDocumentController.kt
@@ -6,7 +6,6 @@ package fi.espoo.evaka.document.childdocument
 
 import fi.espoo.evaka.Audit
 import fi.espoo.evaka.document.DocumentTemplateContent
-import fi.espoo.evaka.document.DocumentType
 import fi.espoo.evaka.document.getTemplate
 import fi.espoo.evaka.pis.listPersonByDuplicateOf
 import fi.espoo.evaka.shared.ChildDocumentId
@@ -19,7 +18,6 @@ import fi.espoo.evaka.shared.domain.EvakaClock
 import fi.espoo.evaka.shared.domain.NotFound
 import fi.espoo.evaka.shared.security.AccessControl
 import fi.espoo.evaka.shared.security.Action
-import fi.espoo.evaka.vasu.VASU_MIGRATION_COMPLETED
 import org.springframework.web.bind.annotation.DeleteMapping
 import org.springframework.web.bind.annotation.GetMapping
 import org.springframework.web.bind.annotation.PathVariable
@@ -108,12 +106,7 @@ class ChildDocumentController(
                                 ?.takeIf { it.contains(Action.ChildDocument.READ) }
                                 ?.let { ChildDocumentSummaryWithPermittedActions(document, it) }
                         }
-                        .filter {
-                            VASU_MIGRATION_COMPLETED ||
-                                user.isAdmin ||
-                                it.data.type !in
-                                    listOf(DocumentType.MIGRATED_VASU, DocumentType.MIGRATED_LEOPS)
-                        }
+                        .filter { user.isAdmin || !it.data.type.isMigrated() }
                 }
             }
             .also { Audit.ChildDocumentRead.log(targetId = childId) }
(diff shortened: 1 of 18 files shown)


service/src/main/kotlin/fi/espoo/evaka/document/childdocument/ChildDocumentController.kt
```diff
@@ -100,11 +102,18 @@ class ChildDocumentController(
...
-                            ?.let { ChildDocumentSummaryWithPermittedActions(document, it) }
-                    }
+                    documents
+                        .mapNotNull { document ->
+                            permittedActions[document.id]
+                                ?.takeIf { it.contains(Action.ChildDocument.READ) }
+                                ?.let { ChildDocumentSummaryWithPermittedActions(document, it) }
+                        }
+                        .filter {
+                            VASU_MIGRATION_COMPLETED ||
```
Wnt, line 112: This filter implementation could be extracted somewhere and the same implementation could be used in both `getDocument()` and `getDocuments()`. And I think the variable `VASU_MIGRATION_COMPLETED` should be removed.

service/src/main/kotlin/fi/espoo/evaka/document/childdocument/ChildDocumentQueriesCitizen.kt
```diff
@@ -29,6 +31,10 @@ fun Database.Read.getChildDocumentCitizenSummaries(
             )
         }
         .toList<ChildDocumentCitizenSummary>()
+        .filter {
+            VASU_MIGRATION_COMPLETED ||
```
Wnt, line 35: Also here I think `VASU_MIGRATION_COMPLETED` is unnecessary. And you could extract and reuse the filter implementation.

service/src/main/kotlin/fi/espoo/evaka/vasu/DocumentMigrator.kt
```diff
@@ -0,0 +1,550 @@
...
+        language =
+            when (vasu.language) {
+                VasuLanguage.FI -> DocumentLanguage.FI
+                VasuLanguage.SV -> DocumentLanguage.SV
+            },
+        confidential = true,
+        legalBasis =
+            when (vasu.type) {
+                CurriculumType.DAYCARE ->
+                    when (vasu.language) {
```
Wnt, line 180: Looks like you are always returning translations.lawVasu so you don't need to check the language here
```suggestion
                    translations.lawVasu
```

service/src/main/kotlin/fi/espoo/evaka/vasu/DocumentMigrator.kt
```diff
@@ -0,0 +1,550 @@
...
+                        VasuLanguage.FI -> translations.lawVasu
+                        VasuLanguage.SV -> translations.lawVasu
+                    }
+                CurriculumType.PRESCHOOL ->
+                    when (vasu.language) {
```
Wnt, line 185: Same here 
```suggestion
                    translations.lawLeops
```

Wnt (changes requested):

service/src/main/kotlin/fi/espoo/evaka/vasu/DocumentMigrator.kt
Joosakur: Ah yeah, that was from time when translations were hard coded here, cleaned it up now

service/src/main/kotlin/fi/espoo/evaka/document/childdocument/ChildDocumentController.kt
Joosakur: The idea I was going for here was that once the migration is considered ready, the constant value could be flipped to make the results visible to non-admins, and soon after if everything went fine all these filters could be permanently removed. But yeah maybe that's a bit unnecessary if it's just a global constant and not an env variable which would allow different cities and envs to have the go-live at different points in time.

What about the extraction then. I guess the DocumentType enum could have a boolean property telling if it's migrated. That would replace the `it.data.type !in listOf(DocumentType.MIGRATED_VASU, DocumentType.MIGRATED_LEOPS)` condition which is the only fully shared part, assuming that the `VASU_MIGRATION_COMPLETED` is removed.

Joosakur: Implemented that change

Wnt (approved):

PR #4929, opened 2024-03-21
List placement types and service need options under finance basics

vesahonkasaari:
Service need options are expandable but empty, voucher value listings coming there next

### DIFF
--- a/frontend/src/employee-frontend/components/finance-basics/PlacementTypeItem.tsx
+++ b/frontend/src/employee-frontend/components/finance-basics/PlacementTypeItem.tsx
@@ -1,9 +1,10 @@
-// SPDX-FileCopyrightText: 2017-2022 City of Espoo
+// SPDX-FileCopyrightText: 2017-2024 City of Espoo
 //
 // SPDX-License-Identifier: LGPL-2.1-or-later
 
-import React, { useCallback, useState } from 'react'
+import React from 'react'
 
+import { useBoolean } from 'lib-common/form/hooks'
 import { PlacementType } from 'lib-common/generated/api-types/placement'
 import { ServiceNeedOption } from 'lib-common/generated/api-types/serviceneed'
 import { CollapsibleContentArea } from 'lib-components/layout/Container'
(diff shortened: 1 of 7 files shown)


frontend/src/employee-frontend/components/finance-basics/ServiceNeedsSection.tsx
```diff
@@ -0,0 +1,51 @@
...
+import { renderResult } from '../async-rendering'
+
+import PlacementTypeItem from './PlacementTypeItem'
+import { serviceNeedsQuery } from './queries'
+
+export default React.memo(function ServiceNeedsSection() {
+  const { i18n } = useTranslation()
+
+  const [open, setOpen] = useState(true)
+  const toggleOpen = useCallback(() => setOpen((isOpen) => !isOpen), [setOpen])
```
Joosakur, line 22: The React state setter can be called with a function, so you don't need to directly depend on the current value in the scope: `setOpen(prev => !prev)`

We also have a custom hook `useBoolean` which can be used instead of `useState(true)` and which exports a toggle function out of the box.

frontend/src/employee-frontend/components/finance-basics/ServiceNeedsSection.tsx
```diff
@@ -0,0 +1,51 @@
...
+      open={open}
+      toggleOpen={toggleOpen}
+      data-qa="service-needs-section"
+      data-isloading={isLoading(data)}
+    >
+      {renderResult(data, (serviceNeedsList) => (
+        <>
+          {serviceNeedsList
+            .map((serviceNeed) => serviceNeed.validPlacementType)
+            .filter((value, index, array) => array.indexOf(value) === index)
```
Joosakur, line 39: What does this filter do? Ah, I guess it's distinct. This might actually have a minor issue in that if some placement type doesn't yet have any service need options, then the whole placement type will be missing from here, which could affect new cities in the future. You could consider using `import { placementTypes } from 'lib-customizations/employee'` instead which also improves readability.

frontend/src/employee-frontend/components/finance-basics/ServiceNeedsSection.tsx
```diff
@@ -0,0 +1,51 @@
+// SPDX-FileCopyrightText: 2017-2022 City of Espoo
```
Joosakur, line 1: Doesn't matter much but the new files could have `-2024`. Maybe on tech cafe we could discuss if all these could be simply changed to `2017-Present`.

frontend/src/employee-frontend/components/finance-basics/ServiceNeedItem.tsx
```diff
@@ -0,0 +1,34 @@
...
+        opaque
+        title={<H4>{serviceNeed}</H4>}
+        open={open}
+        toggleOpen={toggleOpen}
+        paddingHorizontal="0"
+        paddingVertical="0"
+      >
+        <>{/* TODO: voucher value listing here */}</>
+      </CollapsibleContentArea>
+      <div className="separator small" />
```
Joosakur, line 31: While there seems to be some existing css definition for `.separator`, I don't see any for `.small` and probably this should anyway be replaced by a styled component.

frontend/src/employee-frontend/components/finance-basics/PlacementTypeItem.tsx
```diff
@@ -0,0 +1,66 @@
...
+
+  return (
+    <CollapsibleContentArea
+      opaque
+      title={<H3>{i18n.placement.type[placementType]}</H3>}
+      open={open}
+      toggleOpen={toggleOpen}
+      paddingHorizontal="0"
+    >
+      <>
```
Joosakur, line 38: these <> fragments seem unnecessary here

frontend/src/employee-frontend/components/finance-basics/PlacementTypeItem.tsx
```diff
@@ -0,0 +1,66 @@
...
+        {serviceNeedsList
+          .filter(
+            (value) =>
+              value.validPlacementType == placementType && value.defaultOption
+          )
+          .map((serviceNeed) => (
+            <ServiceNeedItem
+              key={serviceNeed.nameFi + ' (oletus)'}
```
Joosakur, line 46: id is generally a better key when available since it's guaranteed to be unique

frontend/src/employee-frontend/components/finance-basics/ServiceNeedItem.tsx
vesahonkasaari: Guess that should indeed be just separator if it's not large. I modeled this usage from FeeThresholdsItem in the same context, but what would be a better way to do this? (Looking at git blame this seems to be so 2022...)

Joosakur: I would at least check if the HorizontalLine under lib-components would work, name sounds promising. If not, there also seems to be five different `const Separator = styled.div...` components and a bunch of `styled.hr` components so maybe some of them could be made reusable. With this classname approach the biggest issue is that it's quite difficult to find out where that css is defined, so any suitable styled component, whether imported from somewhere or defined locally in this file, would be better.

vesahonkasaari: HorizontalLine seems to do the trick, thanks.

frontend/src/employee-frontend/components/finance-basics/ServiceNeedsSection.tsx
```diff
@@ -14,29 +14,36 @@ import { renderResult } from '../async-rendering'
...
 import { serviceNeedsQuery } from './queries'
+import { useBoolean } from 'lib-common/form/hooks'
+import {PlacementType} from "../../../lib-common/generated/api-types/placement";
 
 export default React.memo(function ServiceNeedsSection() {
   const { i18n } = useTranslation()
 
-  const [open, setOpen] = useState(true)
-  const toggleOpen = useCallback(() => setOpen((isOpen) => !isOpen), [setOpen])
+  const [open, useOpen] = useBoolean(false)
```
Joosakur, line 23: just FYI: you can avoid coming up with a name for the second argument (useOpen) by destructuring further:
`const [open, { toggle: toggleOpen }] = useBoolean(false)`

Joosakur (approved):

PR #4930, opened 2024-03-21
Add Jamix customerId field into daycare groups

Wnt:
This is first part of Jamix meal ordering system integration
## After this change
It's possible to define Jamix customerId per daycare group level. This change also adds a front-end feature flag to control the visibility of Jamix integration related configuration and functions.
image


service/src/main/kotlin/fi/espoo/evaka/daycare/DaycareGroupQueries.kt
```diff
@@ -59,19 +58,19 @@ fun Database.Transaction.updateGroup(
...
-        .bind("name", name)
-        .bind("startDate", startDate)
-        .bind("endDate", endDate)
-        .execute()
-        .let { if (it != 1) throw NotFound("Group $groupId not found") }
+    createUpdate {
+            sql(
+                """
+UPDATE daycare_group
+SET name = ${bind(name)}, start_date = ${bind(startDate)}, end_date = ${bind(endDate)}, jamix_customer_id = ${bind(jamixCustomerId)}
```
Joosakur, line 68: Optional nitpick: if we don't want to mix nulls and empty strings in db, we could consider using `${bind(jamixCustomerId?.takeIf { it.isNotBlank() })}` or alternatively making the column NOT NULL with old rows migrated to empty string.

Joosakur (approved):
some integration test is failing

service/src/main/kotlin/fi/espoo/evaka/daycare/DaycareGroupQueries.kt
Wnt: I changed the data type into nullable Integer

PR #4931, opened 2024-03-21
re-sync personal messaging accounts with daycare ACL

Joosakur:
#### Summary

data fix after https://github.com/espoon-voltti/evaka/pull/4880 is in prod


Joosakur:
reviewed as part of https://github.com/espoon-voltti/evaka/pull/4880

PR #4932, opened 2024-03-21
Employee discussion reservation surveys

tmkrepo:
## Employee side implementation for creating and managing discussion surveys for individual groups:
- discussion survey frontend is behind feature flag `discussionReservations`

### Features left out:
- visualization improvements of discussion surveys in event calendars
- event notifications for discussion surveys

### DB:
- data model changes:
  - added `event_type` to explicitly distinguish discussion surveys from old calendar events

### Backend:
- endpoints for
  - fetching discussion surveys for group
  - fetching reservation calendar day details for populating calendar with holiday and event information
- integration tests

### Frontend:
- new form UI for viewing, editing and creating discussion surveys and associated event times
- an expandable calendar for visualizing discussion survey event times and reservations
- e2e testing for basic functions


frontend/src/employee-frontend/components/unit/tab-calendar/discussion-surveys/times-calendar/CalendarEventTimeReservation.tsx
```diff
@@ -0,0 +1,363 @@
...
+        )}
+        {deleteConfirmModalVisible && (
+          <InfoModal
+            type="warning"
+            title={t.reservationModal.deleteConfirmation.title}
+            icon={faQuestion}
+            reject={{
+              action: () => setDeleteConfirmModalVisible(false),
+              label: t.reservationModal.deleteConfirmation.cancelButton
+            }}
```
Joosakur, line 348: Consider using ConfirmedMutation component here. It is meant for the common use-case of "click button -> show confirmation modal -> if user confirms the run the mutation" so if these two don't include anything special, you would get the same functionality with less code.

frontend/src/employee-frontend/components/unit/tab-calendar/discussion-surveys/times-calendar/CalendarEventTimeInput.tsx
```diff
@@ -0,0 +1,50 @@
...
+  return (
+    <InputContainer
+      spacing="xxs"
+      alignItems="center"
+      justifyContent="center"
+      data-qa="time-input-container"
+    >
+      <TimeInputF
+        bind={startTime}
+        info={{ text: '' }}
```
Joosakur, line 37: check if the empty info is necessary or if the rop should be simply omitted

frontend/src/employee-frontend/components/unit/tab-calendar/discussion-surveys/survey-editor/DiscussionSurveyForm.tsx
```diff
@@ -0,0 +1,304 @@
...
+            mutation={updateCalendarEventMutation}
+            text={t.discussionReservation.saveSurveyButton}
+            disabled={!isBasicInfoValid}
+            onSuccess={() =>
+              navigate(
+                `/units/${unitId}/groups/${groupId}/discussion-reservation-surveys/${eventData.id}`
+              )
+            }
+            onClick={() => {
+              const values = form.value()
```
Joosakur, line 241: If I remember correctly, calling `.value()` throws if the form is not valid, so you might want to check `.isValid()` before and possibly return cancelMutation, even if the disabled prop might mostly already handle that scenario

frontend/src/employee-frontend/components/unit/tab-calendar/discussion-surveys/survey-editor/DiscussionSurveyForm.tsx
```diff
@@ -0,0 +1,304 @@
...
+  const t = i18n.unit.calendar.events
+
+  const [cancelConfirmModalVisible, setCancelConfirmModalVisible] =
+    useState(false)
+
+  const navigate = useNavigate()
+
+  const { times, title, description, attendees } = useFormFields(form)
+
+  const basicInfoValidationErrors = useMemo(
```
Joosakur, line 85: Seems like a more correct way to do this might be to wrap the form in `validated()`. Now some of the field level validation comes from the form and some of the validation is added on top ad hoc, which can be confusing and error prone. But there might also be valid reasons why that doesn't work here.

frontend/src/employee-frontend/components/unit/tab-calendar/discussion-surveys/survey-editor/DiscussionSurveyEditor.tsx
```diff
@@ -0,0 +1,118 @@
...
+    [today]
+  )
+
+  const visibleCalendarRange = useMemo(() => {
+    const previousMonday = today.subDays(today.getIsoDayOfWeek() - 1)
+
+    return new FiniteDateRange(previousMonday, calendarHorizonDate)
+  }, [calendarHorizonDate, today])
+
+  const extendHorizon = () => {
```
Joosakur, line 70: since other things are memoized, this should probably also be wrapped in useCallback

frontend/src/employee-frontend/components/unit/tab-calendar/discussion-surveys/DiscussionSurveyView.tsx
```diff
@@ -0,0 +1,465 @@
...
+      ),
+    [today]
+  )
+
+  const visibleCalendarRange = useMemo(() => {
+    const previousMonday = today.subDays(today.getIsoDayOfWeek() - 1)
+    return new FiniteDateRange(previousMonday, calendarHorizonDate)
+  }, [calendarHorizonDate, today])
+
+  const extendHorizon = () => {
```
Joosakur, line 249: useCallback

Joosakur (commented):
Looks pretty good 👍 Here's some comments from the first round of cursory review. Next I'll try to do some manual testing, maybe on Monday.

frontend/src/employee-frontend/components/unit/tab-calendar/discussion-surveys/times-calendar/CalendarEventTimeReservation.tsx
tmkrepo: Modal behavior here has a "special" detail: confirmation is prompted conditionally based on the change detection to the reservation information, e.g. confirmation on deletion is only prompted when deleting a reserved event time. As I understand, `ConfirmedMutation` will always prompt for confirmation and would result in less cognizant behavior.

frontend/src/employee-frontend/components/unit/tab-calendar/discussion-surveys/times-calendar/CalendarEventTimeInput.tsx
tmkrepo: Leaving the info prop out will lead to an internal default being used resulting in visible validation texts which is not desired in the cramped day cell components.

Joosakur:
I guess citizen side is coming later, but I noticed that on their calendar there's now a notification icon on every date between event min and max. Might want to filter those out, e.g. by not even returning events of this type from backend to citizen yet.
Screenshot 2024-03-25 at 9 37 45

Joosakur:
Tree-menu should be sorted, first by group name and then by child name. Also when reserving a time on behalf of a citizen the dropdown with children should be sorted.

PR #4933, opened 2024-03-21
Lapsen muut huoltajat ja sijaishuoltajat saa nähdä hakemukset/päätökset

Gekkio:

### DIFF
--- a/frontend/src/citizen-frontend/applications/ChildApplicationsBlock.tsx
+++ b/frontend/src/citizen-frontend/applications/ChildApplicationsBlock.tsx
@@ -86,7 +86,13 @@ const ChildHeading = styled(H2)`
 `
 
 export default React.memo(function ChildApplicationsBlock({
-  data: { childId, childName, applicationSummaries, permittedActions }
+  data: {
+    childId,
+    childName,
+    applicationSummaries,
+    permittedActions,
+    decidableApplications
+  }
 }: ChildApplicationsBlockProps) {
   const navigate = useNavigate()
   const t = useTranslation()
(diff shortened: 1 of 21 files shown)


frontend/src/citizen-frontend/decisions/decisions-page/Decisions.tsx
```diff
@@ -60,9 +60,13 @@ export default React.memo(function Decisions() {
     )[]
     firstName: string
     lastName: string
+    canDecide: UUID[]
```
Joosakur, line 63: could improve naming to clarify what ids these are

frontend/src/citizen-frontend/applications/ChildApplicationsBlock.tsx
```diff
@@ -246,7 +247,8 @@ export default React.memo(function ChildApplicationsBlock({
                   {applicationStatus === 'WAITING_CONFIRMATION' &&
                     permittedActions[applicationId]?.includes(
                       'READ_DECISIONS'
-                    ) && (
+                    ) &&
+                    ownedByCurrentUser && (
```
Joosakur, line 251: should this be also based on the canDecide or something more generic like that instead of kind of hard coding it here that the application owner can do this? 🤔

Joosakur (approved):

frontend/src/citizen-frontend/applications/ChildApplicationsBlock.tsx
Gekkio: It's a bit tricky since we have here data and permissions about applications but accepting/rejecting is an action done on the individual decisions (Action.Citizen.Decision). But I'll think about it a bit more...

Joosakur (approved):
👏

PR #4952, opened 2024-03-28
Meal times per unit

Wnt:
## After this change
It's possible to define meal times for daycare units. The inputs are only visible when front-end feature flag `jamixIntegration` is set to `true`.

image

This change also fixes a bug where `UnitEditor` didn't reflect the state API returns after submit.


akheron (approved):

Gekkio (approved):
Looks good and I have just one suggestion:

Consider using composition instead of flat prefixed properties. In other words, instead of adding 5 new fields to `DaycareFields` and similar classes, add `data class Mealtimes(val eveningSnack: ..., ...)` and add one `val mealtimes: Mealtimes` field. This of course doesn't apply to the database schema, but everywhere else you could get rid of the repeated prefix with composition, just like we're already doing with many other things in the daycare data (`val visitingAddress: VisitingAddress`, `val unitManager: UnitManager`, etc...)

PR #4967, opened 2024-04-03
List voucher values for each service need

vesahonkasaari:


service/src/main/kotlin/fi/espoo/evaka/invoicing/controller/FinanceBasicsController.kt
```diff
@@ -131,6 +134,26 @@ class FinanceBasicsController(
         }
         Audit.FinanceBasicsFeeThresholdsUpdate.log(targetId = id)
     }
+
+    @GetMapping("/voucher-values")
+    fun getVoucherValues(
+        db: Database,
+        user: AuthenticatedUser,
```
Joosakur, line 141: `AuthenticatedUser.Employee` to avoid extra clients

frontend/src/lib-customizations/defaults/employee/i18n/fi.tsx
```diff
@@ -4319,7 +4320,52 @@ export const fi = {
       }
     },
     serviceNeeds: {
-      title: 'Palveluntarpeet'
+      title: 'Palveluntarpeet',
+      voucherValues: 'Palvelusetelien arvot',
+      validity: 'Voimassaoloaika',
+      // baseValue: 'Perusarvo, 3v tai yli',
```
Joosakur, line 4326: probably forgotten comment

frontend/src/employee-frontend/components/finance-basics/PlacementTypeItem.tsx
```diff
@@ -44,6 +47,7 @@ export default React.memo(function PlacementTypeItem({
           <ServiceNeedItem
             key={serviceNeed.id}
             serviceNeed={serviceNeed.nameFi + ' (oletus)'}
+            voucherValuesList={voucherValuesMap[serviceNeed.id]}
```
Joosakur, line 50: If there is no voucher value for this service need yet, then I suppose this might give undefined instead of empty list? This is one of the rare cases where type system can't be trusted with its current config. Maybe `voucherValuesMap[serviceNeed.id] ?? []` could work.

vesahonkasaari: Good catch, that would seem to be an appropriate fix

Joosakur (approved):

PR #4993, opened 2024-04-08
Näytetään viestin vastaanottajamäärä ja varoitetaan jos enemmän kuin kaksi

Joosakur:
#### Summary

- refactor MessageEditor from lib-components to employee-frontend since it is not used elsewhere and does not seem to be close to reusable anyway
- fetch and display recipient count (number of guardians)
- if 3 or more recipients, display warning and require confirmation before sending
- not implemented for employee-mobile in this PR


Gekkio (approved):
E2E test?

PR #5031, opened 2024-04-17
Add constraint triggers for guardian blocklist

Gekkio:


service/src/main/kotlin/fi/espoo/evaka/pis/service/GuardianQueries.kt
```diff
@@ -51,7 +56,8 @@ fun Database.Transaction.addToGuardianBlocklist(childId: ChildId, guardianId: Pe
         .execute()
 }
 
-fun Database.Transaction.deleteFromGuardianBlocklist(childId: ChildId, guardianId: PersonId) {
+/** Unblocks a child <-> VTJ guardian relationship by removing it from the blocklist if it exists */
+fun Database.Transaction.unblockGuardian(childId: ChildId, guardianId: PersonId) {
```
Joosakur, line 60: I guess no easy way to trigger vtj-refresh here?

Joosakur (approved):

service/src/main/kotlin/fi/espoo/evaka/pis/service/GuardianQueries.kt
Gekkio: Yeah and then it wouldn't really be a query anymore but a more complicated and expensive operation (= something that doesn't belong to `GuardianQueries.kt`)

PR #5060, opened 2024-04-24
Turvakieltoesto ei koske päätöstiedostoja joissa ei ole henkilöiden osoitetietoja

Joosakur:

### DIFF
--- a/service/src/integrationTest/kotlin/fi/espoo/evaka/decision/DecisionControllerIntegrationTest.kt
+++ b/service/src/integrationTest/kotlin/fi/espoo/evaka/decision/DecisionControllerIntegrationTest.kt
@@ -1,0 +1,288 @@
+// SPDX-FileCopyrightText: 2017-2024 City of Espoo
+//
+// SPDX-License-Identifier: LGPL-2.1-or-later
+
+package fi.espoo.evaka.decision
+
+import fi.espoo.evaka.FullApplicationTest
+import fi.espoo.evaka.application.ApplicationStateService
+import fi.espoo.evaka.application.ApplicationStatus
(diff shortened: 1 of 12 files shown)


service/src/integrationTest/kotlin/fi/espoo/evaka/decision/DecisionControllerIntegrationTest.kt
```diff
@@ -0,0 +1,288 @@
...
+
+    @Autowired lateinit var decisionController: DecisionController
+
+    @Autowired lateinit var applicationStateService: ApplicationStateService
+
+    @Autowired lateinit var asyncJobRunner: AsyncJobRunner<AsyncJob>
+
+    @Test
+    fun `PDF without contact info can be downloaded by service worker`() {
+        val serviceWorker =
```
Gekkio, line 54: You can reduce some duplication by saving the DevEmployee, and converting it to a AuthenticatedUser directly:

```
val serviceWorker = DevEmployee(roles = setOf(UserRole.SERVICE_WORKER))
db.transaction { it.insert(serviceWorker) }
...
downloadPdf(serviceWorker.user, decisionId)
```

Basically there should almost never be a reason to instantiate AuthenticatedUser.Employee manually and duplicate the id and roles, because you're probably adding the same user to the db using DevEmployee. Also applies to citizen users, although you need to pass the auth level: `val citizenUser = somedevperson.user(CitizenAuthLevel.Strong)`

Joosakur: ah nice! this was helpful in another PR too

PR #5061, opened 2024-04-24
Yksikölle erilliset vuorohoidon aukioloajat

reynders:
Jos yksikölle on valittu ilta/viikonloppuhoito -valinta, näytetään yksikön tiedoissa kyseiselle hoitotyypille kellonajat. Näitä kellonaikoja voidaan myöhemmin käyttää esimerkiksi varaustoiminnallisuudessa näyttämään vain palveluntarpeen tyypin mukaiset varausajat.


service/src/main/resources/db/migration/V407__daycare_shiftcare_operation_times.sql
```diff
@@ -0,0 +1,27 @@
...
+UPDATE daycare
+SET shift_care_operation_times = array [
+    CASE WHEN 1 = ANY (shift_care_operation_days) THEN '(00:00,23:59)'::timerange_non_nullable_range END,
+    CASE WHEN 2 = ANY (shift_care_operation_days) THEN '(00:00,23:59)'::timerange_non_nullable_range END,
+    CASE WHEN 3 = ANY (shift_care_operation_days) THEN '(00:00,23:59)'::timerange_non_nullable_range END,
+    CASE WHEN 4 = ANY (shift_care_operation_days) THEN '(00:00,23:59)'::timerange_non_nullable_range END,
+    CASE WHEN 5 = ANY (shift_care_operation_days) THEN '(00:00,23:59)'::timerange_non_nullable_range END,
+    CASE WHEN 6 = ANY (shift_care_operation_days) THEN '(00:00,23:59)'::timerange_non_nullable_range END,
+    CASE WHEN 7 = ANY (shift_care_operation_days) THEN '(00:00,23:59)'::timerange_non_nullable_range END
+    ];
```
Joosakur, line 27: At this point shift_care_operation_times is always a null-filled array --> generated shift_care_operation_times is always an empty array --> this update won't do anything. 

Maybe we should fill those '(00:00,23:59)' for units where `round_the_clock` is true? Also we should consider renaming that as `has_shift_care_times` or something. Also there should maybe be a constraint, that when it is false then shift_care_operation_days is empty. Or better yet, it could maybe be a generated column too. Currently this allows for combinations that don't make sense.

reynders: That initialization part is a no-op, true. Other than that I think whatever we do with shift_care_operation_time should be done to operation_times too. And although it is unclear, shift_care_operation_times are being reset if the round_the_clock -flag is set to false

reynders: I added a test that the shift care times reset when round the clock is unset

Joosakur (approved):

PR #5062, opened 2024-04-24
Ruokatilauksissa käytettävän erityisruokavalion valinta lapsen tietoihin

Wnt:
Tämä muutos lisää henkilökunnan käyttöliittymään lapsen "Lisätietoja" kohtaan erityisruokavaliovalinnan:
image

Tietokannan puolelle tämä muutos lisää uuden tietokantataulun `special_diets` ja nullable relaation siihen `child` tauluun.

Tämän muutoksen myötä henkilökunnan API:n tulee kaksi uutta endpointtia:
- `GET /diets` - palauttaa listan eVakan erityisruokavalioista
- `PUT /diets` - työkalu jolla ylläpitotiimi voi syöttää Jamix APIn palauttaman erityisruokavaliolistauksen eVakaan

### DIFF
--- a/service/src/main/kotlin/fi/espoo/evaka/daycare/ChildQueries.kt
+++ b/service/src/main/kotlin/fi/espoo/evaka/daycare/ChildQueries.kt
@@ -69,11 +69,6 @@ fun Database.Transaction.updateChild(child: Child) {
         .bind("medication", child.additionalInformation.medication)
         .bind("languageAtHome", child.additionalInformation.languageAtHome)
         .bind("languageAtHomeDetails", child.additionalInformation.languageAtHomeDetails)
-        .bind(
-            "dietId",
-            if (child.additionalInformation.specialDiet != null)
-                child.additionalInformation.specialDiet.id
-            else null
-        )
+        .bind("dietId", child.additionalInformation.specialDiet?.id)
         .execute()
 }


service/src/main/kotlin/fi/espoo/evaka/specialdiet/SpecialDietQueries.kt
```diff
@@ -0,0 +1,35 @@
...
+import fi.espoo.evaka.shared.db.Database
+import org.jdbi.v3.core.mapper.PropagateNull
+
+data class SpecialDiet(@PropagateNull val id: Int?, val name: String, val abbreviation: String)
+
+fun Database.Transaction.setSpecialDiets(specialDietList: List<SpecialDiet>) {
+    val newSpecialDietIds = specialDietList.map { it.id }
+    execute {
+        sql("UPDATE child SET diet_id = null WHERE diet_id != ALL (${bind(newSpecialDietIds)})")
+    }
```
Joosakur, line 16: Fine for now, but for a longer term plan this could be unsafe with severe results if some children suddenly no longer have any diet and no one will notice, even if jamix removing them is unlikely. Not sure how to approach that. Might need to set some diet_missing flag which can then be discovered and handled by users through some report / mega warning. Should be discussed with UX/POs.

service/src/main/kotlin/fi/espoo/evaka/daycare/ChildQueries.kt
```diff
@@ -64,5 +69,11 @@ fun Database.Transaction.updateChild(child: Child) {
         .bind("medication", child.additionalInformation.medication)
         .bind("languageAtHome", child.additionalInformation.languageAtHome)
         .bind("languageAtHomeDetails", child.additionalInformation.languageAtHomeDetails)
+        .bind(
+            "dietId",
+            if (child.additionalInformation.specialDiet != null)
+                child.additionalInformation.specialDiet.id
+            else null
```
Joosakur, line 76: simplify to `child.additionalInformation.specialDiet?.id`?

service/src/main/kotlin/fi/espoo/evaka/specialdiet/SpecialDietController.kt
```diff
@@ -0,0 +1,91 @@
...
+import fi.espoo.evaka.shared.domain.EvakaClock
+import fi.espoo.evaka.shared.security.AccessControl
+import fi.espoo.evaka.shared.security.Action
+import org.springframework.web.bind.annotation.*
+
+data class JamixSpecialDiet(val modelId: Int, val fields: JamixSpecialDietFields)
+
+data class JamixSpecialDietFields(val dietName: String, val dietAbbreviation: String)
+
+fun cleanupJamixDietList(specialDietList: List<JamixSpecialDiet>): List<SpecialDiet> {
```
Joosakur, line 19: consider adding a simple unit test for this to verify different cases

service/src/main/kotlin/fi/espoo/evaka/specialdiet/SpecialDietController.kt
```diff
@@ -0,0 +1,91 @@
...
+                cleanupJamixDietAbbreviationString(it.fields.dietAbbreviation)
+            )
+        }
+        .filterNot { it.name.contains("POISTA") }
+        .filterNot { it.name.isEmpty() && it.abbreviation.isEmpty() }
+}
+
+fun cleanupJamixDietNameString(s: String): String {
+    return s.replace("tsekattava", "", true)
+        .replace("tsek", "", true)
```
Joosakur, line 34: Wondering if there's any allergies that could contain "tsek" within them. Would `"\\stsek"` work?

service/src/main/kotlin/fi/espoo/evaka/specialdiet/SpecialDietController.kt
```diff
@@ -0,0 +1,91 @@
...
+        .replace(Regex("\\s+"), " ")
+        .trim()
+}
+
+fun cleanupJamixDietAbbreviationString(s: String): String {
+    return s.replace("ätarkasta", "", true).replace(Regex("\\s+"), " ").trim()
```
Joosakur, line 40: just double checking that there's no typo in "ätarkasta"

service/src/main/kotlin/fi/espoo/evaka/specialdiet/SpecialDietController.kt
Wnt: I was planning to add tests for this once I move this code to it's final location. e.g. when we are doing the Jamix GET /diets API call directly from eVaka. For now this is just a quickly made tool which implementation details might change before production use.

service/src/main/kotlin/fi/espoo/evaka/specialdiet/SpecialDietController.kt
Wnt: Good point. The current list returned by Jamix API doesn't seem to contain any. But something we need to consider when this tool evolves into a real diet list sync integreation between Jamix & eVaka

service/src/main/kotlin/fi/espoo/evaka/specialdiet/SpecialDietController.kt
Wnt: No typo. My guess is that by prepending the abbreviation with "ä" all of these get sorted to the end of list in some Jamix system and do not clutter the list of actually used values.

Joosakur (approved):

PR #5066, opened 2024-04-25
Tallennetaan ja näytetään talouden aikajanalla tieto päämiessuhteen lisääjästä/muokkaajasta

Joosakur:
Screenshot 2024-04-26 at 12 42 41


service/src/main/kotlin/fi/espoo/evaka/pis/CreatorsAndModifiers.kt
```diff
@@ -0,0 +1,90 @@
...
+
+enum class CreateSource : DatabaseEnum {
+    USER,
+    APPLICATION,
+    DVV;
+
+    override val sqlType = "create_source"
+}
+
+sealed class Creator {
```
reynders, line 21: I'm happy with "creator" but maybe "origin" or "source" would have been more common terms for this

reynders (approved):

service/src/main/kotlin/fi/espoo/evaka/pis/CreatorsAndModifiers.kt
Joosakur: There's also Modifier which relates to modified_at/by while Creator relates to created_at/by, so I think this is at least consistent, even if a bit weird sounding otherwise.

PR #5080, opened 2024-04-29
Näytetään uuden Varda-integraation virheet Varda-virheet -raportilla

akheron:


service/src/main/kotlin/fi/espoo/evaka/reports/VardaErrorsReport.kt
```diff
@@ -58,17 +58,35 @@ FROM varda_service_need vsn
...
+    NULL AS service_need_id,
+    NULL AS service_need_validity,
+    NULL AS service_need_option_name,
+    child_id,
+    errored_at AS updated,
+    coalesce(last_success_at, created_at) AS created,
+    ARRAY[error] AS errors,
+    NULL AS reset_timestamp
+FROM varda_state
+WHERE error IS NOT NULL
```
Joosakur, line 78: partial index was for errored_at so should maybe use that here too

Joosakur (approved):

service/src/main/kotlin/fi/espoo/evaka/reports/VardaErrorsReport.kt
akheron: Good catch, fixed.

PR #5204, opened 2024-05-24
Korjaus null pointer exceptioniin palse-arvojen luonnissa

vesahonkasaari:
Palse-arvon lisääminen palvelutarpeelle jolla ei ollut ennestään yhtään palse-arvoa epäonnistui, koska servicellä tuli null pointer exception

### DIFF
--- a/service/src/main/kotlin/fi/espoo/evaka/invoicing/controller/FinanceBasicsController.kt
+++ b/service/src/main/kotlin/fi/espoo/evaka/invoicing/controller/FinanceBasicsController.kt
@@ -177,13 +177,17 @@ class FinanceBasicsController(
                     if (serviceNeedOption == null)
                         throw BadRequest("Invalid service need option ID")
 
-                    val currentVoucherValues =
-                        tx.getVoucherValuesByServiceNeedOption()[body.serviceNeedOptionId]!!
+                    val currentVoucherValues = tx.getVoucherValuesByServiceNeedOption()
 
-                    val latest = currentVoucherValues.maxByOrNull { it.voucherValues.range.start }
+                    if (currentVoucherValues.containsKey(body.serviceNeedOptionId)) {
+                        val latest =
+                            currentVoucherValues[body.serviceNeedOptionId]!!.maxBy {
+                                it.voucherValues.range.start
+                            }
 
-                    if (latest != null && latest.voucherValues.range.end == null)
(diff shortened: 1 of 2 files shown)


service/src/main/kotlin/fi/espoo/evaka/invoicing/controller/FinanceBasicsController.kt
```diff
@@ -176,13 +176,17 @@ class FinanceBasicsController(
...
+                        val latest =
+                            currentVoucherValues[body.serviceNeedOptionId]!!.maxBy {
+                                it.voucherValues.range.start
+                            }
 
-                    if (latest != null && latest.voucherValues.range.end == null)
-                        tx.updateVoucherValueEndDate(latest.id, body.range.start)
+                        if (latest.voucherValues.range.end == null)
+                            tx.updateVoucherValueEndDate(latest.id, body.range.start)
+                    }
```
Joosakur, line 189: I wonder if this can be done without the unsafe `!!`. Also maxBy appears to be deprecated. How about something like
```
currentVoucherValues.getOrDefault(body.serviceNeedOptionId, emptyList())
    .maxByOrNull { it.voucherValues.range.start }
    ?.also { latest ->
        if (latest.voucherValues.range.end == null)
            tx.updateVoucherValueEndDate(latest.id, body.range.start)
    }
```

Also it's a bit confusing that `endDate` here is exclusive, unlike in almost all other places, so I would instead use
`tx.updateVoucherValueEndDate(latest.id, body.range.start.minusDays(1))`
and in the query
`SET validity = daterange(lower(validity), ${bind(endDate)}, '[])`

vesahonkasaari: That !! shouldn't really be unsafe, it's just working around the compiler not being smart enough to realize that the containsKey() check means that the map access necessarily returns a non-null value. But I see your points in general, let's iterate.

Joosakur: Yep true, in this case it isn't, but when reading code it always forces one to stop and think whether this manual guarantee is still valid, so if the same thing can be easily written in a way that allows compiler to guarantee type-safety, it's generally better, especially in terms of avoiding future regression bugs.

Joosakur (approved):

PR #5242, opened 2024-06-05
Lähetetään varoitus yksikönjohtajille ERV nollaantumisesta

Wnt:
## Ennen tätä muutosta
Kun Jamix erityisruokavalioiden synkronoiinti aiheutti jonkin eVakan lapsen erityisruokavalion asetuksen tyhjäksi, siitä jäi vain systeemin lokille viesti joka kertoi kuinka monen lapsen erityisruokavalio oli nollaantunut.

## Tämän muutoksen jälkeen
Kunkin yksikön johtajille lähetetään sähköpostilla varoitus jossa on eVaka lapsen tunniste ja aikaisemmin asetettu ruokavalio. Sähköpostit lähetetään vain sellaisista lapsista joilla on voimassaoleva sijoitus.

### DIFF
--- a/service/src/main/kotlin/fi/espoo/evaka/children/ChildrenQueries.kt
+++ b/service/src/main/kotlin/fi/espoo/evaka/children/ChildrenQueries.kt
@@ -5,6 +5,7 @@
 package fi.espoo.evaka.children
 
 import fi.espoo.evaka.shared.ChildId
+import fi.espoo.evaka.shared.DaycareId
 import fi.espoo.evaka.shared.PersonId
 import fi.espoo.evaka.shared.db.Database
 import java.time.LocalDate
@@ -64,3 +65,19 @@ SELECT child_id FROM foster_parent WHERE parent_id = ${bind(userId)} AND valid_d
             )
         }
         .toList()
+
+fun Database.Read.getActivePlacementUnitsForChildren(
+    today: LocalDate,
+    childIds: Set<ChildId>
+): Map<DaycareId, List<ChildId>> =
+    createQuery {
+            sql(
+                """
+SELECT unit_id, child_id
+FROM placement
+WHERE child_id = ANY (${bind(childIds)}) AND start_date <= ${bind(today)} AND end_date >= ${bind(today)}
+"""
+            )
+        }
+        .toList { column<DaycareId>("unit_id") to column<ChildId>("child_id") }
+        .groupBy({ it.first }, { it.second })
--- a/service/src/main/kotlin/fi/espoo/evaka/children/ChildrenQueries.kt
+++ b/service/src/main/kotlin/fi/espoo/evaka/children/ChildrenQueries.kt
@@ -75,7 +75,7 @@ fun Database.Read.getActivePlacementUnitsForChildren(
                 """
 SELECT unit_id, child_id
 FROM placement
-WHERE child_id = ANY (${bind(childIds)}) AND start_date <= ${bind(today)} AND end_date >= ${bind(today)}
+WHERE child_id = ANY (${bind(childIds)}) AND daterange(start_date, end_date, '[]') @> ${bind(today)}
 """
             )
         }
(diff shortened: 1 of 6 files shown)


service/src/main/kotlin/fi/espoo/evaka/children/ChildrenQueries.kt
```diff
@@ -64,3 +65,19 @@ SELECT child_id FROM foster_parent WHERE parent_id = ${bind(userId)} AND valid_d
...
+fun Database.Read.getActivePlacementUnitsForChildren(
+    today: LocalDate,
+    childIds: Set<ChildId>
+): Map<DaycareId, List<ChildId>> =
+    createQuery {
+            sql(
+                """
+SELECT unit_id, child_id
+FROM placement
+WHERE child_id = ANY (${bind(childIds)}) AND start_date <= ${bind(today)} AND end_date >= ${bind(today)}
```
Joosakur, line 78: I think range operations are usually preferred
`AND daterange(start_date, end_date, '[]') @> ${bind(today)}`

service/src/main/kotlin/fi/espoo/evaka/jamix/JamixService.kt
```diff
@@ -92,38 +105,99 @@ class JamixService(
 
     fun syncDiets(db: Database.Connection, clock: EvakaClock, job: AsyncJob.SyncJamixDiets) {
         if (client == null) error("Cannot sync diet list: JamixEnv is not configured")
-        fetchAndUpdateJamixDiets(client, db)
+        val warningEmails =
+            fetchAndUpdateJamixDiets(client, db, clock, emailEnv.sender(Language.fi))
         fetchAndUpdateJamixTextures(client, db)
+        warningEmails.forEach { emailClient.send(it) }
```
Joosakur, line 111: This doesn't look transactionally safe. If for example `fetchAndUpdateJamixTextures` fails then the whole job is run again, but the diet deletion has already been committed, so emails won't be sent on a retry either. Also need to be careful of not sending already sent emails again. I believe the correct way is to plan email sending async jobs as part of the same transaction that nulls the diets.

Joosakur (changes requested):

service/src/main/kotlin/fi/espoo/evaka/jamix/JamixService.kt
```diff
@@ -106,11 +106,15 @@ class JamixService(
         if (client == null) error("Cannot sync diet list: JamixEnv is not configured")
         val warningEmails =
             fetchAndUpdateJamixDiets(client, db, clock, emailEnv.sender(Language.fi))
-        fetchAndUpdateJamixTextures(client, db)
-        warningEmails.forEach { emailClient.send(it) }
+        try {
+            fetchAndUpdateJamixTextures(client, db)
+        } finally {
+            warningEmails.forEach { emailClient.send(it) }
```
Joosakur, line 112: Now any error in the texture sync is silently ignored, so that probably isn't what we want. Also this still has plenty of error scenarios that aren't properly handled and exactly-once-delivery can not be guaranteed with this approach. For that each email needs to be it's own separately retryable job, and those need to be planned within the same transaction that `fetchAndUpdateJamixDiets` uses. 

On the other hand this is a feature that triggers very rarely, so maybe it doesn't have to be bullet proof. Then again, if it ever does trigger it is quite crucial. So I don't know, could quickly discuss this after daily with others.

Wnt: I added error logging and a try-catc around `emailClient.send(it)`

Joosakur: Ok, maybe it's good enough for now. I'll add some ticket on the bottom of tech backlog to make it less of a hax later if there's time.

Joosakur (approved):

PR #5245, opened 2024-06-06
Tilapäiset ilmoitukset kuten häiriötiedotteet kuntalaiselle ja työntekijän mobiiliin

Joosakur:


frontend/src/employee-mobile-frontend/common/TopBar.tsx
```diff
@@ -68,40 +74,87 @@ export default React.memo(function TopBar({
 }: Props) {
   const { i18n } = useTranslation()
 
+  const notificationResult = useQueryResult(currentSystemNotificationQuery())
+
+  // only shown on main screen
+  const hideSystemNotification = invertedColors
```
Gekkio, line 80: I think the notification gets refetched basically on every page change, because TopBar is remounted on pretty much every page and the hook is unaware when we're not going to show the data. :thinking:

frontend/src/employee-mobile-frontend/common/TopBar.tsx
```diff
@@ -68,40 +74,87 @@ export default React.memo(function TopBar({
...
+
+  const [showNotificationModal, setShowNotificationModal] = useState(false)
```
Gekkio, line 82: `useBoolean` would give you on/off functions for free

Gekkio (approved):

frontend/src/employee-mobile-frontend/common/TopBar.tsx
Joosakur: for some reason it doesn't seem to, but I'll add the query enabled option to be sure

frontend/src/employee-mobile-frontend/common/TopBar.tsx
Joosakur: At least when there's no need for fancier things like `toggle`, I personally find the standard hook more readable and with less naming to think about, and `useBoolean` seems like an over-abstraction. 🤔  For example here the name for the `on`-function would naturally be `showNotificationModal`, but then what should the boolean value be called?

PR #5311, opened 2024-06-24
Vahvistetaan hakemuksen poistaminen käsittelystä

Joosakur:

### DIFF
--- a/frontend/src/lib-components/molecules/modals/InfoModal.tsx
+++ b/frontend/src/lib-components/molecules/modals/InfoModal.tsx
@@ -54,14 +54,22 @@ export default React.memo(function InfoModal({ children, ...props }: Props) {
           <Button
             appearance="inline"
             data-qa="modal-okBtn"
-            onClick={props.resolve.action}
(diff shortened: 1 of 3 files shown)


frontend/src/lib-components/molecules/modals/InfoModal.tsx
```diff
@@ -54,14 +54,22 @@ export default React.memo(function InfoModal({ children, ...props }: Props) {
           <Button
             appearance="inline"
             data-qa="modal-okBtn"
-            onClick={props.resolve.action}
+            onClick={(e) => {
+              e.preventDefault()
+              e.stopPropagation()
```
akheron, line 59: Why add `preventDefault` and `stopPropagation` calls here?

Joosakur: When I was clicking the modal buttons and the application row was behind them, that also got clicked and opened. This helped, but maybe only one of them is necessary.

reynders (approved):

PR #5317, opened 2024-06-25
Ei näytetä suljettuja yksiköitä toimintojen avauksissa

Joosakur:

### DIFF
--- a/service/src/main/kotlin/fi/espoo/evaka/daycare/DaycareQueries.kt
+++ b/service/src/main/kotlin/fi/espoo/evaka/daycare/DaycareQueries.kt
@@ -418,12 +418,13 @@ WHERE daycare_id = ${bind(daycareId)}
         }
         .toList()
 
-fun Database.Read.getUnitFeatures(): List<UnitFeatures> =
+fun Database.Read.getUnitFeatures(today: LocalDate): List<UnitFeatures> =
     createQuery {
             sql(
                 """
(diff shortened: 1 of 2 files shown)


service/src/main/kotlin/fi/espoo/evaka/daycare/DaycareQueries.kt
```diff
@@ -418,12 +418,13 @@ WHERE daycare_id = ${bind(daycareId)}
...
         .toList()
 
-fun Database.Read.getUnitFeatures(): List<UnitFeatures> =
+fun Database.Read.getUnitFeatures(today: LocalDate): List<UnitFeatures> =
     createQuery {
             sql(
                 """
 SELECT id, name, enabled_pilot_features AS features, provider_type, type
 FROM daycare
+WHERE closing_date IS NULL OR closing_date > ${bind(today)}
```
reynders, line 427: Nitpicking - shows closed units that are closed today

reynders (approved):

service/src/main/kotlin/fi/espoo/evaka/daycare/DaycareQueries.kt
Joosakur: Hmm I don't even know if that means that the unit is still open today or not. Usually evaka's end dates are inclusive so the unit would still be open on the closing date, but not sure.

PR #5321, opened 2024-06-26
Henkilökunta voi editoida ainoastaan omia läsnäolojaan

reynders:

### DIFF
--- a/service/src/main/kotlin/fi/espoo/evaka/attendance/RealtimeStaffAttendanceController.kt
+++ b/service/src/main/kotlin/fi/espoo/evaka/attendance/RealtimeStaffAttendanceController.kt
@@ -79,6 +79,16 @@ class RealtimeStaffAttendanceController(private val accessControl: AccessControl
                                     staffForAttendanceCalendar.map { it.id }
                             )
                             .groupBy { it.employeeId }
+
+                    val allowedToEdit =
+                        accessControl.checkPermissionFor(
+                            tx,
+                            user,
+                            clock,
+                            Action.Employee.UPDATE_STAFF_ATTENDANCES,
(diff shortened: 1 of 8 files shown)


service/src/main/kotlin/fi/espoo/evaka/attendance/RealtimeStaffAttendanceController.kt
```diff
@@ -205,6 +216,13 @@ class RealtimeStaffAttendanceController(private val accessControl: AccessControl
                         body.entries.mapNotNull { it.id }
                     )
                     body.entries.map { entry ->
+                        accessControl.requirePermissionFor(
```
Gekkio, line 219: This doesn't need to be executed in a loop multiple times, because all required data is always available. Maybe it could be moved to right after the Unit.UPDATE_STAFF_ATTENDANCES check?

service/src/main/kotlin/fi/espoo/evaka/attendance/RealtimeStaffAttendanceController.kt
```diff
@@ -101,7 +101,17 @@ class RealtimeStaffAttendanceController(private val accessControl: AccessControl
                                                 att.departedAutomatically
                                             )
                                         },
-                                plannedAttendances = plannedAttendances[employeeId] ?: emptyList()
+                                plannedAttendances = plannedAttendances[employeeId] ?: emptyList(),
+                                allowedToEdit =
+                                    accessControl
+                                        .checkPermissionFor(
```
Gekkio, line 107: Running this in a loop can cause bad performance issues. Use the batch version of checkPermissionFor outside the loop (so you get a `Map<EmployeeId, AccessControlDecision>`) and lookup from the map instead

Gekkio (approved):

PR #5339, opened 2024-07-02
Käytetään kunnallisen tiedotteen otsikkoa sähköpostiherätteen kuvauksessa

reynders:

### DIFF
--- a/service/src/main/kotlin/fi/espoo/evaka/emailclient/EvakaEmailMessageProvider.kt
+++ b/service/src/main/kotlin/fi/espoo/evaka/emailclient/EvakaEmailMessageProvider.kt
@@ -298,10 +298,13 @@ $unsubscribeEn
     override fun assistanceNeedPreschoolDecisionNotification(language: Language): EmailContent =
         assistanceNeedDecisionNotification(language) // currently same content
 
+    override fun messageNotification(language: Language, thread: MessageThreadStub): EmailContent =
+        messageNotification(language, thread, false)
+
     override fun messageNotification(
         language: Language,
         thread: MessageThreadStub,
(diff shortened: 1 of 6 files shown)


service/src/main/kotlin/fi/espoo/evaka/emailclient/EvakaEmailMessageProvider.kt
```diff
@@ -319,7 +323,9 @@ $unsubscribeEn
                     else Triple("tiedote", "allmänt meddelande", "bulletin")
             }
         return EmailContent.fromHtml(
-            subject = "Uusi $typeFi eVakassa / Nytt $typeSv i eVaka / New $typeEn in eVaka",
+            subject =
+                if (showMessageSubjectInEmailSubject) thread.title
```
Gekkio, line 327: This needs HTML escaping (with `HtmlEscape.escapeHtml5`), because it's directly embedded as raw HTML and the data comes from a user and is considered untrusted

Gekkio (approved):

service/src/main/kotlin/fi/espoo/evaka/emailclient/IEmailMessageProvider.kt
```diff
@@ -74,7 +74,11 @@ interface IEmailMessageProvider {
 
     fun missingHolidayReservationsNotification(language: Language): EmailContent
 
-    fun messageNotification(language: Language, thread: MessageThreadStub): EmailContent
+    fun messageNotification(
+        language: Language,
+        thread: MessageThreadStub,
+        showMessageSubjectInEmailSubject: Boolean
```
Gekkio, line 80: Note that this breaks other builds since we're changing the API interface

PR #5483, opened 2024-08-12
Korjataan laskugeneroinnin suorituskyky, ja parannellaan koodia

Gekkio:

### DIFF
--- a/service/src/main/kotlin/fi/espoo/evaka/invoicing/service/DraftInvoiceGenerator.kt
+++ b/service/src/main/kotlin/fi/espoo/evaka/invoicing/service/DraftInvoiceGenerator.kt
@@ -82,29 +82,18 @@ class DraftInvoiceGenerator(
 ) {
     fun generateDraftInvoices(
         tx: Database.Read,
-        decisions: Map<PersonId, List<FeeDecision>>,
-        permanentPlacements: Map<ChildId, List<Pair<FiniteDateRange, PlacementStub>>>,
-        temporaryPlacements: Map<PersonId, List<Pair<FiniteDateRange, PlacementStub>>>,
-        period: FiniteDateRange,
-        daycareCodes: Map<DaycareId, AreaId>,
-        operationalDaysByChild: Map<ChildId, Set<LocalDate>>,
-        businessDays: Set<LocalDate>,
-        feeThresholds: FeeThresholds,
-        absences: List<AbsenceStub>,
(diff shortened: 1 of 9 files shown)


service/src/main/kotlin/fi/espoo/evaka/invoicing/service/InvoiceGenerator.kt
```diff
@@ -126,7 +120,10 @@ class InvoiceGenerator(
                 .toSet() +
                 permanentPlacements.keys +
                 temporaryPlacements.values.flatMap { pairs -> pairs.map { it.second.child.id } }
-        val operationalDaysByChild = tx.getOperationalDatesForChildren(range, allChildren)
+        val operationalDaysByChild =
+            tx.getOperationalDatesForChildren(range, allChildren).mapValues {
+                DateSet.of(it.value.map(LocalDate::toFiniteDateRange))
```
Joosakur, line 125: might be more readable to add an overload for DateSet.of which accepts LocalDates directly

service/src/main/kotlin/fi/espoo/evaka/invoicing/service/DraftInvoiceGenerator.kt
```diff
@@ -113,7 +112,13 @@ class DraftInvoiceGenerator(
                         data.areaIds,
                         data.businessDays,
                         data.feeThresholds,
-                        absencesByChild,
+                        getChildAbsences =
+                            memoize { child: ChildId ->
+                                (absencesByChild[child] ?: emptyList()).associateBy(
```
Joosakur, line 117: I think there can be up to 2 absence types per date so `associate` kind of randomly looses one? Not sure if the same issue exists in the original code too.

Gekkio: Only billable absences are fetched so there can be only one per date. It's just not obvious from the type `List<AbsenceStub>`, which is one of the reasons why I've changed most functions to accept absence data using types that either make this either explicit (e.g. `Map<LocalDate, AbsenceType>`) or remove the need for this detail completely (e.g. `hasPlannedAbsence: (date: LocalDate) => Boolean`)

Joosakur: ah gotcha, all good then

Joosakur (approved):

PR #5589, opened 2024-08-30
Henkilökuntalistaan lisätty "piilotetaan deaktivoidu"-filtteri

reynders:


frontend/src/employee-frontend/components/employees/EmployeesPage.tsx
```diff
@@ -71,6 +79,18 @@ export default React.memo(function EmployeesPage() {
               icon={faSearch}
               width="L"
             />
+            <FixedSpaceRow>
+              <Gap size="xs" />
+              <Checkbox
```
Joosakur, line 84: This doesn't seem to make sense. Maybe both InputField and Checkbox should be inside the same FixedSpaceRow and Gap removed? Then they are placed next to each other with an automatic gap in between (gap size and vertical alignment can be controlled via props).

Joosakur (approved):

PR #5693, opened 2024-09-19
Tuen päätöksen tukipalveluiden voimassaoloaika valinnaiseksi

reynders:

### DIFF
--- a/service/src/main/kotlin/fi/espoo/evaka/assistanceneed/decision/AssistanceNeedDecision.kt
+++ b/service/src/main/kotlin/fi/espoo/evaka/assistanceneed/decision/AssistanceNeedDecision.kt
@@ -24,6 +24,7 @@ data class AssistanceNeedDecision(
     val decisionNumber: Long? = null,
     @Nested("child") val child: AssistanceNeedDecisionChild?,
     val validityPeriod: DateRange,
+    val endDateNotKnown: Boolean,
     val status: AssistanceNeedDecisionStatus,
     val language: OfficialLanguage,
     val decisionMade: LocalDate?,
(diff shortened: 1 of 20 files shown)


service/src/main/resources/db/migration/V438__assistance_need_decision_no_end_date.sql
```diff
@@ -0,0 +1,7 @@
+ALTER TABLE assistance_need_decision
+    ADD COLUMN end_date_not_known BOOLEAN DEFAULT FALSE NOT NULL,
```
Joosakur, line 2: default should be dropped afterwards, we now prefer always explicitly setting values and avoiding defaults

service/src/main/resources/db/migration/V438__assistance_need_decision_no_end_date.sql
```diff
@@ -0,0 +1,7 @@
...
+    ADD CONSTRAINT check$assistance_need_decision_validity_period_end_date
+        CHECK (
+            'ASSISTANCE_SERVICES_FOR_TIME' <> ANY(assistance_levels) OR
+            (end_date_not_known = FALSE OR upper(validity_period) IS NULL)
+            );
```
Joosakur, line 7: I think there's some error in the logic here. For example it always allows end date to be missing irrespective of the other conditions. All these three conditions are combined with OR so currently it's enough that any one of them is TRUE.

service/src/main/kotlin/fi/espoo/evaka/assistanceneed/decision/AssistanceNeedDecision.kt
```diff
@@ -24,6 +24,7 @@ data class AssistanceNeedDecision(
     val decisionNumber: Long? = null,
     @Nested("child") val child: AssistanceNeedDecisionChild?,
     val validityPeriod: DateRange,
+    val endDateNotKnown: Boolean = false,
```
Joosakur, line 27: probably wouldn't use defaults here either

Joosakur (approved):

PR #5837, opened 2024-10-18
Näytetään ryhmitystä odottavat lapset vain jos sijoitus on seuraavan kuuden kuukauden aikana

Joosakur:

### DIFF
--- a/service/src/main/kotlin/fi/espoo/evaka/daycare/controllers/DaycareController.kt
+++ b/service/src/main/kotlin/fi/espoo/evaka/daycare/controllers/DaycareController.kt
@@ -504,7 +504,7 @@ class DaycareController(
                                 unitId,
                             )
                         ) {
-                            getMissingGroupPlacements(tx, unitId)
+                            getMissingGroupPlacementsWithinSixMonths(tx, unitId, clock.today())
                         } else {
                             emptyList()
                         }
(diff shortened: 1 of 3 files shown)


service/src/main/kotlin/fi/espoo/evaka/daycare/controllers/DaycareController.kt
```diff
@@ -619,6 +619,12 @@ class DaycareController(
             .also { Audit.UnitView.log(targetId = AuditId(unitId)) }
     }
 
+    private fun getMissingGroupPlacementsWithinSixMonths(
+        tx: Database.Read,
+        unitId: DaycareId,
+        today: LocalDate,
+    ) = getMissingGroupPlacements(tx, unitId).filter { it.gap.start <= today.plusMonths(6) }
```
terolaakso, line 626: `getMissingGroupPlacements` is only used in these two places. How about including the 6 month limit to the SQL instead?

Joosakur: Moved the filtering now to the query function. I suppose it could also go within the SQL itself, but I'm a bit unsure how to do that since it's based on `unnest(p.ranges) AS gap`. I could of course wrap the result in another CTE and add a WHERE-clause to that, but the readability downside might not be worth the performance upside.

service/src/main/kotlin/fi/espoo/evaka/placement/PlacementService.kt
```diff
@@ -644,6 +648,7 @@ LEFT JOIN service_need_option default_sno ON default_sno.default_option AND defa
                 )
             }
             .toList<MissingGroupPlacement>()
+            .filter { it.gap.start <= today.plusMonths(6) }
```
terolaakso, line 651: On line 618 it checks that the result includes only missing group placements starting from `evakaLaunch`. I think similar check could be introduced for `sixMonthsFromToday`. Like

```
AND daterange(p.start_date, p.end_date, '[]') && daterange(NULL, ${bind(sixMonthsFromToday)}, '[]')
```
Perhaps those two dates could be used to create a single `daterange` as well, i.e. from evakaLaunch until sixMonthsFromToday?

But I agree that doing the filtering in Kotlin is more readable.

terolaakso (approved):

PR #6083, opened 2024-12-04
Näytetään hakemuksen muokkaajan nimi hakemusnäkymässä (nyt näkyy vain aikaleima)

kechpaja-at-gofore:
Lisätä muokkaajan nimi hakemusnäkymään muokkausajan alle:

2024-12-04-151136_723x342_scrot

Muutin myös `create` ja `updated` -kenttien nimet standardien mukaisiksi.


service/src/main/kotlin/fi/espoo/evaka/application/ApplicationQueries.kt
```diff
@@ -1005,14 +1026,16 @@ fun Database.Transaction.updateForm(
...
+            "UPDATE application SET document = ${bindJson(transformedForm)}, modified_at = ${bind(now)}, modified_by = ${bind(modifiedBy)} WHERE id = ${bind(id)}"
         )
     }
 }
 
 fun Database.Transaction.resetCheckedByAdminAndConfidentiality(
     id: ApplicationId,
     form: ApplicationForm,
+    now: HelsinkiDateTime,
+    modifiedBy: EvakaUserId,
```
Joosakur, line 1038: Many of the functions here including this one aren't really called by the user. Instead they are sort of secondary side-effects that result from some primary action such as changing status and only update some metadata-like fields an not really the application itself, even though it is an update to the application table. Now one user request ends up updating these fields multiple times which is at least unnecessary although not really harmful. 

One thing to also consider is that if we keep the separate status_modified fields, then should the normal modified_by be changed when only status changes? If not, then these queries which are essentially part of the status update process should probably not update it either.

No right or wrong answers really, just something you could think about if you didn't yet. Otherwise looks good!

kechpaja-at-gofore: > Now one user request ends up updating these fields multiple times which is at least unnecessary although not really harmful.

I agree that this might be worth changing at some point, but I don't think it can be fixed without going too far out of scope for this change. 

WRT the status_modified fields: I think any changes to them are out of scope for this change as well. I could be convinced that they should be removed, but that would need to be a separate work item, and would require rethinking a number of things. As things are, and especially given the possibility of removing status_modified_*, it makes sense to update modified_at every time.

Thanks!

Joosakur: fair enough

Joosakur (approved):

PR #6086, opened 2024-12-05
Lisätään kotikunta

patari:
Tallennetaan VTJ:stä kotikunta.


service/src/main/kotlin/fi/espoo/evaka/pis/service/PersonService.kt
```diff
@@ -323,7 +324,7 @@ class PersonService(private val personDetailsService: IPersonDetailsService) {
                         origin = PersonAddressDTO.Origin.VTJ,
                         streetAddress = person.streetAddress,
                         postalCode = person.postalCode,
-                        city = person.city,
+                        city = person.postOffice,
```
Joosakur, line 327: could consider renaming this field too (if it isn't stored in any json blobs or such that would require migration)

Joosakur (approved):

service/src/main/kotlin/fi/espoo/evaka/pis/service/PersonService.kt
patari: fixed

